PR #7, opened 2017-02-28
Remove cleanup function

Emantor:
This removes the cleanup function and uses on_deactivate instead.
Do we deactivate the targets drivers after a successful test?

### DIFF
--- a/labgrid/target.py
+++ b/labgrid/target.py
@@ -207,8 +207,6 @@ class Target:
     def cleanup(self):
         """Clean up conntected drivers and resources in reversed order"""
         for drv in reversed(self.drivers):
-            if hasattr(drv, 'cleanup') and callable(getattr(drv, 'cleanup')):
-                drv.cleanup()
+            self.deactivate(drv)
         for res in reversed(self.resources):
-            if hasattr(res, 'cleanup') and callable(getattr(res, 'cleanup')):
-                res.cleanup()
+            self.deactivate(res)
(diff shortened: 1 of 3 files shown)


labgrid/target.py
```diff
@@ -203,12 +203,3 @@ def deactivate(self, client):
...
         client.state = BindingState.bound
-
-    def cleanup(self):
-        """Clean up conntected drivers and resources in reversed order"""
-        for drv in reversed(self.drivers):
-            if hasattr(drv, 'cleanup') and callable(getattr(drv, 'cleanup')):
-                drv.cleanup()
-        for res in reversed(self.resources):
-            if hasattr(res, 'cleanup') and callable(getattr(res, 'cleanup')):
-                res.cleanup()
```
jluebbe: This should be kept and deactivate the drivers instead.

jluebbe: This should be kept and deactivate the drivers instead.

labgrid/environment.py
```diff
@@ -33,7 +33,3 @@ def __attrs_post_init__(self):
     def get_target(self, role: str='main') -> Target:
         """Returns the specified target."""
         return self.targets[role]
-
-    def cleanup(self):
```
jluebbe: This should stay, as the cleanup function on drivers should stay as well (see below).

Emantor:
Pushed a new version, pls review again.

jluebbe:
Looks good now. I'll wait for some test results before merging this.

Emantor:
I ran my tests overnight, SSHDriver cleanup function was not called before and throws an error now after the test ends, I'll fix that soon™.

PR #8, opened 2017-02-28
coveralls: exclude lines via .coveragerc and use it for travis

Emantor:
Remove labgrid/protocols/* from coverage.


Emantor:
I updated the branch to only ignore the suggested lines from [here](http://coverage.readthedocs.io/en/coverage-4.2/excluding.html#advanced-exclusion), should be ready to merge now.

PR #14, opened 2017-03-02
remote/client: with 'places' also print aliases and comment

saschahauer:
Signed-off-by: Sascha Hauer <s.hauer@pengutronix.de>

### DIFF
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -159,7 +159,12 @@ class ClientSession(ApplicationSession):
                 print("Place '{}':".format(name))
                 place.show(level=1)
             else:
-                print(name)
+                line = "{}".format(name)
+
+                if place.aliases:
+                    line += " ({})".format(' '.join(place.aliases))
+
+                print("{0:<40s} {1}".format(line, place.comment))
 
     def _match_places(self, pattern):
         result = set()


labgrid/remote/client.py
```diff
@@ -159,7 +159,16 @@ def print_places(self):
                 print("Place '{}':".format(name))
                 place.show(level=1)
             else:
-                print(name)
+                aliases = []
+                line = "{}".format(name)
+
+                for alias in place.aliases:
+                    aliases.append("{}".format(alias))
```
Emantor: Leave this out, see next comment.

labgrid/remote/client.py
```diff
@@ -159,7 +159,16 @@ def print_places(self):
...
+
+                if aliases:
+                    line += " ({})".format(' '.join(aliases))
```
Emantor: replace aliases with place.aliases

labgrid/remote/client.py
```diff
@@ -159,7 +159,16 @@ def print_places(self):
                 print("Place '{}':".format(name))
                 place.show(level=1)
             else:
-                print(name)
+                aliases = []
```
Emantor: remove aliases variable, we don't need to reformat from a set to a list, we can use it directly.

PR #18, opened 2017-03-03
support using remote resources in targets and extend the client tool

jluebbe:


labgrid/target.py
```diff
@@ -126,10 +126,30 @@ def bind_driver(self, client):
...
+            errors = []
+            suppliers = []
+            for requirement in requirements:
+                try:
+                    suppliers.append(self.get(requirement))
+                except NoSupplierFoundError as e:
+                    errors.append(e)
+            if not suppliers:
+                if len(errors) == 1:
+                    raise errors[0]
```
Emantor: Why do we use a errors list here?

jluebbe: If we only have a single error, I wanted to pass that to the caller. This way, the more specific exception (NoDriverFoundError/NoResourceFoundError) class is still available.

Emantor: Ah yes, they are both subclasses. NVM then.

PR #31, opened 2017-03-17
strategy/common: open ConsoleProtocol on creation

Emantor:
If we do not open the ConsoleProtocol implementation class before starting the
first transition we have a race between the device sending out it's first
bytes (which we want to match) and the ConsoleProtocol implementation
connection. Circumvent this by opening the implementation after creating the
strategy.

### DIFF
--- a/labgrid/strategy/bareboxstrategy.py
+++ b/labgrid/strategy/bareboxstrategy.py
@@ -34,6 +34,10 @@ class BareboxStrategy(Strategy):
 
     def __attrs_post_init__(self):
         super().__attrs_post_init__()
+        # Open the ConsoleProtocol to not race the output after reset
+        # FIXME: If we have more than one Driver that implements this protocol
+        # we need to iterate over all of them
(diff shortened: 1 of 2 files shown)


Emantor:
I'm not entierely sure we want to solve this in the ``strategy`` class.

labgrid/strategy/common.py
```diff
@@ -25,6 +26,10 @@ def __attrs_post_init__(self):
             raise BindingError(
                 "Strategies can only be created on a valid target"
             )
+        # Open the ConsoleProtocol to not race the output after reset
+        # FIXME: If we have more than one Driver that implements this protocol
+        # we need to iterate over all of them
+        self.target.activate(self.target.get_driver(ConsoleProtocol))
```
jluebbe: We should instead do this in the early steps in a specific strategy. The bound drivers are also available there.

Emantor:
Amended, rebased and ready for review.

PR #32, opened 2017-03-17
Fix/dev-requirements

Emantor:
Do we need the txaio library as well? It is used in remote/remote.py:22-24


jluebbe:
While we don't need autobahn for local testing, it seems useful to always have it available for now. txaio is pulled in by autobahn.

PR #37, opened 2017-03-21
[RDY] Topic/docs2

Emantor:


README.rst
```diff
@@ -29,12 +29,6 @@ Contributing
 
 Quickstart
 ----------
-Install the required libraries:
-
-.. code-block:: bash
-
-   $ sudo apt install libow-dev
```
jluebbe: We should document this requirement somewhere else.

doc/components.rst
```diff
@@ -5,55 +5,71 @@
...
+   :members:
+   :show-inheritance:
 
 Drivers
 =======
 
 CommandProtocol
-------------
+---------------
+.. autoclass:: labgrid.protocol.CommandProtocol
```
jluebbe: The CommandProtocol is not a driver.

doc/usage.rst
```diff
@@ -1,6 +1,29 @@
 Usage
 =====
 
+Installation
+------------
+
+The default installation is available via Pypi:
```
jluebbe: PyPI instead of Pypi.

doc/usage.rst
```diff
@@ -1,6 +1,29 @@
...
+
+    $ git clone https://github.com/labgrid-project/labgrid
+    $ cd labgrid && python3 setup.py install
+
+Extra requires
+~~~~~~~~~~~~~~
+Labgrid supports different extras:
+
+- onewire: install onewire support, requires onewire from pypi >=0.0.2
+- coordinator: installs required depencies to start a crossbar coordinator
```
jluebbe: Add an example how to enable both extras.

doc/components.rst
Emantor: I started to group the Drivers by Protocol, but since we can just list their base classes I will remove that.

README.rst
Emantor: It's now documented in the usage/installation/extras requires section.

Emantor:
Squashed it some more, fixed the suggestions.

PR #72, opened 2017-03-31
shell/test_rt: improve re pattern

jremmet:
pattern failed if min was a small number aligned with whitespaces


jluebbe:
Thanks for the fix!

Issue #27, opened 2017-03-15
agetty issue breaks first login


tretter:
I'm using the SerialDriver backed by the NetworkSerialPort to connect to a target console. The test tries to login as 'root' without a password.

Most of the time the login is successful, but sometimes it runs into a timeout and the test fails. The debug output contains the following messages, indicating that the test tried to login as 'orot' and failing to supply a password.

```
DEBUG: Write bytes (b'root\n')
DEBUG: Reading 1 (min 1) bytes with 5 timeout
DEBUG: Read bytes (b'o') or timeout reached
DEBUG: Reading 16 (min 1) bytes with 4.89918851852417 timeout
DEBUG: Read bytes (b'rot\r\r\nPassword: ') or timeout reached
DEBUG: Reading 1 (min 1) bytes with 4.79821252822876 timeout
DEBUG: Read bytes (b'') or timeout reached
```

I would not expect that the driver reorders the characters that are sent via the serial console.

Emantor:
We directly hand over those bytes to the RFC2217 implementation of pyserial, which in turn sends those out through its own network socket. As far as I can tell there is no reordering done in their code path as well, maybe it's a ser2net bug?

jluebbe:
Could you check the TCP packets using wireshark? Are they really transferred in the wrong order?

tretter:
I checked the TCP packets for the write and the read.

The data in the packet from my machine is "root.", but the data of the packet to my machine is "orot...".

Emantor:
After further discussion this issue is unrelated to @labgrid-project, closing.

tretter:
The issue is a bug in agetty, which potentially reorders the characters on the login prompt (https://bugzilla.kernel.org/show_bug.cgi?id=194905).

However, labgrid should be able to handle this situation and as a workaround send the username more slowly.

jluebbe:
I've implemented a [workaround](/jluebbe/labgrid/tree/wip/txdelay), which introduces a delay before sending each byte. That seems to solve this problem.
Before merging that workaround, we'll test it internally some more to see how robust it is.
If it works well I'll probably add a way to override the txdelay for specific transmissions (such as sending the user name in the ShellDriver).

jluebbe:
The workaround has been merged (see the 'txdelay' parameter for the ConsoleProtocol driver in the docs: http://labgrid.readthedocs.io/en/latest/configuration.html#drivers).

The original bug was fixed in util-linux and is contained in v2.30-rc1.

PR #77, opened 2017-04-05
driver/shelldriver: Retry login to shell

jremmet:
Some boards have issues on login. They get wrong user names the first
time. For this case also send a empty password to get login into a sane
state. After that just retry and leave a info message.

### DIFF
--- a/labgrid/driver/shelldriver.py
+++ b/labgrid/driver/shelldriver.py
@@ -88,6 +88,16 @@ class ShellDriver(CommandMixin, Driver, CommandProtocol):
         return self._run(cmd, timeout=timeout)
 
     @step()
+    def _send_login(self):
+        self.console.sendline(self.username)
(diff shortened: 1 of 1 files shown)


jremmet:
For me this work around #27.
Maybe we need a configuration switch?

jluebbe:
I'd like to avoid workarounds in the low-level drivers for special cases like this. Instead I tried to solve similar problems on one of our boards with this work-in-progess change: [jluebbe/labgrid wip/txdealy](/jluebbe/labgrid/tree/wip/txdelay).
Does that help for your problem as well?
Also, if you problem is related to logging in, you may be affected by #27.

jremmet:
Yes it looks like #27. 
With 'txdelay: 0.001' your changes works reliable. (tested with ttyUSB0) With the 0.05 from the example I was afraid that the sum of wait from all bytes may to big.

jluebbe:
While it would nice to only use txdelay for the login, that should be done in a separate pull request.

PR #78, opened 2017-04-05
minor improvements and cleanups

jluebbe:


labgrid/driver/serialdriver.py
```diff
@@ -51,13 +51,13 @@ def _read(self, size: int=1, timeout: int=0):
...
-        self.logger.debug("Reading %s (min %s) bytes with %s timeout",
-                          reading, size, timeout)
         self.serial.timeout = timeout
         res = self.serial.read(reading)
-        self.logger.debug("Read bytes (%s) or timeout reached", res)
-        if not res:
-            raise TIMEOUT("Timeout exceeded")
+        if res:
+            self.logger.debug("Read %i bytes: %s, timeout %.2f, min=%i",
+                              len(res), res, timeout, size)
```
ejoerns: Maybe you could interchange the `timeout` and `min` fields here. Otherwise one could assume that `min` refers to a minimum timeout instead of the minimum bytes to read.

jluebbe: I've changed it to "requested size" in the message.

doc/configuration.rst
```diff
@@ -794,36 +794,93 @@ The skeleton for an environment consists of:
...
 can simply be added (at the correct indentation level, one level deeper) to the
 environment configuration.
 
 Exporter Configuration
 ----------------------
-A labgrid exporter is configured by adding resources to groups and exporting
-those groups.
+The exporter is configured using a YAML file with a similar syntax to the
+environment configs used for pytest or instantiating the :any:`Environment`
+object.
```
ejoerns: A bit tricky to read here, I would say at least "configured *by* using a YAML file" and "pytest or *by* instantiating" and/or putting "(with a similar syntax to the environment configs used for pytest)" in brackets

doc/configuration.rst
```diff
@@ -794,36 +794,93 @@ The skeleton for an environment consists of:
...
+
+The simplest case is with one group called "group1" containing a single
+:any:`USBSerialPort`:
+
+.. code-block:: yaml
+
+   group1:
+     USBSerialPort:
+       match:
+	 '@sys_name': '3-1.3'
```
ejoerns: tab?

doc/configuration.rst
```diff
@@ -794,36 +794,93 @@ The skeleton for an environment consists of:
...
+
+To reduce the amount of repeated declarations when many similar resources
+needed to be exported, the `Jinja2 template engine <http://jinja.pocoo.org/>`_
```
ejoerns: s/needed/need/

doc/configuration.rst
```diff
@@ -794,36 +794,93 @@ The skeleton for an environment consists of:
...
+       {model: netio, host: netio4, index: {{ idx - 8 }}}
+       # elif 13 <= idx <= 16
+       {model: netio, host: netio5, index: {{ idx - 12 }}}
+       # endif
+   # endfor
+
+Use ``#`` for line statements (like the for loops in the example) and ``##``
+for line comments.
+Statements like ``{{ 4000 + idx }}`` are expanded based on variables in the
+Jinja2 templage.
```
ejoerns: s/templage/template

ejoerns:
```patch
diff --git a/doc/configuration.rst b/doc/configuration.rst
index 13f5c84..ec2ba81 100644
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -830,8 +830,8 @@ environment configuration.
 
 Exporter Configuration
 ----------------------
-The exporter is configured using a YAML file with a similar syntax to the
-environment configs used for pytest or instantiating the :any:`Environment`
+The exporter is configured by using a YAML file (with a syntax similar to the
+environment configs used for pytest) or by instantiating the :any:`Environment`
 object.
 To configure the exporter, you need to define one or more `resource groups`,
 each containing one or more `resources`.
@@ -857,10 +857,10 @@ The simplest case is with one group called "group1" containing a single
    group1:
      USBSerialPort:
        match:
-        '@sys_name': '3-1.3'
+         '@sys_name': '3-1.3'
 
 To reduce the amount of repeated declarations when many similar resources
-needed to be exported, the `Jinja2 template engine <http://jinja.pocoo.org/>`_
+need to be exported, the `Jinja2 template engine <http://jinja.pocoo.org/>`_
 is used as a preprocessor for the configuration file:
 
 .. code-block:: yaml
@@ -883,4 +883,4 @@ is used as a preprocessor for the configuration file:
 Use ``#`` for line statements (like the for loops in the example) and ``##``
 for line comments.
 Statements like ``{{ 4000 + idx }}`` are expanded based on variables in the
-Jinja2 templage.
+Jinja2 template.
```

PR #81, opened 2017-04-07
tests: add tests for the onewire components

Emantor:

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,4 +1,3 @@
-from unittest.mock import MagicMock
 from importlib.util import find_spec
 
 import attr
@@ -23,8 +22,8 @@ def serial_port(target):
 
 
 @pytest.fixture(scope='function')
-def serial_driver(target, serial_port, monkeypatch):
-    serial_mock = MagicMock
+def serial_driver(target, serial_port, monkeypatch, mocker):
+    serial_mock = mocker.MagicMock
     import serial
     monkeypatch.setattr(serial, 'Serial', serial_mock)
     s = SerialDriver(target)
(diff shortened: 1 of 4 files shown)


tests/conftest.py
```diff
@@ -30,6 +30,17 @@ def serial_driver(target, serial_port, monkeypatch):
...
+def onewire_port(target):
+    return OneWirePIO(target, "testhost", "path")
+
+@pytest.fixture(scope='function')
+def onewire_driver(target, onewire_port, monkeypatch):
+    onewire_mock = MagicMock
+    import onewire
+    monkeypatch.setattr(onewire, 'Onewire', onewire_mock)
+    s = OneWirePIODriver(target)
+    return s
```
jluebbe: Either move these fixtures to test_onewire.py or configure the mock in the test itself. Also use pytest-mock instead of unittest.mock.

Emantor:
Amended, also removed unittests imports from other tests.

tests/conftest.py
```diff
@@ -7,9 +6,9 @@
 from py.path import local
 
 from labgrid import Target, target_factory
-from labgrid.driver import SerialDriver
+from labgrid.driver import SerialDriver, OneWirePIODriver
```
jluebbe: These imports are unused (OneWirePIO below as well).

tests/test_onewire.py
```diff
@@ -0,0 +1,24 @@
...
+    s = OneWirePIODriver(target)
+    return s
+
+def test_onewire_resource_instance(target):
+    o = OneWirePIO(target, "testhost", "path")
+    assert (isinstance(o, OneWirePIO))
+
+def test_onewire_driver_instance(target, onewire_driver):
+    o = target.get_driver(OneWirePIODriver)
+    isinstance(o, OneWirePIODriver)
```
jluebbe: Also add tests for calling set and get using the mock.

Emantor:
And ready again, now with 100% module coverage for the onewiredriver.

PR #83, opened 2017-04-10
add resources for ethernet switch ports

jluebbe:


labgrid/resource/ethernetport.py
```diff
@@ -0,0 +1,195 @@
...
+    def _start(self):
+        import asyncio
+
+        if self.poll_task:
+            return
+
+        @asyncio.coroutine
+        def poll_step(self):
+            self.logger.debug("polling neighbor table")
+            self.neighbors = self._get_neigh()
```
Emantor: `self.neighbors` gets assigned here and is used only once later, maybe remove the instance variable and use it as a local variable in the later code flow.

labgrid/resource/ethernetport.py
```diff
@@ -0,0 +1,195 @@
...
+                extra.setdefault('macs', {})[mac] = self.neighbors.get(mac, [])
+            extra.update(switch.ports.get(resource.interface, {}))
+            if resource.extra != extra:
+                resource.extra = extra
+                self.logger.debug("new information for {}: {}".format(resource, extra))
+
+
+@target_factory.reg_resource
+@attr.s
+class SNMPEthernetPort(ManagedResource):
```
Emantor: Please add an example Docstring.

labgrid/remote/client.py
```diff
@@ -710,6 +710,53 @@ def bootstrap(self):
         target.activate(drv)
         drv.load(self.args.filename)
 
+    def _get_ip(self, place):
+        target = self._get_target(place)
+        from ..resource import EthernetPort
+        try:
+            resource = target.get_resource(EthernetPort)
+        except KeyError:
+            print("resource not found")
```
Emantor: Slight nitpick: Return a more useful error here.

PR #87, opened 2017-04-18
remote: verify place name for create and delete

jluebbe:


labgrid/remote/client.py
```diff
@@ -284,6 +284,8 @@ def print_place(self):
     def add_place(self):
         """Add a place to the coordinator"""
         name = self.args.place
+        if not name:
+            raise UserError("missing place name".format(name))
```
ejoerns: `format(name)` seems to be misplaced, here.

labgrid/remote/client.py
```diff
@@ -295,6 +297,8 @@ def add_place(self):
     def del_place(self):
         """Delete a place from the coordinator"""
         name = self.args.place
+        if not name:
+            raise UserError("missing place name".format(name))
```
ejoerns: dito

PR #94, opened 2017-05-23
ShellDriver: implement file transfer support via XMODEM

rohieb:
This implementation starts the respective sender or receiver on the target and then uses tehmaze's [xmodem][] Python implementation to transfer files over the serial console. It can work with busybox's `rx` (only for receiving on the target) or `lrz`/`lsz` from [lrzsz][] on the target, and autodetects which commands are available.

XMODEM is a very simple, but also very historic protocol with some quirks. See the comment in `labgrid.driver.ShellDriver._put()` for some insight and workarounds.

Unfortunately, after the peer on the target has been started, there is no way to stop it if the transfer fails and get back to a shell prompt, since the peer on the target interprets everything on the console as XMODEM messages. lsz's `--min-bps` option does not seem to work very good (but are included nevertheless), so powercycling the target is possibly the best way to get rid of such situations.

[xmodem]: https://github.com/tehmaze/xmodem
[lrzsz]: https://ohse.de/uwe/software/lrzsz.html

### DIFF
--- a/labgrid/driver/shelldriver.py
+++ b/labgrid/driver/shelldriver.py
@@ -1,7 +1,9 @@
 # pylint: disable=no-member
 """The ShellDriver provides the CommandProtocol, ConsoleProtocol and
  InfoProtocol on top of a SerialPort."""
+import io
 import logging
+import os
 import re
 import shlex
 from time import sleep
@@ -9,18 +11,21 @@ from time import sleep
 import attr
 from pexpect import TIMEOUT
 
+import xmodem
+
 from ..factory import target_factory
-from ..protocol import CommandProtocol, ConsoleProtocol, InfoProtocol
+from ..protocol import (CommandProtocol, ConsoleProtocol, FileTransferProtocol,
+                        InfoProtocol)
 from ..step import step
-from ..util import gen_marker, Timeout
-from .common import Driver
+from ..util import Timeout, gen_marker
 from .commandmixin import CommandMixin
+from .common import Driver
 from .exception import ExecutionError
 
 
 @target_factory.reg_driver
 @attr.s
-class ShellDriver(CommandMixin, Driver, CommandProtocol):
+class ShellDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
     """ShellDriver - Driver to execute commands on the shell
     ShellDriver binds on top of a ConsoleProtocol.
 
@@ -45,6 +50,9 @@ class ShellDriver(CommandMixin, Driver, CommandProtocol):
         )  #pylint: disable=attribute-defined-outside-init,anomalous-backslash-in-string
         self.logger = logging.getLogger("{}:{}".format(self, self.target))
         self._status = 0  #pylint: disable=attribute-defined-outside-init
+
+        self._xmodem_cached_rx_cmd = ""
+        self._xmodem_cached_sx_cmd = ""
 
     def on_activate(self):
         if self._status == 0:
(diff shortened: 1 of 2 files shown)


labgrid/driver/shelldriver.py
```diff
@@ -203,3 +205,127 @@ def _put_ssh_key(self, key):
...
+        if not tmpfile:
+            raise ExecutionError("Could not make temporary file on target")
+        tmpfile = tmpfile[0]
+
+        if self._run("which rx")[0]:
+            cmd = "rx '{}'".format(tmpfile)
+        elif self._run("which lrz")[0]:
+            # redirect stderr to prevent lrz from printing "ready to receive
+            # $file", which will confuse the XMODEM instance
+            cmd = "lrz -X -y -c --o-sync '{}' 2>/dev/null".format(tmpfile)
```
jluebbe: We shouldn't need -o-sync. Do we need -b?

labgrid/driver/shelldriver.py
```diff
@@ -203,3 +205,127 @@ def _put_ssh_key(self, key):
...
+        self._run("rm -f '{}'".format(tmpfile))
+
+    @step(title="get", args=['remotefile', 'localfile'])
+    def _get(self, remotefile, localfile):
+        if self._run("which lsz")[0]:
+            # redirect stderr to prevent lsz from printing "Give XMODEM receive
+            # cmd now", which will confuse the XMODEM instance
+            cmd = "lsz -b -X -m 1200 -M 10 '{}' 2>/dev/null".format(remotefile)
+        else:
+            raise ExecutionError("No XMODEM sender (lsz) available on target")
```
jluebbe: Also do caching as above.

rohieb:
These commits should address your feedback, feel free to sqash them, or ask for more ;)

labgrid/driver/shelldriver.py
```diff
@@ -274,23 +274,31 @@ def _put(self, localfile, remotefile):
...
         except FileNotFoundError:
             raise ExecutionError("Could not stat '{}'".format(localfile))
 
+        def  _target_cleanup(tmpfile):
+            self._run("rm -f '{}'".format(tmpfile))
+
         with open(localfile, "rb") as ifh:
+
+            # We first write to a temp file, which we'll `dd` onto the destination file later
+            tmpfile = self._run("mktemp")[0]
```
jluebbe: Check the return value or use _run_check.

jluebbe:
Please also squash the changes into a single commit.

labgrid/driver/shelldriver.py
rohieb: Right, done.

labgrid/driver/shelldriver.py
```diff
@@ -203,3 +208,147 @@ def _put_ssh_key(self, key):
...
+        self.console.expect(self.prompt, timeout=30)
+
+        # truncate the file to get rid of CPMEOF padding
+        dd_cmd = "dd if='{}' of='{}' bs=1 count={}".format(tmpfile, remotefile, file_size)
+        self.logger.debug("dd command: " + dd_cmd)
+        _, _, ret = self._run(dd_cmd)
+
+        _target_cleanup(tmpfile)
+        if ret != 0:
+            raise ExecutionError("Could not truncate destination file: dd returned %d", ret)
```
rohieb: Oops, that should have been a `"%d" % ret`. Fixed.

rohieb:
Forgot to add a Signed-off-by line.

rohieb:
I've now also included a way to upload and run scripts on the target, as a convenience for the user; and extended the documentation a bit.

rohieb:
OK, the feature set should be complete by now. I've added functionality to upload arbitrary data from byte arrays instead of files, and also rebased to current master.

PR #100, opened 2017-06-12
doc: UBootStrategy has state 'uboot', not 'barebox'

rohieb:
See labgrid/strategy/ubootstrategy.py, line 28


jluebbe:
Please add a signed of by, looks good otherwise.

rohieb:
Thanks, fixed.

PR #105, opened 2017-06-20
bootloader reset

Emantor:
These commits implement reset functionality for bootloader drivers.

### DIFF
--- a/labgrid/driver/bareboxdriver.py
+++ b/labgrid/driver/bareboxdriver.py
@@ -104,6 +104,15 @@ class BareboxDriver(CommandMixin, Driver, CommandProtocol, LinuxBootProtocol):
             raise ExecutionError(cmd)
         return res[0]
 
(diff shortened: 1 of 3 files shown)


labgrid/driver/bareboxdriver.py
```diff
@@ -104,6 +104,14 @@ def run_check(self, cmd: str):
             raise ExecutionError(cmd)
         return res[0]
 
+    @Driver.check_active
```
jluebbe: add @step() decorator

labgrid/driver/ubootdriver.py
```diff
@@ -142,6 +142,14 @@ def _check_prompt(self):
             self._status = 0
             raise
 
+    @Driver.check_active
+    def reset(self):
```
jluebbe: add @step() decorator

PR #110, opened 2017-07-11
Add imports key for remote client

Emantor:
TODO:
- [x] add Tests

### DIFF
--- a/labgrid/config.py
+++ b/labgrid/config.py
@@ -128,3 +128,16 @@ class Config:
 
     def get_targets(self):
         return self.data.get('targets', {})
+
+    def get_imports(self):
+        """Helper function that returns the list of all imports
+
+        Returns:
+            list: List of files which should be imported
+        """
+        imports = []
+
+        for user_import in self.data.get('imports', []):
+            imports.append(self.resolve_path(user_import))
+
+        return imports
--- a/labgrid/environment.py
+++ b/labgrid/environment.py
@@ -1,4 +1,5 @@
 import attr
+import os
 import yaml
 
 from .exceptions import NoConfigFoundError
@@ -23,6 +24,14 @@ class Environment:
             raise NoConfigFoundError(
                 "{} is not a valid yaml file".format(self.config_file)
             )
+        for user_import in self.config.get_imports():
+            from importlib.machinery import SourceFileLoader
+            import sys
+
+            module_name = os.path.basename(user_import)[:-3]
+
+            module = SourceFileLoader(module_name, user_import).load_module()
+            sys.modules[module_name] = module
 
     def get_target(self, role: str='main') -> Target:
         """Returns the specified target.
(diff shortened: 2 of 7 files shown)


labgrid/environment.py
```diff
@@ -23,6 +24,9 @@ def __attrs_post_init__(self):
             raise NoConfigFoundError(
                 "{} is not a valid yaml file".format(self.config_file)
             )
+        for user_file in self.config.get_imports():
+            importlib.import_module(user_file)
+            importlib.invalidate_caches()
```
jluebbe: We don't need invalidate_caches, as the file should already exist when starting labgrid.

labgrid/environment.py
```diff
@@ -23,6 +24,9 @@ def __attrs_post_init__(self):
             raise NoConfigFoundError(
                 "{} is not a valid yaml file".format(self.config_file)
             )
+        for user_file in self.config.get_imports():
+            importlib.import_module(user_file)
```
jluebbe: Don't we need to resolve the path relative to the environment file, before passing it to importlib? We should use the same behavior as images or normal paths in the config.

labgrid/config.py
```diff
@@ -128,3 +128,11 @@ def set_option(self, name, value):
...
     def get_targets(self):
         return self.data.get('targets', {})
+
+    def get_imports(self):
+        """Helper function that returns the list of all imports
+
+        Returns:
+            list: List of files which should be imported
+        """
+        return self.data.get('imports', [])
```
jluebbe: Resolve the path here to be consistent with the other config APIs.

labgrid/environment.py
```diff
@@ -23,6 +24,15 @@ def __attrs_post_init__(self):
             raise NoConfigFoundError(
                 "{} is not a valid yaml file".format(self.config_file)
             )
+        for user_file in self.config.get_imports():
+            from importlib.machinery import SourceFileLoader
+            import sys
+
+            file_path = self.config.resolve_path(user_file)
```
jluebbe: This is no longer needed when the Config resolves the paths.

man/labgrid-device-config.1
```diff
@@ -87,6 +87,10 @@ Path to the mxs\-usb\-loader binary
 .B \fBimx\-usb\-loader\fP
 Path to the imx\-usb\-loader binary
 .UNINDENT
+.SH IMPORTS
+.sp
+The \fBimports\fP key is a list of files which are imported by the environment
+after loading the configuration.
```
jluebbe: Add a note that paths relative to the config file are supported.

tests/test_environment.py
```diff
@@ -50,3 +50,26 @@ def test_instance_invalid_yaml(self, tmpdir):
...
+        targets:
+          main:
+            drivers: {}
+        imports:
+          - {}
+        """.format("{}",str(i))
+        )
+        e = Environment(str(p))
+        assert (isinstance(e, Environment))
+        assert "myimport" in sys.modules
```
jluebbe: Also check that the Test class exists.

Emantor:
Adressed your comments, should be ready to merge.

PR #114, opened 2017-07-29
fixtures: check if getattr returns the _xml attribute

Emantor:
In pytest plugins are marked available as long as they are registered,
registration however does not indicate usage. Check if junit xml output is
enabled and only than report additional attributes.


Emantor:
My apologies, i was sure that hasplugin() would only supply used plugins, not registered ones. Better test next time :(

PR #118, opened 2017-08-08
Add remote client state support

Emantor:
This adds support for strategy states to the labgrid client.

TODO:
- [x] Rework on_deactivate to actually work recursively

### DIFF
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -500,6 +500,20 @@ class ClientSession(ApplicationSession):
         manager.loop = self.loop
 
     def _get_target(self, place):
+        target = self.env.get_target(place.name)
+        if target:
+            if self.args.state:
+                if self.args.verbose >= 2:
+                    from labgrid.stepreporter import StepReporter
+                    StepReporter()
+                from labgrid.strategy import Strategy
+                from labgrid.driver import SerialDriver
+                strategy = target.get_driver(Strategy)
+                print("Transitioning into state {}".format(self.args.state))
+                strategy.transition(self.args.state)
+                serial = target.get_active_driver(SerialDriver)
+                target.deactivate(serial)
+            return target
         self._prepare_manager()
         target = Target(place.name, env=self.env)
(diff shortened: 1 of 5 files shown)


Emantor:
CI requires #119 to pass the python 3.6 Tests.

labgrid/remote/client.py
```diff
@@ -820,13 +820,15 @@ def main():
     subparser = subparsers.add_parser('io',
                                       help="change (or get) a digital IO status")
     subparser.add_argument('action', choices=['high', 'low', 'get'])
+    subparser.add_argument('-v', '--verbose', action='count', default=0)
```
Emantor: Sure, I was not sure whether we prefer it for the individual subparsers here, since two already implement `--verbose` with different actions.

jluebbe: Maybe it's useful to always enable the step reporting when using state support? Do we need to switch these two verbose mechanisms separately?

Emantor: Without the StepReporter there is no feedback for the user, on the other hand the StepReporter can be quiet verbose. I think it is useful to debug why a state transition is not working, but if your strategy is working as expected, you hardly need the extra information from the StepReporter

labgrid/target.py
```diff
@@ -244,9 +244,9 @@ def activate(self, client):
 
         # activate recursively and resolve conflicts
         for supplier in client.suppliers:
+            supplier.resolve_conflicts(client)
```
Emantor: I dropped this commit. I had it backwards and thought we wanted to deactivate the _suppliers_ not the clients.

labgrid/target.py
```diff
@@ -273,6 +273,9 @@ def deactivate(self, client):
         client.on_deactivate()
         client.state = BindingState.bound
 
+        for supplier in client.suppliers:
+            self.deactivate(supplier)
+
```
Emantor: Yes, implemented that as suggested.

PR #124, opened 2017-08-29
multi-instance named resources and drivers

jluebbe:
This adds names to Resources and Drivers, which allows us to have multiple instances of the same class in one Target.

PR #123 needs to be merged first.

PR #125, opened 2017-08-29
implement autoinstaller tool

jluebbe:
This implements a new stand-alone tool for labgrind, which uses an environment config file with embedded python code to automatically install software to multiple targets.

PR #124 needs to be merge before this.

### DIFF
--- a/labgrid/autoinstall/main.py
+++ b/labgrid/autoinstall/main.py
@@ -1,0 +1,221 @@
+"""The autoinstall.main module runs an installation script automatically on multiple targets."""
+import ast
+import argparse
+import logging
+import sys
+import multiprocessing
+import textwrap
+from time import sleep
+
+from .. import Environment, target_factory
+from ..stepreporter import StepReporter
+from ..exceptions import NoResourceFoundError
+
+
+class Handler(multiprocessing.Process):
+    def __init__(self, env, args, name):
+        super().__init__(name=name)
+        self.env = env
+        self.args = args
+        self.config = env.config
+        self.name = name
+
+        self.context = {
+            'name': self.name,
+            'env': self.env,
+            'config': self.config,
+        }
+        self.context.update(target_factory.resources)
+        self.context.update(target_factory.drivers)
+
+    def _get_function(self, name, context):
+        snippet = self.config.data['autoinstall'].get(name)
+        if not snippet:
+            return None
+
+        code = 'def {}():\n{}'.format(name, textwrap.indent(snippet, ' '))
+        tree = ast.parse(code, filename=self.env.config_file)
+        ast.increment_lineno(tree, snippet.start_mark.line)
+        co = compile(tree, filename=self.env.config_file, mode='exec')
+
+        stage = {}
+        exec(co, context, stage)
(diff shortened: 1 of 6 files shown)


labgrid/autoinstall/main.py
```diff
@@ -0,0 +1,203 @@
...
+
+    def run(self):
+        self.log = logging.getLogger(self.name)
+        self.log.info("creating new handler (PID {})".format(self.pid))
+
+        try:
+            self.target = self.env.get_target(self.name)
+            self.context['target'] = self.target
+        except:
+            self.log.exception("handler creation failed")
```
Emantor: Should be "target creation failed"

labgrid/autoinstall/main.py
```diff
@@ -0,0 +1,203 @@
+"""The autoinstall.main module runs an installation script automatically on multiple targets."""
+import ast
+import argparse
+import logging
+import sys
+import traceback
```
Emantor: Unnecessary import, not used in this file, same for:
- `attr`
- `from contextlib import closing`

labgrid/autoinstall/main.py
```diff
@@ -0,0 +1,203 @@
...
+    def configure(self):
+        if not 'autoinstall' in self.env.config.data:
+            self.log.error("no 'autoinstall' section found in '{}'".format(self.env.config_file))
+            return False
+
+        if not 'handler' in self.env.config.data['autoinstall']:
+            self.log.error("no 'handler' definition found in '{}'".format(self.env.config_file))
+            return False
+
+        self.hostname = gethostname()
```
Emantor: hostname variable is not used anywhere

labgrid/autoinstall/main.py
```diff
@@ -0,0 +1,203 @@
+"""The autoinstall.main module runs an installation script automatically on multiple targets."""
+import ast
+import argparse
+import logging
+import sys
+import traceback
+import multiprocessing
+import textwrap
+from contextlib import closing
+from socket import gethostname
```
Emantor: If we don't use `self.hostname` in the manager class, we can delete this import

labgrid/autoinstall/main.py
```diff
@@ -0,0 +1,203 @@
...
+    def _get_initial_resource(self):
+        cls = self.config.data['autoinstall'].get('initial-resource')
+        if not cls:
+            return None
+
+        return self.target.get_resource(self.context[cls], await=False)
+
+    def run(self):
+        self.log = logging.getLogger(self.name)
+        self.log.info("creating new handler (PID {})".format(self.pid))
```
Emantor: [Logging Optimizations](https://docs.python.org/3/howto/logging.html#optimization) mentions that values should be passed via lazy interpolation, to avoid the formatting cost if the message is not emitted, e.g. 
```
self.log.info("creating new handler (PID %s)", self.pid)
```
instead of
```
self.log.info("creating new handler (PID {})".format(self.pid))
```

jluebbe:
I've addressed your comments and added some basic tests.

PR #128, opened 2017-09-11
add some sanity checks / give more output

Farom:


Emantor:
You are missing the SOBs, otherwise looks good to me.

PR #130, opened 2017-09-14
add NetworkPowerDriver-model "netio_kshell"

jonau:
added model to interface with all netio_4x-models over telnet(kshell)

Signed-off-by: Johannes Nau <johannes.nau@jumo.net>


jluebbe:
Thanks for your contribution!

PR #134, opened 2017-09-21
Automatic print command output to local stdout / stderr

jonau:
To gather information about a test-failure it is sometime necessary to print the output of a command executed at the target. To reduce boilerplate in the testcases i propose to add a option to the command protocol to automatically print the output to the stdout/stderr.

### DIFF
--- a/labgrid/driver/bareboxdriver.py
+++ b/labgrid/driver/bareboxdriver.py
@@ -1,6 +1,7 @@
 # pylint: disable=no-member
 import logging
 import re
+import sys
 import shlex
 from time import sleep
 
@@ -57,7 +58,7 @@ class BareboxDriver(CommandMixin, Driver, CommandProtocol, LinuxBootProtocol):
 
     @Driver.check_active
     @step(args=['cmd'])
-    def run(self, cmd: str, *, step):
+    def run(self, cmd: str, *, step, print=False):
         """
         Runs the specified command on the shell and returns the output.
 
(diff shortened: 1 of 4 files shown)


jonau:
Maybe it is useful to also print the command together with the command promt out, so we can see the which command was used in the test-results. This would help the developers to reproduce the test failure more easily.

Emantor:
Calling pytest with `-vv` should already print the information for all drivers except the `SSHDriver`.
The step reporter should clearly show which cmd is run and what the output of the command is.
The SSHDriver is missing a `result=True` in the step-decorator for the run command, that should be fixed.

jonau:
I am aware of the step-reporting abilities of labgrid, which is great for debugging. I thought of the following application for the PR:
If you write a test to check for boot-errors you would do something like this:
```python
stdout, stderr, returncode = \
    ssh.run("dmesg | grep -i 'error\|warn\|fail\|deprecated'", print=True)
assert returncode == 0
assert len(stderr) == 0
assert len(stdout) == 0
```
However for addressing the test failure the developer want to know what is the issue here, so you print out the result of the grep:
```python
# only if you want to show the command
# print("dmesg | grep -i 'error\|warn\|fail\|deprecated'")
stdout, stderr, returncode = \
    ssh.run("dmesg | grep -i 'error\|warn\|fail\|deprecated'")
print("\n".join(stdout))
assert returncode == 0
assert len(stderr) == 0
assert len(stdout) == 0, "See stdout for dmesg-messages"
```
with my patch this would be become:
```python
stdout, stderr, returncode = \
    ssh.run("dmesg | grep -i 'error\|warn\|fail\|deprecated'", print=True)
assert returncode == 0
assert len(stderr) == 0
assert len(stdout) == 0, "See stdout for dmesg-messages"
```

In this situation you can not rely on the step reporter as the output of the step reporter is not as easy to read as a direct print out and cluttered with information not relevant to the problem. Also the output goes only to terminal - which is not available in an automatic test environment - and is not shown in the test-report.

As it is easy enough to make this printouts by yourself this PR is not critically. However I thought it would be a reasonable addition.

PR #135, opened 2017-09-21
add ssh commandline to supress warning

jonau:
On every ssh-connection we get a "Warning: Permanently added '...' (RSA) to the list of known hosts."
This will add a option to the ssh client to suppress the output of warning messages.


Emantor:
Thanks!

PR #136, opened 2017-09-21
add transitions to reboot device into bootloader or shell

jonau:
In our testsuite we have tests, which require to reboot the device. With the current barebox-strategy this is not possible.
This will add two more transition targets, which will force a reboot to the bootloader / shell.


Emantor:
Strategies should be defined in your own test files because they have to be customized for your individual use case. The provided strategies only serve as a starting point.

PR #143, opened 2017-10-12
Documentation Updates

SmithChart:
These patches provide multiple enhancements for the documentation.

Those are mostly about the "getting started" part and help users on a Debian current distributions to copy-and-paste their way through this section.


Emantor:
Thanks!

PR #147, opened 2017-10-29
ubootdriver: make password prompt configurable

Emantor:
Signed-off-by: Rouven Czerwinski <rouven@czerwinskis.de>


Emantor:
I made the boot-expression configurable as well and added both to the documentation, should be ready now.

PR #148, opened 2017-11-01
ubootdriver: wait for a short time period right after stopping autoboot

mbodmer:
Signed-off-by: Marc Bodmer <marc.bodmer@securiton.ch>

### DIFF
--- a/labgrid/driver/ubootdriver.py
+++ b/labgrid/driver/ubootdriver.py
@@ -2,6 +2,7 @@
 """The U-Boot Module contains the UBootDriver"""
 import logging
 import re
+import time
 
 import attr
 from pexpect import TIMEOUT
@@ -171,6 +172,7 @@ class UBootDriver(CommandMixin, Driver, CommandProtocol, LinuxBootProtocol):
                 raise Exception("Password entry needed but no password set")
         else:
             self.console.write(self.interrupt.encode('ASCII'))
+            time.sleep(1)
             self._check_prompt()
         for command in self.init_commands:  #pylint: disable=not-an-iterable
             self._run_check(command)


mbodmer:
Hi, I am not sure my fix is the correct way to handle these kind of problems with a serial connection to the target. I have had problems to enter uboot on my target (examples/test_ubloot_strategy.py). The marker strings were not completely transferred and this I have always landed in a timeout.

  DEBUG: Read 11 bytes: b'autoboot:  ', timeout 28.98, requested size 1
  DEBUG: Write 1 bytes: b'\n'
  DEBUG: Write 20 bytes: b"echo 'MFOK''LHTPOQ'\n"
  DEBUG: Read 8 bytes: b'2 \x08\x08\x08 1 ', timeout 2.00, requested size 1
  DEBUG: Read 1 bytes: b'\x08', timeout 2.00, requested size 1
  DEBUG: Read 5 bytes: b'\x08\x08 0 ', timeout 1.99, requested size 1
  DEBUG: Read 1 bytes: b'\r', timeout 1.99, requested size 1
  DEBUG: Read 1 bytes: b'\n', timeout 1.99, requested size 1
  DEBUG: Read 1 bytes: b'=', timeout 1.99, requested size 1
  DEBUG: Read 5 bytes: b'> ech', timeout 1.98, requested size 1
  DEBUG: Read 1 bytes: b'o', timeout 1.98, requested size 1
  DEBUG: Read 10 bytes: b" 'MFOK''LH", timeout 1.98, requested size 1
  DEBUG: Read 1 bytes: b'T', timeout 1.98, requested size 1

to wait for a short period (I have added a 1s sleep...) after stopping autoboot, this behaviour disappears.

regards, Marc

Emantor:
Hey, I stumbled upon the exact same problem (and fix) while working with a TP-Link U-Boot router. Apparently U-Boot does not buffer the input after the password prompt, which barebox does. The fix looks good to me, I'll discuss with @jluebbe whether we need a more sophisticated workaround.

jluebbe:
This problem could also be the result of buffer overflows on the target side. As the only the end of the marker string is lost, I don't think this is related to the autoboot interruption.
Could you try if using txdelay helps? (see http://labgrid.readthedocs.io/en/latest/configuration.html#serialdriver)

mbodmer:
Yes, SerialDriver: txdelay : 0.05 seems to fix this problem. Thanks for this hint.

PR #152, opened 2017-11-06
added a simple power driver for YEPKIT YKUSH switchable USB hubs - th…

tgbloks:
…is currently "highjacks" the NetworkPowerPort resource by using the host entry as serial for the USB hub. 
works for me as it is but i might be willing to tweak some stuff if you are interested in getting it into your repo ;-)

### DIFF
--- a/labgrid/resource/ykushpowerport.py
+++ b/labgrid/resource/ykushpowerport.py
@@ -1,0 +1,17 @@
+import attr
+
+from ..factory import target_factory
+from .common import Resource
+
+
+@target_factory.reg_resource
+@attr.s(cmp=False)
+class YKUSHPowerPort(Resource):
+    """This resource describes a YEPKIT YKUSH switchable USB hub.
+
+    Args:
+        serial (str): serial of the YKUSH device 
+        index (int): port index"""
+    serial = attr.ib(validator=attr.validators.instance_of(str))
+    index = attr.ib(validator=attr.validators.instance_of(int),
+                    convert=lambda x: int(x))
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -108,6 +108,26 @@ The example describes port 0 on the remote power switch
 Used by:
   - `NetworkPowerDriver`_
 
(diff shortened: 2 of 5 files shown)


jluebbe:
Thanks for the PR! I have some comments, though...

This should have it's own resource and driver instead of reusing NetworkPowerPort. For an example of a standalone resource look at resource/onewireport.py. You'd use serial and index as parameters.
For the driver, copy the NetworkPowerDriver and call directly to pykush. Do the import in the ``__attrs_post_init__`` method to avoid the module load overhead if it's not used. Also please map to the ``YKUSH_PORT_STATE_*`` constants defined by the pykush library.

I'd suggest YKUSHPowerPort for the resource and YKUSHPowerDriver for the driver.

tgbloks:
no worries - i totally agree with your proposal (as it sounds just like the "nice" long term solution in my head ;-)
what you see in this PR is just the minimal effort i needed to invest to test labgrid with my available HW
(i very much like what i see so far - thx!) and since i could not find a better place to discuss this (any plans for a mailing list / irc channel?) i started here.
i will update the PR once i had time to implement and test the changes...

labgrid/resource/ykushpowerport.py
```diff
@@ -0,0 +1,17 @@
...
+@attr.s(cmp=False)
+class YKUSHPowerPort(Resource):
+    """This resource describes a YEPKIT YKUSH switchable USB hub.
+
+    Args:
+        serial (str): serial of the YKUSH device 
+        index (str): port index"""
+    serial = attr.ib(validator=attr.validators.instance_of(str))
+    index = attr.ib(validator=attr.validators.instance_of(str),
+                    convert=lambda x: str(int(x)))
```
jluebbe: If you use ``convert=lambda x: int(x)``, you should be able to remove the int() conversions in the on/off functions.

tgbloks:
should be ok now...

jluebbe:
Thanks for your contribution!

PR #154, opened 2017-11-15
Mount .ssh directory instead of authorized_keys file

### LINKED ISSUE #93: ShellDriver.put_ssh_key not working on read-only filesystem
Doing `ShellDriver.put_ssh_key()` on a read-only file system only results in a stacktrace for me:

```
   INFO: Key not on target, mounting...
  File "test-xmodem.py", line 14, in <module>
    sd.put_ssh_key("../projectroot/root/.ssh/development_rsa.pub")
  File "/ptx/work/dude/WORK_B/rhi/realtimeci-DistroKit/labgrid-tests/env/lib/python3.5/site-packages/labgrid-0.1.1.dev1+g0897126-py3.5.egg/labgrid/binding.py", line 87, in wrapper
  File "/ptx/work/dude/WORK_B/rhi/realtimeci-DistroKit/labgrid-tests/env/lib/python3.5/site-packages/labgrid-0.1.1.dev1+g0897126-py3.5.egg/labgrid/driver/shelldriver.py", line 206, in put_ssh_key
    self._put_ssh_key(key)
  File "/ptx/work/dude/WORK_B/rhi/realtimeci-DistroKit/labgrid-tests/env/lib/python3.5/site-packages/labgrid-0.1.1.dev1+g0897126-py3.5.egg/labgrid/step.py", line 192, in wrapper
    _result = func(*_args, **_kwargs)
  File "/ptx/work/dude/WORK_B/rhi/realtimeci-DistroKit/labgrid-tests/env/lib/python3.5/site-packages/labgrid-0.1.1.dev1+g0897126-py3.5.egg/labgrid/driver/shelldriver.py", line 200, in _put_ssh_key
    self._run_check('mount --bind /tmp/keys ~/.ssh/authorized_keys')
  File "/ptx/work/dude/WORK_B/rhi/realtimeci-DistroKit/labgrid-tests/env/lib/python3.5/site-packages/labgrid-0.1.1.dev1+g0897126-py3.5.egg/labgrid/step.py", line 192, in wrapper
    _result = func(*_args, **_kwargs)
  File "/ptx/work/dude/WORK_B/rhi/realtimeci-DistroKit/labgrid-tests/env/lib/python3.5/site-packages/labgrid-0.1.1.dev1+g0897126-py3.5.egg/labgrid/driver/shelldriver.py", line 114, in _run_check
    raise ExecutionError(cmd)
labgrid.driver.exception.ExecutionError: mount --bind /tmp/keys ~/.ssh/authorized_keys
```

… which, in hindsight, is obvious, because on the target console:
```
root@DistroKit:~ mount --bind /tmp/keys ~/.ssh/authorized_keys
mount: mount point /root/.ssh/authorized_keys does not exist
root@DistroKit:~ mkdir ~/.ssh
mkdir: can't create directory '/root/.ssh': Read-only file system
root@DistroKit:~ mount -o rw,remount /
root@DistroKit:~ mkdir ~/.ssh
root@DistroKit:~ ls -la ~/.ssh
drwxr-xr-x    2 root     root             6 May 22  2017 .
drwxr-xr-x    3 1059     3000            26 May 22  2017 ..
```

On most freshly installed targets, `/root/.ssh` will most probably not exist, so on a read-only system, `mkdir ~/.ssh` in [ShellDriver.py, line 179](https://github.com/labgrid-project/labgrid/blob/master/labgrid/driver/shelldriver.py#L179) will fail. I'm guessing the read-write mount is only done a few lines later with `mount --bind`, and if this works like I imagine, I think this issue could be solved by doing a `mount --bind` on the whole `/root/` directory instead of only the `authorized_keys` file.

Emantor:
Previously we did a bind mount on just the authorized key file, this breaks if
we have a read only file system. Create a directory in /tmp and bind mount the
whole ~/.ssh directory instead.

Fixes #93

### DIFF
--- a/labgrid/driver/shelldriver.py
+++ b/labgrid/driver/shelldriver.py
@@ -175,25 +175,23 @@ class ShellDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
             \s+(?P<key>[a-zA-Z0-9/+=]+) # Match Keystring
             \s+(?P<comment>.*) # Match comment""", re.X
         )
-        if self._status == 1:
-            with open(key) as keyfile:
-                keyline = keyfile.readline()
-                self.logger.debug("Read Keyline: %s", keyline)
-                match = regex.match(keyline)
-                if match:
-                    new_key = match.groupdict()
-                else:
-                    raise IOError(
-                        "Could not parse SSH-Key from file: {}".
-                        format(keyfile)
-                    )
-            self.logger.debug("Read Key: %s", new_key)
-            auth_keys, _, exitcode = self._run("cat ~/.ssh/authorized_keys")
-            self.logger.debug("Exitcode: %s", exitcode)
-            if exitcode != 0:
-                self._run("mkdir ~/.ssh")
-                self._run("touch ~/.ssh/authorized_keys")
-            result = []
+        with open(key) as keyfile:
+            keyline = keyfile.readline()
+            self.logger.debug("Read Keyline: %s", keyline)
+            match = regex.match(keyline)
+            if match:
+                new_key = match.groupdict()
+            else:
+                raise IOError(
+                    "Could not parse SSH-Key from file: {}".
+                    format(keyfile)
+                )
+        self.logger.debug("Read Key: %s", new_key)
+        auth_keys, _, read_keys = self._run("cat ~/.ssh/authorized_keys")
+        self.logger.debug("Exitcode trying to read keys: %s, keys: %s", read_keys, auth_keys)
+        result = []
+        _, _, test_write = self._run("touch ~/.test")
+        if read_keys == 0:
             for line in auth_keys:
                 match = regex.match(line)
                 if match:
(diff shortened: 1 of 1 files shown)


labgrid/driver/shelldriver.py
```diff
@@ -208,12 +205,11 @@ def _put_ssh_key(self, key):
...
                     return
-            self.logger.info("Key not on target, mounting...")
-            self._run_check('echo "{}" > /tmp/keys'.format(keyline))
-            self._run_check('chmod 600 /tmp/keys')
-            self._run_check('mount --bind /tmp/keys ~/.ssh/authorized_keys')
-            self._run_check('chmod 700 ~/.ssh')
-            self._run_check('chmod 644 ~/.ssh/authorized_keys')
+            self.logger.debug("Key not on target, mounting...")
+            self._run_check('mkdir -m 700 /tmp/labgrid-ssh/')
+            self._run_check('echo "{}" > /tmp/labgrid-ssh/authorized_keys'.format(keyline))
```
jluebbe: Shouldn't we append the key instead of replacing the other keys?

labgrid/driver/shelldriver.py
```diff
@@ -190,9 +190,6 @@ def _put_ssh_key(self, key):
             self.logger.debug("Read Key: %s", new_key)
             auth_keys, _, exitcode = self._run("cat ~/.ssh/authorized_keys")
             self.logger.debug("Exitcode: %s", exitcode)
-            if exitcode != 0:
-                self._run("mkdir ~/.ssh")
```
jluebbe: We should still try to run mkdir -p ~/.ssh, so it works more often or RW root filesystems.
If it fails, return a message which explains that this directory needs to exist.

Emantor:
I pushed an update, the functions checks for writeability and existing keys, uses normal writes and creates the dir if necessary. In case we have no writeable target we preserve existing keys as well.

labgrid/driver/shelldriver.py
```diff
@@ -208,12 +206,27 @@ def _put_ssh_key(self, key):
...
+            self.logger.debug("Key not on target, no ssh directory and writeable, concatenating...")
+            self._run("mkdir ~/.ssh/")
+            self._run_check("chmod 600 ~/.ssh/")
+            self._run_check('echo "{}" >> ~/.ssh/authorized_keys'.format(keyline))
+            self._run_check("rm ~/.test")
+            return
+
+        self.logger.debug("Key not on target and not writeable, using bind mount...")
+        self._run_check('mkdir -m 700 /tmp/labgrid-ssh/')
+        self._run("cat ~/.ssh/authorized_keys > /tmp/labgrid-ssh/authorized_keys")
```
jluebbe: We should use 'cp -a' to keep other files in this directory accessible.

jluebbe: We should use 'cp -a' to keep other files in this directory accessible.

labgrid/driver/shelldriver.py
```diff
@@ -208,12 +206,27 @@ def _put_ssh_key(self, key):
...
-            self._run_check('chmod 644 ~/.ssh/authorized_keys')
+
+        if test_write == 0 and read_keys == 0:
+            self.logger.debug("Key not on target and writeable, concatenating...")
+            self._run_check('echo "{}" >> ~/.ssh/authorized_keys'.format(keyline))
+            self._run_check("rm ~/.test")
+            return
+
+        if test_write == 0:
+            self.logger.debug("Key not on target, no ssh directory and writeable, concatenating...")
```
jluebbe: read_keys==0 only means that authorized_keys does not exist, the .ssh dir may exist anyway. Also, why are you concatenating in this case?

labgrid/driver/shelldriver.py
```diff
@@ -208,12 +206,27 @@ def _put_ssh_key(self, key):
...
+            self._run_check('echo "{}" >> ~/.ssh/authorized_keys'.format(keyline))
+            self._run_check("rm ~/.test")
+            return
+
+        self.logger.debug("Key not on target and not writeable, using bind mount...")
+        self._run_check('mkdir -m 700 /tmp/labgrid-ssh/')
+        self._run("cat ~/.ssh/authorized_keys > /tmp/labgrid-ssh/authorized_keys")
+        self._run_check('echo "{}" >> /tmp/labgrid-ssh/authorized_keys'.format(keyline))
+        self._run_check('chmod 600 /tmp/labgrid-ssh/authorized_keys')
+        self._run_check('mount --bind /tmp/labgrid-ssh/ ~/.ssh/')
```
jluebbe: Also give a useful error message if this fails (likely because ~/.ssh/ doesn't exist.

labgrid/driver/shelldriver.py
Emantor: Done.

labgrid/driver/shelldriver.py
Emantor: Concatenating is wrong here, copy paste error from above. I added a check for the directory.

labgrid/driver/shelldriver.py
Emantor: I'm sending a warning now

Emantor:
Addressed your comments, lets try again 👍

PR #167, opened 2017-12-08
support for ssh password

tgbloks:
* the networkservice resource has a new password property - if it is set
  the sshdriver will login with sshpass (needs to be installed) instead
  of the key file

### DIFF
--- a/labgrid/driver/sshdriver.py
+++ b/labgrid/driver/sshdriver.py
@@ -30,9 +30,11 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
         self.logger = logging.getLogger("{}({})".format(self, self.target))
 
     def on_activate(self):
-        self.ssh_prefix = "-i {}".format(os.path.abspath(self.keyfile)
+        self.ssh_prefix = "-o LogLevel=ERROR"
+        self.ssh_prefix += " -i {}".format(os.path.abspath(self.keyfile)
                                          ) if self.keyfile else ""
-        self.ssh_prefix += " -o LogLevel=ERROR"
+        self.ssh_prefix += " -o PasswordAuthentication=no" if (
+                not self.networkservice.password) else ""
         self.control = self._check_master()
         self.ssh_prefix += " -F /dev/null"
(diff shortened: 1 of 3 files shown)


labgrid/driver/sshdriver.py
```diff
@@ -104,7 +114,7 @@ def run(self, cmd):
         returns:
         (stdout, stderr, returncode)
         """
-        complete_cmd = "ssh -x {prefix} {user}@{host} {cmd}".format(
+        complete_cmd = "ssh -x{prefix} {user}@{host} {cmd}".format(
```
Emantor: As above I'm wondering about the change here.

labgrid/driver/sshdriver.py
```diff
@@ -30,8 +30,10 @@ def __attrs_post_init__(self):
         self.logger = logging.getLogger("{}({})".format(self, self.target))
 
     def on_activate(self):
-        self.ssh_prefix = "-i {}".format(os.path.abspath(self.keyfile)
+        self.ssh_prefix = " -i {}".format(os.path.abspath(self.keyfile)
```
Emantor: Is this change with the change below necessary? Looks like an unnecessary change to me.

labgrid/driver/sshdriver.py
tgbloks: i found these changes necessary to prevent 2 spaces following each other in the cmd line.
(2 spaces cause split(' ') to add an empty element in the list which will be treated by ssh as the hostname...)

tgbloks:
was waiting with the documentation till we agreed on the rest ;-)

labgrid/driver/sshdriver.py
Emantor: Ah, I see. In that case I would like you too shuffle the `self.ssh_prefix += " -o LogLevel=ERROR"` line to the top and handle the extra missing space there. Otherwise we would have to touch the get and put functions as well. I tested your changes against a local device, works fine :)

tgbloks: damn - i missed put and get.
should work now with one change less ;-)

PR #168, opened 2017-12-10
add fastboot continue command

tgbloks:
added the "continue" command for fastboot.
This particulary useful to run a "special fastboot" uboot version
which always starts in fastboot mode, enableling flashing without
any external trigger. After exiting the fastboot mode with the
"continue" command uboot will continue to boot normally.


tgbloks:
better now?

PR #170, opened 2017-12-13
resource/udev: remove devices only on unbind

Emantor:
Before this change every action which is not in one of the previous cases would
remove the device and set availability to false. This includes the 'bind'
action. Remove the device only if the device unbinds.


tgbloks:
this solves problems where my fastboot device was not recognized every time

Emantor:
Cursory google search does not say anything about such a guarantee, pushin a version with remove in a moment...

PR #171, opened 2017-12-13
configurable USB ids for fastboot

tgbloks:
replaced the static USB ids in the fastboot resource with configurable properties

This is necessary for us since we use different USB IDs.
Once we agreed on this i will update the documentation.
Might be interesting for other USB resources as well...

### DIFF
--- a/labgrid/resource/udev.py
+++ b/labgrid/resource/udev.py
@@ -40,7 +40,7 @@ class UdevManager(ResourceManager):
 class USBResource(ManagedResource):
     manager_cls = UdevManager
 
-    match = attr.ib(validator=attr.validators.instance_of(dict), hash=False)
+    match = attr.ib(default={}, validator=attr.validators.instance_of(dict), hash=False)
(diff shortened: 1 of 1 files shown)


labgrid/resource/udev.py
```diff
@@ -40,7 +40,7 @@ def poll(self):
 class USBResource(ManagedResource):
     manager_cls = UdevManager
 
-    match = attr.ib(validator=attr.validators.instance_of(dict), hash=False)
+    match = attr.ib(default={}, validator=attr.validators.instance_of(dict), hash=False)
```
jluebbe: I'm not sure about this, you could also have an explicit empty match argument if you want to match everything. It seems a little bit dangerous to make match everything implicit.

tgbloks: just wanted to put it up for discussion. Im ok with putting the empty match in my yaml file but it looked to me that the USB resources already have enough internal checks which reduces the need of a mandatory match.
if you prefere to keep i will revert that change.

PR #177, opened 2018-01-08
changed "-f" option for SSH master to "-n"

tgbloks:

this fixes some "strange" ssh behavior for me where the ssh master seems to got to background too early (im missing 2 debug messages in those cases) sometimes (about every 20 tries) on some machines and i then get connection refused errors when trying to connect via the control socket.

### DIFF
--- a/labgrid/driver/sshdriver.py
+++ b/labgrid/driver/sshdriver.py
@@ -52,7 +52,7 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
         )
         # use sshpass if we have a password
         sshpass = "sshpass -e " if self.networkservice.password else ""
-        args = ("{}ssh -f {} -x -o ConnectTimeout=30 -o ControlPersist=300 "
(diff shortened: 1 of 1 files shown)


labgrid/driver/sshdriver.py
```diff
@@ -62,7 +62,7 @@ def _start_own_master(self):
         env = os.environ.copy()
         if self.networkservice.password:
             env['SSHPASS'] = self.networkservice.password
-        self.process = subprocess.Popen(args, env=env)
+        self.process = subprocess.Popen(args, env=env, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
Emantor: Why do you change to PIPEs here?

tgbloks: ups - leftover from testing. prevents ssh from messing up my terminal in case something goes wrong but it might more interesting to get the output back now- i will revert it. just need to make sure its still working without before ...

tgbloks:
pipes reverted - still works ;-)

jluebbe:
I've added the PR description to the commit message and force pushed.

PR #183, opened 2018-01-27
Fix documentation compilation

Emantor:
This PR fixes the errors during documentation compilation. The last commit is a very dirty hack, but works for now. I'll dig into apidoc some more if we want to drop that commit.


jluebbe:
We don't want to have all modules automaticall loaded via `__init__.py`. Can we list those files in the sphinx config explicitly instead?

Emantor:
I removed the importing commit, the documentation builds without errors now. We also enable Warnings as errors.

I will fix the missing autodoc imports in a seperate PR.

PR #185, opened 2018-02-05
remote/client: Added missing speed configuration for microcom console

leopck:
microcom console was missing capability to adjust speed even though
resource.speed exist.


jluebbe:
Thanks for your contribution!

PR #189, opened 2018-02-09
Add requirements updates from pyup

Emantor:
I squashed the pyup pull request and added my own fixes. Afterwards we can enable pyup for labgrid.


Emantor:
Ah, how do we solve the DCO checks? Is there a workaround for a bot feature like this?

Emantor:
According to probot/dco#21 there is an issue for DCO and bot support, but nobody started work on it yet.

dev-requirements.txt
```diff
@@ -1,22 +1,15 @@
...
-pyudev
-yapf
-requests
-autobahn
-xmodem>=0.4.5
+PyYAML==3.12
+pyModbusTCP==0.1.5
+pysnmp==4.4.4
+pysnmp-mibs==0.1.6
+pytest==3.4.0
```
jluebbe: pytest is already part of requirements.txt.

dev-requirements.txt
```diff
@@ -1,22 +1,15 @@
...
-pytest-mock
-pytest-pylint
-pytest-runner
-onewire>=0.2
-pyudev
-yapf
-requests
-autobahn
-xmodem>=0.4.5
+PyYAML==3.12
```
jluebbe: PyYAML is needed by default.

Emantor:
Updated.

Emantor:
I split out the optional features and documented them in the installation section as well.

PR #191, opened 2018-02-12
driver/smallubootdriver: add driver for small UBoots

SmithChart:
This patch adds a driver for really small (and thus stripped down)
UBoot drivers.

This UBoots do not have:
* A real password prompt
* An echo command
* The possibility to print the return code of a command

This driver copes with these limitations.

Tested against the stock-uboot on a TP-Link WR841N v11.

Signed-off-by: Chris Fiege <chris@tinyhost.de>

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -618,6 +618,58 @@ Arguments:
     defaults to "U-Boot 20\d+"
   - bootstring (str): optional, regex to match on Linux Kernel boot
 
+SmallUBootDriver
+~~~~~~~~~~~~~~~~
+A SmallUBootDriver interfaces with stripped-down UBoot variants that are
+sometimes used in cheap consumer electronics.
+
+SmallUBootDriver is meant as a driver for UBoot with only little
+functionality compared to standard a standard UBoot.
+Especially is copes with the following limitations:
+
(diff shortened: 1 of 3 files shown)


doc/configuration.rst
```diff
@@ -580,6 +580,31 @@ Arguments:
   - boot_expression (str): optional, regex to match the uboot start string
     defaults to "U-Boot 20\d+"
 
+SmallUBootDriver
+~~~~~~~~~~~
+A SmallUBootDriver interfaces with stripped-down UBoot variants that are
+sometimes used in cheap consumer electronics.
```
jluebbe: Please describe when you would use this compared to the normal UBootDriver. Also update the UBootDriver docs to be explicit about which commands it needs to work.

jluebbe: Please describe when you would use this compared to the normal UBootDriver. Also update the UBootDriver docs to be explicit about which commands it needs to work.

SmithChart: I have added more documentation on the minimal features in UBoot needed for the SmallUBootDriver.

SmithChart:
Ping

I updated the documentation according to your request @jluebbe. Could you give it a look and decide if it can be merged now?

PR #192, opened 2018-02-12
 driver/powerdriver: Add SerialPortPowerdriver 

SmithChart:
Adding a powerdriver that controls an external relais board connected to
the control signals of a USB-UART-dongle.

This is a poor-mans solution for a power switch that can be used for
desktops with only a single place.

### DIFF
--- a/labgrid/driver/powerdriver.py
+++ b/labgrid/driver/powerdriver.py
@@ -131,6 +131,41 @@ class NetworkPowerDriver(Driver, PowerResetMixin, PowerProtocol):
 
 @target_factory.reg_driver
 @attr.s(cmp=False)
+class DigitalOutputPowerDriver(Driver, PowerResetMixin, PowerProtocol):
+    """
+    DigitalOutputPowerDriver uses a DigitalOutput to control the power
+    of a DUT.
+    """
+    bindings = {"output": DigitalOutputProtocol, }
+    delay = attr.ib(default=1.0, validator=attr.validators.instance_of(float))
+
+    def __attrs_post_init__(self):
+        super().__attrs_post_init__()
+
+    @Driver.check_active
+    @step()
+    def on(self):
+        self.output.set(True)
+
+    @Driver.check_active
+    @step()
+    def off(self):
+        self.output.set(False)
+
+    @Driver.check_active
+    @step()
+    def cycle(self):
+        self.off()
+        time.sleep(self.delay)
+        self.on()
+
+    @Driver.check_active
+    @step()
+    def get(self):
+        return self.output.get()
+
+@target_factory.reg_driver
+@attr.s(cmp=False)
 class YKUSHPowerDriver(Driver, PowerResetMixin, PowerProtocol):
     """YKUSHPowerDriver - Driver using a YEPKIT YKUSH switchable USB hub
         to control a target's power - https://www.yepkit.com/products/ykush"""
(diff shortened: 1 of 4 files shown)


labgrid/driver/powerdriver.py
```diff
@@ -187,3 +188,70 @@ def cycle(self):
...
+    some kind of relay board to the flow control pins.
+
+    The serial port should NOT be used for serial communication at the same
+    time. This will probably reset the flow-control signals.
+
+    Usable signals are DTR and RTS.
+    """
+
+    port = attr.ib(validator=attr.validators.instance_of(str))
+    signal= attr.ib(validator=attr.validators.instance_of(str))
```
Emantor: Please bind to a SerialPort Resource instead of taking the raw /dev/path here.

labgrid/driver/powerdriver.py
```diff
@@ -187,3 +188,70 @@ def cycle(self):
...
+    an USB-UART-dongle) to control some external power switch. You may connect
+    some kind of relay board to the flow control pins.
+
+    The serial port should NOT be used for serial communication at the same
+    time. This will probably reset the flow-control signals.
+
+    Usable signals are DTR and RTS.
+    """
+
+    port = attr.ib(validator=attr.validators.instance_of(str))
```
Emantor: Please bind to a SerialPort Resource instead of taking the raw /dev/path here.

labgrid/driver/powerdriver.py
Emantor: IMO extending the existing resources with an optional `power_signal` ist more straightforward.

SmithChart:
I am not sure about the specific reason: I tested using the same port for both - and it did behave strange. I did not invest time to investigate the behavior.

I will have ha look at integrating a PowerProtocoll into the SerialPortDriver.

jluebbe:
@SmithChart We shouldn't make the normal SerialDriver implement the PowerProtocol, as this would require using named drivers for the common case of a separate power driver. 

You could make this a a ConsolePIODriver, which binds to a ConsoleProtocol to export a control line as a PIO, then use the existing DigitalOutputPowerDriver for power control. We'd need to add a method to set/get control lines to ConsoleProtocol for this to work though. What do you think?

SmithChart:
> You could make this a a ConsolePIODriver, which binds to a ConsoleProtocol to export a control line as a PIO, then use the existing DigitalOutputPowerDriver for power control. We'd need to add a method to set/get control lines to ConsoleProtocol for this to work though. What do you think?

I am not sure if it is a good idea to actually dig this feature so deep into the labgrid infrastructure.
Binding the driver to a serial port as suggested by @Emantor makes total sense to me.

Adding features to the ConsolePrrotocol  just to support this featrue seems to softens the rather sharp outline of the architecture. This feature is meant to help a beginner to build a simple setup to start with labgrid.

Maybe it would be a better idea to alter the driver to a DigitalOutput driver and use the DigitalOutputPowerDriver on top of that.

SmithChart:
I had discusses this change with @jluebbe face-to-face. 

I splitted this driver into two parts. A DigitalOutputDriver and a DigitalOutputPowerSwitch.

For more details see the commit messages.

SmithChart:
Ping

I rebased my commits. Is there anything else missing?

SmithChart:
Ping

Still waiting for feedback here.

Issue #176, opened 2017-12-31
SerialDriver driver activates U-boot prompt.


kjeldflarup:
Sometimes the SerialDriver stops u-boot. It seems to press the anykey, and gets the u-boot prompt.
Looking in _await_login this comment "# TODO use step timeouts" indicates that this is a known problem.

The interruption of u-boot is caused by _await_login calling self.console.sendline("") as the first thing.  
It is of course target dependent, but my board always prints login:, no need to send a newline to get it. Should the target already have passed that point, then say 500 ms silence would indicate that I need to send a newline, to get the current prompt, whether it would be the logged in prompt, the login prompt OR a u-boot prompt.

Thus _await_login should loop on self.console.expect and only send a newline when receiving a timeout and the serial port has been quiet during that period.
I had this behavior implemented in the Ruby code which I used before starting to use labgrid.

Furthermore it should be able to detect the U-boot prompt, and recover from that too. But if the interruption can be prevented, this is not necessary.

jluebbe:
This behavior is in ShellDriver, not in SerialDriver. It's there to trigger printing the current prompt, so that it can continue without waiting.
To avoid activating the ShellDriver on a U-Boot prompt, the robust way is to use the UBootDriver and a strategy to explicitly capture the board in U-Boot and then trigger the kernel boot.

kjeldflarup:
Explicit capture of UBoot, would mean that the test of UBoot is not complete. We might oversee a misconfiguration in Uboot.

jluebbe:
This should be fixed via PR #193.

PR #193, opened 2018-02-12
 driver/shelldriver: Add 'enable console' feature

SmithChart:
This feature can be used for embedded devices that do not - by default -
start a console on the serial terminal.  The 'console_ready'-regex is
used to identify when the kernel is ready to start a console. Afterwards
<enter> is send to the device. The shelldriver now waits for a prompt or login.

Tested with a 4.4-kernel on a LEDE-based device.

If 'console_ready' is not set _await_login() should behave like before.

Depends on #192.

### DIFF
--- a/labgrid/driver/shelldriver.py
+++ b/labgrid/driver/shelldriver.py
@@ -11,6 +11,8 @@ from time import sleep
 import attr
 from pexpect import TIMEOUT
 
+import time
+
 import xmodem
 
 from ..factory import target_factory
@@ -44,6 +46,9 @@ class ShellDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
     password = attr.ib(default="", validator=attr.validators.instance_of(str))
     keyfile = attr.ib(default="", validator=attr.validators.instance_of(str))
     login_timeout = attr.ib(default=60, validator=attr.validators.instance_of(int))
+    console_ready = attr.ib(default="", validator=attr.validators.instance_of(str))
+    await_login_timeout = attr.ib(default=2, validator=attr.validators.instance_of(int))
+
 
     def __attrs_post_init__(self):
         super().__attrs_post_init__()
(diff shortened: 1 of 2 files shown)


labgrid/driver/shelldriver.py
```diff
@@ -102,24 +103,38 @@ def run(self, cmd, timeout=30.0):
...
+                    self.console.sendline(self.password)
+                    self.console.expect(self.prompt, timeout=5)
+                else:
+                    raise Exception("Password entry needed but no password set")
+            self._check_prompt()
+        elif index == 2:
+            # we have just activated a console here
+            # lets start over again and see if login or prompt will appear now
+            self.console.sendline("")
+            self._await_login()
```
jluebbe: _await_login() will send newline anyway, why do we need two? I'm somewhat skeptical regarding the recursion in this case, an explicit loop over the main expect might be clearer.

jluebbe:
@kjeldflarup What do you think about this? If we wait per default for 1 second of silence before trying to poke the login promt with a newline, would that work for your case?

If we see the console_ready string, we know it's okay to poke it immediately.

kjeldflarup:
@jluebbe Yes I think so, because we have bootdelay=0.
If boot delay was higher, then I believe the prompt would count down every second, so that shouldn't trigger a newline, but it will be close. Perhaps set the silence to 1.1 second.

After U-boot has started to print text on console, it is possible to send a newline, which will be buffered, and read when reaching "Hit any key to stop autoboot:" However there will not be any 1 sec delays during uboot initialization and that. Only pitfall could be if a 1 sec delay is detected before U-boot initializes, and the delay between timeout and the newline reaching the target is long enough to make the target ready to receive the newline. 
I don't think the latter is a problem. It is quite hypothetical.

labgrid/driver/shelldriver.py
SmithChart: You are right. _await_login() will send a newline anyway. (Even with the new silence-timeout schema discussed laster.) I'll remove L136.

Regarding the explicit loop:
I am fine with that. I just (mostly unconscious) copied this style from for example the ubootstrategy.py:50.
I will change this too.

SmithChart:
@jluebbe: 
I hope I've hit all the changes we discussed. I am now using an explicit loop instead of recursion.
This should make this function more readable.

Additionally I added the detection of silence before sending a newline to the DUT to deal with the issue @kjeldflarup described in Issue #176.
The driver should now wait for await_login_timeout of silence from the DUT before sending a newline.

I have tested this changes on my TP-Link Router setup here: I can still boot into the shell on the serial and run commands.

SmithChart:
Ping

I have rebased this branch. Is there anything else missing?

SmithChart:
Ping

Still waiting for feedback here.

PR #194, opened 2018-02-13
 config: Refuse to load imports if it's no list

SmithChart:
Using a single string as imports-key is syntactical correct for python
since a string is iterable. But using a string will not work logically and
throws a somewhat meaningless error.

This should improve the error messages of user-input parsing.

Also add imports to config-example.


Emantor:
The build error is due to changes in sphinx in version 1.7.0. With PR #189 we should be fine since we pin to a working version, i'll investigate the doc build error on thursday.

PR #197, opened 2018-02-19
ubootdriver: make bootstring configurable and update docstring

### LINKED ISSUE #184: Entering kernel is not detected by driver
I have a pytest that fails during verification of kernel present. Looking at the 'ubootdriver.py' it looks like that the verification is depended on writing Linux version upon kernel boot.

Is there any way to configure this detection method as  my Linux version 4.9.0-adi does not show this upon entering the kernel.

`    @Driver.check_active
    @step()
def await_boot(self):
        """Wait for the initial Linux version string to verify we succesfully
        jumped into the kernel.
        """
        self.console.expect(r"Linux version \d")`

Also with multiple kernels on different targets I would like to be able to detect the correct kernels etc. as part of the driver.

Currently this is blocking all my tests :(

Emantor:
Fixes #184

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -615,6 +615,7 @@ Arguments:
     defaults to "enter Password: "
   - boot_expression (str): optional, regex to match the uboot start string
     defaults to "U-Boot 20\d+"
+  - bootstring (str): optional, regex to match on Linux Kernel boot
 
 BareboxDriver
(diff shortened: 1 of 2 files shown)


GOMSPACEMartin:
@Emantor See you already added a fix, so I will pass on a PR this time around ;)

PR #201, opened 2018-02-23
implement driver priorities and the ResetProtocol

Emantor:
Initial starting point for the driver priorities. I wrote three tests and tweaked the `get_driver` function so far, so the tests pass. I have not looked a other parts of the chain yet.
TODO:
- [x] update docstrings for the DigitalOutputResetDriver

### DIFF
--- a/tests/test_target.py
+++ b/tests/test_target.py
@@ -277,3 +277,68 @@ def test_get_by_string(target):
 
     with pytest.raises(KeyError):
         target.get_driver("nosuchdriver")
+
+# Test priorities
+
+def test_get_by_diff_priority(target):
+    class AProtocol(abc.ABC):
+        pass
+
+    @attr.s
+    class A(Driver, AProtocol):
+        priorities = {AProtocol: -10}
+
+    @attr.s
+    class C(Driver, AProtocol):
+        priorities = {AProtocol: 10}
+        pass
+
+    a = A(target, None)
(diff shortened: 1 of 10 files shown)


Emantor:
I implemented the priorities in the various PowerDrivers with a new mixin. This way the powerdrivers provide the `reset` Funktion by emulating a Powercycle. If a real ResetDriver is available it is prefered due to a higher priority.

tests/test_target.py
```diff
@@ -277,3 +277,66 @@ class C(Resource):
...
+        priorities = {AProtocol: 10}
+        pass
+
+    a = A(target, None)
+    c = C(target, None)
+    target.activate(a)
+    target.activate(c)
+
+    with pytest.raises(NoDriverFoundError):
+        target.get_driver(AProtocol)
```
jluebbe: Check the error message as well.

tests/test_target.py
```diff
@@ -277,3 +277,66 @@ class C(Resource):
...
+    class C(Driver, AProtocol):
+        pass
+
+    a = A(target, None)
+    c = C(target, None)
+    target.activate(a)
+    target.activate(c)
+
+    with pytest.raises(NoDriverFoundError):
+        target.get_driver(AProtocol)
```
jluebbe: Check the error message here, too.

PR #206, opened 2018-02-28
driver/quartushpsdriver: add support for Quartus HPS

BastianStender:
Adds the driver itself, an example strategy and documentation.

The example strategy can be used like this:

``` yaml
targets
  my-target:
    resources:
      RemotePlace:
        name: 'my-target'
    drivers:
      - QuartusHPSDriver: {}
      - SerialDriver:
          txdelay: 0.01
      - NetworkPowerDriver:
          delay: 5.0
      - BareboxDriver:
          prompt: 'barebox@[^:]+:[^ ]+ '
          autoboot: "abort autoboot"
          interrupt: "\x03"
      - ShellDriver:
          prompt: 'root@[^:]+:[^ ]+'
          login_prompt: ' login: '
          username: 'root'
      - QuartusHPSStrategy:
          image_xload: /path/to/barebox-xload.img
          image: /path/to/barebox.img

imports:
- examples/strategy/quartusstrategy.py
```

### DIFF
--- a/labgrid/driver/quartushpsdriver.py
+++ b/labgrid/driver/quartushpsdriver.py
@@ -1,0 +1,82 @@
+# pylint: disable=no-member
+import attr
+import subprocess
+import os.path
+import re
+
+from ..factory import target_factory
+from ..resource.remote import NetworkAlteraUSBBlaster
+from ..resource.udev import AlteraUSBBlaster
+from ..step import step
+from .common import Driver, check_file
+from .exception import ExecutionError
+
+
+@target_factory.reg_driver
+@attr.s(cmp=False)
+class QuartusHPSDriver(Driver):
+    bindings = {
+        "interface": {AlteraUSBBlaster, NetworkAlteraUSBBlaster},
+    }
+
+    image = attr.ib(default=None, validator=attr.validators.optional(attr.validators.instance_of(str)))
+    cable_number = attr.ib(default=None, validator=attr.validators.optional(attr.validators.instance_of(str)))
+
+    def __attrs_post_init__(self):
+        super().__attrs_post_init__()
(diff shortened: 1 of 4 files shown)


BastianStender:
Should I include the configuration snippet from the commit message in the docs or is it self-explanatory enough? If yes, where should I include it?

labgrid/driver/quartushpsdriver.py
```diff
@@ -0,0 +1,83 @@
...
+from ..resource.remote import NetworkAlteraUSBBlaster
+from ..resource.udev import AlteraUSBBlaster
+from ..step import step
+from .common import Driver, check_file
+from .exception import ExecutionError
+
+
+@target_factory.reg_driver
+@attr.s(cmp=False)
+class QuartusHPSDriver(Driver):
```
Emantor: This needs to implement a Protocol. BootstrapProtocol would be the obvious choice, however your driver implements an additional feature: choosing the address of where to load the file. I am not sure that we want to extend the existing protocol or define a new one for this.

BastianStender:
Pushed changes:
- add FlashProtocol
- let QuartusHPSDriver implement FlashProtocol
- adjust docs accordingly

I have no strong opinion if more elaborate strategies should be in another repository. @jluebbe ?

jluebbe:
We discussed this internally and decided that having the FlashProtocol with this simple implementation doesn't help us much currently. Also, it would cause more churn when we add a more general FlashProtocol later. So we prefer to use the QuartusHPSDriver directly for now.

BastianStender:
Pushed changes:
- remove FlashProtocol

Should the `QuartusHPSStrategy` be part of Labgrid or do you want it in a separate repo?

PR #208, opened 2018-03-01
Add Support for the Pengutronix USB-SD-Mux Hardware

ejoerns:
This adds support for the quite useful SD card juggler from Pengutronix (https://www.pengutronix.de/de/2017-10-23-usb-sd-mux-automated-sd-card-juggler.html) finally allowing to use it for automated testing with labgrid.

I also added a lg-command `sd-mux` to switch the USB-SD-Mux via `labgrid-client` but kept it out of this series because not being sure about how generic this is.

Despite the actual `/dev/sgX` device the Muxer uses is a `scsi_generic` one, I added the muxer as an USB device because it actually has a USB cable.

For now I realize this by matching the actual SCSI-Device and scanning if the parent matches the USB-SD-Mux USB vender and model ID's. To make this work with recent labgrid UBS resource implementation I had to adapt the DEVNUM and BUSNAME lookup a little.

Another approach that came to my mind was matching the actual USB device and then scanning its childs to find out the `/dev/sgX` device which I found to be more complex and not required as long as we do not intend to handle the SD-Cards block devices together with the sg device of the same Muxer to provide a more abstract 'switchable SD-Card'.

The series requires to set up the `usbsdmux ` tool as described here: https://github.com/pengutronix/usbsdmux

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -205,6 +205,18 @@ A USBMassStorage resource describes a USB memory stick or similar device.
 
 Used by:
   - `USBStorageDriver`_
+  - `NetworkUSBStorageDriver`_
+
+NetworkUSBMassStorage
+~~~~~~~~~~~~~~~~~~~~~
+A NetworkUSBMassStorage resource describes a USB memory stick or similar
+device available on a remote computer.
+
+Used by:
+  - `NetworkUSBStorageDriver`_
+
+The NetworkUSBMassStorage can be used in test cases by calling the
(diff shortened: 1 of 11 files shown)


ejoerns:
I currently implement the ntework usb storage driver part for this to for actually being able to write images to the muxed SD card.

For being able to use this appropriately and not having to match sg to block devices manually it might be useful to combine the sd-muxing with the block device on a higher level.

ejoerns:
I've updated the series to now include full support for the USBSDMux hardware.

PR #210, opened 2018-03-02
Document how to use multiple drivers with names in the yaml configuration files.

Emantor:
Add documentation for the bindings key in the environment configuration files. I also added a missing resource to configuration.

### DIFF
--- a/doc/usage.rst
+++ b/doc/usage.rst
@@ -100,6 +100,9 @@ To access the target's console, the correct driver object can be found by using
 When using the ``get_driver`` method, the driver is automatically activated.
 The driver activation will also wait for unavailable resources when needed.
 
+For more information on the environment configuration files and the usage of
+multiple drivers, see :ref:`configuration:Environment Configuration`.
+
 pytest Plugin
 -------------
 Labgrid includes a `pytest <http://pytest.org>`_ plugin to simplify writing tests which
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -38,7 +38,8 @@ import sphinx_rtd_theme
 extensions = ['sphinx.ext.autodoc',
               'sphinxcontrib.napoleon',
               'sphinx.ext.coverage',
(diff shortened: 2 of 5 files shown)


doc/usage.rst
```diff
@@ -1,6 +1,76 @@
 Usage
 =====
 
+Environment YAML files
```
jluebbe: Why did you move this section before the Target section? Especially the support for named resources and drivers is not relevant for most setups and should be explained later.

jluebbe:
I've pushed a change to use real resources and drivers to make the example clearer.

doc/usage.rst
Emantor: I wanted to unify the seperate explanations and both of the next sections depend on knowledge of the environment yaml configuration syntax. I don't think that seperating the multiple drivers part benefits the reader, we can also explain it to the reader here which also makes it simpler to look up.

jluebbe: The library part actually explains how to create Targets, Drivers and Resources manually, so the reader knows how these fit together.

Emantor: @ejoerns we'd like a third opinion on this, please provide yours.

ejoerns: The Environment configuration should have an extra chapter to mention all the dirty details and what can be done with it. But I would also agree on @jluebbe's concern that placing it at the very top of the 'usage' chapter leads the user in the wrong direction.
May it will be possible to add short notes wit examples (as done so far) in both the library and the pytest chapter on usage of Environment files and link to a separate chapter for more details. As Environment configs are neither specific to the library nor to the pytest case, I would give it the same level they have as @Emantor already did but maybe place it somewhere below the pytest chapter in doc/usage.rst or simply extend the already existing  'Environment Configuration' chapter in doc/configuration.rst.

Emantor:
I incorporated your changes and moved the multiple drivers decription into the Configuration chapter, as suggested by @ejoerns.

PR #211, opened 2018-03-02
Some unrelated fixes

ejoerns:


labgrid/environment.py
```diff
@@ -22,7 +22,7 @@ def __attrs_post_init__(self):
             self.config = Config(self.config_file)
         except:
             raise NoConfigFoundError(
-                "{} is not a valid yaml file".format(self.config_file)
```
jluebbe: The Config class already returns NoConfigFoundError or InvalidConfigError, so we shouldn't catch those here only to give less information.

tests/test_crossbar.py
```diff
@@ -110,14 +110,14 @@ def test_place_aquire(place):
 
 def test_place_add_no_name(crossbar):
     with pexpect.spawn('python -m labgrid.remote.client create') as spawn:
-        spawn.expect("missing place name")
+        spawn.expect("missing place name. Set with -p <place> or via env var $PLACE")
```
jluebbe: This fails on travis. Maybe because you need to escape some parts which are interpreted as regular expressions?

labgrid/environment.py
ejoerns: Ok, that was my second feeling on these lines thus I will remove the re-raise.

tests/test_crossbar.py
ejoerns: Yes, that's my assumption, too when I looking at the test results. Will fix it an re-push it.

Emantor: The test expects an EOF after missing place name, thats why it fails.

jluebbe: expect(pattern) discards any unmatched output before pattern matches, so I don't think expect(pexpect.EOF) is the problem. Also, according to the backtrace, the error is before the EOF line.

Emantor: I read it as 'somepatter''EOF', my bad.

ejoerns:
I've fixed and extended the `NoConfigFoundError` patch and adapted the test. Also fixed the expect error by simply not changing the test but only the full message.

This updated series also contains 3 new patches that reflect some minor documentation issues I came across when working on labgrid.

PR #212, opened 2018-03-02
Make the ExecutionError readable for the run_check case

Emantor:
If you use the run_check function and get an ExecutionError, it will only contain the command that was run, but not the stdout or stderr. Add those and refactor run_check into the mixin.

### DIFF
--- a/labgrid/driver/commandmixin.py
+++ b/labgrid/driver/commandmixin.py
@@ -22,3 +22,34 @@ class CommandMixin:
             sleep(sleepduration)
         if timeout.expired:
             raise ExecutionError("Wait timeout expired")
+
+    def _run_check(self, cmd: str, timeout=30):
+        """
+        Internal function which runs the specified command on the shell and
+        returns the output if successful, raises ExecutionError otherwise.
+
+        Args:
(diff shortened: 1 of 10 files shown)


Emantor:
Rebased and added tests

labgrid/driver/commandmixin.py
```diff
@@ -21,4 +21,22 @@ def wait_for(self, cmd, pattern, timeout=30.0, sleepduration=1):
         while not any(pattern in s for s in self.run_check(cmd)) and not timeout.expired:
             sleep(sleepduration)
         if timeout.expired:
-            raise ExecutionError("Wait timeout expired")
+            raise ExecutionError("Wait timeout expired", '')
+
+    @Driver.check_active
+    @step(args=['cmd'], result=True)
+    def run_check(self, cmd: str, timeout=30):
```
jluebbe: This removes the _run_check method, which can be used internally during activation and will break _put_ssh_key of the ShellDriver.

Emantor:
Fixed the removal of the _run_check function.

Emantor:
Moved _run_check into the mixin.

PR #217, opened 2018-03-07
Increase coverage

Emantor:
Various small new tests to increase coverage


tests/test_util.py
```diff
@@ -0,0 +1,20 @@
+from labgrid.util import diff_dict, flat_dict
+
+def test_diff_dict():
+    dict_a = {"a": 1
+              ,"b": 2}
```
jluebbe: ‽

Emantor:
Done :+1: ‽

PR #218, opened 2018-03-08
Use --hostname to override export host name

kjeldflarup:
When exporting USBSerialPort the exporter exports the host as the hostname. This may however not always resolve. 

setting LG_HOSTNAME to ex an IP address will override gethostname() in the exporter
This variable can be used with ex. USBSerialPort, USBGenericExport, USBSigrokExport
Tested with USBSerialPort only.

### DIFF
--- a/labgrid/remote/exporter.py
+++ b/labgrid/remote/exporter.py
@@ -390,8 +390,8 @@ def main():
         '--name',
         dest='name',
         type=str,
-        default=gethostname(),
-        help='public name of this exporter'
+        default=None,
+        help='public name of this exporter (defaults to the system hostname)'
     )
     parser.add_argument(
         '-d',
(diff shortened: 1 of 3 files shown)


jluebbe:
There is already the --name parameter to set the name used to register resources. I've added a commit on top which implements an --hostname parameter instead of an environment variable. So far I've only tested it without real hardware.

kjeldflarup:
Could I get a copy of Your commit?

jluebbe:
@kjeldflarup It's in this PR branch. See https://github.com/labgrid-project/labgrid/pull/218/commits/144e3b810bc70e528d152eca036f7f45b2b22d06

kjeldflarup:
I have testet this on my Raspberry Pi and found it working.

PR #219, opened 2018-03-08
Add simplerest model for NetworkPowerPort

kjeldflarup:
Adds a generic powercontroller for NetworkPowerPort  using http get rest calls.
It uses a "Tweak" in NetworkPowerPort declaration where anything can be written in the "host" field.

### DIFF
--- a/labgrid/driver/power/simplerest.py
+++ b/labgrid/driver/power/simplerest.py
@@ -1,0 +1,25 @@
+""" Simple rest interface for Power Port. Used for ex. misc Raspberry Pi configs
+    Author: Kjeld Flarup <kfa@deif.com>
+
+    The URL given in hosts in exporter.yaml must replace {value} with '0' or '1'
+    It is optional whether to use {index} or not.
+
+  NetworkPowerPort:
+      model: simplerest
+      host: 'http://172.17.180.53:9999/relay/{index}/{value}'
+      index: 0
+"""
+
+import requests
+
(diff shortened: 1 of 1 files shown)


jluebbe:
If I read it correctly, this only works for power controllers which use 'on'/'off' while set and '0'/'1' for get? The "networkpowerurl" model name is probably too generic for this behavior. Is this for some specific hardware? Or do you have several different models which all use this REST API?

kjeldflarup:
The primary target is Raspberry Pi, which can implement power relay in many ways. Making a rest server using flask is a simple task.
A typical setup is to connect a solid state relay directly to a GPIO pin.

Regarding on/off, You are right, that ought to be 0/1, which my rest server also supports.

jluebbe:
Ah, OK. Maybe rename it to simplerest. networkpowerurl is too generic for this.
I'll then squash and merge it.

kjeldflarup:
I have renamed to simplerest, but it just occurred to me that the --hostname in https://github.com/labgrid-project/labgrid/pull/218 might come in handy.  
I will thus test with storing the URL in "index" instead. That might also be more appropriate.  
But is there a way to use --hostname, or should I just write:  
`host: !template $LG_IP`

kjeldflarup:
Another caveat is that index is parsed as an integer. So I will have to stick with tweaking "host"
That has already been pushed to the PR

PR #223, opened 2018-03-20
Add an example with local and remote resources as well as strategy-based tests

jluebbe:


labgrid/factory.py
```diff
@@ -80,6 +80,23 @@ def _convert_to_named_list(self, data):
             assert 'cls' in item
         return result
 
+    def normalize_role(self, name, config):
+        role = config.get('role', name)
```
Emantor: This function is named `normalize_role` but does not do anything with the role. Rename to `normalize_config` and remove the role parameter?

labgrid/remote/client.py
```diff
@@ -809,6 +823,21 @@ def create():
...
+def find_role_by_place(config, place):
+    for role, role_config in config.items():
+        resources, _ = target_factory.normalize_role(role, role_config)
+        remote_places = resources.get('RemotePlace', {})
+        remote_place = remote_places.get(place)
+        if remote_place:
+            return role
+
+def find_any_role_with_place(config):
+    for role, role_config in config.items():
```
Emantor: As above, iterating over the role is not necessary.

labgrid/remote/client.py
```diff
@@ -809,6 +823,21 @@ def create():
     loop.run_until_complete(ready.wait())
     return session[0]
 
+def find_role_by_place(config, place):
+    for role, role_config in config.items():
```
Emantor: If we don't have to pass the role, this can be simplified as well.

labgrid/factory.py
jluebbe: Renamed to normalize_config and dropped the name parameter.

labgrid/remote/client.py
jluebbe: We still need the loop, but we don't need to pass the role name to normalize_config.

labgrid/remote/client.py
jluebbe: The role name is part of the return value here as well.

Issue #203, opened 2018-02-27
Character decoding


kjeldflarup:
SOMEBODY thought that it was a good idea to use a unicode centerdot 0xb7 in a product name.

Now this gives me a failure in the SSHDriver: 
```
>       stdout = stdout.decode("utf-8").split('\n')
E       UnicodeDecodeError: 'utf-8' codec can't decode byte 0xb7 in position 90: invalid start byte
```
The ShellDriver may have the same problem, as well as barebox and uboot  
Changing to iso-8859-1 solves my problem, for now.

Possible solutions.

1. Add an encoding parameter to the drivers.
2. Add an encoding parameter to the ressource
3. Add an encoding parameter to the run command
4. Change the error handling in the decode command to ignore or replace

The most optimal solution to the problem is to be able to set the encoding for the individual target. The easiest place may be the SSHDriver directly. But the info may actually be more accurate to set in the resource in the exporter, thus the same test case can connect to two differently configured targets.

A more simple solution would be to write ("utf-8", 'ignore')  
This though has the drawback, that stray binary output no longer causes the test to fail. This is clearly a tradeoff.  
The run command could be added an ignore parameter. (OR codecignore to be less confusing), in this way the failing test cases could be modified only.

jluebbe:
I've merged your PR #214, so you can configure the handling in the run() method. We can still allow specifying the default encoding on the driver later (which I'd prefer to the resource as it's not related to the physical interface).
Additionally, we could add add raw_run() methods (if needed) which wouldn't do any en-/decoding.
Also, the middle dot would need to be encoded as 0xC2 0xB7 (according to http://www.fileformat.info/info/unicode/char/b7/index.htm), so possibly the product name is buggy. ;)

kjeldflarup:
Came across this issue else where  
The thing is that UTF-8 is a bad default iso-8859-1 (aka latin_1) is much better as default, because it is 8 bit, and all 256 characters would be legal, so no error could ever happen. Only use UTF-8 if special characters matters.

PR #237, opened 2018-04-30
multiple: add _run function, use _run in _run_check for the mixin

### LINKED ISSUE #230: Missing authorized_keys causes "ShellDriver.run can not be called"
If I remove .ssh/authorized_keys on my target I get an error:

This is the full call path of the error
```
target = Target(name='main', env=Environment(config_file='labgrid.yaml'))

    def strategy(target):
>       shell = target.get_driver('ShellDriver')

conftest.py:6: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../../labgrid/labgrid/target.py:191: in get_driver
    self.activate(found[0])
../../labgrid/labgrid/target.py:433: in activate
    client.on_activate()
../../labgrid/labgrid/driver/shelldriver.py:69: in on_activate
    self._put_ssh_key(self.keyfile)
../../labgrid/labgrid/step.py:209: in wrapper
    _result = func(*_args, **_kwargs)
../../labgrid/labgrid/driver/shelldriver.py:252: in _put_ssh_key
    self._run_check("chmod 700 ~/.ssh/")
../../labgrid/labgrid/driver/commandmixin.py:37: in _run_check
    stdout, stderr, exitcode = self.run(cmd, timeout=timeout)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = ShellDriver(target=Target(name='main', env=Environment(config_file='labgrid.yaml')), name=None, state=<BindingState.bo...800', keyfile='/userdata/labgridtools/ssh/labgrid_rsa.pub', login_timeout=120, console_ready='', await_login_timeout=2)
_args = ('chmod 700 ~/.ssh/',), _kwargs = {'timeout': 30}

    @wraps(func)
    def wrapper(self, *_args, **_kwargs):
        if self.state is not BindingState.active:
            raise StateError(
                "{} can not be called ({} is in state {})".format(
>                   func.__qualname__, self, self.state.name)
            )
E           labgrid.binding.StateError: ShellDriver.run can not be called (ShellDriver(target=Target(name='main', env=Environment(config_file='labgrid.yaml')), name=None, state=<BindingState.bound: 1>, prompt='~ # ', login_prompt=' login: ', username='root', password='****', keyfile='/userdata/labgridtools/ssh/labgrid_rsa.pub', login_timeout=120, console_ready='', await_login_timeout=2) is in state bound)
```

The error can be prevented in two ways:

1) Change run to _run
https://github.com/labgrid-project/labgrid/blob/aeb049d99b585d336e34f768d636c87a908262f7/labgrid/driver/commandmixin.py#L37

2) Change the order, so state is changed before calling on_activate
https://github.com/labgrid-project/labgrid/blob/aeb049d99b585d336e34f768d636c87a908262f7/labgrid/target.py#L432-L434

Emantor:
Commit 1deea1c71728f50644ffd7b7e709577f43e1d4c2 introduced a bug where the
internal _run_check function used the run function guarded by a check_active
decorator. Use _run for the run_check function, refactor _run functions for
drivers who did not have it and adjust the tests to mock the correct funtion.

Fixes #230

### DIFF
--- a/labgrid/driver/bareboxdriver.py
+++ b/labgrid/driver/bareboxdriver.py
@@ -62,6 +62,9 @@ class BareboxDriver(CommandMixin, Driver, CommandProtocol, LinuxBootProtocol):
     @Driver.check_active
     @step(args=['cmd'])
     def run(self, cmd: str, *, step, timeout: int = 30):  # pylint: disable=unused-argument
+        return self._run(cmd, step=step, timeout=timeout)
+
+    def _run(self, cmd: str, *, step, timeout: int = 30):  # pylint: disable=unused-argument
         """
         Runs the specified command on the shell and returns the output.
 
--- a/labgrid/driver/sshdriver.py
+++ b/labgrid/driver/sshdriver.py
@@ -99,11 +99,10 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
         return self._start_own_master()
 
     @Driver.check_active
-    @step(args=['cmd'])
+    @step(args=['cmd'], result=True)
     def run(self, cmd, codec="utf-8", decodeerrors="strict", timeout=None): # pylint: disable=unused-argument
         return self._run(cmd, codec=codec, decodererrors=decodeerrors)
 
-    @step(args=['cmd'])
     def _run(self, cmd, codec, decodeerrors): # pylint: disable=unused-argument
         """Execute `cmd` on the target.
 
--- a/labgrid/driver/sshdriver.py
+++ b/labgrid/driver/sshdriver.py
@@ -101,6 +101,10 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
     @Driver.check_active
     @step(args=['cmd'])
     def run(self, cmd, codec="utf-8", decodeerrors="strict", timeout=None): # pylint: disable=unused-argument
+        return self._run(cmd, codec=codec, decodererrors=decodeerrors)
+
+    @step(args=['cmd'])
+    def _run(self, cmd, codec, decodeerrors): # pylint: disable=unused-argument
         """Execute `cmd` on the target.
 
         This method runs the specified `cmd` as a command on its target.
--- a/labgrid/driver/commandmixin.py
+++ b/labgrid/driver/commandmixin.py
@@ -34,7 +34,7 @@ class CommandMixin:
         Returns:
             List[str]: stdout of the executed command
         """
-        stdout, stderr, exitcode = self.run(cmd, timeout=timeout)
+        stdout, stderr, exitcode = self._run(cmd, timeout=timeout)
         if exitcode != 0:
             raise ExecutionError(cmd, stdout, stderr)
         return stdout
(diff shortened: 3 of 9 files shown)


labgrid/driver/bareboxdriver.py
```diff
@@ -59,8 +59,11 @@ def on_deactivate(self):
         self._status = 0
 
     @Driver.check_active
-    @step(args=['cmd'])
     def run(self, cmd: str, *, step, timeout: int = 30):
+        return run(cmd, step=step, timeout=timeout)
+
+    @step(args=['cmd'])
```
jluebbe: Steps on internal methods aren't really useful.

labgrid/driver/sshdriver.py
```diff
@@ -105,6 +105,10 @@ def _check_master(self):
     @Driver.check_active
     @step(args=['cmd'])
     def run(self, cmd, codec="utf-8", decodeerrors="strict"):
+        return _run(cmd, codec=codec, decodererrors=decodeerrors)
+
+    @step(args=['cmd'])
```
jluebbe: Here as well.

Emantor:
Added a commit which removes the step decorators from _run functions. FWIW: there are more cases where we have step decorators on internal functions, but thats something for another PR.

Emantor:
I updated this branch and fixed the conflicts.

jluebbe:
I've reworked the patches a bit (fix infinite recursion, move steps instead of deleting) and added tests for run() in addition to run_check().

kjeldflarup:
I just found the need to add `codec="utf-8", decodeerrors="strict"` to run_check, but a PR on that would clearly clash with this PR. 
Hope that this PR will be ready soon.

jluebbe:
@kjeldflarup Would having a run_bytes() which worked with bytes instead of strings help you? Then you could handle the decoding youself.

kjeldflarup:
@jluebbe We already have codecs on run. I just got the nice idea to use run_check as much as possible.
I don't think it is worth it to invent more methods than necessary.

jluebbe:
I'd like to understand your usecase better:
How does that look in you test cases? Do you use different codes and error settings on one target depending on the test? If it's always the same, would a default codec at the ShellDriver be more useful?

kjeldflarup:
We had the issue before: https://github.com/labgrid-project/labgrid/issues/203, perhaps reopen.
Adding codec to run gave a useful solution. Adding it to run_check seems like a natural completion of this path. 
Changing in the driver ressources also was a solution.

PR #240, opened 2018-05-17
ubootstrategy: from shell to u-boot by reboot instead of powercycle

nefethael:


jluebbe:
This way, the system state in the bootloader can differ depending on what happened before. For example the SoC reset reason might be different for power-on-reset compared to reboot. If you need this behavior in your tests, I think it would be better to have that in a custom Strategy (possibly based on the GraphStrategy).

Emantor:
Closing this based on @jluebbe explanation above. If you have questions regarding strategies, don't hesitate to ask us on IRC or via Matrix (which bridges to the same channel)

PR #244, opened 2018-05-22
shelldriver: Add public check_prompt method with customizable timeouts

nefethael:
Some processes can flood console stdout, resulting in _check_prompt failure with low expect timeouts.
This PR allows check_prompt to be called with custom timeouts. 

### DIFF
--- a/labgrid/driver/shelldriver.py
+++ b/labgrid/driver/shelldriver.py
@@ -176,7 +176,14 @@ class ShellDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
         """
         return self._status
 
-    def _check_prompt(self):
+    @step(args=['marker_timeout','prompt_timeout'])
+    def check_prompt(self, marker_timeout=2, prompt_timeout=1):
+        """
+            Check if we have a valid prompt
+        """
+        return self._check_prompt(marker_timeout, prompt_timeout)
+
+    def _check_prompt(self, marker_timeout=2, prompt_timeout=1):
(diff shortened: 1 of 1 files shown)


jluebbe:
_check_prompt currently only exists as an internal function. In which case do you have an valid prompt and get output from the target at the same time?

nefethael:
@jluebbe when using [utp_com](https://github.com/freescale/utp_com) to send command to imx6 board [imx-uuc](https://github.com/codeauroraforum/imx-uuc) daemon, I have a valid serial prompt but somehow messages from uuc are catched by shelldriver and it fails to recover the prompt within short timeout. With longer check, I can get back a valid prompt.

jluebbe:
Hmm, this PR doesn't change the timeouts for the existing users for _check_timeout, only when used from the new check_timeout. When you use run(), the existing timeout is still used. How does that fix your problem?

nefethael:
Hi Jan,

I switched job mission, so I can't tell exactly, but I used it kind this way:

....
    uboot_strat.transition('uboot')
    uboot_strat.transition('shell')
    uboot_strat.shell.check_prompt(30, 30)
    uboot_strat.shell.run('ls /media/usbstick')
...

Basically, as far as I remember, after installing new rootfs, I reboot and a long U-boot script runs some stuff, partitionning etc, and prompt appears after a long period. 

Regards

nefethael:
@jluebbe I'll not be able to reproduce so I'm closing this PR

PR #245, opened 2018-05-23
ser2net fix

jluebbe:
When many udev events happen very quickly (such as when a whole USB bus is reconnected), the intermediate disconnected state may be missed. This patch makes sure to restart ser2net when needed by comparing the current parameters to the ones used to start running ser2net process.


Emantor:
Do we have any idea why the underlying disconnects are not recognised?

This workaround looks fine, but it would be better if we could fix the underlying issue as well.

jluebbe:
The exporter misses the intermediate states because it doesn't process the the udev events (which is done in resource/udev.py) and can only see the current state.

rohieb:
The first commit also fixes #236.

Emantor:
Thanks for mentioning :+1:

PR #247, opened 2018-05-25
Fix Priorities for get_active_driver

Emantor:
`get_active_driver` is just a special case of `get_driver`, refactor the code to
reflect that by adding an active keyword parameter. Also assert if activate and
active are both set.

This also fixes a bug where `get_active_driver` did not respect driver
priorities.

### DIFF
--- a/labgrid/target.py
+++ b/labgrid/target.py
@@ -137,16 +137,9 @@ class Target:
             self.await_resources(found)
         return found[0]
 
-    def get_driver(self, cls, *, name=None, activate=True):
-        """
-        Helper function to get a driver of the target.
-        Returns the first valid driver found, otherwise None.
+    def _get_driver(self, cls, *, name=None, activate=True, active=False):
+        assert not (activate == True and active == True)
 
-        Arguments:
-        cls -- driver-class to return as a resource
-        name -- optional name to use as a filter
-        activate -- activate the driver (default True)
-        """
         found = []
         other_names = []
         if type(cls) is str:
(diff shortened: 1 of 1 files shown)


labgrid/target.py
```diff
@@ -147,6 +147,8 @@ def get_driver(self, cls, *, name=None, activate=True):
         name -- optional name to use as a filter
         activate -- activate the driver (default True)
         """
+        assert activate == active
```
jluebbe: This breaks the default behavior (search for drivers in any state and activate them).

labgrid/target.py
```diff
@@ -200,35 +204,7 @@ def get_active_driver(self, cls, *, name=None):
...
-            else:
-                raise NoDriverFoundError(
-                    "no active driver matching {} found in target {}".format(cls, self)
-                )
-        elif len(found) > 1:
-            raise NoDriverFoundError(
-                "multiple active drivers matching {} found in target {}".format(cls, self)
-            )
-        return found[0]
+        return self.get_driver(cls, name=name, activate=False, active=True)
```
jluebbe: This would fail the assert above.

tests/test_target.py
```diff
@@ -58,13 +58,13 @@ class B(Driver):
     assert "have other names" in excinfo.value.msg
     with pytest.raises(NoDriverFoundError) as excinfo:
         target[B, "adriver"]
-    assert "no active driver" in excinfo.value.msg
+    assert "no driver" in excinfo.value.msg
```
jluebbe: I'd prefer to keep the explicit exception messages for clarity (below as well).

labgrid/target.py
kjeldflarup: Yes I encountered that!

kjeldflarup:
Now it does not make any errors, but which driver do I then get?

Emantor:
You get the driver with the highest priority. I'll refactor as suggested by @jluebbe.

labgrid/target.py
Emantor: Yes, that should be !=...

tests/test_target.py
```diff
@@ -64,7 +64,7 @@ class B(Driver):
     target.activate(a2)
     with pytest.raises(NoDriverFoundError) as excinfo:
         target[A]
-    assert "multiple active drivers" in excinfo.value.msg
+    assert "multiple drivers" in excinfo.value.msg
```
jluebbe: Why is this change still needed?

Emantor: I missed the multiple drivers line, good catch.

Issue #250, opened 2018-05-28
Creating unique temporary directories on DUT


kjeldflarup:
We would like to make a generic mechanism in labgrid, which can be used for creating a mktemp(1) directory on target and copy files to that.

 def test_hello(request,target):
    command = target.get_active_driver( "CommandProtocol")
    tmpdir = command.mkdtemp()
    tmpdir.put(os.path.join(request.fspath, 'src/hello'))
    stdout = command.run_check(tmpdir.path + '/' + 'hello')

mkdtemp() execute mktemp -d on DUT, and returns an object with a put method, which sets the destination directory to the temporary directory.

Now should this feature be added to CommandProtocol or FileTransferProtocol?  
The driver needed to do this, must have both!

The Class for the returned object, should of course also be in the labgrid three. 
labgrid/util seems to be the most appropriate place.

The returned object is quite limited. It would have been great with at clone of pytest's tmpdir with py.path.local support. But as far as I can see, none of the code can be reused. Thus all needs to be rewritten from scratch.

Implementing this as a function scope fixture would also make sense, but there is no obvious place to put new generic fixtures in the labgrid tree.

PR #251, opened 2018-05-28
SSHFS: Add helper class to mount filesystem on target through ssh

nefethael:

### DIFF
--- a/labgrid/driver/sshdriver.py
+++ b/labgrid/driver/sshdriver.py
@@ -104,7 +104,7 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
 
     @Driver.check_active
     @step(args=['cmd'])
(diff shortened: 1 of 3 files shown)


kjeldflarup:
Would be nice with an example in the examples section.

A use case could be to combine it with mktemp -d on the DUT, and mount a tmpdir fixture from pytest.

See also https://github.com/labgrid-project/labgrid/issues/250

nefethael:
@kjeldflarup you're right, i'll add example some day but here how i use it:

in test.py:
`sshfs = SSHFS(cmd.target, 'mydir', '/tmp/mymount') `

in local.yaml:
```
....
    - NetworkService:
        address: myhost.local
        username: john
        password: doe
        name: 'mydir'
....
images:
  mydir: '/home/john/doe/mywanteddir'
```

jluebbe:
This should go into https://github.com/labgrid-project/labgridhelper.

PR #253, opened 2018-05-31
UsbStick2 & ImageCreator

nefethael:
Split raw image creation and virtual usbstick handling proposal with configfs system to create gadget.


jluebbe:
We should port this functionality into the labgridhelper library.

Issue #246, opened 2018-05-24
SSHDriver cannot execute run_check


kjeldflarup:
This code fails:
```
@pytest.fixture(scope='session')
## Returns a SSHDriver object to do ssh operations
def sshcommand(strategy):
    strategy.transition("ssh")
    return strategy.ssh

def test_run_check(sshcommand):
    sshcommand.run_check('echo ssh')
```

```
E               TypeError: got an unexpected keyword argument 'timeout'
/usr/lib/python3.5/inspect.py:2910: TypeError
```

The problem seems to be that CommandMixin calls run in this way:
```
stdout, stderr, exitcode = self.run(cmd, timeout=timeout)
```
But only ShellDriver implements timeout.

BastianStender:
1deea1c ("driver/multiple: move run_check function") introduced this bug.

BastianStender:
`UBootDriver` and `FakeCommandDriver` are missing the timeout keyword too.

PR #254, opened 2018-05-31
Add timeout parameter to run on drivers, to support run_check

kjeldflarup:
Closes: https://github.com/labgrid-project/labgrid/issues/246

Signed-off-by: Kjeld Flarup <kfa@deif.com>


kjeldflarup:
Using pylint directive is also a kind of documentation. I have updated the commit with that!

PR #255, opened 2018-06-01
Bug fixes, linting, drop python3.4 and Travis rework

BastianStender:
This PR includes several bug fixes. These bugs were found while using pylint. To be more consistent in coding style fix most of the linting warnings and all errors. Some warnings were left out, because the fix is either not obvious or time consuming. This is work tbd.

While at it python3.4 support was dropped.

Finally Travis-CI should check for linting errors to prevent those bugs from now on.

I think more warnings and errors should be ignored, but we need a separate discussion for that. Some warnings are caused by pylint not recognizing [attr](https://github.com/python-attrs/attrs) behavior. Discussion happens in #257.


Emantor:
Do you have any idea why coverages takes a nosedive?

BastianStender:
> Do you have any idea why coverages takes a nosedive?

Yes, this is mostly caused by "drop python3.4 support" and "linting: do miscellaneous linting". These commits remove lines, either by dropping the `@asyncio.coroutine` decorator line on tested code or by 
reformatting (joining lines of tested code or adding line breaks in untested code). See code coverage drops in:

- [labgrid/driver/smallubootdriver.py](https://codecov.io/gh/labgrid-project/labgrid/pull/255/diff?src=pr&el=tree#diff-bGFiZ3JpZC9kcml2ZXIvc21hbGx1Ym9vdGRyaXZlci5weQ==)
- [labgrid/remote/exporter.py](https://codecov.io/gh/labgrid-project/labgrid/pull/255/diff?src=pr&el=tree#diff-bGFiZ3JpZC9yZW1vdGUvZXhwb3J0ZXIucHk=)

BastianStender:
- rebased
- drop "factory: normalize config with correct driver class" in favor of merged #258
- sort "strategy/graphstrategy: simplify dict iteration in graph()" after "linting: strategy/graphstrategy: no empty list as arg default" and prefix with "linting: "

BastianStender:
- tox.ini: use [leading minus](http://tox.readthedocs.io/en/latest/config.html#confval-commands=ARGVLIST) instead of `ignore_errors=True` in order _not_ to influence the tox exit code

labgrid/autoinstall/main.py
```diff
@@ -73,7 +73,7 @@ def run(self):
             self.context['target'] = self.target
             if self.target is None:
                 raise KeyError
-        except:
+        except KeyError:
```
jluebbe: Other exceptions can be generated by env.get_target() during target creation, which should also be caught by this catch all.

labgrid/target.py
```diff
@@ -97,15 +97,15 @@ def await_resources(self, resources, timeout=None, avail=True):
...
+    def get_resource(self, cls, *, name=None, wait_for=True):
         """
         Helper function to get a resource of the target.
         Returns the first valid resource found, otherwise None.
 
         Arguments:
         cls -- resource-class to return as a resource
         name -- optional name to use as a filter
-        await -- wait for the resource to become available (default True)
+        wait_for -- wait for the resource to become available (default True)
```
jluebbe: I'd prefer wait_avail instead of wait_for. wait_for suggests that you need to pass the thing you want to wait for.

labgrid/remote/exporter.py
```diff
@@ -381,13 +381,9 @@ def onChallenge(self, challenge):
         self.loop.stop()
 
     async def acquire(self, group_name, resource_name):
-        resource = self.groups[group_name][resource_name]
```
jluebbe: This (and the one below) should actually be TODOs. We want to be able to perform local actions when a resource is acquired, and this would be the correct place when we get around to that.

labgrid/driver/shelldriver.py
```diff
@@ -61,13 +61,14 @@ def on_activate(self):
             self._inject_run()
         if self.keyfile:
             self._put_ssh_key(self.keyfile)
-        self._run("dmesg -n 1")  # Turn off Kernel Messages to the console
+        # Turn off Kernel Messages to the console
+        self._run("dmesg -n 1")  # pylint: disable=missing-kwoa
```
jluebbe: Please disable missing-kwoa and ￼unused-argument for the whole file. The annotations are too noisy.

labgrid/remote/common.py
```diff
@@ -103,8 +103,8 @@ class Place:
     acquired = attr.ib(default=None)
     acquired_resources = attr.ib(default=attr.Factory(list))
     allowed = attr.ib(default=attr.Factory(set), convert=set)
-    created = attr.ib(default=attr.Factory(lambda: time.time()))
-    changed = attr.ib(default=attr.Factory(lambda: time.time()))
+    created = attr.ib(default=attr.Factory(lambda: time.time()))  # pylint: disable=unnecessary-lambda
```
jluebbe: This message seems correct, we should just use Factory(time.time).

labgrid/remote/common.py
```diff
@@ -16,27 +16,27 @@ def __attrs_post_init__(self):
 
     @property
     def avail(self):
-        return self.data['avail']
+        return self.data['avail']  # pylint: disable=unsubscriptable-object
```
jluebbe: Disable this for the whole file.

BastianStender:
- dropped tox related changes as advised
- moved accidental raw string transformation from `driver/serialdigitaloutput: raise error on unexpected arg in get()` to `linting: do miscellaneous linting`
- autoinstall/main: catch `Exception` instead of `KeyError` in `Handler.run()`
- replace `wait_for` with `wait_avail`
- remote/exporter: comment unused variable and add `TODO`
- driver/shelldriver: annotate whole file for pylint and remove specific annotations
- remote/common: annotate whole file for pylint and remove specific annotations
- add commit `remote/common: remove unnecessary lambda in Place` and remove `unnecessary-lambda` annotation

PR #257, opened 2018-06-04
pylintrc: adjust configuration to labgrid needs

BastianStender:
> no-member

Pretty useless, because attr is used pretty much everywhere and pylint does not recognize it. Ignore this.

> arguments-differ

Not sure if we should check this, e.g. almost every driver implements a run method with different arguments.

> useless-super-delegation

The [docs](https://labgrid.readthedocs.io/en/latest/development.html#writing-a-driver) state that this is needed:
"The minimum requirement is a call to `super().__attrs_post_init__()`." Why do we override it to call the method on super?

> no-self-use

I think wrapping functionality in methods that could be functions is still a good thing for the sake of logic. Ignore this.

> protected-access

Not sure if we should check this, there are quite some occurrences across labgrid. Maybe this is by design or we need to define exceptions in file headers for this.

> duplicate-code

Detects only by-design boilerplate code, ignore it.

> max-attributes=10

Increase this, 10 should be okay too.

> min-public-methods=0

Decrease this to fit labgrid's inheritance model.

> max-parents=10

Increase this, having a lot of parents is needed by design.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -65,7 +65,7 @@ confidence=
 # --enable=similarities". If you want to run only the classes checker, but have
 # no Warning level messages displayed, use"--disable=all --enable=classes
 # --disable=W"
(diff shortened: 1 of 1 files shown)


Emantor:
We definitely want arguments differ. This has been a source of problem in recent changes, it would greatly help to keep arguments the same across drivers.

BastianStender:
Okay, nice. I have not added it the commit because I was unsure - so no need to change. What about the other changes?

Emantor:
> no-member

We don't get around disabling this for attr.s.

> useless-super-delegation:
https://github.com/python-attrs/attrs/issues/167

attrs does not provide inheritenace class initiliazation, so we have to do it ourselves.

> no-self-use

we should at least make those @staticmethod

> protected-access

We should fix those occurences and/or provide a public interface. 

> duplicate-code

If it is only boilerplate code it sounds good, otherwise it needs to be fixed too.

IMO increasing the attribute numbers seems reasonable.

@jluebbe what is your opinion here?

jluebbe:
I agree with @Emantor here.

BastianStender:
Oh, I did not know that attrs requires explicit super delegation and had no idea Python has a `@staticmethod` decorator - nice.

Check the linting results of PR #255 [here](https://travis-ci.org/labgrid-project/labgrid/jobs/387660495#L1794). If this is okay with you, I will modify this PR accordingly.

PR #259, opened 2018-06-05
bareboxdriver: refactor login expect loop and add password support

jluebbe:
The fake console had to be extended to respond to echo commands.
Also clarify the equivalent comment in the ShellDriver.


jluebbe:
@BastianStender I've updated the pull request with the requested changes.

PR #261, opened 2018-06-08
Using imx-usb-loader for MXS devices

rohieb:
Since [recently](https://git.pengutronix.de/cgit/barebox/commit/?id=ca929ac265), imx-usb-loader can also load code into i.MX23/28 (MXS) devices.

Make it possible to override the use of mxs-usb-loader when bootstrapping a device by instantiating a IMXUSBDriver in a YAML file, like this:

``` yaml
targets:
  main:
    resources:
      RemotePlace:
        name: my-remote-place
    drivers:
      IMXUSBDriver: {}
```

### DIFF
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -708,34 +708,38 @@ class ClientSession(ApplicationSession):
         place = self.get_acquired_place()
         args = self.args.filename
         target = self._get_target(place)
+        from ..protocol.bootstrapprotocol import BootstrapProtocol
         from ..driver.usbloader import IMXUSBDriver, MXSUSBDriver
         from ..driver.openocddriver import OpenOCDDriver
         from ..resource.remote import (NetworkMXSUSBLoader, NetworkIMXUSBLoader,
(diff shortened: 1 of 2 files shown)


labgrid/remote/client.py
```diff
@@ -708,34 +708,38 @@ def bootstrap(self):
...
-                args = dict(arg.split('=', 1) for arg in self.args.bootstrap_args)
-                try:
-                    drv = target.get_driver(OpenOCDDriver)
-                except NoDriverFoundError:
-                    drv = OpenOCDDriver(target, name=None, **args)
-                drv.interface.timeout = self.args.wait
-                break
+        try:
+            drv = target.get_driver(BootstrapProtocol)
+        except:
```
BastianStender: Why are you using a bare except here? I think `NoDriverFoundError` is applicable here too.

rohieb: Indeed, I seem to have overlooked that.

rohieb:
Fixed the wording in the commit messages, imx-usb-loader fixes are still unreleased on branch upstream/next.

PR #264, opened 2018-06-11
linting: declare methods without self-use static and annotate legitimate methods

BastianStender:
Fixes this linting warning:

> no-self-use (R0201):
> Method could be a function. Used when a method doesn’t use its bound instance, and so could be written as a function.

### DIFF
--- a/labgrid/factory.py
+++ b/labgrid/factory.py
@@ -82,7 +82,8 @@ class TargetFactory:
             assert 'cls' in item
         return result
 
-    def normalize_config(self, config):
+    @staticmethod
+    def normalize_config(config):
         resources = {}
         drivers = {}
         for item in TargetFactory._convert_to_named_list(config.get('resources', {})):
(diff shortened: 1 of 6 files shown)


labgrid/remote/client.py
```diff
@@ -925,7 +925,7 @@ def create():
 
 def find_role_by_place(config, place):
     for role, role_config in config.items():
-        resources, _ = target_factory.normalize_config(role_config)
+        resources, _ = target_factory.__class__.normalize_config(role_config)
```
jluebbe: The __class__ should be unnecessary here. You can use staticmethods on instances as well.

tests/test_factory.py
```diff
@@ -54,7 +54,7 @@ def test_convert_dict(self):
         BarPort:
           name: bar
         """)
-        l = target_factory._convert_to_named_list(data)
+        l = target_factory.__class__._convert_to_named_list(data)
```
jluebbe: This shouldn't be necessary here as well.

BastianStender:
Pushed changes:
- use static methods on instances without `__class__` indirection

PR #272, opened 2018-06-25
Allow use of $LG_ENV environment variable instead of --lg-env arg

esben:
Depending on the exact workflow, it can be convenient to be able to set
an environment variabel once, and just run pytest without '--lg-env'
argument while testing.

Both --lg-env and --env-config overrules $LG_ENV, making this change
backwards compatible.

### DIFF
--- a/tests/test_fixtures.py
+++ b/tests/test_fixtures.py
@@ -1,3 +1,4 @@
+import os
 import pexpect
 import pytest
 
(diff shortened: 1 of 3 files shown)


tests/test_fixtures.py
```diff
@@ -41,6 +42,14 @@ def test_env_old_fixture(short_env, short_test):
         spawn.close()
         assert spawn.exitstatus == 0
 
+def test_env_env_fixture(short_env, short_test):
+    env=os.environ.copy()
```
rohieb: Hmm, I wonder if this is necessary. Can you just set env["LG_ENV"] without importing the outer environment? I could imagine that importing it could leak some unneeded variables that could influence the test.

esben: When no env argument is given, pexpect.spawn() executes with os.environ, so I am not introducing any leakage that wasn't there before. Also, I really do need at least some of the existing os.environ, as executing without fx. PATH really does not work as expected.

PR #273, opened 2018-06-27
Improve usability

BastianStender:
Miscellaneous changes to improve overall usability:
- do not run tests if init failed
- raise additional error on failed serial open
- show steps in GraphStrategy
- use check_output instead of check_call for more informative exceptions and debug log command as well as output

### DIFF
--- a/labgrid/pytestplugin/fixtures.py
+++ b/labgrid/pytestplugin/fixtures.py
@@ -4,6 +4,7 @@ import pytest
 
 from .. import Environment
 from ..exceptions import NoResourceFoundError
+from ..remote.client import UserError
 from ..resource.remote import RemotePlace
 
 # pylint: disable=redefined-outer-name
@@ -115,5 +116,11 @@ def env(request):
 def target(env):
     """Return the default target `main` configured in the supplied
     configuration file."""
-    target = env.get_target()
+    try:
+        target = env.get_target()
+        if target is None:
+            raise UserError("Using target fixture without 'main' target in config")
+    except UserError as e:
+        pytest.exit(e)
+
     return target
--- a/labgrid/driver/serialdriver.py
+++ b/labgrid/driver/serialdriver.py
@@ -84,7 +84,12 @@ class SerialDriver(ConsoleExpectMixin, Driver, ConsoleProtocol):
     def open(self):
         """Opens the serialport, does nothing if it is already closed"""
         if not self.status:
-            self.serial.open()
+            try:
+                self.serial.open()
+            except serial.SerialException as e:
+                raise serial.SerialException(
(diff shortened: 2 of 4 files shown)


labgrid/pytestplugin/fixtures.py
```diff
@@ -113,5 +114,9 @@ def env(request):
 def target(env):
     """Return the default target `main` configured in the supplied
     configuration file."""
-    target = env.get_target()
+    try:
+        target = env.get_target()
+    except UserError as e:
```
jluebbe: Why do you limit this to UserErrors? We probably can't recover on other errors as well.

labgrid/driver/fastbootdriver.py
```diff
@@ -43,7 +45,10 @@ def on_deactivate(self):
     @Driver.check_active
     @step(title='call', args=['args'])
     def __call__(self, *args):
-        subprocess.check_call(self._get_fastboot_prefix() + list(args))
+        cmd = self._get_fastboot_prefix() + list(args)
+        self.logger.debug(' '.join(cmd))
+        stdout = subprocess.check_output(cmd, stderr=subprocess.STDOUT)
+        self.logger.debug(stdout)
```
jluebbe: Hmm. Currently, if fastboot fails and you're not running in pytest with stdout capture, you just have the error message from fastboot on the console. That's gone with your change. Do we want to pass the output to a higher log level if an exception happens?

labgrid/pytestplugin/fixtures.py
BastianStender: Yes, that is true. Do we want to also exit on `target is None`? What's the reasoning behind `get_target()` returning `None` anyway? Shouldn't this be an exception instead? Or is returning `None` the workaround to handle a unsuccessful `get_target()` during test runtime?

labgrid/driver/fastbootdriver.py
BastianStender: Why is stdout output gone with my change? The exception should contain that, see my commit message.

BastianStender:
Pushed changes:
- add strategy/graphstrategy: do not allow unknown states in via
- add strategy/graphstrategy: raise exception on impossible paths with via

labgrid/pytestplugin/fixtures.py
jluebbe: get_target() returning None can also be treated as fatal here, but with an explicit error message. Using pytest.exit here should be ok for now, as this fixture is used very early.

BastianStender:
Pushed changes:
- drop check_output() changes
- pytestplugin/fixtures: raise UserError on `target is None` in target fixture

BastianStender:
Pushed changes:
- rebased on current master

PR #274, opened 2018-06-27
doc/development: explain default paths in graph strategy

BastianStender:

### DIFF
--- a/doc/development.rst
+++ b/doc/development.rst
@@ -317,6 +317,11 @@ A root state is a state that has no dependencies.
         def state_B(self):
             pass
 
+Graph Stategies allow to specify multiple paths to states.
+The first argument of `@GraphStrategy.depends()` becomes part of the default
+path.
+If a different path should be followed the keyword argument `via` can be used
+on `transition()`.
 
 .. code-block:: python
 
@@ -325,6 +330,11 @@ A root state is a state that has no dependencies.
         graph_strategy.transition('B')  #  returns: ['A1', 'B']
         graph_strategy.transition('B')  #  returns: []
         graph_strategy.transition('B', via=['A2'])  #  returns: ['Root', 'A2', 'B']
+
+`pytest fixtures <https://docs.pytest.org/en/latest/fixture.html>`_ can be used
+to `transition()` into the designated state before a test is executed.
+Use `transition('my_state', via=...)` to follow a (non-default) path, e.g. an
+alternative boot method.
 
 .. code-block:: python
 
--- a/doc/development.rst
+++ b/doc/development.rst
@@ -294,6 +294,7 @@ All states **HAVE TO**:
 
 Every Graph Strategy graph has to have exactly one root state.
 A root state is a state that has no dependencies.
+If `invalidate()` is overridden the super method must be called.
 
 .. code-block:: python
 


doc/development.rst
```diff
@@ -317,6 +317,8 @@ A root state is a state that has no dependencies.
         def state_B(self):
             pass
 
+The first argument of `@GraphStrategy.depends()` is the default path. If a
+different path should be followed the keyword argument `via` can specify it.
```
jluebbe: This should mention that the via argument is to .transition() and not to the .depends() decorator.

doc/development.rst
```diff
@@ -326,6 +328,10 @@ A root state is a state that has no dependencies.
         graph_strategy.transition('B')  #  returns: []
         graph_strategy.transition('B', via=['A2'])  #  returns: ['Root', 'A2', 'B']
 
+In case shortcuts can be followed after a specific state was reached
+introduce an instance variable and decide on that which path to follow. This
+should be done in a `pytest fixture <https://docs.pytest.org/en/latest/fixture.html>`_.
+
```
jluebbe: Why do you recommend to move some system state and logic from the Strategy to a fixture? Or am I misunderstanding this?

BastianStender: Sorry, that was phrased improperly. The idea is to introduce system state to the Graph Strategy and use that in the fixture to cut corners with `transition('my_tate', via=..)`. I will rephrase this.

doc/development.rst
BastianStender: True.

BastianStender:
Pushed changes:
- use semantic linefeeds
- clarify that the `via` argument is part of `.transition()` and not `.depends()`
- clarify fixture and following shortcuts
- add Graph Stategies shortcut idea

BastianStender:
Pushed changes:
- call super when overriding GraphStrategy.invalidate()

BastianStender:
Pushed changes:
- clarify that the 1st argument of `GraphStrategy.depends()` is then **part of** the default path

doc/development.rst
```diff
@@ -470,6 +470,23 @@ To remove this limitation, we should have a common way to make files available
...
+itself should be able to decide which is the best path to follow.
+
+  - Remember states that were executed
+  - Introduce a speed rating for each state (default could be `1`)
+  - Add "states that must have been executed before" argument for each state
+    (default should be an empty list)
+
+The Graph Strategy could now determine the best path (based on the speed
+rating) that is already unlocked (all required states were executed before)
+when `transition()` is called.
```
jluebbe: This would seem to require that the user transitions to some special unlock states in the correct order. That would go against the idea that the user should only specify the intended state and not need to know about how to get there.

Is the intention of this to have a way to skip repeated state transitions that take a long time?

doc/development.rst
```diff
@@ -326,6 +332,16 @@ A root state is a state that has no dependencies.
...
 
+Use `pytest fixtures <https://docs.pytest.org/en/latest/fixture.html>`_ to
+`transition()` into the designated state before a test is executed.
+Use `transition('my_state', via=...)` to follow a (non-default) path, e.g. a
+shortcut.
+This can be achieved by setting an instance variable in a state of a Graph
+Strategy (e.g. during board provisioning) and letting the fixture decide which
+path to follow depending on that variable.
+This variable must be set to its initial value in the `invalidate()` method of
+a Graph Strategy.
```
jluebbe: Hmm, I'm still against moving logic (deciding which path to take) from the strategy to a fixture. The fixture should only request a state and the strategy should find the (best) path to that state. The idea described in the next commit seems to follow that separation more cleanly. So we should just describe how to use the current graph strategy here and not how to work around it.

doc/development.rst
BastianStender: The idea is that the "unlock states" are part of the default path. Yes, skipping states is the main reason for this proposal.

doc/development.rst
BastianStender: The actual information which path to follow is currently always in the fixture, because the fixture calls `transition(.., via=..)`. There is no other logic in the GraphStrategy to follow a better path if not explicitly defined with `via=`. Or am I misunderstanding you?

BastianStender:
Pushed changes:
- drop Graph Strategy shortcut idea
- drop shortcuts as an example and (wrong) best practices

BastianStender:
@jluebbe, @fscherf: are you okay with the remaining documentation changes?

PR #276, opened 2018-06-29
drivers/quartushpsdriver: do not cache cable number

BastianStender:
The Quartus Tools work on JTAG cable numbers only. The numbers are not
stable, see this scenario:

- two devices are connected and detected
- we resolve the usb path (with jtagconfig) to JTAG cable number '2'
- the device with cable number '1' disappears
- the device originally identified by cable number '2' is now cable
  number '1'
- we (accidentally) run quartus_hps on non-existent cable number '2'

It is also possible that the wrong device is flashed, this is racy by
design. We cannot solve the raciness, but can mitigate it by not caching
the cable number.

### DIFF
--- a/labgrid/driver/quartushpsdriver.py
+++ b/labgrid/driver/quartushpsdriver.py
@@ -23,10 +23,6 @@ class QuartusHPSDriver(Driver):
         default=None,
         validator=attr.validators.optional(attr.validators.instance_of(str))
     )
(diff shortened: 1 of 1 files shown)


jluebbe:
I've moved the `self._get_cable_number()` call out of the command construction. This should make reading possible tracebacks easier.

PR #278, opened 2018-07-02
driver/power: add support for Gude Expert Power Control 8316

fblampe:
This driver adds support for the Gude Expert Power Control 8316,
using the same implementation as gude24.py, except for the port
limit of 8 and a workaround for a bug in version 1.2.0 of the
firmware.

Signed-off-by: Felix Lampe <flampe@arri.de>


fblampe:
Thanks for reviewing it so quickly!

PR #279, opened 2018-07-02
strategy/graphstrategy: pytest.exit() on strategy errors

BastianStender:
In case the strategy fails to reach the designated state for a test it
does not make sense to continue.


BastianStender:
This solution is only valid if the GraphStrategy is used by pytest. That is not always the case (e.g. with `labgrid-client`). I will leave this PR open for reference and discussion.

jluebbe:
As discussed internally, a better approach would be to handle this at the pytest level, either in a fixture or possibly in the labgrid plugin.

PR #280, opened 2018-07-03
strategy/graphstrategy: add force kwarg to transition()

BastianStender:
Calling `transition(.., force=True)` forces the Graph Strategy to
execute all states on the path from root node to the designated state.


BastianStender:
As internally discussed this does not fit the Graph Strategy. I will use `invalidate()` instead.

PR #281, opened 2018-07-03
driver/fastbootdriver: add run argument

strumtrar:
The fastboot protocol defines a "oem" command, that allows vendor specific
subcommands. barebox supports "exec", which executes arbitrary shell commands.

Add support for this to the fastbootdriver.

Signed-off-by: Steffen Trumtrar <s.trumtrar@pengutronix.de>

### DIFF
--- a/labgrid/driver/fastbootdriver.py
+++ b/labgrid/driver/fastbootdriver.py
@@ -60,6 +60,11 @@ class AndroidFastbootDriver(Driver):
         self("flash", partition, filename)
 
     @Driver.check_active
+    @step(args=['cmd'])
+    def run(self, cmd):
+        self("oem", "exec", "--", cmd)
+
+    @Driver.check_active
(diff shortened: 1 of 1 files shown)


labgrid/driver/fastbootdriver.py
```diff
@@ -59,6 +59,11 @@ def flash(self, partition, filename):
         check_file(filename, command_prefix=self.fastboot.command_prefix)
         self("flash", partition, filename)
 
+    @Driver.check_active
+    @step(args=['cmd'])
+    def run(self, cmd):
+        self("oem exec -- ", cmd)
```
jluebbe: The fixed string should be split into individual arguments, otherwise it will likely fail when used locally instead of via ssh.

strumtrar:
@jluebbe Updated according to review feedback

PR #284, opened 2018-07-05
docker: Fixes #260. Integrate separate docker repo into labgrid repo

### LINKED ISSUE #260: Automatic regression test via Docker Hub
A few times we have experienced that Labgrid got broken for our use. Thus we came up with this idea.

Setup a docker hub project like this one https://hub.docker.com/r/deifwpt/labgrid-client/~/settings/automated-builds/ which is linked to a repository.

This is our own, but it is only linked to our own repository where we put our Dockerfile. If the Docker file is moved to labgrid, then any changes in labgrid/master will cause a new docker build.

This docker build, can then be used to do nightly regression tests in our CI system.

We would not use the image for our "production" tests. We would setup a separate docker image and manually rebuild that when we need the latest labgrid.

This of course means that the dockerfiles made by https://github.com/krevsbech  would be moved to labgrid

krevsbech:
As proposed in issue #260 integrate the docker images into the project.
This enable regression testing of the python setup files as docker hub can
be configured to automatically build upon commit in the Labgrid repo.

Signed-off-by: Kasper Revsbech <krev@triax.com>

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,16 @@
+FROM python:3.6
+MAINTAINER krev@triax.com
+
+# Tools
+RUN apt-get update && \
+    apt-get install -y ser2net && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
+
+WORKDIR /opt/labgrid
+
+# Ensure that labgrid is build from current dir
+COPY ./ ./
+RUN pip install crossbar
+RUN pip install -e /opt/labgrid
+RUN git clone https://github.com/vishnubob/wait-for-it.git /opt/wait-for-it
--- a/docker/README.rst
+++ b/docker/README.rst
@@ -1,0 +1,66 @@
+Labgrid Docker files
+====================
+
+Purpose
+-------
+This folder docker files to build one shared Docker image for the following Labgrid instances/containers.
+In staging the a docker-compose file describe how to use the image to start up the following instances:
+
+- **labgrid coordinator** (Crossbar) This instance simply configures a Labgrid coordinator instance.
+- **labgrid client** This instance contains the Labgrid client tools.
+- **labgrid exporter** This instance contains the Labgrid exporter tools.
+
+
+Desired architecture
+--------------------
+The docker image in this repository is thought to be used in a distributed setting, where a coordinator, client and exporter are utilized, as described in the `docs <https://labgrid.readthedocs.io/en/latest/getting_started.html#setting-up-the-distributed-infrastructure>`_
+
+Build
+-----
+The docker image must be build from the repository root as it copy in the labgrid code, hence in order to buiĺd from "./labgrid/":
+
+  .. code-block:: bash
+
+     $ docker build -f docker/Dockerfile .
+
+Staging - Test environment
+--------------------------
+In *staging* a docker compose based staging environment resides. The staging environment contains a coordinator, an exporter and a client instance. Moreover a simple example configuration is provided to expose the USB serial on the Host. **Note** It is assumed that /dev/ttyUSB0 is available on the host to run the staging. It is also assumed that the serial is connected to the DUT.
+
+Each container can be started individually with the appropriate debug TTY attached:
+
+Coordinator
+...........
+
+  .. code-block:: bash
+
+     $ docker-compose up coordinator
+
+Exporter
+........
+
+  .. code-block:: bash
+
+     $ docker-compose up exporter
+
+Client
+......
+
+  .. code-block:: bash
+
+     $ docker-compose up -d client
+
+As the client require an interactive shell, this can be achieved as described below, where an interactive shell is opened and the client is used to list the resources seen be the coordinator:
+
+  .. code-block:: bash
+
+     $ docker-compose exec client /bin/bash
+     $ labgrid-client -x ws://coordinator:20408/ws resources
+
+If debug TTY's are not needed the staging environment is simply started by:
+
+  .. code-block:: bash
+
+     $ docker-compose up -d
+
+The client can still be accessed by exec as described above.
--- a/docker/staging/docker-compose.yml
+++ b/docker/staging/docker-compose.yml
@@ -1,0 +1,43 @@
+version: '3.3'
+services:
+  coordinator:
+    build:
+      context: "../../"
+      dockerfile: "docker/Dockerfile"
+    networks:
+      - labgrid_net
+    tty: true
+    command: "crossbar start"
+  client:
+    build:
+      context: "../../"
+      dockerfile: "docker/Dockerfile"
+    networks:
+      - labgrid_net
+    tty: true
+    stdin_open: true
+    command: "/bin/bash"
+  exporter:
+    build:
+      context: "../../"
+      dockerfile: "docker/Dockerfile"
+    volumes:
+      - "./exporter-conf:/opt/exporter-conf"
+    devices:
+      - "/dev/ttyUSB0:/dev/ttyUSB0"
+    networks:
+      - labgrid_net
+    command: /opt/exporter-conf/entrypoint.sh
+    depends_on:
+      - coordinator
+    tty: true
+    stdin_open: true
+    privileged: true
+
+networks:
+  labgrid_net:
+    driver: bridge
+    driver_opts:
+      com.docker.network.bridge.name: labgrid_bridge
+    ipam:
+      driver: default
--- a/docker/staging/exporter-conf/config.yaml
+++ b/docker/staging/exporter-conf/config.yaml
@@ -1,0 +1,6 @@
+default-group:
+  location: default
+  NetworkSerialPort:
+    host: exporter
+    port: 3001
+    speed: 115200
--- a/docker/staging/exporter-conf/entrypoint.sh
+++ b/docker/staging/exporter-conf/entrypoint.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+SCRIPTPATH="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+ser2net -c ${SCRIPTPATH}/ser2net.conf
+# wait-for-it is a simple script that waits until the coordinator service is up before starting the exporter
+/opt/wait-for-it/wait-for-it.sh coordinator:20408 -- labgrid-exporter -n default -x ws://coordinator:20408/ws ${SCRIPTPATH}/config.yaml
--- a/docker/staging/exporter-conf/ser2net.conf
+++ b/docker/staging/exporter-conf/ser2net.conf
@@ -1,0 +1,1 @@
+3001:telnet:0:/dev/ttyUSB0:115200


docker/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM python:3.6
+
```
esben: What about a MAINTAINER statement?

docker/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+# Ensure that labgrid is build from current dir
+COPY ./ ./
+RUN pip install crossbar
+RUN pip install -e /opt/labgrid
+RUN git clone https://github.com/vishnubob/wait-for-it.git /opt/wait-for-it
+
+RUN echo "stty rows 33 cols 204" >> /etc/bash.bashrc
+
+
```
esben: Remove these two trailing blank lines.

docker/staging/docker-compose.yml
```diff
@@ -0,0 +1,45 @@
+version: '3.3'
+services:
+  coordinator:
+    build:
+      context: "../../"
+      dockerfile: "docker/Dockerfile"
+    networks:
+      - labgrid_net
+    tty: true
+    entrypoint: "crossbar start"
```
esben: Why not use `command` instead, so it is possible to start image with alternative command (for debugging purposes)?

docker/staging/docker-compose.yml
```diff
@@ -0,0 +1,45 @@
...
+  client:
+    build:
+      context: "../../"
+      dockerfile: "docker/Dockerfile"
+    networks:
+      - labgrid_net
+    tty: true
+    stdin_open: true
+    entrypoint: "/bin/bash"
```
esben: Why use `entrypoint` instead of `command` here?

docker/staging/docker-compose.yml
```diff
@@ -0,0 +1,45 @@
...
+    build:
+      context: "../../"
+      dockerfile: "docker/Dockerfile"
+    volumes:
+      - "./exporter-conf:/opt/exporter-conf"
+    devices:
+      - "/dev/ttyUSB0:/dev/ttyUSB0"
+    networks:
+      - labgrid_net
+    entrypoint: /opt/exporter-conf/entrypoint.sh
```
esben: As above. I think it would better with `command`.

docker/staging/docker-compose.yml
```diff
@@ -0,0 +1,45 @@
...
+
+networks:
+  labgrid_net:
+    driver: bridge
+    driver_opts:
+      com.docker.network.bridge.name: labgrid_bridge
+    ipam:
+      driver: default
+      config:
+        - subnet:  172.10.238.0/24
```
esben: Ouch, would it be possible to avoid hardcoding IP addresses?  It can conflict with existing use of private IP addresses (the company I work for uses this range internally...).

krevsbech:
Please review again, as fixup has been conducted answering review from @esben

Emantor:
Is it really necessary to add the docker file to the labgrid repostiroy? Can't we simply add a webhook which triggers a build of the docker images via the labgrid repostiroy on push?
What is the distinct advantage we gain from adding the Dockerfile to the repository?
ATM @krevsbech modify the docker repo at will, while he has no access to the main labgrid repository.
Providing a webhook which triggers an automated build should not be a problem.

I think providing a webhook to a labgrid docker hub automated build may be the main thing to add here, other automated builds can than depend on that "main" docker hub automated build.

esben:
Having the Dokerfile in the labgrid repository makes it much simpler to fork the whole toolchain of modifying labgrid and have a Docker image built automatically.

Having Dockerfile in repository also makes it much more likely that a Docker image can be built with an old version. When (not if) changes are made to labgrid that requires changes to Dockerfile, it is crucial to have them in same repository.

Finally, having Dockerfile in labgrid projects will allow future work of using that as an integrated part of labgrid CI, so that changes that break Docker images are caught before they are merged.

Keeping Dockerfile out of main repo just not gain you anything positive, other than giving different people merge power over different parts of the project (which is IMHO not entirely a good thing). Keep in mind that anybody will able to change labgrid, including the Dockerfile.  Just fork it and do whatever you like.  It is open source after all :)

docker/staging/docker-compose.yml
```diff
@@ -0,0 +1,43 @@
...
+    build:
+      context: "../../"
+      dockerfile: "docker/Dockerfile"
+    volumes:
+      - "./exporter-conf:/opt/exporter-conf"
+    devices:
+      - "/dev/ttyUSB0:/dev/ttyUSB0"
+    networks:
+      - labgrid_net
+    command: /opt/exporter-conf/entrypoint.sh
```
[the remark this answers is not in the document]
krevsbech: As I see it, staging is an example setup that can be used for inspiration of deployment and sanity test. Hence I don't aim to make i to generic, moreover I think the entrypoint.sh is tightly coupled to the conf file. If one want to create custom he/she most properly have to change all three files

docker/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM python:3.6
+MAINTAINER krev@triax.com
+
+# Tools
+RUN apt-get update && \
+    apt-get install -y ser2net
```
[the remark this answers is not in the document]
krevsbech: Thanks for the tip :)

docker/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+
+WORKDIR /opt/labgrid
+
+# Ensure that labgrid is build from current dir
+COPY ./ ./
+RUN pip install crossbar
+RUN pip install -e /opt/labgrid
+RUN git clone https://github.com/vishnubob/wait-for-it.git /opt/wait-for-it
+
+RUN echo "stty rows 33 cols 204" >> /etc/bash.bashrc
```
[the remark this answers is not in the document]
krevsbech: You are right this is strange. This is because the inhered images are configured for very small terminals, making interactive sessions cumbersome, but this is a hack that I will remove

krevsbech:
> Having the Dokerfile in the labgrid repository makes it much simpler to fork the whole toolchain of modifying labgrid and have a Docker image built automatically.
> 
> Having Dockerfile in repository also makes it much more likely that a Docker image can be built with an old version. When (not if) changes are made to labgrid that requires changes to Dockerfile, it is crucial to have them in same repository.
> 
> Finally, having Dockerfile in labgrid projects will allow future work of using that as an integrated part of labgrid CI, so that changes that break Docker images are caught before they are merged.
> 
> Keeping Dockerfile out of main repo just not gain you anything positive, other than giving different people merge power over different parts of the project (which is IMHO not entirely a good thing). Keep in mind that anybody will able to change labgrid, including the Dockerfile. Just fork it and do whatever you like. It is open source after all :)

For me the main motivation is the one that breaking changes will be found pre pr, and that we can even define that a successful docker build is a part of your PR checks. Then as @esben  mention there is a description of how to "run" the framework that is versioned along the framework. 

Regarding the maintenance your are right that I don't have a commit bit in the main repository and should not have one. But as a all PR's are public I can still review and so can other people with an "opinion" as @esben

Emantor:
Would git submodules work for your approach? Does docker hub support the necessary commands for that?
That way we could create a coupling of the two repositories, add the build support for docker hub, but you could still independently manage the docker files.

docker/staging/docker-compose.yml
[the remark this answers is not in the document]
esben: True, the current entrypoint.sh is currently coupled with the configuration. So it is fair to keep it there (for now).
Looking forward, I think it would make sense to make this image generic so that it can be used together with custom configuration files mounted to a config path in the image.

esben:
Although I don't belong to the git-submodules-are-evil crowd (I am using submodules a lot for many purposes), but I think it is a bit like using a sledgehammer to crack a nut to use a submodule to manage a handful of small text files like this.

Is it really considered a big problem or burden to carry the Docker file, a small shell script and a couple of example configuration files in this repo?

Emantor:
I think your are doing a great job of maintaining the Dockerfiles and I don't want to impede your workflow by introducing a PR merge middle man (me or another maintainer) who can't comment on possible technical problems since he/she does not use the Dockerfiles or Docker itself.

esben:
@Emantor Please keep in mind that those that actually does use Docker (together with labgrid) would benefit from having it in this repo, and the benefit IMHO does outweigh the possible impediment introduced to the workflow.

I don't think it is supposed to be a lot of maintenance to keep the Dockerfile (and related files).  They will hopefully only need to be modified occasinally.

krevsbech:
> I think your are doing a great job of maintaining the Dockerfiles and I don't want to impede your workflow by introducing a PR merge middle man (me or another maintainer) who can't comment on possible technical problems since he/she does not use the Dockerfiles or Docker itself.

That is a fair concern. However, I see a number of actors that can qualify docker related changes. Hence i see the blocking risk as of less concern than the upsides of having the files in repo. But in the end it is your call. If you still believe it is not the right place to put the files I will close the PR and migrate the proposed changes to the dedicated docker repo.

esben:
@Emantor: For what it is worth, I will be helpful with reviewing Docker related PR's here.

docker/README.rst
```diff
@@ -0,0 +1,66 @@
...
+- **labgrid exporter** This instance contains the Labgrid exporter tools.
+
+
+Desired architecture
+--------------------
+The docker image in this repository is thought to be used in a distributed setting, where a coordinator, client and exporter are utilized, as described in the `docs <https://labgrid.readthedocs.io/en/latest/getting_started.html#setting-up-the-distributed-infrastructure>`_
+
+Build
+-----
+The docker image must be build from the repository root as it copy in the labgrid code, hence in order to buiĺd from "./labgrid/":
```
jluebbe: s/copy in/copies/
Also the l in build has an accent.

docker/README.rst
```diff
@@ -0,0 +1,66 @@
...
+
+  .. code-block:: bash
+
+     $ docker build -f docker/Dockerfile .
+
+Staging - Test environment
+--------------------------
+In *staging* a docker compose based staging environment resides. The staging environment contains a coordinator, an exporter and a client instance. Moreover a simple example configuration is provided to expose the USB serial on the Host. **Note** It is assumed that /dev/ttyUSB0 is available on the host to run the staging. It is also assumed that the serial is connected to the DUT.
```
jluebbe: How would this be used in practice, if I need to use different exported devices? Would I write a new Dockerfile based on this image or a different docker-compose file?
Also, please use semantic newlines (see https://labgrid.readthedocs.io/en/latest/development.html#documentation).

docker/staging/docker-compose.yml
```diff
@@ -0,0 +1,43 @@
...
+    depends_on:
+      - coordinator
```
jluebbe: Hmm, doesn't the client depend on the coordinator, too?

docker/staging/docker-compose.yml
```diff
@@ -0,0 +1,43 @@
+version: '3.3'
+services:
+  coordinator:
+    build:
+      context: "../../"
+      dockerfile: "docker/Dockerfile"
+    networks:
+      - labgrid_net
+    tty: true
+    command: "crossbar start"
```
jluebbe: The coordinator will write persistent data to it's working directory (a yaml file of the configured places). Is that kept over docker container restarts?

docker/staging/exporter-conf/config.yaml
```diff
@@ -0,0 +1,6 @@
+default-group:
+  location: default
+  NetworkSerialPort:
```
jluebbe: Could you use a USBSerialPort instead? labgrid-exporter would then manage the ser2net process itself.

jluebbe:
With the point that having the Docker file in this repo makes it easier to keep the same workflow when forking the repo and ensures that the Dockerfile matches the labgrid version, I'm not against going this way. I've asked some questions above so I can understand better how this is going to work and will be used.

docker/README.rst
esben: You could either create derived images with the required configurations, or you could map in the configuration using docker volumes.

docker/staging/docker-compose.yml
esben: Yes, it should.

krevsbech:
Replaced by #320

PR #286, opened 2018-07-10
Fix spellcheck and missing timeout in sshdriver

kjeldflarup:
Fix errors introduced with commit https://github.com/labgrid-project/labgrid/commit/1f89ed9f4e10b0773870105ee31df8a208593121

Signed-off-by: Kjeld Flarup <kfa@deif.com>


labgrid/driver/sshdriver.py
```diff
@@ -101,9 +101,9 @@ def _check_master(self):
     @Driver.check_active
     @step(args=['cmd'], result=True)
     def run(self, cmd, codec="utf-8", decodeerrors="strict", timeout=None): # pylint: disable=unused-argument
-        return self._run(cmd, codec=codec, decodererrors=decodeerrors)
+        return self._run(cmd, codec=codec, decodeerrors=decodeerrors)
```
esben: Shouldn't we pass on the timeout value to `_run()` here?

labgrid/driver/sshdriver.py
```diff
@@ -101,9 +101,9 @@ def _check_master(self):
...
 
-    def _run(self, cmd, codec, decodeerrors): # pylint: disable=unused-argument
+    def _run(self, cmd, codec="utf-8", decodeerrors="strict", timeout=None): # pylint: disable=unused-argument
```
esben: If timeout value is passed on from `run()` to `_run()`, I don't see why any default values are needed here.

labgrid/driver/sshdriver.py
kjeldflarup: Timeout is actually ignored. commandmixin just happen to pass it to sshdriver.
The new is that commandmixin now calls _run instead of run, so actually it could be removed here, but that would not be backwards compatible.

labgrid/driver/commandmixin.py
```diff
@@ -34,14 +34,14 @@ def _run_check(self, cmd: str, timeout=30):
         Returns:
             List[str]: stdout of the executed command
         """
-        stdout, stderr, exitcode = self._run(cmd, timeout=timeout)
+        stdout, stderr, exitcode = self._run(cmd, timeout=timeout, codec=codec, decodeerrors=decodeerrors)
```
BastianStender: Where are `codec` and `decodeerrors` supposed to come from? I guess you confused different methods here.

BastianStender: Where are the variables `codec` and `decodeerrors` supposed to come from? I guess you missed updating
`def _run_check(self, cmd: str, timeout=30):` with these arguments, right?

BastianStender: You should either transform the `_run()` method of `BareboxDriver`, `UBootDriver` and `SmallUBootDriver` or go without CommandMixin's convenience function `run_check` on targets with decode errors.

kjeldflarup: You're right. I hope that these run functions will be stable soon and not have more special options.
I will add these but as unused parameters, because I don't see any use for codec's in these drivers. 
Better test functions would help a lot. In my PR for the remote_tmpdir I actually have a test case which covers the SSH driver better.

BastianStender:
I fixed this while already on it:
- driver/commandmixin: fix missing arguments and force kwarg for `timeout`, `codec`, `decodeeror`
- driver/sshdriver: adjust remotepath kwarg in put()
- use same `_run()` signature everywhere

kjeldflarup:
Thanks @BastianStender

BastianStender:
Squashed the fixup. @jluebbe could you review/merge this?

PR #299, opened 2018-08-27
Feature Flags

### LINKED ISSUE #249: Feature: Selecting test cases based on available hardware/fixtures
To make the test environment really scalable, it would be great, if I could define for each test case, which dependencies it has. For example, when having a complete rack with switchable power supplies, IO boards and externel peripherals, I could run all tests. When running the tests on the desk, without having the full setup, I would define in a config, which HW I have and hereby filter the tests, that are executed (automatically).

Emantor:
- [x] Docs
- [x] Tests

Fixes #249

### DIFF
--- a/doc/usage.rst
+++ b/doc/usage.rst
@@ -301,6 +301,69 @@ For this example, you should get a report similar to this:
 
   ========================== 3 passed in 29.77 seconds ===========================
 
+Feature Flags
+~~~~~~~~~~~~~
+Labgrid includes support for feature flags on a global and target scope.
+They will be concatenated and compared to a pytest mark on the test to decide
+whether the test can run with the available features.::
+
+   import pytest
+
+   @pytest.mark.lg_feature("camera")
+   def test_camera(target):
+      [...]
+
+together with an example environment configuration:
+
+.. code-block:: yaml
+
+  targets:
+    main:
+      features:
+        - camera
+      resources: {}
+      drivers: {}
+
+would run the above test, however the following configuration would skip the
+test because of the missing feature:
+
+.. code-block:: yaml
+
+  targets:
+    main:
+      features:
+        - console
+      resources: {}
+      drivers: {}
+
+This is also reported in the pytest execution as a skipped test with the reason
+being the missing feature.
+
+For tests with multiple required features, pass them as a list to pytest:::
+
+   import pytest
+
+   @pytest.mark.lg_feature(["camera", "console"])
+   def test_camera(target):
+      [...]
+
+Features do not have to be set per target, they can also be set via the global
+features key:
+
+.. code-block:: yaml
+
+  features:
+    - camera
+  targets:
+    main:
+      features:
+        - console
+      resources: {}
+      drivers: {}
+
+This yaml would combine both the global and the target features.
+
+
 Test Reports
 ~~~~~~~~~~~~
 
--- a/labgrid/config.py
+++ b/labgrid/config.py
@@ -196,3 +196,6 @@ class Config:
             images[name] = self.resolve_path(image)
 
         return images
+
+    def get_features(self):
+        return set(self.data.get('features', {}))
--- a/labgrid/environment.py
+++ b/labgrid/environment.py
@@ -43,6 +43,15 @@ class Environment:
 
         return self.targets[role]
 
+    def get_features(self):
+        return self.config.get_features()
+
+    def get_target_features(self):
+        flags = set()
+        for target, value in self.config.get_targets().items():
+            flags = flags | set(value.get('features', {}))
+        return flags
+
     def cleanup(self):
         for target in self.targets:
             self.targets[target].cleanup()
--- a/labgrid/pytestplugin/reporter.py
+++ b/labgrid/pytestplugin/reporter.py
@@ -1,12 +1,10 @@
 import logging
 import sys
 import pytest
-import os
 from _pytest.capture import safe_text_dupfile
 
 from .. import Environment
 from ..step import steps
-from ..consoleloggingreporter import ConsoleLoggingReporter
 
 logging.basicConfig(
     level=logging.INFO,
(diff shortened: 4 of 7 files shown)


labgrid/resource/udev.py
```diff
@@ -292,44 +292,24 @@ class USBSDMuxDevice(USBResource):
     def __attrs_post_init__(self):
         self.match['ID_VENDOR_ID'] = '0424'
         self.match['ID_MODEL_ID'] = '4041'
+        self.control_path = None
```
jluebbe: This commit seems to be in the wrong PR.

labgrid/config.py
```diff
@@ -196,3 +196,6 @@ def get_images(self):
             images[name] = self.resolve_path(image)
 
         return images
+
+    def get_feature_flags(self):
+        return self.data.get('feature flags', [])
```
jluebbe: No spaces here, just use 'features' instead.

labgrid/environment.py
```diff
@@ -43,6 +43,12 @@ def get_target(self, role: str = 'main') -> Target:
 
         return self.targets[role]
 
+    def get_feature_flags(self):
+        return self.config.get_flags()
+
+    def get_target_feature_flags(self, name):
+        return self.config.get_targets().get(name).get('feature flags')
```
jluebbe: No spaces here as well.

labgrid/environment.py
```diff
@@ -43,6 +43,12 @@ def get_target(self, role: str = 'main') -> Target:
 
         return self.targets[role]
 
+    def get_feature_flags(self):
+        return self.config.get_flags()
```
jluebbe: This is broken at this point in the series. Also, 'environment: return flags for all targets' changes this code completely.

tests/test_fixtures.py
```diff
@@ -43,12 +43,14 @@ def test_env_old_fixture(short_env, short_test):
         assert spawn.exitstatus == 0
 
 def test_env_env_fixture(short_env, short_test):
-    env=os.environ.copy()
-    env['LG_ENV'] = short_env
+    env=os.environ
+    env['LG_ENV'] = str(short_env)
+    print(env)
```
jluebbe: Leftover debug print? Please also describe what was broken and how this change fixes the problem.

labgrid/pytestplugin/reporter.py
```diff
@@ -33,6 +35,26 @@ def pytest_configure(config):
         logging.getLogger().setLevel(logging.DEBUG)
     if lg_log:
         ConsoleLoggingReporter(lg_log)
+    env_config = config.option.env_config
```
jluebbe: Please describe why this change is needed.

labgrid/environment.py
```diff
@@ -44,10 +44,15 @@ def get_target(self, role: str = 'main') -> Target:
...
-        return self.config.get_targets().get(name).get('feature flags')
+        return self.config.get_feature_flags()
+
+    def get_target_feature_flags(self):
+        flags = []
+        for target, value in self.config.get_targets().items():
+            flags += value.get('feature flags', [])
+            if flags:
+                flags = list(set(flags))
+        return flags
```
jluebbe: Why not keep the flags as a set from the beginning (and also return a set)?

doc/usage.rst
```diff
@@ -296,6 +296,52 @@ For this example, you should get a report similar to this:
...
+Feature Flags
+~~~~~~~~~~~~~
+Labgrid includes support for feature flags on a global and target scope.
+They will be concatenated and compared to a pytest mark on the test to decide
+whether this combination does support the flags.::
+
+   import pytest
+
+   @pytest.mark.lg_feature("camera")
+   def test camera(target):
```
jluebbe: Replace ' ' in the function name with '_'.

doc/usage.rst
```diff
@@ -296,6 +296,52 @@ For this example, you should get a report similar to this:
 
   ========================== 3 passed in 29.77 seconds ===========================
 
+Feature Flags
+~~~~~~~~~~~~~
+Labgrid includes support for feature flags on a global and target scope.
+They will be concatenated and compared to a pytest mark on the test to decide
+whether this combination does support the flags.::
```
jluebbe: "whether the test can run with the available features"?

Emantor:
I pushed a new version which should address your review comments.

I merged several environments commits and dropped the fix for the environment fixture test.

labgrid/environment.py
jluebbe: This is not yet addressed. Also, the commit message no longer matches the method names.

labgrid/config.py
```diff
@@ -196,3 +196,6 @@ def get_images(self):
             images[name] = self.resolve_path(image)
 
         return images
+
+    def get_features(self):
```
jluebbe: The commit message doesn't match the method name.

labgrid/pytestplugin/reporter.py
```diff
@@ -21,6 +21,40 @@ def bold(text):
...
+
+    have_feature = set(env.get_features() + env.get_target_features())
+
+    for item in items:
+        want_feature = []
+        marker = item.get_marker("lg_feature")
+        if marker:
+            arg = marker.args[0]
+            if isinstance(arg, str):
+                want_feature.append(arg)
```
jluebbe: As you're not looping here, you can just assign want_feature directly. No need to create it as a empty list first.

labgrid/pytestplugin/reporter.py
```diff
@@ -21,6 +21,40 @@ def bold(text):
...
+            elif isinstance(arg, list):
+                want_feature += arg
+            else:
+                raise Exception("Unsupported feature argument type")
+        missing_feature = []
+        for feature in want_feature:
+            if feature not in have_feature:
+                missing_feature.append(feature)
```
jluebbe: You can simplify this to set operations.
```
>>> {1,2,3} - {1,2}
{3}
```

labgrid/pytestplugin/reporter.py
```diff
@@ -21,6 +21,40 @@ def bold(text):
...
+        if len(missing_feature) > 0:
```
jluebbe: Drop the len().

labgrid/pytestplugin/reporter.py
```diff
@@ -21,6 +21,40 @@ def bold(text):
 def under(text):
     return "\033[4m{}\033[0m".format(text)
 
+@pytest.hookimpl()
+def pytest_collection_modifyitems(config, items):
```
jluebbe: As this has nothing to do with the reporter, this should go into a separate file. Maybe `collection.py`?

Emantor: Maybe its a good idea to move the `pytest_configure` function as well into the __init__.py?

labgrid/pytestplugin/reporter.py
Emantor: I need to initialize the variable, otherwise I'll get an error while calculating the missing features.

labgrid/pytestplugin/reporter.py
Emantor: Done.

labgrid/config.py
Emantor: Fixed.

labgrid/pytestplugin/reporter.py
Emantor: Fixed.

Emantor:
Another update:
- [x] fix usage rst examples
- [x] move to set operations for the feature calculation
- [x] operate on sets in the environment
Please review again.

PR #301, opened 2018-08-29
Docker integration

dutsii:
Added Docker support


labgrid/util/dict.py
```diff
@@ -37,3 +37,19 @@ def filter_dict(d, cls, warn=False):
...
+        d (dict): The dictionary to recursively search through
+        key (str): The key to search for
+    """
+    if key in d:
+        return d[key]
+    for k, v in d.items():
+        if isinstance(v,dict):
+            value = find_dict(v,key)
+            if value is not None:
+                return value
```
krevsbech: Minor: Add missing new line

dutsii:
Fixed according to review

labgrid/driver/dockerdriver.py
```diff
@@ -0,0 +1,86 @@
...
+import logging
+
+from labgrid.factory import target_factory
+from labgrid.driver.common import Driver
+from labgrid.resource.docker import DockerDaemon, DockerConstants
+
+
+@target_factory.reg_driver
+@attr.s(cmp=False)
+class DockerDriver(Driver):
```
Emantor: This Driver does not implement any of the Protocols.
It should implement at least one, I guess the PowerProtocol would be a good fit. This would allow one of the example strategies to start and stop the container.
If none of the available protocols fit this use case, please provide a valid use case and outline an example protocol.

Emantor:
I think an example under the examples/directory with a README on how to setup the example would be great to understand how the Driver works.

labgrid/resource/docker.py
```diff
@@ -0,0 +1,127 @@
...
+
+@target_factory.reg_resource
+@attr.s(cmp=False)
+class DockerNetworkService(ManagedResource, NetworkService):
+    """
+    The docker network service is a managed resource mapping a container name to a network service.
+    """
+    manager_cls = DockerManager
+    container_name = attr.ib(default=None, validator=attr.validators.instance_of(str))
+    address = attr.ib(default="", validator=attr.validators.in_([""]))
```
jluebbe: Hmm, you're overriding the address attribute from `NetworkService` here.
I think this could work more cleanly by using the same approach we have in the `RemotePlaceManager`/`RemotePlace` and related resources:
You wouldn't define `DockerNetworkServices` explicitly, but the `DockerManager` would create `NetworkService` resources as needed and keep them up-to-date (see how `RemotePlaceManager` creates resources dynamically with `make_resource()`). This way, you could also keep the `avail` status of the `NetworkService` correct as the container is started and stopped.

labgrid/driver/dockerdriver.py
```diff
@@ -0,0 +1,106 @@
+import attr
+import docker
```
jluebbe: Please import this where needed so labgrid also works without the docker package installed (and also keep the startup time lower).

trxanga:
Here at Triax we would like to resume the process of merging
DockerDriver into mainstream labgrid.  Therefore the
PR from August 2018 has been updated such that:

1. Activating the DockerDriver and setting it in (PowerProtocol) state
   `on` will cause creation and activation of a NetworkResource.
2. The necessary `import docker` statements are all executed at the
   very last moment. Therefore it is safe to start up labgrid in an
   environment missing the python docker module (provided, of course,
   that one is not going to actually *use* the docker functionality).
3. Directory `examples/docker` contains a small demo of how to use the
   driver - with companion readme file.  The demo is given a standard
   pytest setup with conftest.py and env.yaml files.
4. A new test for tox is added in `tests/test_docker.py`. The
   test carries out the same steps as the demo, but now with config
   files "embedded" in `test_docker.py`. Besides the test will be
   disabled if docker is absent on the execution host - following the
   pattern used to test the sigrok driver.
5. Additionally `tests/test_docker.py` contains a test where all
   dependencies on docker plus the python socket and time modules have
   been mocked out. This test case will always run - regardless of the
   host - and provides good coverage of the contributed docker driver
   code.
6. Extended explanation of the DockerDriver is available in
   `configuration.rst`.

Note: If using SSHDriver to log in to the created docker container, a
warning message will be spit out at first contact:

    mesg: ttyname failed: Inappropriate ioctl for device

This happens when SSHDriver creates the SSH master connection, maybe
because the connection is not created within an ordinary shell in the
container. The message is not really related to the labgrid docker
code and has no effect on the outcome of tests. The message can be
choked with:

    diff --git a/labgrid/driver/sshdriver.py b/labgrid/driver/sshdriver.py                                                                                                                             
    index 99e48af..1fcec4f 100644                                                                                                                                                                      
    --- a/labgrid/driver/sshdriver.py                                                                                                                                                                  
    +++ b/labgrid/driver/sshdriver.py                                                                                                                                                                  
    @@ -60,7 +60,7 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):                                                                                                    
             env = os.environ.copy()                                                                                                                                                                   
                 env['SSHPASS'] = self.networkservice.password                                                                                                                                         
    -        self.process = subprocess.Popen(args, env=env)                                                                                                                                            
    +        self.process = subprocess.Popen(args, env=env, stderr=open("/dev/null", "w"))                                                                                                             
                                                                                                                                                                                                       
             try:                                                                                                                                                                                      
                 if self.process.wait(timeout=30) != 0:

But that's probably not a good idea to do in general, right?

labgrid/resource/docker.py
dutsii: See comments from trxanga

labgrid/driver/dockerdriver.py
dutsii: See comments from trxanga

jluebbe:
@trxanga Thanks for the update, this looks very good. I've had a quick look at the code an have a few high-level questions.

The configuration parameters are split between the DockerDaemon resource (docker_daemon_url) and the DockerDriver (image_uri, container_name, host_config, network_services, ...). Is there a specific reason for this specific split?
An alterative might be to call rename the resource to DockerContainer and move the related options there? I'm not sure if this is a better design, though.

Regarding the Strategy: It might help to let the strategy deactivate the SSHDriver before switching to the off state, as that should give it time to clean up.

In the _update_resource() method, you try to connect to the remote socket. Isn't this to much overhead?

trxanga:
@jluebbe 

*Q1: Why not gather together all docker config options (e.g.                                                                                                                                                 
in new DockerContainer)?*

It is correct that the docker driver - as illustrated in
`examples/docker` and `tests/test_docker.py` - currently only
communicates with one docker daemon that resides at the host machine.
So it could make sense to combine the config parameters.

But already now the code is able to communicate with more than one
docker daemon. This ability is not presently used, but we plan to use
it in the near future.

The plan is to equip a remote PC with special test HW and then make
the HW accessible from within a docker container running at that PC.
The gain is easier maintenance: If only the driver is present and the
PC runs a docker daemon, it's straightforward to execute tests on it
even if other testing activities are carried out while the PC is not
used by labgrid. (The special test HW is expensive!).

So we would like to keep the full freedom of communicating with any
number of docker daemons while - for each daemon - using any image
that we please. And for such purposes it is ideal to separate
"pointers" to daemons (DockerDaemon config) from what images each such
daemon should run (DockerDriver config).

This is the reason for the split.

*Q2: Why not deactivate SSHDriver from within DockerShellStrategy?*

Yes, the ssh driver should be given time to close down properly before
removing the docker container. But I now realize that the code for
DockerShellStrategy was misleading. Through inheritance from
ShellStrategy it made the user believe that DockerShellStrategy would
handle *both* docker creation *and* shell access. It does not,
however. Actually it is really the other way around: shell access
depends on a running docker container. So cleanup should rather be
done within ShellStrategy, not something related to docker.

I have rewritten DockerShellStrategy (in particular renamed to
DockerStrategy), but don't know what to do in ShellStrategy so it
handles closing down of any underlying connection provider. In
particular, I don't understand "bindings" in ShellStrategy. The
binding to PowerProtocol is OK since it abstracts away the particular
driver used to control power. But why bind to ShellDriver? This
prevents ShellStrategy from being used with e.g. SSHDriver. Wouldn't
it be clearer if strategies always bind to protocols (here
CommandProtocol)?

I see, there is a problem with this, however: Drivers have
activate/deactivate methods, but protocols don't. So exactly the SSH
close down that you ask for cannot really be formulated in a strategy
like ShellStrategy.

*Q3: Why trying to socket connect to container? SSHDriver already does so.*

When the docker container runs on the developer PC, the socket connect
call in `_update_resource()` is actually blazingly fast. It is
correct, though, that a noticable overhead may appear when the docker
driver is going to talk with remote daemons. But we feel "polling" the
remote port is important and really belongs with the DockerDriver. It
should be the duty of DockerDriver to ensure that the NetworkService
actually *is* available when declared to be so. Or: the container
really hasn't come up before all services in the container have
started and become operational.

Currently, the sshdriver contains appropriate machinery for waiting
for the port to become available.  But if the sshdriver were changed
or replaced with some other driver that would rather abort if not
being able to connect immediately we would prefer if the blame would
not hit the DockerDriver.

One could also argue, though, that just being able to connect to the
container doesn't really prove taht it is fully up. So perhaps we
should make it optional to wait for the socket connect (with a default
of True)?

*Steps taken*

I have rebased all PR code on top of latest from master, have amended
the former commit with the changes to the docker strategy and then
force pushed to the `docker_integration` branch.

trxanga:
@jluebbe

The motivation provided above for keeping an explicit split between
DockerDaemon and DockerDriver is incorrect. The real motivation is
simply to keep the ability of creating one or more containers
emulating targets on remote machines.

Labgrid exporters will make these docker daemon resources available to
our test environment.

The remote machines might have properties that are not available
on the machine the test is started from, e.g.:

 * They could be more powerful.
 * They could be placed in an environment with specific characteristics.

The former explanation said that the remote docker container should be
used for controlling test resources. But we already have labgrid
exporters for that; so the docker driver shouldn't be used for that
purpose.

jluebbe:
I've fixed the merge conflicts and some minor formatting problems. I intend to merge this as is, and we'll see how and if it needs to change later after we have some more experience with the real use-cases.

PR #303, opened 2018-09-03
SSHDriver: Properly handle SSH timeouts

a3f:
This introduces three new parameters to `NetworkService`:

- `options` (str): optional, ssh(1) options to pass through to SSH
- `timeout` (int): optional, equivalent to options: `-o ConnectionTimeout=<timeout>`
                 (default `30`)
- `shared` (str): optional, specifies whether SSH multiplexing (`ControlMaster` in
                ssh_config(5)) is used (default `True`)

When `shared: false`, the run method's timeout parameter may be used as
well to specify a timeout, in which case it overrides the
`NetworkService`'s.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,9 @@ Release 0.2.0 (unreleased)
 New Features
 ~~~~~~~~~~~~
 
+- `SSHDriver` and `NetworkService` timeout is now configurable.
+- `NetworkService` now accepts extra SSH-specific options through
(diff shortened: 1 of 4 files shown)


jluebbe:
We discussed this internally: The NetworkService is not the correct place to specify how to use SSH connection sharing. Also, this PR mixes connection timeouts with command timeouts which need to be kept separate. For a DUT which needs some time before SSH works, we need to loop in labgrid (with shorter SSH timeouts) to keep things under control.

PR #304, opened 2018-09-05
strategy: add off state to BareboxStrategy and UBootStrategy

jremmet:
This adds the off state which allows to switch the DUT off.

### DIFF
--- a/doc/development.rst
+++ b/doc/development.rst
@@ -256,17 +256,23 @@ Lets take a look at the builtin `BareboxStrategy`. The Status enum for Barebox:
 
    class Status(enum.Enum):
        unknown = 0
-       barebox = 1
-       shell = 2
+       off = 1
+       barebox = 2
+       shell = 3
 
-defines 2 custom states and the `unknown` state as the start point.
-These two states are handled in the transition function:
+defines 3 custom states and the `unknown` state as the start point.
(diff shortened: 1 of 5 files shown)


doc/development.rst
```diff
@@ -276,10 +277,17 @@ These two states are handled in the transition function:
         self.barebox.boot("")
         self.barebox.await_boot()
         self.target.activate(self.shell)
+    elif status == Status.off:
+        if self.status == Status.barebox:
+            self.target.deactivate(self.barebox)
+        elif self.status == Status.shell:
+            self.target.deactivate(self.shell)
+        self.target.activate(self.power)
```
jluebbe: You can call deactivate unconditionally, as it's a no-op if the driver isn't active. This applies below as well.

labgrid/strategy/bareboxstrategy.py
```diff
@@ -13,6 +13,7 @@ class Status(enum.Enum):
     unknown = 0
     barebox = 1
     shell = 2
+    off = 3
```
jluebbe: Please keep the states in their "natural" order. Also sort the state handlers accordingly. So unknown, off, barebox, shell.

jremmet:
Leave power.cycle() by now.

PR #305, opened 2018-09-10
Add Labgrid-project pyserial fork

Emantor:
Add our local pyserial fork in requirements and mention the fork in the installation chapter.


Emantor:
I force pushed an extended version of the note for installation.

PR #308, opened 2018-09-24
SSH Tunneling and ManagedFile

Emantor:
This PR introduces:
- sshmanager to manage ssh connections to exporters and use only one socket connection
- ManagedFile to support upload of image files to exporters
- A proxy infrastructure to use port forwarding for serial connections

Check List:
- [x] Docs
- [x] Tests

### DIFF
--- a/labgrid/util/managedfile.py
+++ b/labgrid/util/managedfile.py
@@ -1,0 +1,81 @@
+import tempfile
+import hashlib
+import subprocess
+import os
+
+import attr
+
+from .helper import get_user
+from .ssh import sshmanager
+from ..resource.common import Resource, NetworkResource
+
+
+@attr.s
+class ManagedFile:
+    """ The ManagedFile allows the synchronisation of a file to a remote host.
+    It has to be created with the to be synced file and the target resource as
+    argument:
+
+    ::
+        from labgrid.util.managedfile import ManagedFile
+
+        ManagedFile("/tmp/examplefile", <your-resource>)
+
+
+    Synchronisation is done with the sync_to_resource method.
+    """
+    local_path = attr.ib(
+        validator=attr.validators.instance_of(str),
+        converter=lambda x: os.path.abspath(str(x))
+    )
+    resource = attr.ib(
+        validator=attr.validators.instance_of(Resource),
+    )
+
+    def __attrs_post_init__(self):
+        if not os.path.isfile(self.local_path):
+            raise FileNotFoundError("Local file {} not found".format(self.local_path))
+
+        username = get_user()
+        hasher = hashlib.sha256()
+        with open(self.local_path, 'rb') as f:
+            for block in iter(lambda: f.read(1048576), b''):
+                hasher.update(block)
+        self.hash = hasher.hexdigest()
+        self.rpath = "/tmp/labgrid/{user}/{hash}/".format(
+            user=username, hash=self.hash
+        )
+
+    def sync_to_resource(self):
+        """sync the file to the host specified in a resource
+
+        Raises:
+            ExecutionError: if the SSH connection/copy fails
+        """
+        if isinstance(self.resource, NetworkResource):
+            host = self.resource.host
+            conn = sshmanager.open(host)
+            conn.run_command("mkdir -p {}".format(self.rpath))
+            conn.put_file(
+                self.local_path,
+                "{}{}".format(self.rpath, os.path.basename(self.local_path))
+            )
+
+    def get_remote_path(self):
+        """Retrieve the remote file path
+
+        Returns:
+            str: path to the file on the remote host
+        """
+        if isinstance(self.resource, NetworkResource):
+            return "{}{}".format(self.rpath, os.path.basename(self.local_path))
+        else:
+            return self.local_path
+
+    def get_hash(self):
+        """Retrieve the hash of the file
+
+        Returns:
+            str: SHA256 hexdigest of the file
+        """
+        return self.hash
--- a/labgrid/util/managedfile.py
+++ b/labgrid/util/managedfile.py
@@ -42,7 +42,7 @@ class ManagedFile:
             for block in iter(lambda: f.read(1048576), b''):
                 hasher.update(block)
         self.hash = hasher.hexdigest()
-        self.rpath = "/tmp/labgrid/{user}/{hash}/".format(
+        self.rpath = "/tmp/labgrid-{user}/{hash}/".format(
             user=username, hash=self.hash
         )
 
--- a/labgrid/util/proxy.py
+++ b/labgrid/util/proxy.py
@@ -1,0 +1,43 @@
+from .ssh import sshmanager
+
+from ..resource.common import NetworkResource, Resource
+
+__all__ = ['proxymanager']
+
+class ProxyManager:
+    """The ProxyManager class is only used inside labgrid.util.proxy (similar
+    to a singleton), don't instanciate this class, use the exported
+    proxymanager instead."""
+    @staticmethod
+    def get_host_and_port(res, force_proxy=False):
+        """ get host and port for a proxy connection from a Resource
+
+        Args:
+            res (Resource): The resource to retrieve the proxy for
+            proxy (:obj:`bool`, optional): whether to always proxy the
+                connection, defaults to False
+
+        Returns:
+            (host, port) host and port for the proxy connection
+
+        Raises:
+            ExecutionError: if the SSH connection/forwarding fails
+        """
+        assert isinstance(res, Resource)
+
+        if not hasattr(res, 'extra'):
+            return res.host, res.port
+
+        if not res.extra.get('proxy') or not res.extra.get("proxy_required"):
+            return res.host, res.port
+
+        if isinstance(res, NetworkResource):
+            proxy_required = res.extra['proxy_required']
+            proxy = res.extra['proxy']
+            if proxy_required or force_proxy:
+                port = sshmanager.request_forward(proxy, res.host, res.port)
+                host = 'localhost'
+                return host, port
+            return res.host, res.port
+
+proxymanager = ProxyManager()
--- a/labgrid/util/ssh.py
+++ b/labgrid/util/ssh.py
@@ -1,0 +1,322 @@
+# pylint: disable=no-member
+import tempfile
+import logging
+import subprocess
+import os
+import socket
+from functools import wraps
+
+import attr
+from ..driver.exception import ExecutionError
+
+from .helper import get_free_port
+
+__all__ = ['sshmanager', 'SSHConnection', 'ForwardError']
+
+
+@attr.s
+class SSHConnectionManager:
+    """The SSHConnectionManager manages multiple SSH connections. This class
+    should not be directly instanciated, use the exported sshmanager from this
+    module instead.
+    """
+    _connections = attr.ib(
+        default=attr.Factory(dict),
+        init=False,
+        validator=attr.validators.optional(attr.validators.instance_of(dict))
+    )
+
+    def __attrs_post_init__(self):
+        self.logger = logging.getLogger("{}".format(self))
+
+    def get(self, host: str):
+        """Retrieve or create a new connection to a given host
+
+        Arguments:
+            host (str): host to retrieve the connection for
+
+        Returns:
+            :obj:`SSHConnection`: the SSHConnection for the host"""
+        host = socket.getfqdn(host)
+        instance = self._connections.get(host)
+        if instance is None:
+            # pylint: disable=unsupported-assignment-operation
+            self.logger.debug("Trying to start new control socket")
+            instance = SSHConnection(host)
+            instance.connect()
+            self._connections[host] = instance
+        return instance
+
+    def add_connection(self, connection):
+        """Add an existing SSHConnection to the manager.
+        This is useful for manually created connections which should also be
+        available via the manager.
+
+        Arguments:
+            connection (:obj:`SSHConnection`): SSHconnection to add to the manager
+        """
+        # pylint: disable=unsupported-assignment-operation
+        assert isinstance(connection, SSHConnection)
+        if connection.host not in self._connections:
+            self._connections[connection.host] = connection
+
+    def remove_connection(self, connection):
+        # pylint: disable=unsupported-assignment-operation
+        assert isinstance(connection, SSHConnection)
+        if connection.isconnected():
+            raise ExecutionError("Can't remove connected connection")
+        del self._connections[connection.host]
+
+    def remove_by_name(self, name):
+        # pylint: disable=unsupported-assignment-operation
+        del self._connections[name]
+
+    def open(self, host):
+        return self.get(host)
+
+    def close(self, host):
+        con = self.get(host)
+        con.disconnect()
+        self.remove_connection(con)
+
+    def request_forward(self, host, dest, port):
+        con = self.get(host)
+        return con.add_port_forward(dest, port)
+
+    def remove_forward(self, host, dest, port):
+        con = self.get(host)
+        con.remove_port_forward(dest, port)
+
+    def put_file(self, host, local_file, remote_file):
+        con = self.get(host)
+        con.put_file(local_file, remote_file)
+
+    def get_file(self, host, remote_file, local_file):
+        con = self.get(host)
+        con.get_file(remote_file, local_file)
+
+    def close_all(self):
+        """Close all open connections and remove them from the manager """
+        for name, connection in self._connections.items():
+            if connection.isconnected():
+                connection.disconnect()
+        names = self._connections.copy().keys()
+        for name in names:
+            self.remove_by_name(name)
+
+
+
+
+@attr.s
+class SSHConnection:
+    """SSHConnections are individual connections to hosts managed by a control
+    socket. In addition to command execution this class also provides an
+    interface to manage port forwardings. These are used in the remote
+    infrastructure to tunnel multiple connections over one SSH link.
+
+    A public identity infrastructure is assumed, no extra username or passwords
+    are supported."""
+    host = attr.ib(validator=attr.validators.instance_of(str))
+    _connected = attr.ib(
+        default=False, init=False, validator=attr.validators.instance_of(bool)
+    )
+    _tmpdir = attr.ib(
+        default=attr.Factory(lambda: tempfile.mkdtemp(prefix="labgrid-connection-")),
+        init=False,
+        validator=attr.validators.instance_of(str)
+    )
+    _forwards = attr.ib(init=False, default=attr.Factory(dict))
+
+    def __attrs_post_init__(self):
+        self._logger = logging.getLogger("{}".format(self))
+        self._ssh_prefix = [ "-o", "LogLevel=ERROR", "-o", "PasswordAuthentication=no" ]
+        self._socket = os.path.join(
+            self._tmpdir, 'control-{}'.format(self.host)
+        )
+
+    def _open_connection(self):
+        """Internal function which appends the control socket and checks if the
+        connection is already open"""
+        self._check_and_start_master()
+
+    def _run_socket_command(self, command, forward=None):
+        "Internal function to send a command to the control socket"
+        complete_cmd = [
+            "ssh", "-x", "-o",
+            "ControlPath={}".format(self._socket), "-O",
+            command,
+        ]
+        if forward:
+            for item in forward:
+                complete_cmd.append(item)
+        complete_cmd.append("{host}".format(host=self.host))
+        res = subprocess.check_call(
+            complete_cmd,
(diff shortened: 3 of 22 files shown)


jluebbe:
I've added four changes as fixups:
- Rename active to connected in the util/ssh.py
- Simplfy the backwards compatibility code in remote/common.py
- Hash the ManagedFile incrementally
- Rename dest_file to remote_file in util/ssh.py for consistency.

I've also fixed some minor things directly, see this git range-diff:
```
 1:  fe67bb47ac60 !  1:  5285b3eded1c multiple: move get_free_ports to labgrid.util
    @@ -47,7 +47,7 @@
     +import socket
     +from contextlib import closing
     +
    -+def get_free_port(()):
    ++def get_free_port():
     +    """Helper function to always return an unused port."""
     +    with closing(socket(AF_INET, SOCK_STREAM)) as s:
     +        s.bind(('', 0))
 2:  1e37d90c9f4d !  2:  2dde7d216917 travis: add key for localhost
    @@ -1,6 +1,8 @@
     Author: Rouven Czerwinski <r.czerwinski@pengutronix.de>
     
    -    travis: add key for localhost
    +    travis: add ssh key for localhost
    +
    +    Also install packages by using the travis apt addon syntax.
     
 3:  6190a12adfda !  3:  d66448405c54 util/ssh: add SSHConnection, SSHConnectionManager and sshmanager
    @@ -28,11 +28,7 @@
     +from socket import gethostname, socket, AF_INET, SOCK_STREAM
      from contextlib import closing
      
    --def get_free_port(()):
    -+def get_free_port():
    -     """Helper function to always return an unused port."""
    -     with closing(socket(AF_INET, SOCK_STREAM)) as s:
    -         s.bind(('', 0))
    + def get_free_port():
     
     diff --git a/labgrid/util/ssh.py b/labgrid/util/ssh.py
     new file mode 100644
 4:  8453700cb1d9 =  4:  acb0294f4b56 doc/development: document sshmanager
 5:  3e32d38cd12f !  5:  ec72183cc419 remote/exporter: add isolated switch for SSH forwards
    @@ -57,3 +57,58 @@
          }
      
          crossbar_url = args.crossbar
    +
    +diff --git a/man/labgrid-exporter.1 b/man/labgrid-exporter.1
    +--- a/man/labgrid-exporter.1
    ++++ b/man/labgrid-exporter.1
    +@@
    + .B \-x\fP,\fB  \-\-crossbar\-url
    + the crossbar url of the coordinator
    + .TP
    ++.B \-i\fP,\fB  \-\-isolated
    ++enable isolated mode (always request SSH forwards)
    ++.TP
    + .B \-n\fP,\fB  \-\-name
    + the public name of the exporter
    + .TP
    + .B \-\-hostname
    + hostname (or IP) published for accessing resources
    + .UNINDENT
    ++.SS \-i / \-\-isolated
    ++.sp
    ++This option enables isolated mode, which causes all exported resources marked
    ++as requiring SSH connection forwarding.
    ++Isolated mode is useful when resources (such as NetworkSerialPorts) are not
    ++directly accessible from the clients.
    ++The client will then use SSH to create a port forward to the resource when
    ++needed.
    + .SS \-n / \-\-name
    + .sp
    + This option is used to configure the exporter name under which resources are
    +
    +diff --git a/man/labgrid-exporter.rst b/man/labgrid-exporter.rst
    +--- a/man/labgrid-exporter.rst
    ++++ b/man/labgrid-exporter.rst
    +@@
    +     display command line help
    + -x, --crossbar-url
    +     the crossbar url of the coordinator
    ++-i, --isolated
    ++    enable isolated mode (always request SSH forwards)
    + -n, --name
    +     the public name of the exporter
    + --hostname
    +     hostname (or IP) published for accessing resources
    + 
    ++-i / --isolated
    ++~~~~~~~~~~~~~~~
    ++This option enables isolated mode, which causes all exported resources marked
    ++as requiring SSH connection forwarding.
    ++Isolated mode is useful when resources (such as NetworkSerialPorts) are not
    ++directly accessible from the clients.
    ++The client will then use SSH to create a port forward to the resource when
    ++needed.
    ++
    + -n / --name
    + ~~~~~~~~~~~
    + This option is used to configure the exporter name under which resources are
 6:  fcf50794659e !  6:  99e9c9c51cb6 remote/common: add proxy and proxy_required properties
    @@ -2,9 +2,12 @@
     
         remote/common: add proxy and proxy_required properties
     
    -    This also plumps them down into the underlying Resource base class, making both
    +    This also passes them down into the underlying Resource base class, making both
         parameters available to all resources.
     
    +    For backwards compatibility with old exporters, it defaults to disabling
    +    proxy mode.
    +
     
     diff --git a/labgrid/remote/common.py b/labgrid/remote/common.py
    @@ -17,13 +20,13 @@
     +    @property
     +    def proxy(self):
     +        """proxy information"""
    -+        return self.data['proxy']
    ++        return self.data.get('proxy')
     +
     +    @property
     +    def proxy_required(self):
     +        """proxy_required indicated whether the exporter requires a proxy
     +        connection"""
    -+        return self.data['proxy_required']
    ++        return self.data.get('proxy_required')
     +
          def asdict(self):
              return {
    @@ -93,8 +96,13 @@
                  new.parent = remote_place
                  new.avail = resource_entry.avail
                  new.extra = resource_entry.extra
    -+            new.proxy = resource_entry.proxy
    -+            new.proxy_required = resource_entry.proxy_required
    ++
    ++            try:
    ++                new.proxy = resource_entry.proxy
    ++                new.proxy_required = resource_entry.proxy_required
    ++            except AttributeError:
    ++                pass
    ++
                  new._remote_entry = resource_entry
                  if not isinstance(new, ManagedResource):
                      self.unmanaged_resources.append(new)
 7:  2860d4f19867 !  7:  226587b3124c util/proxy: add proxymanager for serial port forwardings
    @@ -2,7 +2,8 @@
     
         util/proxy: add proxymanager for serial port forwardings
     
    -    The proxymanager should be used in drivers to request a forawrd over SSH.
    +    The proxymanager should be used in drivers to request a forward over SSH.
    +    If no proxying is requested, a direct connection is used as previously.
     
    @@ -64,6 +65,9 @@
     +        """
     +        assert isinstance(res, Resource)
     +
    ++        if not res.proxy:
    ++            return res.host, res.port
    ++
     +        if isinstance(res, NetworkResource):
     +            if res.proxy_required or proxy:
     +                port = sshmanager.request_forward(res.proxy, res.host, res.port)
 8:  595d57fe2355 =  8:  397c97800ebc client: add port forwarding for serial console
 9:  ebfe7d6bcb14 =  9:  e66d19f746a8 util: add ManagedFile
10:  3d03d1f58609 = 10:  0469a5ac7865 doc/development: document ManagedFile and ProxyManager
11:  82d21f2b964b = 11:  c06d543dbac3 multiple: use ManagedFile for file uploads
12:  cd0ce230c1da <  -:  ------------ remote: ignore attribute error for new proxy attributes
13:  f8e23e47b2bd <  -:  ------------ remote/common: fallback to None for unavailable proxy parameters
14:  b5220290e789 <  -:  ------------ util/proxy: return normal forwards if there is no proxy attribute
15:  e222c583bcf4 = 12:  5b62afab5be6 doc/getting_started: add note about tmp files created by ManagedFile
16:  95ef14330fda ! 13:  e018f3ae5219 pytestplugin/fixtures: close all sshconnections after the test ends
    @@ -1,6 +1,6 @@
     Author: Rouven Czerwinski <r.czerwinski@pengutronix.de>
     
    -    pytestplugin/fixtures: close all sshconnections after the test ends
    +    pytestplugin/fixtures: close all ssh connections after the test ends
     
17:  f263496aa61a = 14:  1f10f023af6f CHANGES: update with new features
 -:  ------------ > 15:  30c59b24c21e fixup! util/ssh: add SSHConnection, SSHConnectionManager and sshmanager
 -:  ------------ > 16:  78a9bd912d8b fixup! remote/common: add proxy and proxy_required properties
 -:  ------------ > 17:  e97914729fb2 fixup! util: add ManagedFile
 -:  ------------ > 18:  f638fd10e4c0 fixup! util/ssh: add SSHConnection, SSHConnectionManager and sshmanager
```

labgrid/util/managedfile.py
```diff
@@ -0,0 +1,80 @@
+import tempfile
+import hashlib
+import subprocess
+import os
+import getpass
+
+import attr
+
+from .ssh import sshmanager
+from labgrid.resource.common import Resource, NetworkResource
```
jluebbe: Don't mix relative with absolute imports.

labgrid/util/ssh.py
```diff
@@ -0,0 +1,324 @@
...
+__all__ = ['sshmanager', 'SSHConnection', 'ForwardError']
+
+
+@attr.s
+class SSHConnectionManager:
+    """The SSHConnectionManager manages multiple SSH connections. This class
+    should not be directly instanciated, use the exported sshmanager from this
+    module instead.
+    """
+    _connections = attr.ib(
```
jluebbe: Why do you use an attribute here?

labgrid/util/ssh.py
```diff
@@ -0,0 +1,324 @@
...
+            host (str): host to retrieve the connection for
+
+        Returns:
+            :obj:`SSHConnection`: the SSHConnection for the host"""
+        host = socket.getfqdn(host)
+        instance = self._connections.get(host)
+        if instance is None:
+            # pylint: disable=unsupported-assignment-operation
+            self.logger.debug("Trying to start new control socket")
+            # instance = self._start_control_socket(host)
```
jluebbe: Remove this comment?

labgrid/remote/common.py
```diff
@@ -39,12 +39,25 @@ def extra(self):
         """extra resource information"""
         return self.data['params'].get('extra', {})
 
+    @property
+    def proxy(self):
```
jluebbe: Why did we decide to add these as top-level members instead of using the existing extra dict?
I remember we discussed this, but don't remember the actual reason.

labgrid/util/proxy.py
```diff
@@ -0,0 +1,46 @@
+from .ssh import sshmanager
+
+from ..resource.common import NetworkResource, Resource
+
+__all__ = ['proxymanager']
+
+class ProxyManager:
+    """The ProxyManager class is only used inside labgrid.util.proxy (similar
+    to a JAVA singleton), don't instanciate this class, use the exported
```
jluebbe: Singletons are not specific to Java. :)

labgrid/util/ssh.py
```diff
@@ -0,0 +1,324 @@
...
+        names = self._connections.copy().keys()
+        for name in names:
+            self.remove_by_name(name)
+
+
+def check_connected(func):
+    """Check if an SSHConnection is connected as a decorator"""
+
+    @wraps(func)
+    def wrapper(cls, *_args, **_kwargs):
```
jluebbe: Why do you use cls instead of self? The value should be a object, not the class.

labgrid/util/ssh.py
```diff
@@ -0,0 +1,324 @@
...
+class SSHConnection:
+    """SSHConnections are individual connections to hosts managed by a control
+    socket. In addition to command execution this class also provides an
+    interface to manage port forwardings. These are used in the remote
+    infrastructure to tunnel multiple connections over one SSH link.
+
+    A public identity infrastructure is assumed, no extra username or passwords
+    are supported."""
+    host = attr.ib(validator=attr.validators.instance_of(str))
+    _connected = attr.ib(
```
jluebbe: Why do you use attrs for internal values?

labgrid/util/ssh.py
```diff
@@ -0,0 +1,324 @@
...
+    def _open_connection(self):
+        """Internal function which appends the control socket and checks if the
+        connection is already open"""
+        self._ssh_prefix += " -o ControlPath={}".format(
+            self._socket
+        ) if self._check_master() else ""
+
+    def _run_socket_command(self, command, forward=""):
+        "Internal function to send a command to the control socket"
+        complete_cmd = "ssh -x -o ControlPath={cpath} -O {command}{forward} {host}".format(
```
jluebbe: By using "{command}{forward}", you require the caller to add the space themselves. Just check if a forward is given and then build the command directly as a list, as this also avoids possible problems with spaces in the arguments.

labgrid/util/ssh.py
```diff
@@ -0,0 +1,324 @@
...
+            stdout=subprocess.DEVNULL,
+            stderr=subprocess.DEVNULL,
+            timeout=2
+        )
+
+        return res
+
+    def _run_command(self, command):
+        "Internal function to run a command over the SSH connection"
+        complete_cmd = "ssh -x -o ControlPath={cpath} {host} {command}".format(
```
jluebbe: Build the command as a list to avoid problems with spaces in the arguments.

labgrid/util/ssh.py
```diff
@@ -0,0 +1,324 @@
...
+            cpath=self._socket,
+            host=self.host,
+            command=command,
+        ).split(' ')
+        res = subprocess.check_call(
+            complete_cmd,
+            stdin=subprocess.DEVNULL,
+            stdout=subprocess.DEVNULL,
+            stderr=subprocess.DEVNULL
```
jluebbe: Wouldn't it be useful to pass stderr on (above as well)?

labgrid/util/ssh.py
```diff
@@ -0,0 +1,324 @@
...
+        check = subprocess.call(
+            args,
+            stdin=subprocess.DEVNULL,
+            stdout=subprocess.DEVNULL,
+            stderr=subprocess.DEVNULL
+        )
+        if check == 0:
+            return ""
+
+        return self._start_own_master()
```
jluebbe: It seems strange that a check function also starts the master?

tests/test_util.py
```diff
@@ -1,7 +1,28 @@
...
+def connection_localhost():
+    con = SSHConnection("localhost")
+    con.connect()
+    yield con
+    con.disconnect()
+
+@pytest.fixture
+def sshmanager_fix():
+    yield sshmanager
+    sshmanager.close_all()
```
jluebbe: Is this still needed after "pytestplugin/fixtures: close all ssh connections after the test ends"?

tests/test_util.py
```diff
@@ -41,3 +62,177 @@ class A:
...
+
+def test_sshconnection_run(connection_localhost):
+    assert connection_localhost.run_command("echo Hello") == 0
+
+def test_sshconnection_port_forward_add_remove(connection_localhost):
+    port = 1337
+    test_string = "Hello World"
+
+    local_port = connection_localhost.add_port_forward('localhost', port)
+    nc_listen = subprocess.Popen(["nc", "-l", "-p", "1337"], stdout=subprocess.PIPE, stdin=None)
```
jluebbe: Why do you use nc instead of opening the socket in python?

tests/test_util.py
```diff
@@ -41,3 +62,177 @@ class A:
...
+    nc_send = subprocess.Popen(["nc", "localhost", "{}".format(local_port)], stdin=subprocess.PIPE,stdout=None)
+    try:
+        nc_send.communicate("Hello World".encode("utf-8"), timeout=1)
+    except subprocess.TimeoutExpired:
+        pass
+    nc_listen.terminate()
+    assert nc_listen.communicate()[0].decode("utf-8") == test_string
+    connection_localhost.remove_port_forward('localhost', port)
+
```
jluebbe: Use a consistent amount of lines between the functions (pylint wants 2).

labgrid/util/ssh.py
```diff
@@ -0,0 +1,324 @@
...
+            )
+
+        if not os.path.exists(control):
+            raise ExecutionError("no control socket to {}".format(self.host))
+
+        self._logger.debug('Connected to %s', self.host)
+
+        return control
+
+    def _disconnect(self):
```
jluebbe: What happens with existing forwards in this case?

tests/test_util.py
```diff
@@ -1,7 +1,28 @@
...
+from labgrid.driver.exception import ExecutionError
+from labgrid.resource.serialport import NetworkSerialPort
+from labgrid.resource.common import Resource, NetworkResource
+
+@pytest.fixture
+def connection_localhost():
+    con = SSHConnection("localhost")
+    con.connect()
+    yield con
+    con.disconnect()
```
jluebbe: Does this remove the forwards as well?

tests/test_util.py
```diff
@@ -41,3 +62,177 @@ class A:
...
+    t = tmpdir.join("test")
+    t.write(
+"""
+Test
+"""
+    )
+    hash = hashlib.sha256(t.read().encode("utf-8")).hexdigest()
+    mf = ManagedFile(t,res)
+    mf.sync_to_resource()
+    assert os.path.isfile("/tmp/labgrid/{}/{}/test".format(getpass.getuser(), hash))
```
jluebbe: Using getpass will import other modules (such as termios).

Basically we just need (in util.py):
```python
def get_user():
    user = os.environ.get("USER")
    if user:
        return user
    import pwd
    return pwd.getpwuid(os.getuid())[0]
```

tests/test_util.py
Emantor: Because I only thought of nc. I'll rework this to use sockets.

tests/test_util.py
Emantor: Will fix.

labgrid/util/managedfile.py
Emantor: Will fix

labgrid/util/ssh.py
Emantor: This makes debugging easier since you can see the state of private attributes as well, i.e. in backtraces. Though these are missing the `init=False` directive.

labgrid/util/ssh.py
Emantor: Will do.

labgrid/util/proxy.py
Emantor: I'll remove the JAVA.

labgrid/util/ssh.py
Emantor: I'll rename `cls` to `obj`

labgrid/util/ssh.py
Emantor: I'll rework this and the other commands into lists

labgrid/util/ssh.py
Emantor: same as above

labgrid/util/ssh.py
Emantor: Since the check function returns the used socket, we can also start the master here if we don't find a socket. The same logic is used in the SSHDriver, I can rework this to be more explicit during the connection process.

tests/test_util.py
Emantor: Yes, closing the SSH Connections as implemented in the mentioned commit only works with the `env` fixture. Since the tests for labgrid itself don't use this fixture, we have to do this explicitly ourselves.

labgrid/util/ssh.py
Emantor: AFAIR forwards are closed if the socket is closed.

tests/test_util.py
Emantor: I can add this in a separate commit.

labgrid/util/ssh.py
jluebbe: See the BindingMixin.check_active decorator. You can do the sme here.

labgrid/util/ssh.py
jluebbe: Or rename the check function to something else that better fit's what it's actually doing. :)

Emantor:
Review Tasklist:
- [ ] Apply Fixups
- [ ] unmix relative and absolute imports in managedfile.py
- [ ] discuss private attr.s
- [ ] fixup private attr.s with init=False
- [ ] remove comment in line 44 of ssh.py
- [ ] justify extra dict vs param change in remote/common
- [ ] remove JAVA in proxy.py line 9
- [ ] rework wrapper in ssh.py line 112 (cf BindingMixin.check_active)
- [ ] rework commands in ssh.py into lists
- [ ] pass stderr on to allow error lookup
- [ ] rename check function in ssh.py
- [ ] rework tests from nc to sockets
- [ ] fix lines in test_util.py
- [ ] verify that forwards are closed on socket close
- [ ] make sure con.disconnect() removes forwards
- [ ] add get_user helper

labgrid/util/ssh.py
Emantor: I changed it to explicitly close the forwards before closing the socket

labgrid/remote/common.py
Emantor: Because the `extra` attribute is only attached to the ResourceEntry and we need use the proxy information after creating the resource.  We need the Resource information attached to the Resource to later decide whether we need to create the tunnel or not.

PR #309, opened 2018-09-25
sshdriver: use -f if ssh is used with password

jremmet:
-n to put the ssh program to background doesn't work with password.
Use -f instead.

### DIFF
--- a/labgrid/driver/sshdriver.py
+++ b/labgrid/driver/sshdriver.py
@@ -53,7 +53,7 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
         )
         # use sshpass if we have a password
         sshpass = "sshpass -e " if self.networkservice.password else ""
-        args = ("{}ssh -n {} -x -o ConnectTimeout=30 -o ControlPersist=300 -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o ServerAliveInterval=15 -MN -S {} -p {} {}@{}").format( # pylint: disable=line-too-long
+        args = ("{}ssh -f {} -x -o ConnectTimeout=30 -o ControlPersist=300 -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o ServerAliveInterval=15 -MN -S {} -p {} {}@{}").format( # pylint: disable=line-too-long
             sshpass, self.ssh_prefix, control, self.networkservice.port,
             self.networkservice.username, self.networkservice.address).split(" ")
 


jluebbe:
From the documentation, it seems we should be able to always use `-f`:
```
     -f      Requests ssh to go to background just before command execution.  This is useful if
             ssh is going to ask for passwords or passphrases, but the user wants it in the back‐
             ground.  This implies -n.  The recommended way to start X11 programs at a remote
             site is with something like ssh -f host xterm.
```
Indeed, we had -f before, but changed to -n in #177. @tgbloks could you comment here? Did you use `-n` together with `sshpass`?

We let ssh put itself into the background to detect when it has connected successfully, but I'm open to changing that if we find a better way.

tgbloks:
i switched to -n because with -f i did not always get sucessful connection (using sshpass).
Hope i will get a chance to test with the latest lagrid code tomorrow...

jremmet:
Here with -n ssh returns an I run into
            if self.process.wait(timeout=30) is not 0:
                raise ExecutionError(

tgbloks:
i rechecked my tests (using sshpass) with the latest labgrid version with -f and -n.
With -n it works fine for me (just as it used to...) but with -f i sometimes (about every 10th time i run my test) get asked for the ssh password during the test (because the ssh master died somehow?).
Sadly i forgot to check the ssh versions i used but i will post them here once i did...

jremmet:
Thanks for looking into it.
Here I use OpenSSH_7.2p2 Ubuntu-4ubuntu2.4, OpenSSL 1.0.2g. and Python 3.5.2.

Does the "ssh -n" call return and end if you use it on the command line? Here it does, and process.wait(timeout=30) returns 1, throwing the ExecutionError.

jremmet:
Still have trouble with "ssh -n". 
If I remove "self.process.wait(timeout=30) is not 0" check it works.

Should I create an Issue for this?

jluebbe:
> Still have trouble with "ssh -n".
> If I remove "self.process.wait(timeout=30) is not 0" check it works.
What is the returned code in that case?

> Should I create an Issue for this?
It seems to me that these issues are related, so we can discuss them together.

@tgbloks Could you verify that the problem is caused by the ssh master dying? You can query the master similar to how _check_master() works.

tgbloks:
iirc i tried to connnect to the master (or looked up the process)when my "unexpected login" problems appeared. Anyway the the timeout problem from @jremmet  makes perfect sense as ssh -n should not return (neither does it when i try manually) -> atm i dont have an explanation why -n works better (or at all) for me while -f should be what we want here. I am fine with switching to -f again - at least until i had some more time to look into my (most likely very specific) problem.

jremmet:
> > Still have trouble with "ssh -n".
> > If I remove "self.process.wait(timeout=30) is not 0" check it works.
> > What is the returned code in that case?

Return value is 1

jremmet:
> iirc i tried to connnect to the master (or looked up the process)when my "unexpected login" problems appeared. Anyway the the timeout problem from @jremmet makes perfect sense as ssh -n should not return (neither does it when i try manually) -> atm i dont have an explanation why -n works better (or at all) for me while -f should be what we want here. I am fine with switching to -f again - at least until i had some more time to look into my (most likely very specific) problem.

Here ssh -n shows the greater and goes to background. The process runs there but process.wait(timeout=30) returns 1
Python 3.5.2

jremmet:
like @tgbloks suggested I rebased on master and switched to -f

bbara:
Hey, any updates on this?
The call with -n also fails on my machine with "stdin: is not a tty".
This makes sense to me, since -n redirects /dev/null:
```
    -n       Redirects stdin from /dev/null (actually, prevents reading from
             stdin).  This must be used when ssh is run in the background.  A com‐
             mon trick is to use this to run X11 programs on a remote machine.  For
             example, ssh -n shadows.cs.hut.fi emacs & will start an emacs on shad‐
             ows.cs.hut.fi, and the X11 connection will be automatically forwarded
             over an encrypted channel.  The ssh program will be put in the back‐
             ground.  (This does not work if ssh needs to ask for a password or
             passphrase; see also the -f option.)
```

> With -n it works fine for me (just as it used to...) but with -f i sometimes (about every 10th time i run my test) get asked for the ssh password during the test (because the ssh master died somehow?).

@tgbloks are you logging in with SSH key + password or user + password?
Because -n actually should not allow any password interaction with the SSH server,
so maybe sshpass interacts with your local ssh-agent or something like that?
Have not tried it yet, but maybe SSH key + password requires different behaviour.

jluebbe:
OK, it seems that we agree that ``-f`` is should work correctly, so I'm going to merge this.
If this results in further problems, I think we should use Python's pty.openpty() to replace what sshpass does and use /dev/null as stdin for the python process.

PR #311, opened 2018-09-26
graphstrategy: support a compatibility format for transitions with via

jluebbe:
GraphStrategies have an additional 'via' keyword argument for
transitions. This change allows specifying this by using `labgrid-client
-s shell:off,barebox` which translates to `transition('shell',
via=['off', 'barebox'])`.


Emantor:
If we move the split function into the `client.py` we have to catch the keyword argument error for other strategies, in the current state you'll get a StateError because we don't detect that there is no support for the via semantic for other strategies. How does the remote client need to know all the transitions to handle the split in the remote client case?

PR #316, opened 2018-09-28
driver/shelldriver: Fix initial login prompt timeout

esben:
When waiting for the initial login prompt, the `before` value is `b''`
on timeout, so with `last_before = None` the timeout had to fire two times
before the `self.console.sendline("")` was executed.


Emantor:
While `before` is initialized to `None`, the first `read` will set it to `b''` in pexpect.

PR #317, opened 2018-09-28
Fix use of deprecated attrs convert() method

esben:
The `convert()` method was deprecated with attrs 17.4.0.
See http://www.attrs.org/en/stable/changelog.html#id44


Emantor:
Thanks!

PR #318, opened 2018-09-28
remote/client: Allow use of $LG_ENV to specify config file

esben:
When using the pytestplugin, $LG_ENV can be used for specifying the
configuration file.  Adding support for same behavior when using
labgrid-client allows setting up $LG_ENV and then using simple CLI commands
without having to bother with --lg-env and --config options.


esben:
Ah, I was not aware that generated man pages should be committed. Fixed now.

PR #319, opened 2018-09-28
driver/sshdriver: Make SSHDriver a preferred CommandProtocol

esben:
For tests where both ShellDriver and SSHDriver are activated, the SSHDriver
should be preferred, as it much faster.  Without this, calling
`target.get_driver('CommandProtocol')` with both drivers active will fail.

For other use cases (ie. where only one CommandProtocol driver is active at a
time), things will work as before.

### DIFF
--- a/labgrid/driver/sshdriver.py
+++ b/labgrid/driver/sshdriver.py
@@ -22,6 +22,7 @@ from .exception import ExecutionError
 class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
     """SSHDriver - Driver to execute commands via SSH"""
     bindings = {"networkservice": NetworkService, }
+    priorities = {CommandProtocol: 10, FileTransferProtocol: 10}
     keyfile = attr.ib(default="", validator=attr.validators.instance_of(str))
     stderr_merge = attr.ib(default=False, validator=attr.validators.instance_of(bool))
 


Emantor:
This is a fine optimization where tests are allowed to run on whichever `CommandProtocol` implementation driver they can use. However in case you also want to reboot the target, the `SSHDriver` is not guaranteed to be usable in all scenarios, i.e. your target needs to be configured by the `ShellDriver` first.

For your test its probably best to use the specific driver and not rely on the Protocol to choose it.

esben:
Any test that cares about which `CommandProtocol` driver it needs should not rely on `CommandProtocol`, but specify the required driver.

With this change, it will be possible to write tests that works with both `ShellDriver` and `SSHDriver`, and simply use `target.get_driver('CommandProtocol')`, and it will work no matter what combination of `ShellDriver` and/or `SSHDriver` resources are available.  Without this change, if both resources are available,`target.get_driver('CommandProtocol')` will simply bail out.  I cannot see how that is helpful.
Without the change, 

There is definitely tests that SSHDriver is not suitable for.  For those tests, you would need to get ShellDriver specifically.

The main point of this change is that without this, you will not be able use fx. `target.get_driver('CommandProtocol')` when multiple `CommandProtocol` drivers are available, so this change fixes that by setting a preference to SSHDriver.  Any test that cares about which CommandProtocol driver to use should make sure to get the

esben:
Ah, I found the reason for the problems I were seeing.  I had a bug in my custom strategy, causing a reboot without deactivation of the SSHDriver.  Fixing that makes sshdriver work as expected :)

A remaining concern is the ControlPersist=300 value.  I know it is a long time, but it could potentially cause unexpected behavior, as the master process could shutdown while SSHDriver is activated.  Maybe it would be better to call _check_master() method everytime run(), put() or get() is called?

esben:
My bad, I commented and closed on wrong PR.

PR #322, opened 2018-10-05
serialdriver: fix warning string

Emantor:
# General

Fix a bug in the serialdriver. It says python instead of pyserial.

## Description
Fix for a copy-and-paste error I did with the pyserial version warning. 

## Task List
- [x] PR has been tested


Emantor:
Indeed, I pushed the correct version based on master.

PR #323, opened 2018-10-05
 config: evaluate symbolic links to the labgrid test configuration files

jgoetzinger:


Emantor:
This fundamentally changes discovery of image files, import files and adjacent configuration.
This is not really a fix but a rather big change which will break existing environments which use symbolic links for the environment files.
From our discussion on IRC its not entirely clear to me why this change is necessary, I think the same can be achieved by using different means and not breaking compatibility.
Please provide a clear outline with example why this change is needed.

jgoetzinger:
Hi,
we do have a usecase for this change, but it has low priority. At the moment I do not have time to look for an other solution which ensures, that existing setups don´t break. For now we are okay with leaving it the way it is. Maybe I will come back to this again when I have more time.

Emantor:
I'll close this for now, feel free to ping me on IRC for a discussion or open an issue with your use case.

PR #326, opened 2018-10-15
Allow import of python modules using imports: in config.yaml

esben:
This extends the imports: mechanism in the config.yaml file with support for using Python
modules without having to care about their filesystem paths.

Say you have installed a Python module which includes a labgrid strategy driver as my.labgrid.strategy module, you can add

```yaml
imports:
  - my.labgrid.strategy
```

to your config.yaml.

This makes it easier to manage drivers when managed as an infrastructure part, and not directly in the test repositories.

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -1502,6 +1502,7 @@ The skeleton for an environment consists of:
      <more tools>
    imports:
      - <import.py>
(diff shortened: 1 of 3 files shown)


jluebbe:
I've added a commit message based on your PR description.

PR #327, opened 2018-10-16
driver/sshdriver: Extend StrictHostKeyChecking=no to scp

esben:
Apply the StrictHostKeyChecking for scp (put() and get()) instead of only
for ssh commands (run()).

### DIFF
--- a/labgrid/driver/sshdriver.py
+++ b/labgrid/driver/sshdriver.py
@@ -30,6 +30,7 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
 
     def on_activate(self):
         self.ssh_prefix = "-o LogLevel=ERROR"
+        self.ssh_prefix += " -o StrictHostKeyChecking=no"
         self.ssh_prefix += " -i {}".format(os.path.abspath(self.keyfile)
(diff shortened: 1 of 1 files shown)


esben:
Yes, I see your point.  But something strange must be going on. I just reproduced the behaviour here.
Without this commit, put() results in a hostkey check (which fails when session is non-interactive).  With the commit, no hostkey check is done.

Maybe the master process terminates before I get to the scp.  From ssh_config(5):

> These sessions will try to reuse the master instance's network connection rather than initiating new ones, but will fall back to connecting normally if the control socket does not exist, or is not listening.

So I am guessing that the I end up hitting the fallback part.
I will try increasing the ControlPersist value, although I the entire test is less than 300 seconds.

esben:
Ah, I found the reason for the problems I were seeing. I had a bug in my custom strategy, causing a reboot without deactivation of the SSHDriver. Fixing that makes sshdriver work as expected :)

A remaining concern is the ControlPersist=300 value. I know it is a long time, but it could potentially cause unexpected behavior, as the master process could shutdown while SSHDriver is activated. Maybe it would be better to call _check_master() method everytime run(), put() or get() is called?

PR #336, opened 2018-10-22
util/ssh: do not use foreign ControlMaster, leave ControlMaster open indefinitely, small fixes

BastianStender:
This is a collection of util/ssh fixes I came up with while using current labgrid master:
- util/ssh: remove unnecessary empty lines
- util/ssh: use self._socket consistently, remove return
- util/ssh: do not use foreign ControlMaster
- util/ssh: master connection must remain open indefinitely
- util/ssh: copy also symlink referents to remote

### DIFF
--- a/labgrid/util/ssh.py
+++ b/labgrid/util/ssh.py
@@ -210,7 +210,8 @@ class SSHConnection:
         """Put a file onto the remote host"""
         subprocess.check_call([
             "rsync", "-e", "ssh -o ControlPath={}".format(self._socket),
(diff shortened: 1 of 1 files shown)


BastianStender:
Please do not merge yet. The ControlMaster issue seems to persist even with these patches.

BastianStender:
Pushed another commit for util/ssh while at it. It makes rsync copy symlink referents to a remote instead of skipping them. Have a look at the commit message for more details.

BastianStender:
#351 solved this with another approach. Will open another PR for "util/ssh: copy also symlink referents to remote"

PR #337, opened 2018-10-24
strategy/ubootstrategy: fix OFF transition

lrntplc:
Barebox does not exist in ubootstrategy.

Signed-off-by: Laurentiu Palcu <laurentiu.palcu@nxp.com>


jluebbe:
Thanks for fixing this, @lrntplc!

PR #339, opened 2018-10-29
remote/exporter: fix _stop() call in USBSerialPortExport

BastianStender:
This works the same way as ResourceExport.stop() does.

It fixes:
`TypeError: _stop() missing 1 required positional argument: 'start_params'`

### DIFF
--- a/labgrid/remote/exporter.py
+++ b/labgrid/remote/exporter.py
@@ -121,7 +121,7 @@ class USBSerialPortExport(ResourceExport):
 
     def __del__(self):
         if self.child is not None:
-            self._stop()
+            self.stop()
 
     def _get_start_params(self):
         return {


labgrid/remote/exporter.py
```diff
@@ -121,7 +121,7 @@ def __attrs_post_init__(self):
 
     def __del__(self):
         if self.child is not None:
-            self._stop()
+            self._stop(self.start_params)
```
Emantor: From reading the log and source code, this was `self.stop()` before.
Unfortunately the log does not provide hints why it was changed to `self._stop()` (which is obviously wrong).

BastianStender: 6c296c3d ("remote/exporter: fix wrong call") was correct at the time it was merged. The `ResourceExport.stop()` method was introduced by 7585a228 ("remote/exporter: detect when the device path changes and restart ser2net"). This commit also changed the signature of `ResourceExport._stop()` without fixing the stop call in USBSerialPortExport's `__del__()`.

The correct way to fix this is therefore calling `stop()` (not `_stop()`) in `__del__()`.

BastianStender:
PR modified as explained above.

PR #341, opened 2018-11-01
Miscellaneous fixes: error message for coordinator connection failure, fix get_resources() docstring

BastianStender:
Some minor fixes.

### DIFF
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -1224,9 +1224,9 @@ def main():
     }
 
     if args.command and args.command != 'help':
-        session = start_session(args.crossbar, os.environ.get("LG_CROSSBAR_REALM", "realm1"), extra)
         exitcode = 0
         try:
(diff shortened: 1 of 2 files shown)


labgrid/remote/client.py
```diff
@@ -922,7 +922,11 @@ def create():
     _, host, port, _, _, _ = parse_url(url)
 
     coro = loop.create_connection(transport_factory, host, port)
-    loop.run_until_complete(coro)
+    try:
+        loop.run_until_complete(coro)
+    except ConnectionError as e:
+        print("Could not connect to coordinator: {}".format(e))
+        exit(1)
```
jluebbe: This is part of the library layer, so we can't just exit(1) here. For labgrid-client you can move the try/except to the start_session call site in main(). RemotePlaceManager already has a similar try/except.

BastianStender: Fixed as suggested.

PR #342, opened 2018-11-01
remote/client: try to use IP/username/password for ssh subcommand from NetworkService

Bastian-Krause:
IP, username and password might be already available in NetworkService, currently used for SSHDriver. Try to use these also for labgrid-client's ssh subcommand.

### DIFF
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -765,8 +765,13 @@ class ClientSession(ApplicationSession):
 
     def _get_ip(self, place):
         target = self._get_target(place)
-        from ..resource import EthernetPort
-        resource = target.get_resource(EthernetPort)
+        from ..resource import EthernetPort, NetworkService
+        try:
+            resource = target.get_resource(EthernetPort)
+        except NoResourceFoundError:
+            resource = target.get_resource(NetworkService)
+            return resource.address
+
         matches = []
         for details in resource.extra.get('macs').values():
             ips = details.get('ips', [])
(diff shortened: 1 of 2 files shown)


BastianStender:
Anything left to do here?

Bastian-Krause:
Ping?

labgrid/remote/client.py
```diff
@@ -790,19 +786,34 @@ def _get_ip(self, place):
...
-        args = [
+        target = self._get_target(place)
+        env = os.environ.copy()
+        try:
+            resource = target.get_resource(NetworkService)
+            username = resource.username
+            # use sshpass if we have a password
+            if resource.password:
+                env['SSHPASS'] = resource.password
+            sshpass = ['sshpass', '-e'] if resource.password else []
```
maximilianriemensberger: You need to check if there the port is 22 or not. Otherwise it won't work with the ForwardedNetworkService.
```
port = resource.port or 22
```

labgrid/remote/client.py
```diff
@@ -790,19 +786,34 @@ def _get_ip(self, place):
...
+
+        except NoResourceFoundError:
+            username = 'root'
+            sshpass = []
+
```
maximilianriemensberger:
```
            port = 22
```

labgrid/remote/client.py
```diff
@@ -790,19 +786,34 @@ def _get_ip(self, place):
...
+        args = sshpass + [
             'ssh',
-            '-l', 'root',
+            '-l', username,
```
maximilianriemensberger:
```
'-p', str(port),
```

PR #351, opened 2018-11-11
fix managed SSH connections

jluebbe:
This fixes some edge-cases in the SSHConnection implementation:
- handing of existing SSH masters
- host key checking
- testing in travis and locally
- premature master timeout (when unused for a some minutes)
- unclear log messages


Emantor:
Thanks!

PR #354, opened 2018-11-12
RFC: Introduce ColoredStepReporter

BastianStender:
The StepReporter output can be quite overwhelming. This is an attempt to make the logs more readable.

Usage:
```
LG_COLOR_SCHEME=dark pytest --lg-colored-steps -vvs ..
LG_COLOR_SCHEME=light pytest --lg-colored-steps -vvs ..
```
The idea is that different events have different colors. The more colorful, the more important. In order to make less important output "blend into the background" two color schemes are available (dark/light terminal background). Default is "dark".

E.g. "expect" generates a lot of output. In my experience this is nice for debugging strategies and unexpected behavior of the DUT, but most of the time it's not in the focus of the test developer.

tbd:
- add more events
- fine tune colors


Bastian-Krause:
I guess adding more events could be a separate PR.

@Emantor, @ejoerns, @jluebbe: Do you have concerns with this PR (or the colors)? I'd like to see this mainline as it improves debugging - at least for some people.

PR #359, opened 2018-11-19
remote/client: add --sort-last-changed for places/resources subcommands

Bastian-Krause:
labgrid does not limit the number of places that can match a given resource. This very flexible perception of places results in places being created but not used (anymore). Also resources might get matchen by different places. At the end you might have server racks full of embedded boards blocking space and resources without being used (anymore). So I came up with two options:

`labgrid-client places --sort-last-changed` lists all places by their last changed date. Combined with `-v` the actual changed date is shown per place.

`labgrid-client resources --sort-by-matched-place-change` lists all resources by "their" last changed date and shows places matching the respective resource and their last changed date.

The first one should come quite self-explanatory. It can be used to delete places not used for a long time. Some words about the latter one:

It shows resources along with places (and their last changed date) matching the respective resource. The places per resource are sorted by changed date beginning with the most recent one. So the first listed place per resource can be interpreted as the "last changed date" of the resource.

The list of resources is sorted by exactly that "last changed date" of the resource.

**Purpose**
This enables the user to
  a) identify resources (e.g. power ports) not used for a long time
  b) identify resources matched by multiple places

Use case a) is especially useful when combined with a match given after the "resources" subcommand to identify resources by exporter/group/class/name or even combined, e.g.:

`labgrid-client resources --sort-by-matched-place-change rack1/*/NetworkPowerPort`

This shows all NetworkPowerPorts in "rack1" sorted by last changed date in descending order.

**Attention:** the changed date might not reflect the last use of a place. In case a place is acquired (locked) for a long time without being released (unlocked) the place might appear unused. Acquired places cannot be deleted, but (silently) reusing resources from other places is dangerous.

### DIFF
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -196,7 +196,12 @@ class ClientSession(ApplicationSession):
 
     async def print_places(self):
         """Print out the places"""
-        for name, place in sorted(self.places.items()):
+        if self.args.sort_last_changed:
+            places = sorted(self.places.items(), key=lambda p: p[1].changed)
+        else:
+            places = sorted(self.places.items())
+
+        for name, place in places:
             if self.args.acquired and place.acquired is None:
                 continue
             if self.args.verbose:
@@ -1029,6 +1034,8 @@ def main():
     subparser = subparsers.add_parser('places', aliases=('p',),
                                       help="list available places")
     subparser.add_argument('-a', '--acquired', action='store_true')
+    subparser.add_argument('--sort-last-changed', action='store_true',
+                           help='sort by last changed date (oldest first)')
     subparser.set_defaults(func=ClientSession.print_places)
 
     subparser = subparsers.add_parser('who',
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -156,6 +156,13 @@ class ClientSession(ApplicationSession):
             for name in sorted(self.places.keys()):
                 print(name)
 
+    def _get_places_by_resource(self, resource_path):
+        """Yield Place objects that match the given resource path"""
+        for place in self.places.values():
+            match = place.getmatch(resource_path)
+            if match:
+                yield place
+
     async def print_resources(self):
         """Print out the resources"""
         match = ResourceMatch.fromstr(self.args.match) if self.args.match else None
@@ -179,7 +186,7 @@ class ClientSession(ApplicationSession):
                     filtered[exporter][group_name][resource_name] = resource
 
         # print the filtered resources
-        if self.args.verbose:
+        if self.args.verbose and not self.args.sort_by_matched_place_change:
             for exporter, groups in sorted(filtered.items()):
                 print("Exporter '{}':".format(exporter))
                 for group_name, group in sorted(groups.items()):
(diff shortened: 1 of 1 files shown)


labgrid/remote/client.py
```diff
@@ -189,14 +196,41 @@ def onChallenge(self, challenge):
...
         else:
+            results = []
             for exporter, groups in sorted(filtered.items()):
                 for group_name, group in sorted(groups.items()):
                     for resource_name, resource in sorted(group.items()):
-                        print("{}/{}/{}".format(exporter, group_name, resource.cls))
+                        resource_path = (exporter, group_name, resource.cls, resource_name)
+                        places = list(self._get_places_by_resource(resource_path))
+                        # most recently changed place first
+                        places = sorted(places, key=lambda p: p.changed, reverse=True)
```
Emantor: The places computation and sort are done unconditionally and are not used in the old code path.
Please move the computation so the previous old code path remains at the same speed as before.

jluebbe: Or just move it into a sperate command?

labgrid/remote/client.py
```diff
@@ -1023,12 +1057,16 @@ def main():
                                       help="list available resources")
     subparser.add_argument('-a', '--acquired', action='store_true')
     subparser.add_argument('-e', '--exporter')
+    subparser.add_argument('--sort-last-changed', action='store_true',
+                           help="sort by matched place's changed date and show place and date")
```
jluebbe: It's unclear which order this uses. Should that be configurable?

jluebbe: Hmm, currently not. "--sort-last-changed" doesn't sound like it would do anything with places, tough.

labgrid/remote/client.py
jluebbe: Does this really have an effect on performance? If not, should we always show the matching places?

labgrid/remote/client.py
Bastian-Krause: Right. I struggled to come up with a good name here. Maybe `--sort-by-matched-place-change`?

labgrid/remote/client.py
Bastian-Krause: I did not touch the output of `labgrid-client resources` to keep it stable (e.g. for scripting purposes, e.g. tab completion).

Regarding performance: for each resource all places' getmatch() is called. This uses fnmatchcase() up to 4 times (exporter, group, class, name). fnmatchcase does caching, but [only the 256 most recent patterns](https://github.com/python/cpython/blob/3.6/Lib/fnmatch.py#L38). So yes, I guess there are differences in performance, especially for larger setups.

Bastian-Krause: Place computation is now done conditionally.

labgrid/remote/client.py
Bastian-Krause: I renamed the option for the resources subcommand to `--sort-by-matched-place-change` and mentioned the sort order in the help text.

Bastian-Krause:
Pushed changes:
- rebased on current master
- renamed resources subcommand option to `--sort-by-matched-place-change`
- mentioned the sort order in help texts of both options
- made the place computation for `labgrid-client resources --sort-by-matched-place-change` conditionally to avoid performance drops in `labgrid-client resources`

Bastian-Krause:
Anything left to do to get this merged?

PR #360, opened 2018-11-22
resource/udev: fix missing else in USBSDMuxDevice update function

Emantor:
The function would try to iterate on the NoneType object of no device was
available. Add the else path to only iterate on the device children if the
device is available

Fixes: 9ef7746 ("resource/udev: use update function for USBSDMuxDevice")

### DIFF
--- a/labgrid/resource/udev.py
+++ b/labgrid/resource/udev.py
@@ -301,6 +301,7 @@ class USBSDMuxDevice(USBResource):
         if not self.device:
             self.control_path = None
             self.disk_path = None
+            return
         for child in self.device.children:
             if child.subsystem == 'block':
                 self.disk_path = child.device_node


labgrid/resource/udev.py
```diff
@@ -301,11 +301,12 @@ def update(self):
         if not self.device:
             self.control_path = None
             self.disk_path = None
```
Bastian-Krause: I'd prefer a `return` here.

Emantor: Can you elaborate on why?

ejoerns: I would assume to reduce unnecessary nesting depth for conditionals.
Sounds legitimate.

Bastian-Krause: Correct, drop the else and return on `not self.device`.

Emantor:
Force Pushed, please check again.

PR #362, opened 2018-11-22
target: strategy/graphstrategy: simplify driver deactivation, deactivate before root state

Bastian-Krause:
Let the graph strategy deactivate all drivers before root state. This leads to more consistent driver handling in strategies. Simplify driver deactivation while at it.


Emantor:
I'll run this through another test suite internally before merging.

Emantor:
Ran without errors, @jluebbe this is ready for merge.

PR #363, opened 2018-11-23
Add write-image subcommand to the client, fix udev detection and check device path

Emantor:
## Description
Add the write-image command to the client. This allows the writing of images to USB mass storage devices.
Also fix the disk_path property of the usbsdmux.
The `NetworkUSBStorageDriver` has been enhanced with a check for the `path` property of the resource, since the USBSDMux can be available, but in `DUT` mode does not expose a usb storage device.

## Task List
- [ ] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/labgrid/driver/networkusbstoragedriver.py
+++ b/labgrid/driver/networkusbstoragedriver.py
@@ -10,6 +10,8 @@ from ..resource.remote import NetworkUSBMassStorage, NetworkUSBSDMuxDevice
 from ..step import step
 from ..util.managedfile import ManagedFile
 from .common import Driver
+from ..driver.exception import ExecutionError
+
 
 @target_factory.reg_driver
 @attr.s(cmp=False)
@@ -35,6 +37,10 @@ class NetworkUSBStorageDriver(Driver):
 
     @step(args=['filename'])
     def write_image(self, filename):
+        if not self.storage.path:
+            raise ExecutionError(
+                "{} is not available".format(self.storage_path)
+            )
         mf = ManagedFile(filename, self.storage)
         mf.sync_to_resource()
         self.logger.info("pwd: %s", os.getcwd())
@@ -50,6 +56,10 @@ class NetworkUSBStorageDriver(Driver):
 
     @step(result=True)
     def get_size(self):
+        if not self.storage.path:
+            raise ExecutionError(
+                "{} is not available".format(self.storage_path)
+            )
         args = ["cat", "/sys/class/block/{}/size" % self.storage.path[5:]]
         size = subprocess.check_output(self.storage.command_prefix + args)
         return int(size)
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -950,6 +950,33 @@ class ClientSession(ApplicationSession):
         for k, v in sorted(data.items()):
             print("{:<16s} {:<10s}".format(k, str(v)))
 
+    def write_image(self):
+        place = self.get_acquired_place()
+        target = self._get_target(place)
+        drv = None
+        from ..resource.remote import NetworkUSBMassStorage, NetworkUSBSDMuxDevice
+        from ..driver import NetworkUSBStorageDriver
+        try:
+            drv = target.get_driver(NetworkUSBStorageDriver)
+        except NoDriverFoundError:
+            for resource in target.resources:
+                if isinstance(resource, (NetworkUSBSDMuxDevice, NetworkUSBMassStorage)):
+                    try:
+                        drv = target.get_driver(NetworkUSBStorageDriver)
+                    except NoDriverFoundError:
+                        drv = NetworkUSBStorageDriver(target, name=None)
+                    drv.storage.timeout = self.args.wait
+                    break
+        if not drv:
+            raise UserError("target has no compatible resource available")
+        target.activate(drv)
+        try:
+            drv.write_image(self.args.filename)
+        except subprocess.CalledProcessError as e:
+            raise UserError("could not write image to network usb storage: {}".format(e))
+        except FileNotFoundError as e:
+            raise UserError(e)
+
 def start_session(url, realm, extra):
     from autobahn.wamp.types import ComponentConfig
     from autobahn.websocket.util import parse_url
@@ -1240,6 +1267,11 @@ def main():
     tmc_subparser.add_argument('channel', type=int)
     tmc_subparser.add_argument('action', choices=['info', 'values'])
     tmc_subparser.set_defaults(func=ClientSession.tmc_channel)
+
+    subparser = subparsers.add_parser('write-image', help="write an image onto mass storage")
+    subparser.add_argument('-w', '--wait', type=float, default=10.0)
+    subparser.add_argument('filename', help='filename to boot on the target')
+    subparser.set_defaults(func=ClientSession.write_image)
 
     # make any leftover arguments available for some commands
     args, leftover = parser.parse_known_args()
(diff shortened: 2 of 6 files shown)


labgrid/driver/networkusbstoragedriver.py
```diff
@@ -23,6 +26,21 @@ class NetworkUSBStorageDriver(Driver):
         },
     }
 
+    def _check_available(func): # pylint: disable=no-self-argument
+        """Check if a mass storage device is available as a decorator"""
+
+        @wraps(func)
+        def wrapper(self, *_args, **_kwargs):
+            if not self.storate.path:
```
Bastian-Krause: s/storate/storage/

labgrid/driver/networkusbstoragedriver.py
```diff
@@ -23,6 +26,21 @@ class NetworkUSBStorageDriver(Driver):
         },
     }
 
+    def _check_available(func): # pylint: disable=no-self-argument
+        """Check if a mass storage device is available as a decorator"""
```
Bastian-Krause: Shouldn't we really check the path existence here?

labgrid/remote/client.py
```diff
@@ -894,6 +894,29 @@ def tmc_channel(self):
...
+        args = self.args.filename
+        target = self._get_target(place)
+        drv = None
+        from ..resource.remote import NetworkUSBMassStorage, NetworkUSBSDMuxDevice
+        from ..driver import NetworkUSBStorageDriver
+        try:
+            drv = target.get_driver(NetworkUSBStorageDriver)
+        except NoDriverFoundError:
+            for resource in target.resources:
+                if isinstance(resource, NetworkUSBSDMuxDevice) or isinstance(resource, NetworkUSBMassStorage):
```
Bastian-Krause: Use shorter `isinstance(resource, (NetworkUSBMassStorage, NetworkUSBSDMuxDevice))`.

labgrid/remote/client.py
```diff
@@ -894,6 +894,29 @@ def tmc_channel(self):
         for k, v in sorted(data.items()):
             print("{:<16s} {:<10s}".format(k, str(v)))
 
+    def write_image(self):
+        place = self.get_acquired_place()
+        args = self.args.filename
```
Bastian-Krause: Where is `args` used?

labgrid/driver/networkusbstoragedriver.py
Emantor: We would have to call into the remote machine to check that. I don't think its worth it to check this here, its better to fix the reporter to only export existing paths.

labgrid/remote/client.py
```diff
@@ -894,6 +894,28 @@ def tmc_channel(self):
...
+        place = self.get_acquired_place()
+        target = self._get_target(place)
+        drv = None
+        from ..resource.remote import NetworkUSBMassStorage, NetworkUSBSDMuxDevice
+        from ..driver import NetworkUSBStorageDriver
+        try:
+            drv = target.get_driver(NetworkUSBStorageDriver)
+        except NoDriverFoundError:
+            for resource in target.resources:
+                if isinstance(resource, NetworkUSBSDMuxDevice, NetworkUSBMassStorage):
```
Bastian-Krause: `TypeError: isinstance expected 2 arguments, got 3`
This should be:
`if isinstance(resource, (NetworkUSBMassStorage, NetworkUSBSDMuxDevice)):`

labgrid/remote/client.py
```diff
@@ -894,6 +894,28 @@ def tmc_channel(self):
...
+                    try:
+                        drv = target.get_driver(NetworkUSBStorageDriver)
+                    except NoDriverFoundError:
+                        drv = NetworkUSBStorageDriver(target, name=None)
+                    drv.storage.timeout = self.args.wait
+                    break
+        if not drv:
+            raise UserError("target has no compatible resource available")
+        target.activate(drv)
+        drv.write_image(self.args.filename)
```
Bastian-Krause:
```suggestion
        try:
            drv.write_image(self.args.filename)
        except subprocess.CalledProcessError as e:
            raise UserError("Could not write image to network usb storage: {}".format(e))
```

labgrid/remote/client.py
```diff
@@ -1181,6 +1203,11 @@ def main():
     tmc_subparser.add_argument('action', choices=['info', 'values'])
     tmc_subparser.set_defaults(func=ClientSession.tmc_channel)
 
+    subparser = subparsers.add_parser('write-image', help="Write an image onto mass storage")
```
Bastian-Krause: The majority of help texts start lower case.

labgrid/remote/client.py
Emantor: Nice catch with the `FileNotFoundError` raised by the `ManagedFile`.

labgrid/remote/client.py
Emantor: Fixed.

Bastian-Krause:
Looks good, only one small improvement.

labgrid/resource/udev.py
```diff
@@ -289,24 +289,51 @@ class USBSDMuxDevice(USBResource):
...
+            if not self.avail:
+                for child in self.device.children:
+                    if child.subsystem == 'block':
+                        self.disk_path = child.device_node
+                    elif child.subsystem == 'scsi_generic':
+                        self.control_path = child.device_node
+        if self.disk_path and self.control_path:
+            self.internal_avail = True
+        else:
+            self.internal_avail = False
```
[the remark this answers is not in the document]
Emantor: Good catch, fixed as suggested.

Emantor:
@Bastian-Krause Testing revealed that the variable would be set to the value of `self.disk_path`, please take a look at the current version.

labgrid/resource/udev.py
```diff
@@ -289,29 +289,48 @@ class USBSDMuxDevice(USBResource):
...
         self.control_path = None
         self.disk_path = None
         super().__attrs_post_init__()
 
-    def update(self):
-        block_found = False
+    # Overwrite the avail attribute with our internal property
+    @property
+    def avail(self):
+        return self.internal_avail
```
jluebbe: Can't we just check self.disk_path and self.control_path here instead of adding another attr above?

labgrid/resource/udev.py
```diff
@@ -289,29 +289,48 @@ class USBSDMuxDevice(USBResource):
...
-        for child in self.device.children:
-            if child.subsystem == 'block':
-                self.disk_path = child.device_node
-                block_found = True
-            if child.subsystem == 'scsi_generic':
-                self.control_path = child.device_node
-        if not block_found:
-            self.disk_path = None
+        else:
+            if not self.avail:
```
jluebbe: Using the property insead of the attr here seems strange. `if not (self.disk_path and self.control_path)`?

labgrid/driver/networkusbstoragedriver.py
```diff
@@ -23,6 +26,21 @@ class NetworkUSBStorageDriver(Driver):
...
+
+        @wraps(func)
+        def wrapper(self, *_args, **_kwargs):
```
jluebbe: Just use a normal function here, the complexity of a decorator isn't work it for the two functions below (just call the check explicitly as the first step in the function).
Also, without the decorator, the backtraces are cleaner.

Emantor:
This needs to be tested again with an SD-MUX.

Bastian-Krause:
> This needs to be tested again with an SD-MUX.

I've tested it successfully just now.

Emantor:
Running this PR with an exporter and an SD-Mux? I should have been more explicit in what needs to be tested.

Bastian-Krause:
Oh, no, I tested only the client-side.

Emantor:
This is ready for review again. I also tested the SD-Mux availability locally and it now works correctly.

PR #367, opened 2018-12-07
serialdriver: ignore SET_CONTROL answers when using rfc2217

lrntplc:
When using non compliant servers we may get a "ValueError: remote
rejected value for option 'control'" exception.

However, since Labgrid is not using HW flow control at all, we can
safely ignore the SET_CONTROL responses.

According to pyserial documentation, this can be achieved by adding the
'ign_set_control' option in the rfc2217 URL.

Signed-off-by: Laurentiu Palcu <laurentiu.palcu@nxp.com>


jluebbe:
Thanks for the contribution, @lrntplc!

PR #368, opened 2018-12-10
docs: improve GraphStrategy docs

fscherf:
Signed-off-by: Florian Scherf <f.scherf@pengutronix.de>


fscherf:
@Emantor, @jluebbe 

Any thoughts?

PR #374, opened 2019-01-11
Pr/20190109 enable dependency managing through pipenv

kkreitmair:
**Description of feature**
The pull request will add the Files **Pipfile** and **Pipfile.lock**. This one are needed for handling dependency management through **pipenv**.
Through the pinned dependency's and subdependency's, as you can see in Pipfile.lock, it is possiblie to get a reproducible installation of labgrid.
Reproducible installations is useful for developing environments or deployment of labgrid.

Everything about pipenv can be read [here](https://pipenv.readthedocs.io/en/latest/).

The **current requirements.txt** files can be kept or fully removed.

**Verifying the functionality:**
```
// Requires
$ apt-get install python3 python3-pip
$ pip3 install pipenv

// after cloning labgrid project
$ cd labgrid
$ pipenv install
$ pipenv run labgrid-client

// To install dev-packages
$ pipenv install --dev
$ pipenv run crossbar -h
```
**Common general Workflow with pipenv**
```
$ pipenv --three  // generating an virtual environment for python3
$ pipenv install <package> // installing it to virtual environment
$ pipenv shell // enterning virtual environment
// use installed packages
// if everything is fine then
$ pipenv lock  // for generate Pipfile.lock (file with pinned dependencies)
$ pipenv exit // leaving the virtual environment
```
**Special handlings for labgrid project**
-  optional dependencies
   Uncomment the dependencies you need  (like onewire) in section packages of Pipfile and than run:
   ```
   $ pipenv install
   $ pipenv lock
   ```
-  not needed dev dependencies
   Comment out the dependencies you need (like sphinx) in section dev-packages of Pipfile and than run:
   ```
   $ pipenv install
   $ pipenv lock
   ```
- generating deb-requirements.txt for dh_virtualenv
  Uncomment the dependencies in section packages of Pipfile and than run:
   ```
   $ pipenv lock -r --dev > deb-requirements.txt
   ```
- installing everything for travis
  Travis dependencies were included to dev-packages section of Pipfile
   ```
   $ pipenv install --dev
   ```

**Open Questions**
- Where to put the documentation of it?

**Checklist**
- [ ] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -1,0 +1,41 @@
+[[source]]
+url = "https://pypi.org/simple"
+verify_ssl = true
+name = "pypi"
+
+[dev-packages]
+coveralls = ">=1.3.0"
+codecov = ">=2.0.15"
+pytest-cov = "==2.6.1"
+pytest-isort = "==0.2.1"
+pytest-mock = "==1.10.0"
+pytest-pylint = "==0.13.0"
+pytest-runner = "==4.2"
+pytest-dependency = "==0.4.0"
+yapf = "==0.25.0"
+docutils = "==0.14"
+graphviz = "==0.10.1"
+Sphinx = "==1.8.3"
+sphinx_rtd_theme = "==0.4.2"
+
+[packages]
+attrs = "==18.2.0"
+pexpect = "==4.6"
+pyserial = {git = "https://github.com/labgrid-project/pyserial.git", ref="v3.4.0.1", editable=true}
+pytest = "==4.0.2"
+pyudev = "==0.21.0"
+requests = "==2.21.0"
+xmodem = "==0.4.5"
+autobahn = "==18.12.1"
+ansicolors = "==1.1.8"
+"Jinja2" = "==2.10"
+PyYAML = "==3.13"
+labgrid = {editable = true, path = "."}
+pyModbusTCP = "==0.1.8"
+setuptools = ">=38.0.0"
+crossbar = "==18.12.1"
+idna = "==2.5"
+onewire = "==0.2"
+pysnmp = "==4.4.8"
+pysnmp-mibs = "==0.1.6"
+xenavalkyrie = "==1.2.1"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,0 +1,1187 @@
+{
+    "_meta": {
+        "hash": {
+            "sha256": "c86acceaaba52e9abedf42702af631136c13418ff6d81ea28c6361fe6ec59a47"
+        },
+        "pipfile-spec": 6,
+        "requires": {},
+        "sources": [
+            {
+                "name": "pypi",
+                "url": "https://pypi.org/simple",
+                "verify_ssl": true
+            }
+        ]
+    },
+    "default": {
+        "ansicolors": {
+            "hashes": [
+                "sha256:00d2dde5a675579325902536738dd27e4fac1fd68f773fe36c21044eb559e187",
+                "sha256:99f94f5e3348a0bcd43c82e5fc4414013ccc19d70bd939ad71e0133ce9c372e0"
+            ],
+            "index": "pypi",
+            "version": "==1.1.8"
+        },
+        "argh": {
+            "hashes": [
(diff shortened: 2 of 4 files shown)


Emantor:
According to pypa/pipenv#3167 and pypa/pipenv#2729, pipenv currently does not support multiple package lists. In you current PR, this is done by modifying the pipenv file, which is not a good solution to the problem.
So the discussion is either whether we should remove the optional dependencies or wait for pipenv to implement support.

Emantor:
I think its fine to include all dependencies.

kkreitmair:
Fine.
In which sections do you think this should be documented?
In my opinion it should appear in **Getting Started** and **Development** section nearby the conventional installation sections.
Do you find a section helpful, which will describe the update workflow for the Pipfiles? Should such a section be integrated in the Development section?

jluebbe:
I think we'll need to keep the requirements files at least for some time. The dh_virtualenv-based Debian packages use them and dh-pipenv is not available in Debian yet.

If you think this is useful even if we keep both in parallel, I'm fine with add pipenv support.

> Fine.
> In which sections do you think this should be documented?
> In my opinion it should appear in **Getting Started** and **Development** section nearby the conventional installation sections.
> Do you find a section helpful, which will describe the update workflow for the Pipfiles? Should such a section be integrated in the Development section?

I'd prefer that in the development section. Maybe with a pointer from getting started.

Pipfile
```diff
@@ -0,0 +1,50 @@
+[[source]]
+url = "https://pypi.org/simple"
+verify_ssl = true
+name = "pypi"
+
+[requires]
+python_version = "3.5"
```
jluebbe: Does this need a >=?

Pipfile
```diff
@@ -0,0 +1,50 @@
...
+crossbar = "==18.12.1"
+idna = "==2.5"
+onewire = "==0.2"
+pysnmp = "==4.4.6"
+pysnmp-mibs = "==0.1.6"
+xenavalkyrie = "==1.1.2"
+graphviz = "==0.8.4"
+Sphinx = "==1.8.1"
+sphinx_rtd_theme = "==0.4.1"
+pyModbusTCP = "==0.1.7"
```
jluebbe: crossbar, idna, onewire, pysnmp(-mibs), xenavalkyrie and pyModbusTCP are normal dependencies.

Pipfile
```diff
@@ -0,0 +1,50 @@
...
+attrs = "==17.4.0"
+pexpect = "==4.6"
+pyserial = {git = "https://github.com/labgrid-project/pyserial.git", ref="v3.4.0.1", editable=true}
+pytest = "==4.0.2"
+pyudev = "==0.21.0"
+requests = "==2.19.1"
+xmodem = "==0.4.5"
+autobahn = "==18.12.1"
+ansicolors = "==1.1.8"
+"Jinja2" = "==2.10"
```
jluebbe: Is there a reason why Jinja2 is quoted?

Pipfile
```diff
@@ -0,0 +1,50 @@
...
+PyYAML = "==3.13"
+"e1839a8" = {editable = true, path = "."}
```
jluebbe: What is the meaning of this "magic" name?

Pipfile
kkreitmair: Pipenv pins every dependency so does the Python version. If this doesn't fit our requirements, we have to remove the **[requires]** section. For now it is not supported by pipenv to specify a range of Python versions.

Pipfile
kkreitmair: Ok. I will change this.

Pipfile
kkreitmair: This is automatically generated by pipenv. I don't know wy this is needed.

Pipfile
kkreitmair: The magic name is automatically generated by pipenv for the labgrid project dependency. For better readability, i will change this to labgrid.

kkreitmair:
Sorry. I had some problems with git.

Emantor (dismissed):
LGTM

Emantor:
We are fine with using requirements, thanks for the effort put into this, we'll stick with requirements for now.

PR #376, opened 2019-01-15
linting: several small improvements, deprecation fix, indentation fix

Bastian-Krause:
**Description**
Usual mix of linting fixes (nothing to worry about), a fix for the deprecated load_module() and an indentation fix (introduced by #354, c47fef7).

**Checklist**
- [x] PR has been tested


labgrid/util/ssh.py
```diff
@@ -298,18 +299,18 @@ def _check_external_master(self):
         if check == 0:
             self._logger.debug("Found existing control socket")
             return True
-        elif b"No such file or directory" in stdout:
+        if b"No such file or directory" in stdout:
```
jluebbe: This and the following change should be in one of the other commits.

Bastian-Krause: Right, I must have messed up here :/

Bastian-Krause:
Force-pushed:
- moved lazy-logging modification to the correct commit
- moved fix for no-else-return indication to correct commit

PR #379, opened 2019-01-19
LED testing via sysfs

metux:
The driver uses CommandProtocol to run shell commands and checks the results.

* get_brightness(name, val):

    read back the current brightness of given LED

* set_brightness(name, val):

    set the brightness of a specific LED, read back and compare

**Description**

**Checklist**
- [ ] Tests for the feature 

- [ ] PR has been tested

### DIFF
--- a/labgrid/driver/__init__.py
+++ b/labgrid/driver/__init__.py
@@ -25,3 +25,4 @@ from .networkusbstoragedriver import NetworkUSBStorageDriver
 from .resetdriver import DigitalOutputResetDriver
 from .serialdigitaloutput import SerialPortDigitalOutputDriver
 from .xenadriver import XenaDriver
+from .leddriver import LedDriver
--- a/labgrid/driver/leddriver.py
+++ b/labgrid/driver/leddriver.py
@@ -1,0 +1,36 @@
+import attr
+
+from ..factory import target_factory
+from ..protocol import LedProtocol, CommandProtocol
+from .common import Driver
+
+@target_factory.reg_driver
+@attr.s(cmp=False)
+class LedDriver(Driver, LedProtocol):
+    bindings = { "command": CommandProtocol }
+
+    def __attrs_post_init__(self):
+        super().__attrs_post_init__()
+
+    def led_fn(self, name):
+        return "/sys/class/leds/"+name+"/brightness"
+
+    @Driver.check_active
+    def write_brightness(self, name, val):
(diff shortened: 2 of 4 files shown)


jluebbe:
I don't think that this should be a driver, because it doesn't need the integration with the Target, Resources and other Drivers. This way you'd have the disadvantages (needs to be configured in the environment, boilerplate code, unergonomic handling of multiple instances), but don't use the advantages (enabling/disabling based on bindings, multiple alternative implementations of the same protocol, priorities, information about hardware resources).
Instead, it is much simpler to just have set/get_brightness helpers, which can be called with a CommandProtocol implementation and LED name (and a value for set_).

metux:
> I don't think that this should be a driver, because it doesn't need the integration with the Target, Resources and other Drivers. This way you'd have the disadvantages (needs to be configured in the environment, boilerplate code, unergonomic handling of multiple instances), but don't use the advantages (enabling/disabling based on bindings, multiple alternative implementations of the same protocol, priorities, information about hardware resources).

IMHO, the same argument would also apply to InfoProtocol/InfoDriver.

jluebbe:
> IMHO, the same argument would also apply to InfoProtocol/InfoDriver.
Yes, but I didn't see that clearly enough when we merged the InfoDriver. With the experience I have today, I'd prefer to move that functionality into a helper library.

metux:
> > IMHO, the same argument would also apply to InfoProtocol/InfoDriver.
> 
> Yes, but I didn't see that clearly enough when we merged the InfoDriver. With the experience I have today, I'd prefer to move that functionality into a helper library.

hmm, I personally think that such stacked drivers w/ very highlevel functionality are a very good idea.
similar to what we're doing in the kernel - I guess you wouldn't propose dropping the LED subsystem and tell folks to finger w/ raw gpio's directly ;-)

actually, I think the term "LED" might be semantically sub-optimal, as it's also used for generic switches (I've been thinking about an generic switch/relais subsystem, next to the LED subsystem, in the kernel - for more semantic correctness and maybe other features like hold-times, oneshot-triggers, ...).

maybe we could have a more in-depth talk about that: +49-151-27565287

--mtx

Emantor:
Hi, we now have the labgrid-project/labgrid-helper repository to collect these helper functions. The InfoDriver functions have already been ported to this repository and are up for removal in the labgrid core.
I'd like to invite you to contribute your LED testing to the labgrid-helper repository instead.

jluebbe:
This functionality belongs in https://github.com/labgrid-project/labgridhelper.

PR #380, opened 2019-01-21
tests: check that the pytest plugin loads correctly and --help still works

jluebbe:
**Description**

This should catch problems like the one fixed in #377.

**Checklist**
- [x] Tests for the feature 
- [x] PR has been tested


Emantor:
Thanks!

PR #381, opened 2019-01-25
remote/client: let fastboot use driver functions

Emantor:
**Description**
Fix the fastboot client command to use the driver functions which use the ManagedFile.

**Checklist**
- [ ] PR has been tested

### DIFF
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -740,6 +740,12 @@ class ClientSession(ApplicationSession):
             drv = AndroidFastbootDriver(target, name=None)
         drv.fastboot.timeout = self.args.wait
         target.activate(drv)
+        if args[0] == 'flash':
+            drv.flash(args[1], args[2])
+            return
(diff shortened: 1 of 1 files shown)


labgrid/remote/client.py
```diff
@@ -740,6 +740,12 @@ def fastboot(self):
             drv = AndroidFastbootDriver(target, name=None)
         drv.fastboot.timeout = self.args.wait
         target.activate(drv)
+        if args[0] == 'flash':
+            drv.flash(args[1], args[2])
+            return
+        if args[0] == 'boot':
+            drv.boot(args[1])
+            return
```
Bastian-Krause: Is there any particular reason for comparing args[0] with 'boot' and 'flash' here again? Can we move getting the driver to start of the method and use the existing ifs instead of comparing again?

jluebbe: I'd prefer to check the arguments before doing things that might fail for other reasons.

PR #384, opened 2019-02-19
docker: coordinator: fix port number

schmidtandreas:
The default port for coordinator, defined in crossbar configuration
file, is 20408.

Signed-off-by: Andreas Schmidt <mail@schmidt-andreas.de>


Emantor:
Thanks!

PR #387, opened 2019-02-27
Allow use of Python3.7 features

trxanga:
Signed-off-by: Anders Gammelgaard <anga@triax.com>

**Description**

Python 3.7 has new asyncio-based features. To use these together with labgrid from within a docker container, it helps to base the container on Python 3.7. This is easy to accomplish if labgrid can use Python 3.7 directly. This seems to be the case already for the tip of master and only needs to be stated explicitly. Besides, the 3.7 version should be used in relevant places - in particular docker. There may be more places, however, than I am aware of.

**Checklist**
Added new tox.ini environment py37 and ``tox -r`` seems to run flawlessly.

Summary for CHANGES.rst file:
- docker images for client, exporter and coordinator will now use Python3.7.

### DIFF
--- a/docker/client/Dockerfile
+++ b/docker/client/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.6
+FROM python:3.7
 MAINTAINER eha@deif.com
 
 RUN mkdir -p /opt/labgrid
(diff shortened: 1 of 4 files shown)


Emantor:
@esben are you fine with the python3.7 update?

esben:
Yes :+1:

Emantor:
Thanks!

PR #390, opened 2019-03-01
Modbus coil power driver

schmidtandreas:
**Description**

This PR add power driver for ModbusTCP Coil.

I tested it on WAGO PFC200-8215 PLC with DigitalOutput as ModbusTCP-Client.
- First test was using ModbusTCPCoilPowerDriver in a local test python script.
- Second test was labgrid-client power on/off.

**Checklist**
- [x] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,11 @@ New Features in 0.3.0
 - labgrid-client ``write-image`` subcommand: labgrid client now has a
   ``write-image`` command to write images onto block devices.
 
+New and Updated Drivers in 0.3.0
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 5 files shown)


jluebbe:
Thanks for your contribution, but it'd like to avoid duplicating the code from labgrid/driver/modbusdriver.py in a second driver. You can use the [DigialOutputPowerDriver](https://labgrid.readthedocs.io/en/latest/configuration.html#digitaloutputpowerdriver) to adapt between the ModbusCoilDriver and the PowerProtocol. To use that with labgrid-client, you can provide an environment yaml file to it, so it doesnt' need to autodetect the drivers to use.

schmidtandreas:
Ahhh, ok. Thanks for advise!

PR #394, opened 2019-03-05
driver/shelldriver: fix return None for run_script() and run_script_f…

yukach:
fix invalid returned None value in the run_script() and run_script_file() of driver/shelldriverand.


Emantor:
Thanks!

PR #395, opened 2019-03-05
Documentation improvements

rohieb:
Two small editorial changes and then one new section:
- getting started: move udev section into exporter section
- configuration: explain options, images, tools, and example keys
- getting started: give a short introduction into the library

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -1579,6 +1579,10 @@ would resolve the qemu_bin path relative to the BASE dir of the YAML file and
 try to use the RemotePlace with the name set in the LG_PLACE environment
 variable.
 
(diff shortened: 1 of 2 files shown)


rohieb:
Rebased to current master to resolve merge conflicts

jluebbe:
The dbus-matching can be used the same way in local environment yaml-files and for configuring the exporter. It would be useful to state that explicitly, though.
Using labgrid as a library is already described in https://labgrid.readthedocs.io/en/latest/usage.html#library, but your changes could probably be integrated there.

rohieb:
After re-reading the existing documentation, I noticed that much of this pull request was already there, I just didn't see it. So just do the docstring fixes and the small cross reference in the man page for now.

PR #399, opened 2019-03-20
use the SVG version of the official logo

rohieb:
All relevant browsers [can render SVG natively][1], as well as most image
viewers. Throw in a width for good measure (and old Microsoft browsers),
and remove the pre-rendered PNG.

[1]: https://caniuse.com/#feat=svg


rohieb:
After discussion with Jan and Rouven, having a PNG is just enough for now.

PR #408, opened 2019-04-01
Add a FileDigitalOutputDriver

bgermann:
**Description**
This PR adds a FileDigitalOutputDriver. It writes arbitrary string representations of booleans to a file and reads string representations from it.

I tested it with a led available in sysfs.

**Checklist**
- [x] Tests for the feature 
- [x] PR has been tested

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,7 @@ Release 0.3.0 (unreleased)
 New Features in 0.3.0
 ~~~~~~~~~~~~~~~~~~~~~
 
+- The new `FileDigitalOutputDriver` respresents a digital signal with a file.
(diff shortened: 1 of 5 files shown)


jluebbe:
What is you use case for this driver? It seems that this could also be covered by using the pins directly as GPIOs?

bgermann:
It could but if you have the GPIO line exported as LED (and use this in another application) you cannot connect to the line because it is blocked by the led subsystem.

jluebbe:
OK. I don't have a good idea why you'd wan't to control a LED from the exporter, though. Or did you connect something else to the LEDs pin? :)
Could you describe your scenario in more detail?

jluebbe:
I've combined the code and documentation in one commit and added some simple tests.

bgermann:
Thank you!

PR #409, opened 2019-04-02
Update crossbar and fix authid

### LINKED ISSUE #406: Correct misuse of wamp authid
We are currently misusing the WAMP authid to distinguish exporters and clients. According to crossbario/autobahn-python#1162 this is an internal property used for the WAMP handshake.

Emantor:
**Description**
Update Crossbar and autobahn to the latest released versions. We are currently seeing a bug in out setup which results in the following log output:
```
Apr 01 09:23:16 labgrid crossbar[27698]: [Router      27705] INTERNAL WARNING: event for publication 1671249291329997 not in event store
Apr 01 09:23:16 labgrid crossbar[27698]: [Router      27705] INTERNAL WARNING: subscription 3503552087680833 for publication 1671249291329997 not in event store
```
A similar bugreport upstream (crossbario/crossbar#1459) has been closed as not reproducible on an updated version, so this bump should fix the bug.
**Checklist**
- [x] PR has been tested

Fixes #406

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,7 @@ Release 0.3.0 (unreleased)
 New Features in 0.3.0
 ~~~~~~~~~~~~~~~~~~~~~
 
+- Crossbar and autobahn have been updated to 19.3.3 and 19.3.5 respectively.
 - The InfoDriver was removed. The functions have been integrated into the
   labgridhelper library, please use the library for the old functionality.
 - labgrid-client ``write-image`` subcommand: labgrid client now has a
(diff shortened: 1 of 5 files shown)


jluebbe:
Does this have any impact on compatibility between older/newer clients?

Emantor:
Testing locally with a few client calls everything seems to work. However I see `INTERNAL WARNING` messages in the crossbar output of my local instance, maybe I am onto the wrong track here and the `WARNINGS` are not the cause for our issues?

Emantor:
The following Configurations have been tested:
- old coordinator with new exporter
- new coordinator with old and new exporter
- new coordinator with new exporter

Tested by using labgrid client to turn power on and off and looking at the `labgrid-client show` output.

PR #410, opened 2019-04-03
setup.py: add labgrid.util.agents

trxanga:
PR #397 introduced new Python module `labgrid.util.agents`. The module wasn't entered into `setup.py`, however.

When I use `tox -r` to test tip of master (currently commit 897b3a5), two tests in `tests/test_agent.py` fail. But with the proposed change all tests pass.

I have no idea why Travis didn't catch this error.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -49,6 +49,7 @@ setup(
         'labgrid.resource',
         'labgrid.strategy',
         'labgrid.util',
+        'labgrid.util.agents',
     ],
     # the following makes a plugin available to pytest
     entry_points={


Emantor:
Because travis does not use the tox to run the tests and . Please provide a signoff, patch looks good to me.

trxanga:
Hi Rouven

I have replaced the commit for PR #410 with one that is Signed-off. Hope this is enough to get the process going, but not sure.

Regards,
  Anders

________________________________

jluebbe:
@trxanga Thanks for the fix!

PR #412, opened 2019-04-08
driver/usbvideodriver: fix: select correct camera for stream

jgoetzinger:
The default device of v4l2src is /dev/video0. If the device is not specified and more than
one camera is connected, the default camera video0 will always be used.
The fix uses the devices video path to get the correct video device.

Tested with 2 cameras:
Without the fix:
labgrid-client -p board-1 video -> returns the stream of board-1
labgrid-client -p board-2 video -> returns the stream of board-1

With the fix:
labgrid-client -p board-1 video -> returns the stream of board-1
labgrid-client -p board-2 video -> returns the stream of board-2

**Checklist**
- [x] PR has been tested

### DIFF
--- a/labgrid/driver/usbvideodriver.py
+++ b/labgrid/driver/usbvideodriver.py
@@ -39,7 +39,7 @@ class USBVideoDriver(Driver):
         caps = self.select_caps(caps_hint)
 
         tx_cmd = self.video.command_prefix + ["gst-launch-1.0"]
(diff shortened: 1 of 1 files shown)


jgoetzinger:
Is there anything I need to do with this pull request?

Emantor:
@jluebbe this is ready for merge.

PR #413, opened 2019-04-10
driver/shelldriver: handle devices that do not have a login prompt

ejoerns:
For example when using a minimal initramfs system, we might not even
have a login prompt to catch. With the current implementation the only
way of handling this is to set 'login_prompt' to a string that will
never be machted which is quite inconvenient.

A better handling is to make the 'login_prompt' optional, respectively
allow passing `login_prompt=""`.

Signed-off-by: Enrico Jorns <ejo@pengutronix.de>

**Checklist**
- [ ] Tests for the feature 
- [x] PR has been tested

Please provide a comment if this is the way to go or if we care for API breakage here (because of moved attribute).

### DIFF
--- a/labgrid/driver/shelldriver.py
+++ b/labgrid/driver/shelldriver.py
@@ -35,8 +35,8 @@ class ShellDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
     """
     bindings = {"console": ConsoleProtocol, }
     prompt = attr.ib(validator=attr.validators.instance_of(str))
-    login_prompt = attr.ib(validator=attr.validators.instance_of(str))
     username = attr.ib(validator=attr.validators.instance_of(str))
+    login_prompt = attr.ib(default="", validator=attr.validators.instance_of(str))
     password = attr.ib(default="", validator=attr.validators.instance_of(str))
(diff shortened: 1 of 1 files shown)


basti2342:
While looking at this I noticed the default values for the ShellDriver's attributes are a little vague, at least `password` . There are cases where a password is requested and an empty password is regarded okay. This is currently not possible. I'd suggest to use `attr.ib(default=None, validator=attr.validators.optional(attr.validators.instance_of(str)`. This allows a string or `None`.

Looking at the introduced default `login_prompt` defaulting to an empty string and check for emptiness before feeding it into expect is might be okay. But IMO it's cleaner to default to `None` as suggested for `password`.

ejoerns:
I agree that there are boards that either have a) a password, b) no password, c) an empty password.
And that the 3rd case seems to be broken because of https://github.com/labgrid-project/labgrid/blob/master/labgrid/driver/shelldriver.py#L142

Thus, yes we should default to `None` there and allow setting empty string explicitly.
I don't expect it to make a difference for the login prompt, but I would not argue against keeping defaults consistent.

Emantor:
Okay, please rebase your PR and switch to None. I'll create an issue for the password entry.

Emantor:
Closing this, shelldriver has been refactored since.

PR #414, opened 2019-04-11
drivers: fastbootdriver: fix flashing and allow booting of preconfigured images

Bastian-Krause:
**Description**
This fixes and improves ways to use preconfigured images in the yaml config for the AndroidFastbootDriver.

**Checklist**
- [ ] Tests for the feature 
- [x] PR has been tested

### DIFF
--- a/labgrid/driver/fastbootdriver.py
+++ b/labgrid/driver/fastbootdriver.py
@@ -4,6 +4,7 @@ import attr
 from ..factory import target_factory
 from ..step import step
 from .common import Driver
+from ..exceptions import InvalidConfigError
 from ..util.managedfile import ManagedFile
 from ..util.helper import processwrapper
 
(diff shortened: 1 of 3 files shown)


Bastian-Krause:
> Please update CHANGES.rst

Done.

Emantor (dismissed):

labgrid/driver/fastbootdriver.py
```diff
@@ -54,7 +57,9 @@ def boot(self, filename):
 
     @Driver.check_active
     @step(args=['partition', 'filename'])
-    def flash(self, partition, filename):
+    def flash(self, partition, filename=None):
```
jluebbe: The other drivers support images don't have the complexity of multiple partitions.
For fastboot, there are no standardized partition names a far as I know.
So for flash(), making the filename configurable but not the partition seems strange.

Maybe something like this would be more useful:
```
…
AndroidFastbootDriver:
  boot_image: "kernel" # referring to a path under images
  flash_images:
    nand: "nand-image"
    nor: "nor-image"
…
```

Then the attribute would need to validate as a dict instance.
A flashall method could then iterate over all images, and flash with only a partition could lookup the correct image.

What do you think?

Bastian-Krause: Yes, I think that would be better. I don't know when I will have the time to implement that.

Bastian-Krause: Done.

Bastian-Krause:
Force-pushed: rebased on current master.

Emantor (approved):

jluebbe (approved):

Emantor:
Did a rebase, the reviews should still apply.

Emantor (approved):

PR #417, opened 2019-04-11
driver: fastbootdriver: echo command before executing it on run()

Bastian-Krause:
**Description**
The output of commands executed via 'fastboot oem exec <cmd>' cannot be
seen on the host. The command itself cannot be seen via serial on the
barebox shell.

This adds the possibility to echo the command before executing it. This
allows to see which commands lead to which (error) messages on the
serial interface without looking at labgrid's log files. This is
especially useful for scenarios without a SerialDriver, but with a
non-regular serial terminal connected for debugging only.

**Checklist**
- [ ] Tests for the feature 
- [x] PR has been tested

### DIFF
--- a/labgrid/driver/fastbootdriver.py
+++ b/labgrid/driver/fastbootdriver.py
@@ -61,7 +61,10 @@ class AndroidFastbootDriver(Driver):
 
     @Driver.check_active
     @step(args=['cmd'])
-    def run(self, cmd):
+    def run(self, cmd, *, echo_cmd=False):
+        if echo_cmd:
+            self("oem", "exec", "--", "echo {}".format(cmd))
+
         self("oem", "exec", "--", cmd)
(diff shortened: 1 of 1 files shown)


Bastian-Krause:
While working with this I stumbled upon..
`error: Command length (65) exceeds maximum size (64)`

This maximum size unfortunately counts "exec -- " (8 characters) adding "echo " (5 characters), so 51 characters left. We could do..
```
echo -n cmd_chunk[0]
echo -n cmd_chunk[1]
...
echo
```
.. or cut everything after the 64th character. But I am not sure it's worth the hassle... Maybe drop this PR altogether?

Emantor:
I wasn't aware of this limitation in fastboot, this effectively more than halves the available command length. I think dropping/closing is best.

PR #431, opened 2019-05-16
driver: fastbootdriver: don't disable sparsing by default

a3f:
**Description**

By default, the sparse threshold is computed by querying the
target's `max-download-size` variable.

With #416, labgrid adopted the behavior of passing `-S 0` as
default value to fastboot. This is problematic as different
fastboot versions interpret `-S 0` differently:

- Fastboot releases containing [542370dead ("Move fastboot over
  to ParseByteCount."](https://github.com/aosp-mirror/platform_system_core/commit/542370dead58e854c477549bf0b3333aaf23b299)) treat `-S 0` and its absence the same:
  The target's `max-download-size` variable is queried.
- Older fastboot releases, e.g. 1:7.0.0+r33-1 shipped by Debian
  Stretch, instead completely disables sparse when `-S 0`, while not
  supplying it queries the target.

To have the fastbootdriver interoperate with both versions,
supply `-S` only, when there's a user set value, and otherwise
do not supply the option at all.

Fixes: #416, 1bf8952 ("driver: fastbootdriver: allow setting sparse option")
Signed-off-by: Ahmad Fatoum &lt;a.fatoum@pengutronix.de&gt;

------

Bug fix has been tested by running `labgrid-client fastboot flash`, which no longer passes around a `-S 0`. I also tested that the old behavior still works, e.g. a config of
```
targets:
  main:
    drivers:
      AndroidFastbootDriver:
        sparse_size: 0M
```

yields the old behavior of no sparsing.

**Checklist**
- [ ] Tests for the feature 
- [x] PR has been tested

Everything else is IMO not applicable, because it's just a bug fix.

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -881,7 +881,9 @@ Implements:
 Arguments:
   - image (str): filename of the image to upload to the device
   - sparse_size (str): optional, sparse files greater than given size (see
-    fastboot manpage -S option for allowed size suffixes)
+    fastboot manpage -S option for allowed size suffixes). The default is the
+    same as the fastboot default, which is computed after querying the target's
+    ``max-download-size`` variable.
 
 OpenOCDDriver
 ~~~~~~~~~~~~~
--- a/labgrid/driver/fastbootdriver.py
+++ b/labgrid/driver/fastbootdriver.py
@@ -18,7 +18,7 @@ class AndroidFastbootDriver(Driver):
     }
 
     image = attr.ib(default=None)
-    sparse_size = attr.ib(default='0', validator=attr.validators.instance_of(str))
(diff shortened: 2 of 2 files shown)


a3f:
For reference, here is the relevant fastboot code: https://github.com/aosp-mirror/platform_system_core/blob/master/fastboot/fastboot.cpp#L824

a3f:
I inadvertently posted the wrong link. The issue I described doesn't exist anymore in the fastboot master branch. It does however exist in 1:7.0.0+r33-1 shipped by Debian Stretch and the fix is a "no-op" for new fastboot versions, so its equally applicable to both versions.

I revised the commit message to reflect this and added a small note about the default to the documentation.

PR #432, opened 2019-05-16
Add agent module, resource and driver for Dedidec Relais 8 device

jluebbe:
**Description**

This PR first adds an agent module to control the device remotely. Then it adds a resource and driver to allow local and remote control via the agent. An example for both variants is also included

**Checklist**
- [x] Tests for the feature 
- [x] PR has been tested

### DIFF
--- a/labgrid/util/agents/deditec_relais8.py
+++ b/labgrid/util/agents/deditec_relais8.py
@@ -1,0 +1,148 @@
+"""
+This module implements the communication protocol to switch the digital outputs
+on the Deditec Relais8 board.
+
+Supported modules:
+
+- Relais8
+
+Supported Functionality:
+
+- Turn digital output on and off
+"""
+import logging
+
+from binascii import unhexlify, hexlify
+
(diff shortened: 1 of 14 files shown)


labgrid/util/agents/deditec_relais8.py
```diff
@@ -0,0 +1,148 @@
...
+        usb.util.release_interface(self._dev, 0)
+
+
+def handle_set(busnum, devnum, number, status):
+    relais8 = Relais8(bus=busnum, address=devnum)
+    relais8.set_output(number, status)
+
+
+def handle_get(busnum, devnum, number):
+    relais8 = Relais8(bus=busnum, address=devnum)
```
Emantor: This will run the initialization sequence for the board each time, but I think there is no other choice with the current design.

jluebbe: Yes, otherwise there can't be multiple clients talking to different outputs on the same device.

PR #436, opened 2019-05-17
driver: fastbootdriver: implement 'getvar' and 'oem getenv' commands

Bastian-Krause:
**Description**
These commands allow reading variables. The value is returned in a
pretty obfuscated way, so clean up is performed and only the value of
the (environment) variable is then returned.

Note: there is no way to distinguish non-existent variables and
variables with an empty value. This is a fastboot limitation.

**Checklist**
- [ ] Tests for the feature 
- [x] PR has been tested


Bastian-Krause:
Rebased and force-pushed.

PR #437, opened 2019-05-17
RKUsbLoader/RKUSBDriver and minor fixes

otavio:
**Description**

**Checklist**
- [ ] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/labgrid/driver/usbloader.py
+++ b/labgrid/driver/usbloader.py
@@ -118,10 +118,17 @@ class RKUSBDriver(Driver, BootstrapProtocol):
             mf = ManagedFile(usb_loader, self.loader)
             mf.sync_to_resource()
 
-        subprocess.check_call(
-            self.loader.command_prefix
-            + [self.tool, 'db', mf.get_remote_path()]
-        )
+        timeout = Timeout(3.0)
+        while True:
+            try:
+                subprocess.check_call(
+                    self.loader.command_prefix +
+                    [self.tool, 'db', mf.get_remote_path()]
+                )
+                break
+            except subprocess.CalledProcessError:
+                if timeout.expired:
+                    raise
 
         if filename is None and self.image is not None:
             filename = self.target.env.config.get_image_path(self.image)
--- a/labgrid/driver/usbloader.py
+++ b/labgrid/driver/usbloader.py
@@ -1,14 +1,16 @@
 # pylint: disable=no-member
 import subprocess
 import attr
+import time
 
 from ..factory import target_factory
 from ..protocol import BootstrapProtocol
-from ..resource.remote import NetworkMXSUSBLoader, NetworkIMXUSBLoader
-from ..resource.udev import MXSUSBLoader, IMXUSBLoader
+from ..resource.remote import NetworkMXSUSBLoader, NetworkIMXUSBLoader, NetworkRKUSBLoader
+from ..resource.udev import MXSUSBLoader, IMXUSBLoader, RKUSBLoader
 from ..step import step
 from .common import Driver
 from ..util.managedfile import ManagedFile
+from ..util.timeout import Timeout
 
 
 @target_factory.reg_driver
@@ -43,7 +45,7 @@ class MXSUSBDriver(Driver, BootstrapProtocol):
         mf.sync_to_resource()
 
         subprocess.check_call(
-            self.loader.command_prefix+[self.tool, "0", mf.get_remote_path()]
+            self.loader.command_prefix + [self.tool, "0", mf.get_remote_path()]
         )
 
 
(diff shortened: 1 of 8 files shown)


labgrid/driver/usbloader.py
```diff
@@ -82,3 +83,54 @@ def load(self, filename=None):
...
+        pass
+
+    def on_deactivate(self):
+        pass
+
+    @Driver.check_active
+    @step(args=['filename'])
+    def load(self, filename=None):
+        # FIXME make sure we always have an environment or config
+        loaders = self.target.env.config.get_tool('loader') or 'loader'
```
jluebbe: Loader seems to be a very generic name. Is this an additional image which needs to be loaded before the actual bootloader? If so, it should probably be set via an attribute on the driver (like the image itself).

labgrid/driver/usbloader.py
```diff
@@ -82,3 +83,54 @@ def load(self, filename=None):
...
+    def on_activate(self):
+        pass
+
+    def on_deactivate(self):
+        pass
+
+    @Driver.check_active
+    @step(args=['filename'])
+    def load(self, filename=None):
+        # FIXME make sure we always have an environment or config
```
jluebbe: Yes, you need to check if you have an environment set (usally ``if self.target.env``).

labgrid/driver/usbloader.py
```diff
@@ -82,3 +83,55 @@ def load(self, filename=None):
...
+            + [self.tool, 'db', mf.get_remote_path()]
+        )
+
+        if filename is None and self.image is not None:
+            filename = self.target.env.config.get_image_path(self.image)
+        mf = ManagedFile(filename, self.loader)
+        mf.sync_to_resource()
+
+        # wait the loader finish the process
+        time.sleep(1)
```
Emantor: Why is this sleep necessary here?

otavio: We can confirm if the `sleep` is required. Worse, we can do a busy loop, trying to load it with a max of 3s for timeout.

chtavares592:
@Emantor I need to use something that wait for the loader image start at the board. So how @otavio told you,  I edited the code and now it does a loop, trying to load it with a max of 3s for the timeout.

labgrid/driver/usbloader.py
```diff
@@ -78,7 +79,61 @@ def load(self, filename=None):
...
+
+        if filename is None and self.image is not None:
+            filename = self.target.env.config.get_image_path(self.image)
+        mf = ManagedFile(filename, self.loader)
+        mf.sync_to_resource()
+
+        for _ in range(30):
+           # The u-boot image is written on the device
+            try:
+                time.sleep(0.1)
```
Emantor: I prefer something along the lines of
```
deadline = time.monotonic() + 3.0
while time.monotonic() < deadline:
```
as a timeout construct.

chtavares592: @Emantor modifications were performed

otavio:
Hello; what is pending for this to move on? @Emantor any other needed change?

chtavares592:
@jluebbe I did rebase on a master and did push, now it's all right.

labgrid/driver/usbloader.py
```diff
@@ -79,6 +80,66 @@ def load(self, filename=None):
...
+        deadline = time.monotonic() + 3.0
+        while time.monotonic() < deadline:
+            try:
+                subprocess.check_call(
+                    self.loader.command_prefix +
+                    [self.tool, 'db', mf.get_remote_path()]
+                )
+                break
+            except subprocess.CalledProcessError:
+                continue
```
jluebbe: If this times out, we will just continue with the next stage, without raising the exception.
With Timeout from labgrid.util, you could do:
```python
    timeout = Timeout(3.0)
    while True:
        try:
            subprocess.check_call(
                self.loader.command_prefix +
                [self.tool, 'db', mf.get_remote_path()]
            )
            break
        except subprocess.CalledProcessError:
            if timeout.expired:
                raise
```

labgrid/driver/usbloader.py
```diff
@@ -79,6 +80,66 @@ def load(self, filename=None):
...
+        deadline = time.monotonic() + 3.0
+        while time.monotonic() < deadline:
+            try:
+                subprocess.check_call(
+                    self.loader.command_prefix
+                    + [self.tool, 'wl', '0x40', mf.get_remote_path()]
+                )
+                break
+            except subprocess.CalledProcessError:
+                continue
```
jluebbe: This has the same issue as above.

chtavares592:
@jluebbe the changes have been applied

jluebbe:
@otavio Looking at this after some time, I noticed that your `rk-usb-loader` doesn't seem to be available? Is that a wrapper/fork of https://github.com/rockchip-linux/rkdeveloptool?

otavio:
Yes. It is this tool.

In our labgrid config we use:
```yaml
targets:
  main:
    resources:
      RKUSBLoader:
        match:
          'sys_name': '1-1.1.3'
      USBMassStorage:
        match:
          'ID_PATH': 'platform-3f980000.usb-usb-0:1.1.3:1.0-scsi-0:0:0:0'
      USBSerialPort:
        match:
          'ID_PATH': 'platform-3f980000.usb-usb-0:1.3:1.0'
        speed: 1500000
    drivers:
      RKUSBDriver:
        image: bootloader
        usb_loader: boot-rom

tools:
  rk-usb-loader: /usr/bin/rkdeveloptool
images:
  bootloader: images/u-boot.img
  boot-rom: images/RV1108_usb_boot.bin
```

otavio:
Please do whatever is need. We can update here.

PR #439, opened 2019-05-22
util: ssh: make ssh_connection_manager reuse existing connections

jgoetzinger:
**Description**
Check if ssh connection exists. Previous "instance = self._connections.get(host)" always results in instance = None.

Signed-off-by: Jonathan Goetzinger <j.goetzinger@eckelmann.de>

**Checklist**
- [x] PR has been tested

### DIFF
--- a/labgrid/util/ssh.py
+++ b/labgrid/util/ssh.py
@@ -37,8 +37,9 @@ class SSHConnectionManager:
 
         Returns:
             :obj:`SSHConnection`: the SSHConnection for the host"""
-        instance = self._connections.get(host)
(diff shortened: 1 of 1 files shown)


Emantor:
Functionality wise both versions should be identical, can you provide a minimum working example how your change works differently?

jgoetzinger:
Never mind. I just realised my mistake. Both implementations indeed have the same functionality. I was facing a problem with the sshmanager and changed a few things.  The problem is somewhere else. I will make a buckreport with my problem later.

PR #442, opened 2019-05-29
resource: common.py only use proxy when 'proxy_required' is true

jgoetzinger:
**Description**
With the new labgrid-client -P option using:

labgrid-client -P labgrid01 -p board-1 video
labgrid-client -P labgrid01 -p board-1 bootstrap
labgrid-client -P labgrid01 -p board-1 fastboot

results in a timeout or ssh connect error.

When the the self.video.command_prefix is evaluated, labgrid tries to
create a second connection to the exporter with the hostname from 
self.extra.proxy instead of the one specified with the -P option. Since
self.extra.proxy is a default variable, it should only be evaluated when
"proxy_required" is True.

Signed-off-by: Jonathan Goetzinger <j.goetzinger@eckelmann.de>

**Checklist**
- [x] PR has been tested

### DIFF
--- a/labgrid/resource/common.py
+++ b/labgrid/resource/common.py
@@ -65,7 +65,8 @@ class NetworkResource(Resource):
         host = self.host
 
         if hasattr(self, 'extra'):
-            host = self.extra.get('proxy')
+            if self.extra.get('proxy_required'): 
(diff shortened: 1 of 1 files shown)


Emantor:
Thanks @jgoetzinger!

Issue #443, opened 2019-06-04
Unable to get expected report as mentioned in "Strategy Fixture Example" while using uboot strategy


bhargavthriler:
Hi, I am new to labgrid, while following the example code mentioned in "Strategy Fixture Example" and replacing the barebox strategy with uboot strategy I'm get the following error
```
@pytest.fixture(scope='function')   
    def shell_command(target, strategy, capsys):
        with capsys.disabled():           
 >        strategy.transition('shell')  
```

Later I even tried executing the barebox example, still I faced the same issue.

```
target = Target(name='main', env=Environment(config_file='strategy-example.yaml'))
strategy = BareboxStrategy(target=Target(name='main', env=Environment(config_file='strategy-example.yaml')), name=None, state=<BindingState.bound: 1>, status=<Status.unknown: 0>)
capsys = <_pytest.capture.CaptureFixture object at 0x7f1033fed0f0>
                                       
    @pytest.fixture(scope='function')     
    def shell_command(target, strategy, capsys):
        with capsys.disabled():           
  >          strategy.transition('shell')  
                                          
conftest.py:27: 
```                                    
Am I doing some mistake, if so please explain.

My current setup is ubuntu 16.04, target platform "imx6"

**Note**: I didn't find any forums to discuss some issues regarding labgrid. If there is one please mention it. Sorry if this not the right place to discuss these issues.

Emantor:
Hi, your error reports contain the line where the error was logged, however they do not contain the error message. Please post the complete stacktrace.

bhargavthriler:
> Hi, your error reports contain the line where the error was logged, however they do not contain the error message. Please post the complete stacktrace.

Here's the complete stacktrace. My target is connected to "ttyUSB0".
```
Labgrid/labgrid-venv/boot-reboot2$ pytest --lg-env local.yaml test_boot.py 
=========================================================================================== test session starts ============================================================================================
platform linux -- Python 3.5.2, pytest-4.0.2, py-1.8.0, pluggy-0.12.0
rootdir: /home/bhargav/Labgrid/labgrid-venv/boot-reboot2, inifile:
plugins: labgrid-0.2.1.dev138+g9091a77, html-1.20.0, metadata-1.8.0
collected 3 items                                                                                                                                                                                          

test_boot.py main: Turn the target example-board OFF and press enterEmain: Turn the target example-board OFF and press enterEmain: Turn the target example-board OFF and press enterE                                                                                                                                                                                     [100%]

[...]

PR #447, opened 2019-06-06
added '-s' option in example command

bhargavthriler:
**Description**
The pytest `-s` option is required to read from stdin.

**Checklist**
- [x] PR has been tested

### DIFF
--- a/doc/usage.rst
+++ b/doc/usage.rst
@@ -333,7 +333,7 @@ For this example, you should get a report similar to this:
 
 .. code-block:: bash
 
-  $ pytest --lg-env strategy-example.yaml -v
+  $ pytest -s --lg-env strategy-example.yaml -v
   ============================= test session starts ==============================
   platform linux -- Python 3.5.3, pytest-3.0.6, py-1.4.32, pluggy-0.4.0
   …


jluebbe:
As the transitions are executed under ``with capsys.disabled():``, the -s switch should not be needed. Is that broken for you?

Emantor:
Yes, see #443

jluebbe:
@Emantor, according to https://docs.pytest.org/en/latest/capture.html#accessing-captured-output-from-a-test-function, it should be possible to disable capturing with `capsys.disabled()`. If that's broken, we should fix it upstream instead of using `-s` as a workaround.

Emantor:
pytest-dev/pytest#2189 seems to indicate that we are not alone. Only `stdout` is not captured, `stdin` looks to be captured even with `capsys.disabled()`. I'll look into pytest and determine whether this can be fixed there or we need to have a workaround within labgrid.

jluebbe:
The capsys docs now refer explicitly to stdout/stderr, so we need to document using `-s` in the example and with `ManualPowerDriver` under pytest.

PR #455, opened 2019-06-21
remote/client: use idle place helper for delete

Emantor:
**Description**
The del_place function used the pure argument instead of going through
the idle place helper, allowing the user to delete acquired places. Use
the idle place helper to ensure that only idle places are deleted

- [ ] PR has been tested

### DIFF
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -373,7 +373,14 @@ class ClientSession(ApplicationSession):
 
     async def del_place(self):
         """Delete a place from the coordinator"""
-        name = self.args.place
+        pattern = self.args.place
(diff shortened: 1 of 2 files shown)


labgrid/remote/client.py
```diff
@@ -373,7 +373,8 @@ def get_acquired_place(self, place=None):
 
     async def del_place(self):
         """Delete a place from the coordinator"""
-        name = self.args.place
+        place = self.get_idle_place()
```
jluebbe: We shouldn't use pattern matching when deleting to avoid accidents.

Emantor: This has been fixed.

PR #458, opened 2019-06-28
driver: bareboxdriver: remove unused startstring attribute

Bastian-Krause:
**Description**
Since 016f605c ("bareboxdriver: refactor login expect loop and add password support")
startstring is not used anymore.

**Checklist**
- [ ] PR has been tested


Emantor:
@jluebbe was the intent to remove this during the refactor?

jluebbe:
> @jluebbe was the intent to remove this during the refactor?

I'm not sure, but it seems it's unused, as it was not reported as a regression in the last year. The logic seems to be robust enough without it.

PR #407, opened 2019-04-01
Add a GpioDigitalOutputDriver

bgermann:
**Description**
This PR is a follow-up on #234. It adds GpioDigitalOutputDriver which uses libgpiod to write a digital signal to a GPIO line.
The minimum requirement to use libgpiod's kernel interface is Linux 4.8.
The python binding is quite new and neither commonly available in Linux distributions nor on PyPI, so import it only on usage.
I tested it in combination with DigitalOutputPowerDriver on a Banana PI which uses a GPIO line to power a connected device.

**Checklist**
- [x] Documentation for the feature
- [ ] Tests for the feature 
- [x] The arguments and description in doc/configuration.rst have been updated
- [x] CHANGES.rst has been updated
- [x] PR has been tested

PR #468, opened 2019-07-08
Implement GPIO support via sysfs with agent interface

leiflm:
**Description**
This is only complementary work for https://github.com/labgrid-project/labgrid/pull/407

**Checklist**
- [x] Tests for the feature 

- [x] PR has been tested

Complements https://github.com/labgrid-project/labgrid/pull/407

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -1088,6 +1088,25 @@ Implements:
 Arguments:
   - delay (float): optional delay in seconds between off and on
 
+GpioDigitalOutputDriver
+~~~~~~~~~~~~~~~~~~~~~~~
+The GpioDigitalOutputDriver writes a digital signal to a GPIO line.
+
+This driver uses the libgpiod python bindings.
+Make sure to have them installed.
+
+Implements:
+  - :any:`DigitalOutputProtocol`
+
+.. code-block:: yaml
+
+   GpioDigitalOutputDriver:
+     offset: 42
+
+Arguments:
+  - chip (str): A GPIO device's path, name, label or number (default: "0")
+  - offset (int): The offset to a GPIO line
+
 SerialPortDigitalOutputDriver
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 The SerialPortDigitalOutputDriver makes it possible to use a UART
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -1092,8 +1092,8 @@ GpioDigitalOutputDriver
 ~~~~~~~~~~~~~~~~~~~~~~~
 The GpioDigitalOutputDriver writes a digital signal to a GPIO line.
 
-This driver uses the libgpiod python bindings.
-Make sure to have them installed.
+This driver configures GPIO lines via `the sysfs kernel interface <https://www.kernel.org/doc/html/latest/gpio/sysfs.html>`.
+While the driver automatically exports the GPIO, it does not configure it in any other way than as an output.
 
 Implements:
   - :any:`DigitalOutputProtocol`
@@ -1101,11 +1101,10 @@ Implements:
 .. code-block:: yaml
 
    GpioDigitalOutputDriver:
-     offset: 42
+     index: 42
 
 Arguments:
-  - chip (str): A GPIO device's path, name, label or number (default: "0")
-  - offset (int): The offset to a GPIO line
+  - index (int): The index of a GPIO line
 
 SerialPortDigitalOutputDriver
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/labgrid/driver/gpiodriver.py
+++ b/labgrid/driver/gpiodriver.py
@@ -1,0 +1,51 @@
+"""All GPIO-related drivers"""
+import attr
+
+from ..factory import target_factory
+from ..protocol import DigitalOutputProtocol
+from ..step import step
+from .common import Driver
+
+
+@target_factory.reg_driver
+@attr.s(cmp=False)
+class GpioDigitalOutputDriver(Driver, DigitalOutputProtocol):
+    """
+    Controls the state of a GPIO using libgpiod.
+
+    Takes a string property 'chip' which refers to the GPIO device.
+    You can use its path, name, label or number (default: 0).
+    The offset to the GPIO line is set by the integer property 'offset'.
+    """
+
+    bindings = {}
+    offset = attr.ib(validator=attr.validators.instance_of(int))
+    chip = attr.ib(default="0", validator=attr.validators.instance_of(str))
+    gpio_chip = None
+    gpio_state = True
+
+    def __attrs_post_init__(self):
+        super().__attrs_post_init__()
+
+    def on_activate(self):
+        import gpiod
+        self.gpio_chip = gpiod.Chip(self.chip)
+        gpio_line = self.gpio_chip.get_line(self.offset)
+        gpio_line.request("labgrid", gpiod.LINE_REQ_DIR_OUT)
+
+    def on_deactivate(self):
+        gpio_line = self.gpio_chip.get_line(self.offset)
+        gpio_line.release()
+        self.gpio_chip.close()
+
+    @Driver.check_active
+    @step()
+    def get(self):
+        return self.gpio_state
+
+    @Driver.check_active
+    @step()
+    def set(self, status):
+        gpio_line = self.gpio_chip.get_line(self.offset)
+        gpio_line.set_value(int(status))
+        self.gpio_state = status
--- a/labgrid/driver/gpiodriver.py
+++ b/labgrid/driver/gpiodriver.py
@@ -3,49 +3,44 @@ import attr
 
 from ..factory import target_factory
 from ..protocol import DigitalOutputProtocol
+from ..resource.base import SysfsGPIO
+from ..resource.remote import NetworkSysfsGPIO
 from ..step import step
 from .common import Driver
+from ..util.agentwrapper import AgentWrapper
 
 
 @target_factory.reg_driver
 @attr.s(cmp=False)
 class GpioDigitalOutputDriver(Driver, DigitalOutputProtocol):
-    """
-    Controls the state of a GPIO using libgpiod.
 
-    Takes a string property 'chip' which refers to the GPIO device.
-    You can use its path, name, label or number (default: 0).
-    The offset to the GPIO line is set by the integer property 'offset'.
-    """
-
-    bindings = {}
-    offset = attr.ib(validator=attr.validators.instance_of(int))
-    chip = attr.ib(default="0", validator=attr.validators.instance_of(str))
-    gpio_chip = None
-    gpio_state = True
+    bindings = {
+        "gpio": {SysfsGPIO, NetworkSysfsGPIO},
+    }
 
     def __attrs_post_init__(self):
         super().__attrs_post_init__()
+        self.wrapper = None
 
     def on_activate(self):
-        import gpiod
-        self.gpio_chip = gpiod.Chip(self.chip)
-        gpio_line = self.gpio_chip.get_line(self.offset)
-        gpio_line.request("labgrid", gpiod.LINE_REQ_DIR_OUT)
+        if isinstance(self.gpio, NetworkSysfsGPIO):
+            host = self.gpio.host
+        else:
+            host = None
+        self.wrapper = AgentWrapper(host)
+        self.proxy = self.wrapper.load('sysfsgpio')
 
     def on_deactivate(self):
-        gpio_line = self.gpio_chip.get_line(self.offset)
-        gpio_line.release()
-        self.gpio_chip.close()
+        self.wrapper.close()
+        self.wrapper = None
+        self.proxy = None
 
     @Driver.check_active
-    @step()
-    def get(self):
-        return self.gpio_state
-
-    @Driver.check_active
-    @step()
+    @step(args=['status'])
     def set(self, status):
-        gpio_line = self.gpio_chip.get_line(self.offset)
-        gpio_line.set_value(int(status))
-        self.gpio_state = status
+        self.proxy.set(self.gpio.index, status)
+
+    @Driver.check_active
+    @step(result=True)
+    def get(self):
+        return self.proxy.get(self.gpio.index)
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -666,9 +666,11 @@ class ClientSession(ApplicationSession):
         from ..resource.modbus import ModbusTCPCoil
         from ..resource.onewireport import OneWirePIO
         from ..resource.remote import NetworkDeditecRelais8
+        from ..resource.remote import NetworkSysfsGPIO
         from ..driver.modbusdriver import ModbusCoilDriver
         from ..driver.onewiredriver import OneWirePIODriver
         from ..driver.deditecrelaisdriver import DeditecRelaisDriver
+        from ..driver.gpiodriver import GpioDigitalOutputDriver
         drv = None
         for resource in target.resources:
             if isinstance(resource, ModbusTCPCoil):
@@ -691,6 +693,13 @@ class ClientSession(ApplicationSession):
                 except NoDriverFoundError:
                     target.set_binding_map({"relais": name})
                     drv = DeditecRelaisDriver(target, name=name)
+                break
+            elif isinstance(resource, NetworkSysfsGPIO):
+                try:
+                    drv = target.get_driver(GpioDigitalOutputDriver, name=name)
+                except NoDriverFoundError:
+                    target.set_binding_map({"gpio": name})
+                    drv = GpioDigitalOutputDriver(target, name=name)
                 break
         if not drv:
             raise UserError("target has no compatible resource available")
--- a/labgrid/remote/exporter.py
+++ b/labgrid/remote/exporter.py
@@ -330,6 +330,28 @@ class EthernetPortExport(ResourceExport):
 exports["SNMPEthernetPort"] = EthernetPortExport
 
 
+@attr.s(cmp=False)
+class GPIOGenericExport(ResourceExport):
+    """ResourceExport for GPIO lines accessed directly from userspace"""
+
+    def __attrs_post_init__(self):
+        super().__attrs_post_init__()
+        local_cls_name = self.cls
+        self.data['cls'] = "Network{}".format(self.cls)
+        from ..resource import udev
+        local_cls = getattr(udev, local_cls_name)
+        self.local = local_cls(target=None, name=None, **self.local_params)
+
+    def _get_params(self):
+        """Helper function to return parameters"""
+        return {
+            'host': self.host,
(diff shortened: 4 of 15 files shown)


leiflm:
I've just noticed, that the output port is not switched as expected.

I'm investigating the issue and will comment as soon as the issue is resolved.

leiflm:
> I've just noticed, that the output port is not switched as expected.

This was a false assumption. The output *is* indeed toggled. It just turns back to its initial state once the instance is destroyed/the line gets released. This happens really quickly with the current code appearing as if it was never toggled in the first place.

> I'm investigating the issue and will comment as soon as the issue is resolved.

Okay, turns out that when lines are released they recover their former state.

I'll modify the code to work around eventual pin default state recovery.

labgrid/resource/remote.py
```diff
@@ -217,3 +217,20 @@ class NetworkDeditecRelais8(RemoteUSBResource):
...
+class RemoteGPIOResource(NetworkResource, ManagedResource):
+    manager_cls = RemotePlaceManager
+
+    chip = attr.ib(validator=attr.validators.optional(attr.validators.instance_of(str)))
+    offset = attr.ib(validator=attr.validators.optional(attr.validators.instance_of(int)))
+
+
+@target_factory.reg_resource
+@attr.s(cmp=False)
+class NetworkLibGPIOD(RemoteGPIOResource):
```
jluebbe: Do you have a specific reason to use two separate classes?

labgrid/resource/udev.py
```diff
@@ -415,3 +415,65 @@ def path(self):
...
+    def path(self):
+        if self.device:
+            return str(self.device.sys_name)
+
+        return None
+
+
+@target_factory.reg_resource
+@attr.s(cmp=False)
+class LibGPIOD(GPIOResource):
```
jluebbe: As above, I don't see cases where we have different kinds of GPIOResources, so this could be combined.

labgrid/resource/remote.py
leiflm: I had two scenarios in mind:
1. Someone implements the deprecated sysfs interface using SSHDriver
2. People running a different OS (bsd/whatever)

labgrid/resource/udev.py
leiflm: From [my answer above](https://github.com/labgrid-project/labgrid/pull/468#discussion_r301462547):
> I had two scenarios in mind:
> 1. Someone implements the deprecated sysfs interface using SSHDriver
> 2. People running a different OS (bsd/whatever)

leiflm: But I *do* see the obvious point of the unnecessary indirection/abstraction.

leiflm:
Since I'm not firm with the Python ecosystem: Where would the distro-specific requirement declaration for `libgpiod` go? Should I just update the [debian file](https://github.com/labgrid-project/labgrid/blob/master/debian/control)?

jluebbe:
> > I've just noticed, that the output port is not switched as expected.
> 
> This was a false assumption. The output _is_ indeed toggled. It just turns back to its initial state once the instance is destroyed/the line gets released. This happens really quickly with the current code appearing as if it was never toggled in the first place.
> 
> > I'm investigating the issue and will comment as soon as the issue is resolved.
> 
> Okay, turns out that when lines are released they recover their former state.
> 
> I'll modify the code to work around eventual pin default state recovery.

Hmm, it seems that this is a limitation of the /dev/gpiochip interface. The current assumption of the DigitalOutputProtocol is that any configured value is kept until overwritten. This is relevant for usage with labgrid-client io, as no process is around to keep the device node open.

https://git.kernel.org/pub/scm/libs/libgpiod/libgpiod.git/tree/TODO#n11 mentions the plan to write a dbus-based daemon to handle that case, adding even more complexity. Maybe using the sysfs interface is the better approach? ;)

jluebbe:
> Since I'm not firm with the Python ecosystem: Where would the distro-specific requirement declaration for `libgpiod` go? Should I just update the [debian file](https://github.com/labgrid-project/labgrid/blob/master/debian/control)?

The documentation section on the driver should mention the requirement. Updating the Debian file is also useful.

labgrid/resource/remote.py
jluebbe: I currently don't see running labgrid on anything else than linux as a requirement.

As an aside, a driver for the sysfs interface could/would also use the agent mechanism.

labgrid/resource/udev.py
jluebbe: Before you remove this abstraction, let's first figure out if using libgpiod can reasonably fit the current model (regarding the revert to default on close behaviour).

leiflm:
> > > I've just noticed, that the output port is not switched as expected.
> > 
> > 
> > This was a false assumption. The output _is_ indeed toggled. It just turns back to its initial state once the instance is destroyed/the line gets released. This happens really quickly with the current code appearing as if it was never toggled in the first place.
> > > I'm investigating the issue and will comment as soon as the issue is resolved.
> > 
> > 
> > Okay, turns out that when lines are released they recover their former state.
> > I'll modify the code to work around eventual pin default state recovery.
> 
> Hmm, it seems that this is a limitation of the /dev/gpiochip interface. The current assumption of the DigitalOutputProtocol is that any configured value is kept until overwritten. This is relevant for usage with labgrid-client io, as no process is around to keep the device node open.
> 
> https://git.kernel.org/pub/scm/libs/libgpiod/libgpiod.git/tree/TODO#n11 mentions the plan to write a dbus-based daemon to handle that case, adding even more complexity. Maybe using the sysfs interface is the better approach? ;)

Yeah. I think that the effort of transforming the libgpiod implementation into an agent was futile. So maybe the state prior to agent implementation (https://github.com/labgrid-project/labgrid/pull/468/commits/8de1391b33a49bda9a8bcac090e43ad96b454c72) could be merged? Should I force-push a state that removes all the commits on top?

jluebbe:
> Yeah. I think that the effort of transforming the libgpiod implementation into an agent was futile. So maybe the state prior to agent implementation ([8de1391](https://github.com/labgrid-project/labgrid/commit/8de1391b33a49bda9a8bcac090e43ad96b454c72)) could be merged? Should I force-push a state that removes all the commits on top?

That would be premature. Compared to that state, you now have network support.
If you keep the libgpiod handle open in the agent (over multiple get/set calls), you would have the same behaviour as using libgpio directly in the driver (just with added remote support). But even that wouldn't be enough for keeping the output state after the process has exited.

Is there some way in libgpiod to make the state persistent?

leiflm:
> > Yeah. I think that the effort of transforming the libgpiod implementation into an agent was futile. So maybe the state prior to agent implementation ([8de1391](https://github.com/labgrid-project/labgrid/commit/8de1391b33a49bda9a8bcac090e43ad96b454c72)) could be merged? Should I force-push a state that removes all the commits on top?
> 
> That would be premature. Compared to that state, you now have network support.

Yes. But with `libgpiod` the remote system has to have the python bindings for libgpiod installed too.
Compared to simply having a SSHDriver, that seems overly complex, doesn't it?

Maybe I could implement the networked variant using sysfs and keep the local implementation using `libgpiod`?

> If you keep the libgpiod handle open in the agent (over multiple get/set calls), you would have the same behaviour as using libgpio directly in the driver (just with added remote support). But even that wouldn't be enough for keeping the output state after the process has exited.

Yeah. That's true. Maybe instead of implementing the `DigitalOutputProtocol` it should only implement
the `ResetProtocol` for the time being?

> Is there some way in libgpiod to make the state persistent?

Citing the libgpiod project author here:
> A common complaint from users about gpioset is that the state of a line is not retained once the program exits.

His reasoning behind that design:
> While this is precisely the way linux character devices work, it's understandable that most users will want some centralized way of controlling GPIOs - similar to how sysfs worked

As all character devices are supposed to behave like this, there must be some general implementation in the kernel? Skimming through the kernel source (`drivers/gpio/gpiolib.c`)  does not show a state recovery besides [unassigning various flags](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/drivers/gpio/gpiolib.c#n2429).

jluebbe:
> > > Yeah. I think that the effort of transforming the libgpiod implementation into an agent was futile. So maybe the state prior to agent implementation ([8de1391](https://github.com/labgrid-project/labgrid/commit/8de1391b33a49bda9a8bcac090e43ad96b454c72)) could be merged? Should I force-push a state that removes all the commits on top?
> >  
> > That would be premature. Compared to that state, you now have network support.
> 
> Yes. But with `libgpiod` the remote system has to have the python bindings for libgpiod installed too.
> Compared to simply having a SSHDriver, that seems overly complex, doesn't it?
>
> Maybe I could implement the networked variant using sysfs and keep the local implementation using `libgpiod`?

I think I'd perfer to just use the sysfs interface via the agent. This avoids the additional dependency and solves the problem of keeping the configured output even after a process exits.
 
> > If you keep the libgpiod handle open in the agent (over multiple get/set calls), you would have the same behaviour as using libgpio directly in the driver (just with added remote support). But even that wouldn't be enough for keeping the output state after the process has exited.
> 
> Yeah. That's true. Maybe instead of implementing the `DigitalOutputProtocol` it should only implement
> the `ResetProtocol` for the time being?

That would be much less useful. :/

> > Is there some way in libgpiod to make the state persistent?
> 
> Citing the libgpiod project author here:
> 
> > A common complaint from users about gpioset is that the state of a line is not retained once the program exits.
> 
> His reasoning behind that design:
> 
> > While this is precisely the way linux character devices work, it's understandable that most users will want some centralized way of controlling GPIOs - similar to how sysfs worked
> 
> As all character devices are supposed to behave like this, there must be some general implementation in the kernel? Skimming through the kernel source (`drivers/gpio/gpiolib.c`) does not show a state recovery besides [unassigning various flags](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/drivers/gpio/gpiolib.c#n2429).

Given that even gpiolib still has that missing daemon as a TODO, the pragmatic approach for labgrid seems to come down to using the sysfs interface via the agent, as that does everything we need and fits our current model better. If there is a daemon in the future (or we decide to build that ourselves), we can still add an additional driver for the same resource.

It should be possible to translate the gpiochip number + offset information the the global index also via sysfs. That would allow us to use the same resource for both alternative drivers.

leiflm:
> Given that even gpiolib still has that missing daemon as a TODO, the pragmatic approach for labgrid seems to come down to using the sysfs interface via the agent, as that does everything we need and fits our current model better. If there is a daemon in the future (or we decide to build that ourselves), we can still add an additional driver for the same resource.

Alright. For the given reasons, I'll change the agent to use the (deprecated) sysfs instead of libgpiod then.

labgrid/driver/gpiodriver.py
```diff
@@ -0,0 +1,46 @@
...
+from .common import Driver
+from ..util.agentwrapper import AgentWrapper
+
+
+@target_factory.reg_driver
+@attr.s(cmp=False)
+class GpioDigitalOutputDriver(Driver, DigitalOutputProtocol):
+
+    bindings = {
+        "gpio_provider": {SysfsGPIO, NetworkSysfsGPIO},
```
jluebbe:
```suggestion
        "gpio": {SysfsGPIO, NetworkSysfsGPIO},
```
_provider doesn't add information here.

labgrid/remote/client.py
```diff
@@ -692,6 +694,13 @@ def digital_io(self):
                     target.set_binding_map({"relais": name})
                     drv = DeditecRelaisDriver(target, name=name)
                 break
+            elif isinstance(resource, NetworkSysfsGPIO):
+                try:
+                    drv = target.get_driver(GpioDigitalOutputDriver, name=name)
+                except NoDriverFoundError:
+                    target.set_binding_map({"gpio_provider": name})
```
jluebbe: This needs to match the change above.

labgrid/remote/exporter.py
```diff
@@ -330,6 +330,29 @@ def _get_params(self):
...
+        from ..resource import udev
+        local_cls = getattr(udev, local_cls_name)
+        self.local = local_cls(target=None, name=None, **self.local_params)
+
+    def _get_params(self):
+        """Helper function to return parameters"""
+        return {
+            'host': self.host,
+            'chip': self.local.chip,
+            'offset': self.local.offset,
```
jluebbe: This doesn't seem to match what the resources use (`index`).

labgrid/util/agents/sysfsgpio.py
```diff
@@ -0,0 +1,68 @@
...
+            binary_value = b'0'
+
+        if binary_value is None:
+            raise ValueError("GPIO value is out of range.")
+
+        os.write(self.gpio_sysfs_value_fd, binary_value)
+
+
+def handle_set(_index, status):
+    gpio_line = GpioDigitalOutput(index=_index)
```
jluebbe: You should cache the instances in a module level dictionary. That would avoid reconfiguring the GPIO for every call.

labgrid/util/agents/sysfsgpio.py
```diff
@@ -0,0 +1,68 @@
...
+        binary_value = None
+        if status is True:
+            binary_value = b'1'
+        elif status is False:
+            binary_value = b'0'
+
+        if binary_value is None:
+            raise ValueError("GPIO value is out of range.")
+
+        os.write(self.gpio_sysfs_value_fd, binary_value)
```
jluebbe: Can you write multiple values without closing the "value" file in between?

doc/configuration.rst
```diff
@@ -1088,6 +1088,24 @@ Implements:
...
+  - :any:`DigitalOutputProtocol`
+
+.. code-block:: yaml
+
+   GpioDigitalOutputDriver:
+     offset: 42
+
+Arguments:
+  - chip (str): A GPIO device's path, name, label or number (default: "0")
+  - offset (int): The offset to a GPIO line
```
jluebbe: These have been replaced by index by now.

labgrid/util/agents/sysfsgpio.py
leiflm: You can. You only need to seek the cursor to read a new value.

I verified this by instantiating the agent manually and toggling the output via `set()`.

leiflm:
About squashing: Sure. Which range? All of my code? All of the libgpiod code? Feel free to do it in a way you like/see fit.

labgrid/resource/remote.py
leiflm: resolved by https://github.com/labgrid-project/labgrid/pull/468/commits/9e7cfbefa99551d7fdeb2ce7dc0b690d99d7bef6

labgrid/resource/udev.py
leiflm: the `udev` variant seized to exist.

labgrid/util/agents/sysfsgpio.py
```diff
@@ -0,0 +1,84 @@
+"""
+This module implements implements switching GPIOs via sysfs GPIO kernel interface.
```
Emantor: nit: double implements.

labgrid/util/agents/sysfsgpio.py
```diff
@@ -0,0 +1,84 @@
...
+        if not os.path.exists(gpio_sysfs_path):
+            raise ValueError("Device not found")
+
+    def __init__(self, **kwargs):
+        index = kwargs['index']
+        self._logger = logging.getLogger("Device: ")
+        GpioDigitalOutput._assert_gpio_line_is_exported(index)
+        gpio_sysfs_path = os.path.join(GpioDigitalOutput._gpio_sysfs_path_prefix, 'gpio{0}'.format(index))
+        gpio_sysfs_direction_path = os.path.join(gpio_sysfs_path, 'direction')
+        direction_fd = os.open(gpio_sysfs_direction_path, flags=os.O_WRONLY)
```
Emantor: this could be
```
self._logger.debug("Configuring GPIO {idx} as output.".format(idx = index))
with open(gpio_sysfs_direction_path, 'wb') as direction_fd:
    direction_fd.write(b'out')
```

labgrid/util/agents/sysfsgpio.py
leiflm: 05a78a2839d0ed12d8f0550b574be6448611bdad

labgrid/util/agents/sysfsgpio.py
leiflm: db91e47da5283d5b8556cc062e86c01d0d271f05

jluebbe:
I've squashed the incremental changes into one commit. The original commits by @bgermann from PR #407 are unchanged.

bgermann:
Thanks!

pmelange:
> > > I've just noticed, that the output port is not switched as expected.
> > 
> > 
> > This was a false assumption. The output _is_ indeed toggled. It just turns back to its initial state once the instance is destroyed/the line gets released. This happens really quickly with the current code appearing as if it was never toggled in the first place.
> > > I'm investigating the issue and will comment as soon as the issue is resolved.
> > 
> > 
> > Okay, turns out that when lines are released they recover their former state.
> > I'll modify the code to work around eventual pin default state recovery.
> 
> Hmm, it seems that this is a limitation of the /dev/gpiochip interface. The current assumption of the DigitalOutputProtocol is that any configured value is kept until overwritten. This is relevant for usage with labgrid-client io, as no process is around to keep the device node open.
> 
> https://git.kernel.org/pub/scm/libs/libgpiod/libgpiod.git/tree/TODO#n11 mentions the plan to write a dbus-based daemon to handle that case, adding even more complexity. Maybe using the sysfs interface is the better approach? ;)

I am currently working on trying to implement libgpio into labgrid instead of using sysfs.  And I believe I have found the solution to the problem of the configured value being kept uptil overwritten.

When creating a line request, do it with ```direction=Direction.AS_IS```.  This will keep the state of the line as is was before being requested, and therefore no longer causing the value of the line to be changed.  Afterwards, it's possible to set the line to output mode without any changes in the line happening. In addition, it is possible to get the current value of the line in the case where a line value needs to be toggled.  Here is a small code snippet based on the toggle-line-value example in the libgpiod code.

```
def toggle_line_value(chip_path, line_offset):
    value_str = {Value.ACTIVE: "Active", Value.INACTIVE: "Inactive"}
    value = Value.ACTIVE

    with gpiod.request_lines(
        chip_path,
        consumer="toggle-line-value",
        config={
            line_offset: gpiod.LineSettings(
                direction=Direction.AS_IS
            )
        },
    ) as request:
        print("requesting the value")
        old_value = request.get_value(line_offset)
        print("old value {}={}".format(line_offset, value_str[old_value]))
        time.sleep(1)
        print("changing to output mode")
        request.reconfigure_lines(config={ line_offset: gpiod.LineSettings(
            direction=Direction.OUTPUT, output_value=old_value) })
        time.sleep(1)
        value = toggle_value(old_value)
        print("setting a new value")
        request.set_value(line_offset, value)
        time.sleep(1)
        print("requestion the value")
        old_value = request.get_value(line_offset)
        print("confirming value {}={}".format(line_offset, value_str[value]))

```

PR #469, opened 2019-07-09
Subprocess Wrapper

Emantor:
**Description**
This PR adds a subprocess wrapper which acts like `subprocess.check_output`, but takes callbacks to send the output of the command to.
`processmanager.register` adds a callback, `processmanager.unregister` removes the callback. Callbacks receive the process output line based.

**Checklist**
- [ ] Tests for the feature 
- [x] PR has been tested

**TODO List**
- [x] Logging callback for pytest
- [x] Print callback for library usage

### DIFF
--- a/labgrid/driver/fastbootdriver.py
+++ b/labgrid/driver/fastbootdriver.py
@@ -8,6 +8,7 @@ from ..resource.udev import AndroidFastboot
 from ..step import step
 from .common import Driver
 from ..util.managedfile import ManagedFile
+from ..util.helper import processwrapper
 
 
 @target_factory.reg_driver
@@ -57,7 +58,7 @@ class AndroidFastbootDriver(Driver):
     @Driver.check_active
     @step(title='call', args=['args'])
     def __call__(self, *args):
-        subprocess.check_call(self._get_fastboot_prefix() + list(args))
+        processwrapper.check_output(self._get_fastboot_prefix() + list(args))
 
     @Driver.check_active
     @step(args=['filename'])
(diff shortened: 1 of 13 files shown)


Emantor:
@jluebbe please take some look at this. I'll add a default callback to the pytest case, where we can use `print` or debug logging to show the command output. I have tested this with a known good testsuite.

jluebbe:
It should also be possible to replace the existing subprocess.check calls with this. That way, we can pass the output through the normal logging functions.

Emantor:
This is already possible, `check_output` and `check_call` only differ by `check_call` returning 0 instead of the command output. However none of the `check_call` sites within labgrid rely on this, all can be replaces with `check_output`.

jluebbe:
> This is already possible, `check_output` and `check_call` only differ by `check_call` returning 0 instead of the command output. However none of the `check_call` sites within labgrid rely on this, all can be replaced with `check_output`.

Could you also replace the `check_call` sites where appropriate?

labgrid/driver/fastbootdriver.py
```diff
@@ -91,9 +92,10 @@ def getvar(self, var):
             raise NotImplementedError('Retrieving a list of all variables is not supported yet')
 
         cmd = ['getvar', var]
-        proc = subprocess.run(self._get_fastboot_prefix() + cmd, stdout=subprocess.PIPE,
-                              stderr=subprocess.STDOUT, check=True, universal_newlines=True)
-        values = AndroidFastbootDriver._filter_fastboot_output(proc.stdout, '{}: '.format(var))
+        output = processwrapper.check_output(self._get_fastboot_prefix() + cmd)
+        values = AndroidFastbootDriver._filter_fastboot_output(
```
jluebbe: Is _filter_fastboot_output able to handle the additional lines from stderr?

Emantor:
I updated this with pty handling, which gives us somewhat better realtime information for programs which output continuous information like rsync.

labgrid/driver/fastbootdriver.py
Emantor: I don't have a device to test this at the moment.

labgrid/util/helper.py
```diff
@@ -30,20 +32,30 @@ class ProcessWrapper:
     def check_output(self, command):
         """Run a command and supply the output to callback functions"""
         res = []
-        process = subprocess.Popen(command, stderr=subprocess.STDOUT,
-                                   stdout=subprocess.PIPE, bufsize=0)
+        mfd, sfd = pty.openpty()
+        for fd in [mfd, sfd]:
```
Emantor: @jluebbe mentioned that here the fd is also manipulated for the child. This could result in different behaviour for the spawned program, I'll reinvestigate again if the slave fd also needs to be set to `O_NONBLOCK`.

Emantor: This works fine without touching `sfd`, I'll amend the Commit.

jluebbe:
As discussed with @Emantor, I've reworked some parts of this:
- keep using bytes instead of strs for compatibility with subprocess
- avoid busy polling via select
- fix line splitting
- pass final line to the callbacks
- squash pty rework with earlier commits

Emantor:
Changes look good to me.

PR #474, opened 2019-07-12
graphstrategy: raise if via is not list or None

Emantor:
**Description**
This prevents debugging sessions where a string is passed accidentally
to via and handled like a list.

- [x] Tests for the feature 
- [x] PR has been tested


Emantor:
Hm, it may be more graceful to just wrap the `string` into a `list` and allow this case.

jluebbe:
> Hm, it may be more graceful to just wrap the `string` into a `list` and allow this case.

Why? Too much magic can also be confusing.

PR #475, opened 2019-07-12
graphstrategy: allow via string argument

Emantor:
**Description**
Convert single strings into a list if they are passed in via.

- [x] Tests for the feature


jluebbe:
I find would find this behavior somewhat surprising. So at least this needs documentation.

Emantor:
On further thought I will go with the previous pull request which echos an error message if via is not a list.

PR #476, opened 2019-07-17
ubootdriver: use the given timeout for run()

tobsan:
**Description**
Other than what's stated in the commit, I've tested this on a local development board which uses NOR flash, so erasing it is slow, and it timed out every time before making the change.

**Checklist**
- [X] PR has been tested

### DIFF
--- a/labgrid/driver/ubootdriver.py
+++ b/labgrid/driver/ubootdriver.py
@@ -95,17 +95,18 @@ class UBootDriver(CommandMixin, Driver, CommandProtocol, LinuxBootProtocol):
 
     @Driver.check_active
     @step(args=['cmd'], result=True)
-    def run(self, cmd, timeout=None): # pylint: disable=unused-argument
+    def run(self, cmd, timeout=30):
         """
         Runs the specified command on the shell and returns the output.
 
         Args:
             cmd (str): command to run on the shell
+            timeout (int): optional, how long to wait for completion
 
         Returns:
             Tuple[List[str],List[str], int]: if successful, None otherwise
         """
-        return self._run(cmd)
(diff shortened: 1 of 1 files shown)


labgrid/driver/ubootdriver.py
```diff
@@ -95,17 +95,21 @@ def _run(self, cmd: str, *, timeout: int = 30, codec: str = "utf-8", decodeerror
 
     @Driver.check_active
     @step(args=['cmd'], result=True)
-    def run(self, cmd, timeout=None): # pylint: disable=unused-argument
+    def run(self, cmd, timeout=None):
```
Emantor: This should be changed to `def run(self, cmd, timeout=30)` to match the other drivers.

labgrid/driver/ubootdriver.py
```diff
@@ -95,17 +95,21 @@ def _run(self, cmd: str, *, timeout: int = 30, codec: str = "utf-8", decodeerror
...
+            timeout (int): optional, how long to wait for completion
 
         Returns:
             Tuple[List[str],List[str], int]: if successful, None otherwise
         """
-        return self._run(cmd)
+        if timeout:
+            return self._run(cmd, timeout)
+        else:
+            return self._run(cmd)
```
Emantor: than this can simply be written as `return self._run(cmd, timeout)`

labgrid/driver/ubootdriver.py
tobsan: Alright, sounds reasonable.

jluebbe:
Thank, @tobsan!

tobsan:
My pleasure!

PR #479, opened 2019-07-26
Fix uboot timeout

jluebbe:
**Description**
PR #476 passed the timeout argument by position, which causes an error as it's a keyword only argument.

**Checklist**
- [x] Tests for the feature


Emantor:
Thanks!

tobsan:
Thanks, this fixed the error I got as well in my "real-world" test.

PR #490, opened 2019-09-05
Add support for per-target options

tobsan:
**Description**
In environments with multiple targets, having only the global options available requires using using prefix/suffix if those options differ between targets. Per-target options solve this issue and relate to the global options the same way per-target feature flags relate to the global feature flags.

**Checklist**
- [X] Tests for the feature 
- [X] CHANGES.rst has been updated
- [X] PR has been tested

### DIFF
--- a/labgrid/config.py
+++ b/labgrid/config.py
@@ -174,6 +174,64 @@ class Config:
         assert isinstance(value, str)
         self.data.setdefault('options', {})[name] = value
 
+    def get_target_option(self, target, name, default=None):
+        """Retrieve an entry from the options subkey under the specified target
+           subkey
+
+        Args:
+            target (str): name of the target
+            name (str): name of the option
+            default (str): A default parameter in case the option can not be
+                found
+
+        Returns:
+            str: value of the option or default parameter
+
+        Raises:
+            KeyError: if the requested key can not be found in the
+                configuration, or if the target can not be found in the
+                configuration.
+        """
+        if target not in self.data['targets']:
+            raise KeyError("No target '{}' found in configuration".format(target))
+
+        try:
+            return str(self.data['targets'][target]['options'][name])
+        except (KeyError, TypeError):
+            # Empty target declarations become None in the target dict, hence
+            # TypeError when we try to subscript it.
+            if default is None:
+                raise KeyError("No option '{}' found in configuration for target '{}'".format(name, target))
+            else:
+                return default
+
+    def set_target_option(self, target, name, value):
+        """Set an entry in the options subkey under the specified target subkey
+
+        Args:
+            target (str): name of the target
+            name (str): name of the option
+            value (str): the new value
+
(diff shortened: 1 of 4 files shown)


tobsan:
I see now that some tests ended up in the wrong place. I'll restructure those.

labgrid/config.py
```diff
@@ -153,6 +153,64 @@ def set_option(self, name, value):
...
+            name (str): name of the option
+            value (str): the new value
+
+        Raises:
+            KeyError: if the requested target can not be found in the
+                configuration
+        """
+        assert isinstance(target, str)
+        assert isinstance(name, str)
+        assert isinstance(value, str)
```
Emantor: I don't think its necessary to restrict options to strings here, since the YAML syntax also allows integers, dictionaries or other data types.

labgrid/config.py
```diff
@@ -153,6 +153,64 @@ def set_option(self, name, value):
...
+        Raises:
+            KeyError: if the requested key can not be found in the
+                configuration, or if the target can not be found in the
+                configuration.
+        """
+        if target not in self.data['targets']:
+            raise KeyError("No target '{}' found in configuration".format(target))
+
+        try:
+            return str(self.data['targets'][target]['options'][name])
```
Emantor: Is there a reason why options are stringified?

tobsan: No, but the global options `get_option()` are stringified so I figured it was best to follow that.

labgrid/config.py
tobsan: Fair point, although they are stringified later as you pointed out. We could change so that both per-target options and global options may hold arbitrary value types.

labgrid/config.py
Emantor: Fair, I wonder why we introduced this restriction in the first place.

Emantor: So i looked into this and we probably did this to ensure a consistent data type going through the config. Retrieving options from the YAML will always give you a string, since we have no way to annotate the type. Callers can easily convert the options back and raise the appropriate error message if the conversion fails. I think we should stay with str for target options as well.

jluebbe:
Hmm, the get_features/get_target_features behave differently and are implemented in the Environment. I'd prefer consistency, and the get_target_option (which resolve a specific option for a specific target) seems the intuitive behaviour. @Emantor, why did you implement get_target_features this way (combine features from all targets)?
Also, currently requesting features from a test if you have the target is rather verbose, but that's a different issue.

tobsan:
@jluebbe I did find that confusing as well. I could update this PR to make both behave similarly, but I have no deeper knowledge of how/if the target features are used somewhere, so I'd rather not break something existing...

tobsan:
Updated to fix the conflict in CHANGES.rst

tobsan:
Rebased onto latest master

PR #494, opened 2019-09-20
Doc/updates

Emantor:
**Description**
Documentation updates for longer standing issues

**Checklist**

### DIFF
--- a/man/labgrid-client.1
+++ b/man/labgrid-client.1
@@ -179,6 +179,14 @@ matches anything.
 \fBwait\fP token                  Wait for a reservation to be allocated
 .sp
 \fBreservations\fP                List current reservations
+.SH ADDING NAMED RESOURCES
+.sp
+If a target contains multiple Resources of the same type, named matches need to
+be used to address the individual resources. In addition to the \fImatch\fP taken by
+\fIadd\-match\fP, \fIadd\-named\-match\fP also takes a name for the resource. The other
+client commands support the name as an optional parameter and will inform the
+user that a name is required if multiple resources are found, but no name is
+given.
 .SH EXAMPLES
 .sp
 To retrieve a list of places run:
--- a/man/labgrid-device-config.1
+++ b/man/labgrid-device-config.1
@@ -67,11 +67,27 @@ Defaults to \(aqrealm1\(aq.
 .SH IMAGES
 .sp
 The \fBimages:\fP top key provides paths to access preconfigured images to flash
-onto the board.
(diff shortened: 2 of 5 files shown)


man/labgrid-device-config.1
```diff
@@ -67,11 +67,27 @@ Defaults to \(aqrealm1\(aq.
...
+.SS IMAGE EXAMPLE
+.sp
+Two configured images, one for the root filesystem, one for the bootloader:
+.INDENT 0.0
+.INDENT 3.5
+.sp
+.nf
+.ft C
+images:
+  root: platform\-v7a/root.img
```
jluebbe: I'd suggest quoting the string, as this can avoid confusing errors due to yaml type inference.

man/labgrid-device-config.1
```diff
@@ -87,11 +103,39 @@ Path to the mxs\-usb\-loader binary
...
+.SS TOOLS EXAMPLE
+.sp
+Configure the tool path for \fBimx\-usb\-loader\fP:
+.INDENT 0.0
+.INDENT 3.5
+.sp
+.nf
+.ft C
+tools:
+  imx\-usb\-loader: /opt/labgrid\-helper/imx\-usb\-loader
```
jluebbe: Here as well.

man/labgrid-device-config.1
```diff
@@ -87,11 +103,39 @@ Path to the mxs\-usb\-loader binary
...
+.ft P
+.fi
+.UNINDENT
+.UNINDENT
 .SH IMPORTS
 .sp
 The \fBimports\fP key is a list of files or python modules which
 are imported by the environment after loading the configuration.
 Paths relative to the configuration file are also supported.
+.SS IMPORTS EXAMPLE
```
jluebbe: We should point out this is mainly useful to load drivers and strategies, as they are imported before instantiating the target.

man/labgrid-client.1
```diff
@@ -179,6 +179,13 @@ matches anything.
 \fBwait\fP token                  Wait for a reservation to be allocated
 .sp
 \fBreservations\fP                List current reservations
+.SH ADDING NAMED RESOURCES
+.sp
+If a target should contains multiple Resources of the same type, named matches need to
```
jluebbe:
```suggestion
If a target contains multiple Resources of the same type, named matches need to
```

Emantor:
Added the quoting and additional documentation.

Issue #486, opened 2019-08-22
Traceback with UnicodeEncodeError on C locale


ukleinek:
If a ResourceMatch has a non-None rename attribute showing a place prints a '→'. If the locale is "C" (or something else that is not unicode-capable) this results in:

    Traceback (most recent call last):
      File "/opt/venvs/labgrid/lib/python3.5/site-packages/labgrid/remote/client.py", line 1529, in main
        session.loop.run_until_complete(args.func(session))
      File "/usr/lib/python3.5/asyncio/base_events.py", line 466, in run_until_complete
        return future.result()
      File "/usr/lib/python3.5/asyncio/futures.py", line 293, in result
        raise self._exception
      File "/usr/lib/python3.5/asyncio/tasks.py", line 239, in _step
        result = coro.send(None)
      File "/opt/venvs/labgrid/lib/python3.5/site-packages/labgrid/remote/client.py", line 352, in print_place
        place.show(level=1)
      File "/opt/venvs/labgrid/lib/python3.5/site-packages/labgrid/remote/common.py", line 178, in show
        print(indent + "  {}".format(match))
    UnicodeEncodeError: 'ascii' codec can't encode character '\u2192' in position 35: ordinal not in range(128)

PR #495, opened 2019-09-20
remote/common: replace → to support C locale

Emantor:
**Description**
UTF-8 is nice, but we shouldn't rely on the user having a UTF-8 locale.
$ grep -R --color='auto' -P -n "[^\x00-\x7F]" labgrid/*
was used to search for all None UTF-8 characters, the only other matches
are in a docstring.

Fixes https://github.com/labgrid-project/labgrid/issues/486


jluebbe:
By now, C.UTF-8 should be available. Nevertheless, we don't need to make it harder than necessary.

PR #508, opened 2019-10-09
setup: reenable fastentrypoints

jluebbe:
This was lost in commit 7021d8db4308f6a75283ba6a2b267b50f700496f, and
significantly increased the startup time. Get some performance back by reenabling fastentrypoints.
    
Also add the linter hints to avoid losing it again.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,6 +2,9 @@
 
 from setuptools import setup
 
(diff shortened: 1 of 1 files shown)


Emantor:
Nice fix!

PR #510, opened 2019-10-10
driver/power: add PDUDaemon support

jluebbe:
**Description**
This addresses the long-standing request to add support for PDUDaemon
ports via the PowerProtocol.

**Checklist**
- [ ] Tests for the feature 
- [x] PR has been tested

### DIFF
--- a/labgrid/driver/powerdriver.py
+++ b/labgrid/driver/powerdriver.py
@@ -279,3 +279,52 @@ class USBPowerDriver(Driver, PowerResetMixin, PowerProtocol):
             if b"off" in status:
                 return False
         raise ExecutionError("Did not find port status in uhubctl output ({})".format(repr(output)))
+
+
+@target_factory.reg_driver
+@attr.s(eq=False)
+class PDUDaemonDriver(Driver, PowerResetMixin, PowerProtocol):
+    """PDUDaemonDriver - Driver using a PDU port available via pdudaemon"""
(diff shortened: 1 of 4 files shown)


labgrid/driver/powerdriver.py
```diff
@@ -279,3 +279,49 @@ def get(self):
...
+            res += "&delay={}".format(self.delay)
+        return res
+
+    def on_activate(self):
+        self._host, self._port = proxymanager.get_host_and_port(self.port, default_port=16421)
+
+    @Driver.check_active
+    @step()
+    def on(self):
+        self._requests.get(self._build_url('on'))
```
Emantor: According to the README, pdudaemon can return 503 as a status if the request is not accepted, we should raise an error/exception if this is the case.

jluebbe: requests will raise an exception in that case. ;)

Emantor: iPyhton disagrees:
```
In [4]: requests.get("https://httpstat.us/503")  
Out[4]: <Response [503]>

In [5]:                                          
```
According to the [docs](https://requests.kennethreitz.org/en/master/user/quickstart/#response-status-codes) `raise_for_status()` needs to be called on the response object.

jluebbe: Thanks! I've fixed this and updated the PR.

PR #511, opened 2019-10-11
Convert drivers from imports to strings

Emantor:
**Description**
This PR fixes the lookup by string during the driver binding by introducing a new dictionary in the TargetFactory and using this to look at all registered resources and drivers.
It then converts the imports used for the bindings to strings.

**Checklist**
- [ ] Tests for the feature 
- [x] PR has been tested

Tested by running tox, which also revealed the lookup problem.

### DIFF
--- a/labgrid/driver/serialdriver.py
+++ b/labgrid/driver/serialdriver.py
@@ -8,10 +8,10 @@ import serial.rfc2217
 
 from ..factory import target_factory
 from ..protocol import ConsoleProtocol
-from ..resource import SerialPort, NetworkSerialPort
 from .common import Driver
 from .consoleexpectmixin import ConsoleExpectMixin
 from ..util.proxy import proxymanager
+from ..resource import SerialPort
 
 
 @target_factory.reg_driver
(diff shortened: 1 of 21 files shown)


labgrid/driver/serialdriver.py
```diff
@@ -38,7 +37,7 @@ class SerialDriver(ConsoleExpectMixin, Driver, ConsoleProtocol):
     def __attrs_post_init__(self):
         super().__attrs_post_init__()
         self.logger = logging.getLogger("{}({})".format(self, self.target))
-        if isinstance(self.port, SerialPort):
+        if isinstance(self.port, target_factory.registered_class_from_string("SerialPort")):
```
jluebbe: Please don't replace static instances in normal code (for readability and performance reasons).

labgrid/factory.py
```diff
@@ -147,6 +152,18 @@ def make_target(self, name, config, *, env=None):
             self.make_driver(target, driver, name, args)
         return target
 
+    def registered_class_from_string(self, string: str):
+        print(self.all_classes)
```
jluebbe: Leftover debug print.

labgrid/factory.py
```diff
@@ -1,23 +1,28 @@
+import inspect
```
jluebbe: s/not bound/not registered/ in the commit message ("bound" has a different meaning in labgrid).

labgrid/target.py
```diff
@@ -110,7 +110,7 @@ def get_resource(self, cls, *, name=None, wait_avail=True):
         found = []
         other_names = []
         if isinstance(cls, str):
-            cls = self._class_from_string(cls)
+            cls = self._bound_class_from_string(cls)
```
jluebbe: Why not registered?

labgrid/factory.py
```diff
@@ -147,6 +152,18 @@ def make_target(self, name, config, *, env=None):
             self.make_driver(target, driver, name, args)
         return target
 
+    def registered_class_from_string(self, string: str):
```
jluebbe: The name is wrong (and too long to be convinient). Currently, this function will find even unregistered parent classes.

Emantor:
Updated. Please take a look again.

Emantor:
> "bound" is wrong here. I'd suggest:
> 
> > With this we can lookup classes which have been registered directly, but their base classes as well.

And I thought I got all cases. Fixed.

PR #517, opened 2019-10-26
usbvideodriver: add support for newer C920 Webcams

Emantor:
**Description**
Newer Logitech C920 versions seem to have a different USB VID. They also
seem to support the newer UVC1.5 standard, which is not implemented by
Linux, therefore the h264 output is unfortunately not supported.
This PR adds a separate pipeline for these devices which use the MJPG
output and VA-API to convert to a h264 stream.

- [x] PR has been tested

### DIFF
--- a/labgrid/driver/usbvideodriver.py
+++ b/labgrid/driver/usbvideodriver.py
@@ -23,6 +23,12 @@ class USBVideoDriver(Driver):
                 ("mid", "video/x-h264,width=1280,height=720,framerate=15/2"),
                 ("high", "video/x-h264,width=1920,height=1080,framerate=10/1"),
                 ])
+        if match == (0x046d, 0x0892):
+            return ("mid", [
+                ("low", "image/jpeg,width=640,height=360,framerate=5/1"),
+                ("mid", "image/jpeg,width=1280,height=720,framerate=15/2"),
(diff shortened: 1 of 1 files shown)


labgrid/driver/usbvideodriver.py
```diff
@@ -34,12 +40,21 @@ def select_caps(self, hint=None):
         raise InvalidConfigError("Unkown video format {} for device {:04x}:{:04x}".format(
             variant, self.video.vendor_id, self.video.model_id))
 
+    def get_pipeline(self):
+        match = (self.video.vendor_id, self.video.model_id)
+        if match == (0x046d, 0x082d):
+            return "v4l2src device={} ! {} ! h264parse ! fdsink"
+        if match == (0x046d, 0x0892):
+            return "v4l2src device={} ! {} ! jpegparse ! vaapijpegdec ! videoconvert ! vaapipostproc ! vaapih264enc ! h264parse ! fdsink"
```
jluebbe: According to @michaelolbrich, the following should work as well and would be preferred.

```suggestion
            return "v4l2src device={} ! {} ! decodebin ! vaapipostproc ! vaapih264enc ! h264parse ! fdsink"
```

Emantor:
@jluebbe updated.

PR #518, opened 2019-10-27
target: poll resources directly

Emantor:
**Description**
Instead of polling the managers for the resources, poll the resources
directly. This brings the usage more in line with how these resources
are polled in the exporter, allowing the usage of the USBSDMuxDevice in
local configurations again.

**Checklist**
- [x] PR has been tested


Emantor:
Fixed this to poll only `ManagedResource`s

PR #525, opened 2019-11-05
docker: refactor Dockerfiles to reduce their size

fboudra:
* Base image switched to debian:buster-slim to reduce the size of the
images (python:3 base images is fat). Each image is now ~500M instead of
~1G.
* Introduce a labgrid-base image since the build-dependencies are common
to all images.
* Use buildkit backend to slightly speed up the docker build time.

Signed-off-by: Fathi Boudra <fathi.boudra@linaro.org>

### DIFF
--- a/docker/base/Dockerfile
+++ b/docker/base/Dockerfile
@@ -1,0 +1,13 @@
+FROM debian:buster-slim
+
+LABEL maintainer="eha@deif.com"
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+COPY ./ /opt/labgrid/
+
+RUN set -e ;\
+    apt update -q=2 ;\
+    apt install -q=2 --yes --no-install-recommends python3 python3-dev python3-pip python3-setuptools python3-wheel git build-essential libsnappy-dev ;\
+    apt clean ;\
+    rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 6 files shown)


Emantor:
Looks good, unfortunately the docker within the travis `xenial` image does not support buildkit. Can you change the dist to `bionic`? Hopefully this resolves the build failure.

fboudra:
@Emantor thanks for the suggestion. Looks good now.

jluebbe:
The docker build is still on the default distro and fails, I'll fix that myself.

jluebbe:
@fboudra buildkit isn't available in travic-ci's bionic image either. It seems that we could live without it for now and still get most of the benefits of your PR?

jluebbe:
@krevsbech, you have some additional changes in https://github.com/krevsbech/labgrid-docker-images/commits/master relative to the https://github.com/labgrid-project/labgrid-docker-images (and this repo as well). Is seems they are still relevant?

I've also created https://hub.docker.com/u/labgrid, which is now building the existing images.

krevsbech:
> @krevsbech, you have some additional changes in https://github.com/krevsbech/labgrid-docker-images/commits/master relative to the https://github.com/labgrid-project/labgrid-docker-images (and this repo as well). Is seems they are still relevant?
> 
> I've also created https://hub.docker.com/u/labgrid, which is now building the existing images.

Yes that's right thanks for reminding me. For me it would be okay I we let this be merged first and then create an additional pr, with the dangling changes on my repo. Okay?

I think we should sack the  labgrid-project/labgrid-docker-images repo all has been moved to here anyway.

Regarding the docker hub. Nice.

jluebbe:
@fboudra I've removed buildkit from your PR for now, let's see if this works now.

jluebbe:
I've rebased the PR on current master

PR #527, opened 2019-11-12
driver: networkusbstorage: fix get_size(), wait for medium

Bastian-Krause:
**Description**
- fix string formatting in get_size()
- use sysfs polling to wait for medium

This is a workaround. The resource should only be "available" once the medium is there.

**Checklist**
- [x] PR has been tested

### DIFF
--- a/labgrid/driver/networkusbstoragedriver.py
+++ b/labgrid/driver/networkusbstoragedriver.py
@@ -89,6 +89,6 @@ class NetworkUSBStorageDriver(Driver):
             raise ExecutionError(
                 "{} is not available".format(self.storage_path)
             )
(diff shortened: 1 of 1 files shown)


Bastian-Krause:
Force-pushed an explanation for the `try .. except ValueError`. Here's the exception which is handled with this:
```
    @step(result=True)
    def get_size(self):
        if not self.storage.path:
            raise ExecutionError(
                "{} is not available".format(self.storage_path)
            )                                                                                                                                    
        args = ["cat", "/sys/class/block/{}/size".format(self.storage.path[5:])]
        size = processwrapper.check_output(self.storage.command_prefix + args)
>       return int(size)
E       ValueError: invalid literal for int() with base 10: b''
```

PR #531, opened 2019-11-18
driver/commandmixin: add step args for run_check

Emantor:
Add the missing step arguments for run_check.

- [ ] PR has been tested

### DIFF
--- a/labgrid/driver/commandmixin.py
+++ b/labgrid/driver/commandmixin.py
@@ -82,7 +82,7 @@ class CommandMixin:
         return stdout
 
     @Driver.check_active
-    @step(args=['cmd'], result=True)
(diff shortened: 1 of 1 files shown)


jluebbe:
Is this necessary? This will make the already verbose StepReporter output even more verbose. These arguments are usually unchanged from their defaults.

Emantor:
Maybe we need to add another gating for verbosity for the step arguments, but sometimes its rather useful to see all the input arguments for a given function.

jluebbe:
Or check if the values are actually given (and/or differ from the defaults).

Bastian-Krause:
> Or check if the values are actually given (and/or differ from the defaults).

Including only non-default values sounds like a good approach to me.

Emantor:
But this is more involved than this PR, closing this.

PR #532, opened 2019-11-18
driver: shelldriver: combine marker/prompt expects and increase timeout in _check_prompt()

Bastian-Krause:
**Description**
Use only a single expect call. This allows using a combined timeout for the marker and the prompt.

Due to heavy load on the target echoing the marker and displaying the prompt might take a while. This is no error, so increase the timeout (3 seconds -> 5 seconds).

**Checklist**
- [x] PR has been tested


Bastian-Krause:
Anything left to be done here?

Emantor:
This is assigned to @jluebbe, implying he wants to merge this.

PR #536, opened 2019-11-28
travis: push docker images

jluebbe:
**Description**
With the newly introduced base image (see #525), it looks like we can't use the docker hub builder anymore. Push the finished images from travis-ci instead.

**Checklist**
- [x] PR has been tested


jluebbe:
@krevsbech, maybe you could take a look at this?

PR #540, opened 2019-12-09
Docker staging

krevsbech:
**Description**

Collect the contributions to the docker part that was residing in my personal repository.
Also add the staging environment to travis to ensure a smoke/sanity check of the docker environment.
When this is merged I propose to delete the labgrid-docker-images as everything should be inside this repository.

**Checklist**
- [x] Tests for the feature 
- [x] PR has been tested


krevsbech:
@henriknj Please assist with review.

krevsbech:
I have asked @andreasdotzler to push his commits again where they are signed off. I will cherry-pick those and update this PR.

docker/staging/dut/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+RUN mkdir /var/run/sshd
+RUN echo 'root:PASSWORD' | chpasswd
+RUN echo 'PermitRootLogin yes' >> /etc/ssh/sshd_config
+
+# SSH login fix. Otherwise user is kicked off after login
+RUN sed 's@session\s*required\s*pam_loginuid.so@session optional pam_loginuid.so@g' -i /etc/pam.d/sshd
+
+ENV NOTVISIBLE "in users profile"
+RUN echo "export VISIBLE=now" >> /etc/profile
```
henriknj: Squash the layers into as few as possible.

docker/staging/dut/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+    rm -rf /var/lib/apt/lists/*
+
+RUN mkdir /var/run/sshd
+RUN echo 'root:PASSWORD' | chpasswd
+RUN echo 'PermitRootLogin yes' >> /etc/ssh/sshd_config
+
+# SSH login fix. Otherwise user is kicked off after login
+RUN sed 's@session\s*required\s*pam_loginuid.so@session optional pam_loginuid.so@g' -i /etc/pam.d/sshd
+
+ENV NOTVISIBLE "in users profile"
```
henriknj: Maybe a comment about this trick?

krevsbech:
Issues raised by @henriknj have been addressed in the resent update. @henriknj please check if you are happy.

krevsbech:
Updated with event guards in the client test, to fix the flickering test.

docker/staging/docker-compose.yml
```diff
@@ -0,0 +1,53 @@
...
+    network_mode: "host"
+    user: "${CURRENT_UID}"
+    stdin_open: true
+    command: bash -c "set -e &&
+      /opt/wait-for-it/wait-for-it.sh 127.0.0.1:20408 -- labgrid-exporter /opt/conf/exporter.yaml"
+  dut:
+    build:
+      context: "./dut"
+    network_mode: "host"
+    privileged: true
```
jluebbe: Why does this need to be priviledged, especially if you have a well known root password?

docker/base/Dockerfile
```diff
@@ -10,4 +10,5 @@ RUN set -e ;\
     apt update -q=2 ;\
     apt install -q=2 --yes --no-install-recommends python3 python3-dev python3-pip python3-setuptools python3-wheel git build-essential libsnappy-dev ;\
     apt clean ;\
-    rm -rf /var/lib/apt/lists/*
+    rm -rf /var/lib/apt/lists/* ;\
+    git clone https://github.com/vishnubob/wait-for-it.git /opt/wait-for-it
```
jluebbe: Is it useful to lock this to a specific git hash?

docker/staging/docker-compose.yml
```diff
@@ -0,0 +1,53 @@
...
+    tmpfs: "/tmp"
+    # Use wait-for-it to ensure exporter service is up, as exporter is assuming exporter to
+    # Use labgrid-client r to ensure the exporter has populated the resource list in the coordinator
+    command: bash -c "set -e &&
+      cd /simple-test &&
+      /opt/wait-for-it/wait-for-it.sh 127.0.0.1:20408 &&
+      while [ -z $$(/usr/local/bin/labgrid-client r) ]; do echo 'Wait one sec on coordinator' && sleep 1; done &&
+      /usr/local/bin/labgrid-client -p example-place unlock --kick ||
+      /usr/local/bin/labgrid-client -p example-place lock &&
+      /usr/local/bin/pytest --lg-env remote.yaml -s &&
```
jluebbe: Maybe add `-vv` as well.

docker/staging/docker-compose.yml
```diff
@@ -0,0 +1,53 @@
...
+    stdin_open: true
+    network_mode: "host"
+    tmpfs: "/tmp"
+    # Use wait-for-it to ensure exporter service is up, as exporter is assuming exporter to
+    # Use labgrid-client r to ensure the exporter has populated the resource list in the coordinator
+    command: bash -c "set -e &&
+      cd /simple-test &&
+      /opt/wait-for-it/wait-for-it.sh 127.0.0.1:20408 &&
+      while [ -z $$(/usr/local/bin/labgrid-client r) ]; do echo 'Wait one sec on coordinator' && sleep 1; done &&
+      /usr/local/bin/labgrid-client -p example-place unlock --kick ||
```
jluebbe: I'm wary of suggesting to used unlock --kick in scripts. And it shouldn't be necessary here as we've just started the coordinator.

docker/staging/docker-compose.yml
krevsbech: Actually the privileged keyword will ensure the container to be privileged in terms of what i can do on the host running the container hence it has nothing to do with the root user in the container. That being said you are right there is no good reason for the DUT to be privileged, hence I will remove it

docker/base/Dockerfile
krevsbech: Good point.

docker/staging/docker-compose.yml
krevsbech: Good point, actually I used it while debugging so fine to leave it

docker/staging/docker-compose.yml
krevsbech: You are right, I will delete it.

krevsbech:
Updated according to comments from @jluebbe. Note that I have added a sleep in the test in docker-compose file. I know it is rather smelly but I have no better suggestion to ensure that the coordinator is up and ready before starting to interact with it.

docker/base/Dockerfile
```diff
@@ -9,5 +9,6 @@ COPY ./ /opt/labgrid/
 RUN set -e ;\
     apt update -q=2 ;\
     apt install -q=2 --yes --no-install-recommends python3 python3-dev python3-pip python3-setuptools python3-wheel git build-essential libsnappy-dev ;\
+    ln -sf $(which python3) /usr/bin/python ;\
```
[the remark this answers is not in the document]
jluebbe: This seems to fixed in the meantime.

docker/README.rst
```diff
@@ -101,3 +101,36 @@ Start it with something like:
...
+
+- **coordinator**
+- **exporter**
+- **client**
+- **dut**
+
+The environment serves both to allow checking if the environment still function after changes, and can act as an example
+how to configure the docker images needed to run a minimal setup.
+
+To use the staging environment to conduct a smoke test first build the images as instructed above:
```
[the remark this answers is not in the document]
jluebbe: This seems to fixed in the meantime.

docker/staging/dut/Dockerfile
jluebbe: This seems to fixed in the meantime.

docker/staging/dut/Dockerfile
jluebbe: This seems to fixed in the meantime.

docker/staging/docker-compose.yml
```diff
@@ -37,6 +37,7 @@ services:
     image: "labgrid-exporter"
     volumes:
       - "./exporter-conf:/opt/conf"
+      - "/run/udev:/run/udev:ro"
```
jluebbe: Don't you need to expose /dev as well?

jluebbe:
@andreasdotzler Can we add `Signed-off-by: Andreas Dotzler <contact@andreasdotzler.info>` to your commit? That would be needed to merge it.

.crossbar/config.yaml
```diff
@@ -52,11 +52,11 @@ workers:
     role: authenticator
 - id: coordinator
   type: guest
-  executable: python
+  executable: python3
   arguments:
   - -mlabgrid.remote.coordinator
   options:
-    workdir: ..
+    workdir: ../.crossbar
```
jluebbe: This is basically the same as `workdir: .`.
I'm not really happy to change the generic config to make things a bit simpler for docker. @Emantor, what's your opinion?

docker/staging/docker-compose.yml
```diff
@@ -0,0 +1,54 @@
...
+      /usr/local/bin/labgrid-client -p example-place unlock"
+    depends_on:
+      - coordinator
+      - exporter
+      - dut
+  exporter:
+    image: "labgrid-exporter"
+    volumes:
+      - "./exporter-conf:/opt/conf"
+      - "/run/udev:/run/udev:ro"
```
jluebbe: How does the exporter access the host's devices (mainly serial ports)?

jluebbe:
I've rebased the PR onto current master and adjusted the commit tiles to the existing style.

.crossbar/config.yaml
krevsbech: I would argue this is not only to make it simpler for docker but also making it more distinct what is configuration and what is files used for run-time persistence.

docker/staging/docker-compose.yml
krevsbech: Then one would need to mount dev as well. I don't think we should do that in the compose file as the scenario tested in the compose does not rely on any physical HW, But I will ensure it is documented in the README

.crossbar/config.yaml
jluebbe: You mean for the keys generated by crossbar? That's something I see as internal to crossbar, so keeping our persistence there as well seems strange. Maybe a separate directory?

krevsbech: Ahh now I just read up on why I actually did this fix, sorry for the non precise answer above.
The ting is that we would like to persist the state of .crossbar in the docker host not i the container that's why we do the bind mount. 
However, the problem is that as off today the workdir for crossbar is /opt and not a distinct folder. (That's what this commit add) By running straight in /opt we have:
1) some confusion as it is not evident "who" uses these files 
2) The problem that e.g. also labgrid is installed there hence it can not be bind mounted from out-side. 
I tried bind mounting the relevant files in question instead of the complete folder. However, then I get a problem with your atomic write function as i tries to run a mv on the file, which you can not on a mounted file. That's why I thought that adding a specific .crossbar folder in /opt would be the best solution.

krevsbech:
I've just force pushed an updated README with information on how to mount udev stuff. Moreover, I just read through the README, and in my opinion it should be rather clear that the docker-compose act's and an example and to be able to have a smoke-test of the environment and that people can not use it directly if they e.g. need to bind a device to the exporter container.

krevsbech:
As agreed with @jluebbe I have updated the PR, where I keep out the crossbar path change. As it is A thing fixing legacy stuff that is not related to docker.

krevsbech:
@jluebbe Squashed commit regarding u-dev as agreed.

PR #541, opened 2019-12-10
remote/client: add PDUDaemon driver support

sephalon:
**Description**
Add PDUDaemon driver support to the client.

Tested with [PDUDaemon in revision 941f14c](https://github.com/pdudaemon/pdudaemon/commit/941f14cdb0a5a36777c1bfc6fed87cd9c1ca8362). Older versions of PDUDaemon will still work, but since the HTTP listener falsely always responds with an error (see commit message in above link), the Labgrid client will raise an exception as well.

- [x] PR has been tested


jluebbe:
Thanks for your contribution, @sephalon!

PR #548, opened 2019-12-20
Add copy-files client command

sephalon:
**Description**

Adds a copy-files client command to copy files onto mass storage. In our workflow, we often need to update individual boot files (i.e. ``BOOT.BIN`` and `image.ub` on Xilinx SoC devices), but not the whole partition or device image. The user interface tries to reproduce the semantics of the UNIX ``cp`` command as much as possible.

As of now, we can only copy individual files, but not recursively on directories; in the future we might want to extend the ``ManagedFile`` class to support this use case as well.

Tested with an USB-SD-Mux on Ubuntu 18.04 LTS with pmount installed.

**Checklist**
- [ ] Tests for the feature
- [x] PR has been tested

### DIFF
--- a/labgrid/driver/usbstoragedriver.py
+++ b/labgrid/driver/usbstoragedriver.py
@@ -2,6 +2,8 @@
 import enum
 import logging
 import os
+import pathlib
+import subprocess
 import time
 import subprocess
 
@@ -42,6 +44,11 @@ class USBStorageDriver(Driver):
         default=None,
         validator=attr.validators.optional(attr.validators.instance_of(str))
     )
+    WAIT_FOR_MEDIUM_TIMEOUT = 10.0 # s
+    WAIT_FOR_MEDIUM_SLEEP = 0.5 # s
+    PMOUNT_MEDIA_DIR = pathlib.PurePath('/media') # pmount compile-time configure option
+    PUMOUNT_MAX_RETRIES = 5
+    PUMOUNT_BUSY_WAIT = 3 # s
 
     def __attrs_post_init__(self):
         super().__attrs_post_init__()
@@ -52,6 +59,69 @@ class USBStorageDriver(Driver):
 
     def on_deactivate(self):
         pass
+
+    def _wait_for_medium(self, partition=None):
+        timeout = Timeout(self.WAIT_FOR_MEDIUM_TIMEOUT)
+        while not timeout.expired:
+            if self.get_size(partition) > 0:
+                break
+            time.sleep(self.WAIT_FOR_MEDIUM_SLEEP)
+        else:
+            raise ExecutionError("Timeout while waiting for medium")
+
+    def _pmount(self, partition):
+        mount_args = ["pmount", self.storage.path + str(partition)]
+        self._wait_for_medium(partition)
+        processwrapper.check_output(self.storage.command_prefix + mount_args)
+
+    def _pumount(self, partition):
+        dev_path = self.storage.path + str(partition)
+        for i in range(self.PUMOUNT_MAX_RETRIES):
+            sync_args = ["sync", dev_path]
+            processwrapper.check_output(self.storage.command_prefix + sync_args)
+
+            umount_args = ["pumount", dev_path]
+            try:
+                processwrapper.check_output(self.storage.command_prefix + umount_args)
+            except subprocess.CalledProcessError as e:
+                if e.returncode == 5:
+                    self.logger.info('umount: %s: target is busy; wait for %s s',
+                                     dev_path, self.PUMOUNT_BUSY_WAIT)
+                    time.sleep(self.PUMOUNT_BUSY_WAIT)
+                    continue
+                raise e
+            break
+
+    @Driver.check_active
+    @step(args=['filenames', 'target_dir', 'partition'])
+    def copy_files(self, filenames, target_dir=pathlib.PurePath("."), partition=1):
+        """
+        Copies the file(s) specified by filename(s) to the
+        bound USB storage partition.
+
+        Args:
+            filenames (List[str]): path(s) to the file(s) to be copied to the bound USB storage
+                partition.
+            target_dir (str): optional, target directory to copy to (defaults to partition root
+                directory)
+            partition (int): optional, copy to the specified partition (defaults to first partition)
+        """
+        dev_path = pathlib.PurePath(self.storage.path + str(partition))
+        mount_path = self.PMOUNT_MEDIA_DIR / dev_path.name
+
+        self._pmount(partition)
+        self.logger.debug('Mount %s to %s', dev_path, mount_path)
+
+        target_path = pathlib.PurePath(mount_path) / target_dir
+        try:
+            for f in filenames:
+                mf = ManagedFile(f, self.storage)
+                mf.sync_to_resource()
+                self.logger.debug("Copy %s to %s", mf.get_remote_path(), target_path)
+                cp_args = ["cp", "-t", str(target_path), mf.get_remote_path()]
+                processwrapper.check_output(self.storage.command_prefix + cp_args)
+        finally:
+            self._pumount(partition)
 
     @Driver.check_active
     @step(args=['filename'])
@@ -74,18 +144,7 @@ class USBStorageDriver(Driver):
         mf = ManagedFile(filename, self.storage)
         mf.sync_to_resource()
 
-        # wait for medium
-        timeout = Timeout(10.0)
-        while not timeout.expired:
-            try:
-                if self.get_size() > 0:
-                    break
-                time.sleep(0.5)
-            except ValueError:
-                # when the medium gets ready the sysfs attribute is empty for a short time span
-                continue
-        else:
-            raise ExecutionError("Timeout while waiting for medium")
+        self._wait_for_medium()
 
         partition = "" if partition is None else partition
         remote_path = mf.get_remote_path()
@@ -146,11 +205,26 @@ class USBStorageDriver(Driver):
         )
 
     @Driver.check_active
-    @step(result=True)
-    def get_size(self):
-        args = ["cat", "/sys/class/block/{}/size".format(self.storage.path[5:])]
-        size = subprocess.check_output(self.storage.command_prefix + args)
-        return int(size)*512
+    @step(args=['partition'], result=True)
+    def get_size(self, partition=None):
+        """
+        Get the size of the bound USB storage root device or partition.
+
+        Args:
+            partition (int or None): optional, get size of the specified partition or None for
+                getting the size of the root device (defaults to None)
+
+        Returns:
+            int: size in bytes
+        """
+        partition = "" if partition is None else partition
+        args = ["cat", "/sys/class/block/{}{}/size".format(self.storage.path[5:], partition)]
+        size = processwrapper.check_output(self.storage.command_prefix + args)
+        try:
+            return int(size)
+        except ValueError:
+            # when the medium gets ready the sysfs attribute is empty for a short time span
+            return 0
 
 
 @target_factory.reg_driver
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -4,6 +4,7 @@ import argparse
 import asyncio
 import contextlib
 import os
+import pathlib
 import subprocess
 import traceback
 import logging
@@ -1268,6 +1269,33 @@ class ClientSession(ApplicationSession):
         for k, v in sorted(data.items()):
             print("{:<16s} {:<10s}".format(k, str(v)))
 
+    def copy_files(self):
+        place = self.get_acquired_place()
+        target = self._get_target(place)
+        drv = None
+        from ..resource.remote import NetworkUSBMassStorage, NetworkUSBSDMuxDevice
+        from ..driver import NetworkUSBStorageDriver
+        try:
+            drv = target.get_driver(NetworkUSBStorageDriver)
+        except NoDriverFoundError:
+            for resource in target.resources:
+                if isinstance(resource, (NetworkUSBSDMuxDevice, NetworkUSBMassStorage)):
+                    try:
+                        drv = target.get_driver(NetworkUSBStorageDriver)
+                    except NoDriverFoundError:
+                        drv = NetworkUSBStorageDriver(target, name=None)
+                    drv.storage.timeout = self.args.wait
+                    break
+        if not drv:
+            raise UserError("target has no compatible resource available")
+        target.activate(drv)
+        try:
+            drv.copy_files(self.args.filenames, self.args.target_directory, self.args.partition)
+        except subprocess.CalledProcessError as e:
+            raise UserError("could not copy files to network usb storage: {}".format(e))
+        except FileNotFoundError as e:
+            raise UserError(e)
+
     def write_image(self):
         place = self.get_acquired_place()
         target = self._get_target(place)
@@ -1742,6 +1770,18 @@ def main():
     tmc_subparser.add_argument('channel', type=int)
     tmc_subparser.add_argument('action', choices=['info', 'values'])
     tmc_subparser.set_defaults(func=ClientSession.tmc_channel)
+
+    subparser = subparsers.add_parser('copy-files', help="copy files onto mass storage device")
+    subparser.add_argument('-n', '--partition', type=int, default=1, choices=range(1, 256),
+                           metavar='1-255', help='partition number to copy to (default: %(default)s)')
+    subparser.add_argument('-t', '--target-directory', type=pathlib.PurePath,
+                           default=pathlib.PurePath('.'), metavar='DIRECTORY',
+                           help='copy all files into DIRECTORY')
+    subparser.add_argument('-w', '--wait', type=float, default=10.0,
+                           help='storage poll timeout in seconds')
+    subparser.add_argument('filenames', type=pathlib.PurePath, nargs='+', metavar='filename',
+                           help='filename to copy onto mass storage device')
+    subparser.set_defaults(func=ClientSession.copy_files)
 
     subparser = subparsers.add_parser('write-image', help="write an image onto mass storage")
     subparser.add_argument('-w', '--wait', type=float, default=10.0)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,6 +18,8 @@ New Features in 0.4.0
 - Network controlled relay providing GET/PUT based REST API
 - Improved LG_PROXY documentation in docs/usage.rst.
 - Exporter now checks /usr/sbin/ser2net for SerialPortExport
+- labgrid-client now has a ``copy-files`` subcommand to copy files onto mass
+  storage devices.
 
 Bug fixes in 0.4.0
 ~~~~~~~~~~~~~~~~~~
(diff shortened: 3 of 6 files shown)


labgrid/driver/networkusbstoragedriver.py
```diff
@@ -49,6 +51,52 @@ def on_activate(self):
...
 
+    @step(args=['filenames', 'target_dir', 'partition'])
+    def copy_files(self, filenames, target_dir, partition=1):
+        if not self.storage.path:
+            raise ExecutionError(
+                "{} is not available".format(self.storage_path)
+            )
+
+        dev_path = pathlib.PurePath(self.storage.path + str(partition))
+        mount_path = pathlib.PurePath('/media') / dev_path.name
```
jluebbe: Is this correct everywhere, or is this configurable / different by distribution? Or is there a straightforward way to get the actual path at runtime?

labgrid/driver/networkusbstoragedriver.py
```diff
@@ -49,6 +51,52 @@ def on_activate(self):
...
+                raise e
+
+        target_path = pathlib.PurePath(mount_path) / target_dir
+        for f in filenames:
+            mf = ManagedFile(f, self.storage)
+            mf.sync_to_resource()
+            self.logger.debug("Copy {} to {}".format(
+                mf.get_remote_path(), target_path))
+            cp_args = ["cp", mf.get_remote_path(), "-t", target_path]
+            processwrapper.check_output(self.storage.command_prefix + cp_args)
```
jluebbe: This whole block should probably go into a try/finally, to ensure that the device is unmounted even in error cases.

labgrid/driver/networkusbstoragedriver.py
```diff
@@ -49,6 +51,52 @@ def on_activate(self):
...
+            except subprocess.CalledProcessError as e:
+                if e.returncode == 5:
+                    self.logger.info(
+                            'umount: {}: target is busy; wait for {} s'.format(
+                                dev_path, self.UMOUNT_BUSY_WAIT))
+                    time.sleep(self.UMOUNT_BUSY_WAIT)
+                    continue
+                else:
+                    raise e
+            break
```
jluebbe: This needs to have a retry limit. Also run `sync` on the mount point, which helps against busy on some filesystems.

labgrid/driver/networkusbstoragedriver.py
```diff
@@ -49,6 +51,52 @@ def on_activate(self):
...
+
+        mount_args = ["pmount", dev_path]
+        self.logger.debug('Mount {} to {}'.format(dev_path, mount_path))
+        try:
+            processwrapper.check_output(
+                self.storage.command_prefix + mount_args)
+        except subprocess.CalledProcessError as e:
+            # Proceed if device already has been mounted with pmount, otherwise
+            # give up
+            if not (e.returncode == 4 and str(mount_path) in str(e.output)):
```
jluebbe: If the device is already mounted, we're in an unexpected case, and should probably abort?

labgrid/driver/networkusbstoragedriver.py
```diff
@@ -49,6 +51,52 @@ def on_activate(self):
...
+                raise e
+
+        target_path = pathlib.PurePath(mount_path) / target_dir
```
jluebbe: Would it be useful to run `mkdir -p` on target_dir as well?

labgrid/driver/networkusbstoragedriver.py
```diff
@@ -49,6 +51,52 @@ def on_activate(self):
     def on_deactivate(self):
         pass
 
+    @step(args=['filenames', 'target_dir', 'partition'])
+    def copy_files(self, filenames, target_dir, partition=1):
+        if not self.storage.path:
+            raise ExecutionError(
+                "{} is not available".format(self.storage_path)
```
Bastian-Krause: `self.storage_path` does not exist. This typo was copied from the NetworkUSBStorageDriver. #556 aims to fix it.

labgrid/driver/networkusbstoragedriver.py
```diff
@@ -49,6 +51,52 @@ def on_activate(self):
...
+            )
```
Bastian-Krause: This was copied from NetworkUSBStorageDriver and it's wrong. See #556.

Bastian-Krause: #556 was replaced by #573. If that is merged simply decorate `copy_files()` with `@Driver.check_active`.

jluebbe: #573 has now been merged.

labgrid/driver/networkusbstoragedriver.py
sephalon: [The mount path is determined by pmount](https://linux.die.net/man/1/pmount) and cannot be configured; therefore it should be consistent over different distributions. I do not see any method to get the path at runtime.

labgrid/driver/networkusbstoragedriver.py
sephalon: Fixed

labgrid/driver/networkusbstoragedriver.py
sephalon: Fixed

labgrid/driver/networkusbstoragedriver.py
sephalon: I agree, let us not make too many assumptions here. Removed.

labgrid/driver/networkusbstoragedriver.py
sephalon: Actually I would like to stay as close as possible to the semantics of the UNIX `cp` command; and that is not something `cp` would do.

The idea was to stick to simple use cases of updating individual files, maybe adding a recursive copy option later. For more complicated stuff, we could add some kind of remote mount command (e.g. via SSHFS).

labgrid/driver/networkusbstoragedriver.py
sephalon: Fixed

labgrid/driver/networkusbstoragedriver.py
```diff
@@ -46,6 +50,53 @@ def on_activate(self):
     def on_deactivate(self):
         pass
 
+    def _pmount(self, dev_path, mount_path):
+        mount_args = ["pmount", dev_path]
+        self.logger.debug('Mount {} to {}'.format(dev_path, mount_path))
```
Bastian-Krause: Please don't use format in logging methods, rather use % style formatting and leave the formatting to logging.

labgrid/driver/networkusbstoragedriver.py
```diff
@@ -46,6 +50,53 @@ def on_activate(self):
...
+
+            umount_args = ["pumount", dev_path]
+            try:
+                processwrapper.check_output(
+                    self.storage.command_prefix + umount_args)
+            except subprocess.CalledProcessError as e:
+                if e.returncode == 5:
+                    self.logger.info(
+                            'umount: {}: target is busy; wait for {} s'.format(
+                                dev_path, self.PUMOUNT_BUSY_WAIT))
```
Bastian-Krause: Same as above.

labgrid/driver/networkusbstoragedriver.py
```diff
@@ -46,6 +50,53 @@ def on_activate(self):
...
+                    time.sleep(self.PUMOUNT_BUSY_WAIT)
+                    continue
+                else:
+                    raise e
```
Bastian-Krause: No `else` needed, simply raise.

labgrid/driver/networkusbstoragedriver.py
```diff
@@ -46,6 +50,53 @@ def on_activate(self):
...
+        self._pmount(dev_path, mount_path)
+
+        target_path = pathlib.PurePath(mount_path) / target_dir
+        try:
+            for f in filenames:
+                mf = ManagedFile(f, self.storage)
+                mf.sync_to_resource()
+                self.logger.debug("Copy {} to {}".format(
+                    mf.get_remote_path(), target_path))
+                cp_args = ["cp", mf.get_remote_path(), "-t", target_path]
```
Bastian-Krause: `man cp` says `cp [OPTION]... -t DIRECTORY SOURCE...`.

labgrid/driver/networkusbstoragedriver.py
```diff
@@ -46,6 +50,53 @@ def on_activate(self):
...
+                processwrapper.check_output(
+                    self.storage.command_prefix + cp_args)
```
Bastian-Krause: No need for a line break here.

labgrid/driver/networkusbstoragedriver.py
```diff
@@ -46,6 +50,53 @@ def on_activate(self):
...
+    def _pmount(self, dev_path, mount_path):
+        mount_args = ["pmount", dev_path]
+        self.logger.debug('Mount {} to {}'.format(dev_path, mount_path))
+        processwrapper.check_output(self.storage.command_prefix + mount_args)
+
+    def _pumount(self, dev_path):
+        for i in range(self.PUMOUNT_MAX_RETRIES):
+            sync_args = ["sync", dev_path]
+            processwrapper.check_output(
+                self.storage.command_prefix + sync_args)
```
Bastian-Krause: No need for a line break here, labgrid uses a max line length of 100 characters.

labgrid/driver/networkusbstoragedriver.py
```diff
@@ -46,6 +50,53 @@ def on_activate(self):
...
+
+        self._pmount(dev_path, mount_path)
+
+        target_path = pathlib.PurePath(mount_path) / target_dir
+        try:
+            for f in filenames:
+                mf = ManagedFile(f, self.storage)
+                mf.sync_to_resource()
+                self.logger.debug("Copy {} to {}".format(
+                    mf.get_remote_path(), target_path))
```
Bastian-Krause: No need for a line break here.  % style formatting, please.

labgrid/remote/client.py
```diff
@@ -1077,6 +1078,35 @@ def tmc_channel(self):
...
+                    except NoDriverFoundError:
+                        drv = NetworkUSBStorageDriver(target, name=None)
+                    drv.storage.timeout = self.args.wait
+                    break
+        if not drv:
+            raise UserError("target has no compatible resource available")
+        target.activate(drv)
+        try:
+            drv.copy_files(self.args.filenames,
+                           self.args.target_directory, self.args.partition)
```
Bastian-Krause: No need for a line break.

labgrid/remote/client.py
```diff
@@ -1077,6 +1078,35 @@ def tmc_channel(self):
...
+        except subprocess.CalledProcessError as e:
+            raise UserError(
+                "could not copy files to network usb storage: {}".format(e))
```
Bastian-Krause: No need for a line break.

labgrid/driver/networkusbstoragedriver.py
Bastian-Krause: Wouldn't it be better to do something similar to `NetworkUSBStorageDriver.write_image()`'s waiting for medium? Maybe move that to a separate method and use it? I guess you should wait for the partition size to be >0 there.

labgrid/driver/networkusbstoragedriver.py
Bastian-Krause: You can actually set the media dir in [configure](https://bazaar.launchpad.net/~fourmond/pmount/main/view/head:/configure.ac#L27), but I guess assuming `/media/` is okay.

sephalon: Ok, I was not aware of that. I added a constant + comment for this path so that we are at least aware of that it can be changed theoretically.

labgrid/driver/networkusbstoragedriver.py
sephalon: I tried this out, but the value of the size sysfs attribute does not seem to have a causal relationship to the partition being ready to be unmounted. Do you have any documentation for this "trick"?

sephalon:
All comments should have been addressed now; please rereview.

sephalon:
I actually used YAPF to reformat the code, but now I realized that the line length limit is not set in `setup.cfg` (defaults to PEP8 then). Is the YAPF configuration abandoned?

labgrid/driver/networkusbstoragedriver.py
Bastian-Krause: Are you sure? This works for me:
```
labgrid-client sd-mux dut
(while [ true ]; do [ "$(cat /sys/class/block/sdc1/size 2>/dev/null || echo 0)" -gt 0 ] && break; done; pmount /dev/sdc1)&
labgrid-client sd-mux host
ls /media/sdc1
```

sephalon: Ok, but this is about **mounting** now, the comment refers to lines 59-76 though, which deal with **unmounting**.

Bastian-Krause: You are right, I misread that. You should then implement that for the mount case.

Bastian-Krause:
> I actually used YAPF to reformat the code at some point, but now I realized that the line length limit is not set in `setup.cfg` (defaults to PEP8 then). Is the YAPF configuration abandoned?

Looks like it. I think the `.pylintrc` is more relevant.

labgrid/driver/networkusbstoragedriver.py
sephalon: Done, please review.

sephalon:
Thanks, fixup looks good to me. Should I squash now or are you doing that?

Bastian-Krause:
I squashed the fixup. I tested this and stumbled upon more bugs. I pushed the fixes as fixups. Please have a look one more time.

labgrid/driver/usbstoragedriver.py
```diff
@@ -111,7 +111,7 @@ def copy_files(self, filenames, target_dir, partition=1):
                 mf = ManagedFile(f, self.storage)
                 mf.sync_to_resource()
                 self.logger.debug("Copy %s to %s", mf.get_remote_path(), target_path)
-                cp_args = ["cp", "-t", target_path, mf.get_remote_path()]
+                cp_args = ["cp", "-t", str(target_path), mf.get_remote_path()]
```
sephalon: Oh, apparently ``subprocess.Popen`` accepts path-like objects only starting from Python 3.6.

sephalon:
LGTM, please squash.

labgrid/driver/usbstoragedriver.py
Emantor: Yes, the woes of python 3.5 support.

Bastian-Krause:
Squashed commits, rebased on current master.

@sephalon Please have another look.

Bastian-Krause (dismissed):

labgrid/driver/usbstoragedriver.py
```diff
@@ -47,6 +54,69 @@ def on_activate(self):
...
+    def copy_files(self, filenames, target_dir, partition=1):
+        """
+        Copies the file(s) specified by filename(s) to the
+        bound USB storage partition.
+
+        Args:
+            filenames (List[str]): path(s) to the file(s) to be copied to the bound USB storage
+                partition.
+            target_dir (str): optional, target directory to copy to (defaults to partition root
+                directory)
```
agners, line 100: This parameter seems not to be optional, but I guess using "/" as a default might make sense?

labgrid/driver/usbstoragedriver.py
```diff
@@ -113,11 +172,26 @@ def write_image(self, filename=None, mode=Mode.DD, partition=None, skip=0, seek=
...
+
+        Args:
+            partition (int or None): optional, get size of the specified partition or None for
+                getting the size of the root device (defaults to None)
+
+        Returns:
+            int: size in bytes
+        """
+        partition = "" if partition is None else partition
+        args = ["cat", "/sys/class/block/{}{}/size".format(self.storage.path[5:], partition)]
```
agners, line 188: Hm, this fails for me right after muxing the SD to the test system:

```
FAILED test_rpi4.py::test_usb_storage_size - subprocess.CalledProcessError: Command '['cat', '/sys/class/block/sdj1/size']' returned non-zero exit status 1.
```
Presumably because the partition table is not read yet?

Emantor: In the current design, `_wait_for_medium` needs to be called first. But since `get_size` is a user callable function, we should add an optional timeout parameter and do the wait inside the function. This will avoid confusion for users calling the function directly, as you did inside the test.

labgrid/driver/usbstoragedriver.py
Emantor: Jep, definitely.

labgrid/driver/usbstoragedriver.py
Emantor: We should probably fix this case on the resource level. It does not make much sense to expose a resource as available which is not accessible yet. But this is out of scope for this PR.
Can you provide a small reproducer for this?

labgrid/driver/usbstoragedriver.py
sephalon: Fixed

labgrid/driver/usbstoragedriver.py
agners: This is how I can reproduce it, but I guess it depends on the environment (SD muxer..):

```
    d = target.get_driver('USBSDWireDriver')
    d.set_mode('dut')
    time.sleep(1)
    d.set_mode('host')
    storage = target.get_driver('USBStorageDriver')
    # Without one of the two below the error appears.
    #storage._wait_for_medium()
    #time.sleep(2)
    storage.copy_files([ "cmdline2.txt" ])
```

Note that while debugging this I also run into the PTY issue (https://github.com/labgrid-project/labgrid/commit/42f08d5ea32ced0fa8e7993e0fce6f5963a4349e), but with that fixed it is now really just the card reader taking its time until the media is ready.

Emantor: This should be fixed with the latest `USBSDWireDriver`, since it overrides the avail property and only signals availability if both the control and storage device are available. Can you confirm that?

sephalon: I have rebased to the latest master; @agners would you please confirm that it works with the fixed `USBSDWireDriver`?

Emantor:
This needs a rebase for CHANGES.rst.

sephalon:
Done, rebased.

labgrid/driver/usbstoragedriver.py
sephalon: I does not seem like @agners is coming back on this anymore, but I cannot reproduce it either since I do not have a USB-SD-Wire device. However as far as I understand this should have been fixed anyway by now – can we just merge this PR and open an issue if this problem pops up again?

sephalon:
Rebased once again.

Emantor:
The only thing I don't like about this PR is the reliance on pmount instead of a standard DBUS interface like udisks2. There is also #820 now, so you may be able to collaborate on this.

sessl3r:
Do you want us to continue on this PR or do you prefer the #820 and we shall not spend any time in this PR as it will not be accepted anyways? Any opinions here?
If it makes sense to continue we are happy to do so!

Emantor:
Closing this, I will open a new issue with the relevant information.

PR #550, opened 2019-12-31
driver/power: Add power driver to support GEMBIRD SiS-PM Control

mbgg:
This patch adds support for SiS-PM (Silver Shield PM) Control
for Linux, aka sismpctl.

Signed-off-by: Matthias Brugger <matthias.bgg@kernel.org>

**Checklist**
- [x] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -37,6 +37,7 @@ New Features in 0.3.0
 - UBootDriver now allows overriding of default boot command (``run bootcmd``)
   via new ``boot_command`` argument.
 - The config file supports per-target options, in addition to global options.
+- Add power driver to support GEMBIRD SiS-PM implementing SiSPMPowerDriver.
 
 Breaking changes in 0.3.0
 ~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -183,6 +183,25 @@ Used by:
    looking for, paths without the interface will fail to match since they use
    the ``usb`` driver.
 
+SiSPMPowerPort
++++++++++++++++++
+A SiSPMPowerPort describes a GEMBIRD SiS-PM as supported by
+`sispmctl <https://sourceforge.net/projects/sispmctl/>`_.
+
+.. code-block:: yaml
+
+   SiSPMPowerPort:
+     match:
+       ID_PATH: platform-1c1a400.usb-usb-0:2
+     index: 1
+
+The example describes port 1 on the hub with the ID_PATH
+"platform-1c1a400.usb-usb-0:2".
+
+- index (int): number of the port to switch
+
+Used by:
+  - `SiSPMPowerDriver`_
 
 ModbusTCPCoil
 ~~~~~~~~~~~~~
@@ -1248,6 +1267,25 @@ Implements:
 .. code-block:: yaml
 
    USBPowerPort:
+     delay: 5.0
+
+Arguments:
+  - delay (float): optional delay in seconds between off and on
+
+SiSPMPowerDriver
+~~~~~~~~~~~~~~~~~~~
+A SiSPMPowerDriver controls a `SiSPMPowerPort`, allowing control of the target
+power state without user interaction.
+
+Binds to:
+  - `SiSPMPowerPort`_
+
+Implements:
+  - :any:`PowerProtocol`
+
+.. code-block:: yaml
+
+   SiSPMPowerPort:
      delay: 5.0
 
 Arguments:
--- a/labgrid/driver/powerdriver.py
+++ b/labgrid/driver/powerdriver.py
@@ -59,6 +59,60 @@ class ManualPowerDriver(Driver, PowerResetMixin, PowerProtocol):
 
 @target_factory.reg_driver
 @attr.s(eq=False)
+class SiSPMPowerDriver(Driver, PowerResetMixin, PowerProtocol):
+    """SiSPMPowerDriver - Driver using a SiS-PM (Silver Shield PM) to control a
+       target's power using the sispmctl tool - http://sispmctl.sourceforge.net/"""
+
+    bindings = {"port": {"SiSPMPowerPort", "NetworkSiSPMPowerPort"}, }
+    delay = attr.ib(default=2.0, validator=attr.validators.instance_of(float))
+
+    def __attrs_post_init__(self):
+        super().__attrs_post_init__()
+        if self.target.env:
+            self.tool = self.target.env.config.get_tool('sispmctl') or 'sispmctl'
+        else:
+            self.tool = 'sispmctl'
+
+    def _get_sispmctl_prefix(self):
+        return self.port.command_prefix + [
+            self.tool,
+            "-U", "{:03d}:{:03d}".format(self.port.busnum, self.port.devnum),
+        ]
+
+    @Driver.check_active
+    @step()
+    def on(self):
+        cmd = ['-o', str(self.port.index)]
+        processwrapper.check_output(self._get_sispmctl_prefix() + cmd)
+
+    @Driver.check_active
+    @step()
+    def off(self):
+        cmd = ['-f', str(self.port.index)]
+        processwrapper.check_output(self._get_sispmctl_prefix() + cmd)
+
+    @Driver.check_active
+    @step()
+    def cycle(self):
+        self.off()
+        time.sleep(self.delay)
+        self.on()
+
+    @Driver.check_active
+    @step()
+    def get(self):
+        cmd = ['-q', '-g', str(self.port.index)]
+        output = processwrapper.check_output(self._get_sispmctl_prefix() + cmd)
+        if output.strip() == b"on":
+            return True
+        if output.strip() == b"off":
+            return False
+        raise ExecutionError("Did not find port status in sispmctl output ({})"
+                             .format(repr(output)))
+
+
+@target_factory.reg_driver
+@attr.s(eq=False)
 class ExternalPowerDriver(Driver, PowerResetMixin, PowerProtocol):
     """ExternalPowerDriver - Driver using an external command to control a target's power"""
     cmd_on = attr.ib(validator=attr.validators.instance_of(str))
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -693,9 +693,10 @@ class ClientSession(ApplicationSession):
         action = self.args.action
         delay = self.args.delay
         target = self._get_target(place)
-        from ..driver.powerdriver import NetworkPowerDriver, PDUDaemonDriver, USBPowerDriver
+        from ..driver.powerdriver import (NetworkPowerDriver, PDUDaemonDriver,
+                                          USBPowerDriver, SiSPMPowerDriver)
         from ..resource.power import NetworkPowerPort, PDUDaemonPort
-        from ..resource.remote import NetworkUSBPowerPort
+        from ..resource.remote import NetworkUSBPowerPort, NetworkSiSPMPowerPort
 
         drv = None
         try:
@@ -714,6 +715,12 @@ class ClientSession(ApplicationSession):
                     except NoDriverFoundError:
                         drv = USBPowerDriver(target, name=None, delay=delay)
                         break
+                elif isinstance(resource, NetworkSiSPMPowerPort):
+                    try:
+                        drv = target.get_driver(SiSPMPowerDriver)
+                    except NoDriverFoundError:
+                        drv = SiSPMPowerDriver(target, name=None, delay=delay)
+                    break
                 elif isinstance(resource, PDUDaemonPort):
                     try:
                         drv = target.get_driver(PDUDaemonDriver)
--- a/labgrid/resource/remote.py
+++ b/labgrid/resource/remote.py
@@ -219,6 +219,15 @@ class NetworkUSBSDWireDevice(RemoteUSBResource):
         self.timeout = 10.0
         super().__attrs_post_init__()
 
+@target_factory.reg_resource
+@attr.s(eq=False)
+class NetworkSiSPMPowerPort(RemoteUSBResource):
+    """The NetworkSiSPMPowerPort describes a remotely accessible SiS-PM power port"""
+    index = attr.ib(default=None, validator=attr.validators.instance_of(int))
+    def __attrs_post_init__(self):
+        self.timeout = 10.0
+        super().__attrs_post_init__()
+
 
 @target_factory.reg_resource
 @attr.s(eq=False)
--- a/labgrid/resource/udev.py
+++ b/labgrid/resource/udev.py
@@ -519,6 +519,27 @@ class USBPowerPort(USBResource):
 
 @target_factory.reg_resource
 @attr.s(eq=False)
+class SiSPMPowerPort(USBResource):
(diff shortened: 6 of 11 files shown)


labgrid/resource/udev.py
```diff
@@ -456,6 +456,40 @@ def __attrs_post_init__(self):
...
+                ("04b4", "fd15")
+                ]:
+            return False
+
+        return super().filter_match(device)
+
+    # Overwrite the avail attribute with our internal property
+    @property
+    def avail(self):
+        return bool(self.index)
```
Bastian-Krause: This seems wrong. How is this supposed to determine if the port is available?

jluebbe:
> This is a first PR to get some feedback on the overall approach. There are some edges, espacially the fact that I'm overwriting the avail property in the SISPMCTLPowerPort. I'd like to get some guidance on this issue as it looks like wrong (also working) to overwrite the property.

As far as I can see from http://sispmctl.sourceforge.net/#mozTocId756141  and https://sourceforge.net/p/sispmctl/git/ci/master/tree/src/main.c#l100, `sispmctl` currently has no way to point to a single device by bus- and device number or topology. This also shows in that you don't actually use any information from udev/USB when calling the tool.

It seems that newer versions support `-D` to select a device by serial number. But, 
when looking at https://sourceforge.net/p/sispmctl/git/ci/master/tree/src/sispm_ctl.c#l58 (`get_serial()`), it needs to talk to the device to get the serial number, so multiple parallel accesses could conflict.

If you can get the same serial number also via the standard USB enumeration (i.e. via `udevadm info`), you could pass that from the exporter and use it for identification. Another alternative is to extend `sispmctl` to accept bus/devnums or the USB topology path (see labgrid/driver/fastbootdriver.py, look for self.fastboot.path), which would also avoid conflicts between parallel accesses to different devices.

Regarding overriding the index, I'm not sure why you need that to make it work. It seems to me that the index would be fixed, so avail would never change?

Bastian-Krause:
> As far as I can see from http://sispmctl.sourceforge.net/#mozTocId756141 and https://sourceforge.net/p/sispmctl/git/ci/master/tree/src/main.c#l100, `sispmctl` currently has no way to point to a single device by bus- and device number or topology. This also shows in that you don't actually use any information from udev/USB when calling the tool.
> 
> It seems that newer versions support `-D` to select a device by serial number. But,
> when looking at https://sourceforge.net/p/sispmctl/git/ci/master/tree/src/sispm_ctl.c#l58 (`get_serial()`), it needs to talk to the device to get the serial number, so multiple parallel accesses could conflict.
> 
> If you can get the same serial number also via the standard USB enumeration (i.e. via `udevadm info`), you could pass that from the exporter and use it for identification.

I've tried this with my SiS-PM, the serial number is not reflected in the udev properties.

> Another alternative is to extend `sispmctl` to accept bus/devnums or the USB topology path (see labgrid/driver/fastbootdriver.py, look for self.fastboot.path), which would also avoid conflicts between parallel accesses to different devices.

Yes, I agree, access via bus/devnum would be ideal.

Another possible work around might be looking up the serial number for a bus/devnum via `sispmctl -s` (it prints bus/devnum/serial) and subsequently using the serial for the actual on/off/get `sispmctl -D` calls. That needs to be done on every call though..

labgrid/resource/udev.py
Bastian-Krause: Am I missing something here? This has not been addressed yet, right?

Bastian-Krause:
Adding the serial number as an attribute was not what I meant actually:

> Another possible work around might be looking up the serial number for a bus/devnum via sispmctl -s (it prints bus/devnum/serial) and subsequently using the serial for the actual on/off/get sispmctl -D calls. That needs to be done on every call though..

The workaround I suggested is looking up the serial number whenever it is required, no caching. This would at least allow addressing via bus/devnum, although it's overly complicated.

@Emantor / @jluebbe have to decide which approach they like better.

labgrid/resource/udev.py
mbgg: No it's not. I'll have a look now. But it might take time to get this fixed by me. I'm not a python developer so bare with me.

mbgg:
> > Another alternative is to extend `sispmctl` to accept bus/devnums or the USB topology path (see labgrid/driver/fastbootdriver.py, look for self.fastboot.path), which would also avoid conflicts between parallel accesses to different devices.
> 
> Yes, I agree, access via bus/devnum would be ideal.
> 

So do I understand correctly that you want sispmctl to be able to access devices by passing it USB Bus:Device (e.g. 002:007). I implemented this here [1]. After this got merged, next step would be to set the command_prefix in powerdriver.py to '-U self.port.busnum:self.port.devnum'. Did I understand that correctly?

[1] https://github.com/mbgg/sispmctl/commit/493ba6fba758d85991503683617e4e3ce39dd5a0

labgrid/resource/udev.py
mbgg: seems like I don't need that. I'm able to discover the device without overwriting avail function. I'll delete this.

mbgg:
Please let me know what else is needed. An entry in the CHANGES.rst, anything else?

labgrid/driver/powerdriver.py
```diff
@@ -56,6 +56,48 @@ def cycle(self):
...
+    """SISPMCTLPowerDriver - Driver using a SiS-PM (Silver Shield PM) to control a
+       target's power using the sispmctl tool - http://sispmctl.sourceforge.net/"""
+
+    bindings = {"port": {"SISPMCTLPowerPort", "NetworkSISPMCTLPowerPort"}, }
+    delay = attr.ib(default=2.0, validator=attr.validators.instance_of(float))
+
+    def __attrs_post_init__(self):
+        super().__attrs_post_init__()
+        if self.target.env:
+            self.tool = self.target.env.config.get_tool('sispmctls') or 'sispmctl'
```
Bastian-Krause, line 71: What's `sispmctls`?

labgrid/resource/udev.py
```diff
@@ -466,6 +466,30 @@ def __attrs_post_init__(self):
...
+    def filter_match(self, device):
+        match = (device.properties.get('ID_VENDOR_ID'), device.properties.get('ID_MODEL_ID'))
+
+        if match not in [
+                ("04b4", "fd10"),
+                ("04b4", "fd11"),
+                ("04b4", "fd12"),
+                ("04b4", "fd13"),
+                ("04b4", "fd15")
+                ]:
```
Bastian-Krause, line 488: Indentation looks incorrect here.

labgrid/driver/powerdriver.py
mbgg: that's a typo. I'll fix this.

labgrid/resource/udev.py
mbgg: right, I'll fix this

mbgg:
> > Everything else looks good to me. Did you test this version?
> 
> Unfortunately I'm not able to start the crossbar with the newest version:

Forget about that comment. I was able to setup crossbar and did test against the latest branch. It works.

mbgg:
anything else you need from me on this one?

mbgg:
I banged my head the whole night to find out why my Gembird is marked as not available.
When from the labgrid-exporter in SISPMCTLPowerPort the filter_match gets called, the match tuple is (none, none). If I return true here, things work as expected.

When calling `labgrid-suggest -d` the device get's found but marked as not available.

`# labgrid-suggest -d 
  DEBUG USBResource           found match: SISPMCTLPowerPort(target=None, name=None, state=<BindingState.idle: 0>, avail=False, timeout=5.0, match={'SUBSYSTEM': 'usb'}, device=None, index=0)
=== existing device ===
  SISPMCTLPowerPort for /devices/platform/soc/1c1b000.usb/usb3/3-1/3-1.4
  === device properties ===
  device node: /dev/bus/usb/003/006
  vendor: Gembird
  vendor (DB): Cypress Semiconductor Corp.
  model: EG-PMS2
  revision: 0101
  === suggested matches ===
  SISPMCTLPowerPort:
    match:
      ID_PATH: platform-1c1b000.usb-usb-0:1.4
  ---
`

It seems to me that the exporter somehow does not pass vendor_id and model_id but TBH I really would need someone who knows the code base and is ready to have a look.

mbgg:
> 
> It seems to me that the exporter somehow does not pass vendor_id and model_id but TBH I really would need someone who knows the code base and is ready to have a look.

Maybe I just needed to write this message to realize that my .yaml file looked like this:

> SISPMCTLPowerPort:
>              match:
>                      '@ID_PATH': platform-1c1b000.usb-usb-0:1.4
>              index: 4

After deleting the bloody '@' it works now...

mbgg:
This is pending for quite some time. Any comments, do you need anything from me?

mbgg:
@Bastian-Krause friendly reminder, it's been a while since this waits for a review :)

labgrid/driver/powerdriver.py
```diff
@@ -56,6 +56,48 @@ def cycle(self):
...
+        else:
+            self.tool = 'sispmctl'
+
+    def _get_sispmctl_prefix(self):
+        prefix = self.port.command_prefix+[
+            self.tool,
+            "-U", "{0:03d}:{1:03d}".format(self.port.busnum, self.port.devnum),
+        ]
+
+        return prefix
```
Bastian-Krause, line 81:
```suggestion
        return self.port.command_prefix + [
            self.tool,
            "-U", "{:03d}:{:03d}".format(self.port.busnum, self.port.devnum),
        ]
```

labgrid/remote/client.py
```diff
@@ -693,9 +693,9 @@ def power(self):
         action = self.args.action
         delay = self.args.delay
         target = self._get_target(place)
-        from ..driver.powerdriver import NetworkPowerDriver, PDUDaemonDriver, USBPowerDriver
+        from ..driver.powerdriver import NetworkPowerDriver, PDUDaemonDriver, USBPowerDriver, SISPMCTLPowerDriver
```
Bastian-Krause, line 696:
```suggestion
        from ..driver.powerdriver import (NetworkPowerDriver, PDUDaemonDriver,
                                          USBPowerDriver, SISPMCTLPowerDriver)
```

labgrid/resource/remote.py
```diff
@@ -207,6 +207,16 @@ def __attrs_post_init__(self):
         super().__attrs_post_init__()
 
 
+@target_factory.reg_resource
+@attr.s(eq=False)
+class NetworkSISPMCTLPowerPort(RemoteUSBResource):
+    """The NetworkSISPMCTLPowerPort describes a remotely accessible GEMBIRD power port"""
```
Bastian-Krause, line 213:
```suggestion
    """The NetworkSISPMCTLPowerPort describes a remotely accessible SiS-PM power port"""
```

CHANGES.rst
```diff
@@ -34,6 +34,7 @@ New Features in 0.3.0
   in some cases.
 - ``labgrid-client write-image`` gained new arguments: ``--partition``,
   ``--skip``, ``--seek``.
+- Add power driver to support GEMBIRD SiS-PM Control implementing SISPMCTLPowerDriver.
```
Bastian-Krause, line 37:
```suggestion
- Add power driver to support GEMBIRD SiS-PM implementing SISPMCTLPowerDriver.
```

labgrid/resource/udev.py
```diff
@@ -466,6 +466,27 @@ def __attrs_post_init__(self):
         self.match['DRIVER'] = 'hub'
         super().__attrs_post_init__()
 
+@target_factory.reg_resource
+@attr.s(eq=False)
+class SISPMCTLPowerPort(USBResource):
+    """This resource describes a SiS-PM (Silver Shield PM) Control power port.
```
Bastian-Krause, line 472:
```suggestion
    """This resource describes a SiS-PM (Silver Shield PM) power port.
```

Bastian-Krause (changes requested):
Sorry for the very late response.

The naming is weird. This product is called "GEMBIRD Silver Shield PM", short "SiS-PM". The tool to control it is called "sispmctl". So:

`SISPMCTLPowerDriver` -> `SiSPMPowerDriver`
`SISPMCTLPowerPortExport` -> `SiSPMPowerPortExport`
`SISPMCTLPowerPort` -> `SiSPMPowerPort`
`NetworkSISPMCTLPowerPort` -> `NetworkSiSPMPowerPort`

doc/configuration.rst
```diff
@@ -176,6 +176,27 @@ The example describes port 1 on the hub with the ID_PATH
...
+.. code-block:: yaml
+
+   SISPMCTLPowerPort:
+     match:
+       ID_PATH: platform-1c1a400.usb-usb-0:2
+     index: 1
+
+The example describes port 1 on the hub with the ID_PATH
+"platform-1c1a400.usb-usb-0:2".
+(use ``udevadm info /sys/bus/usb/devices/...`` to find the ID_PATH value)
```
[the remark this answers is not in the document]
mbgg: Ok, I delete the line and push

Bastian-Krause:
What about the suggested renaming?

mbgg:
> What about the suggested renaming?

Argh, forgot about it. I'll do it these days. Sorry for the noise.

Bastian-Krause:
Alright, looks good otherwise. Please rebase when you're at it. I have my SiS-PM lying around for a final test :)

labgrid/driver/powerdriver.py
```diff
@@ -56,6 +56,46 @@ def cycle(self):
...
+    def off(self):
+        cmd = ['-f', str(self.port.index)]
+        processwrapper.check_output(self._get_sispmctl_prefix() + cmd)
+
+    @Driver.check_active
+    @step()
+    def cycle(self):
+            self.off()
+            time.sleep(self.delay)
+            self.on()
```
Bastian-Krause, line 98: Bad indentation.

Emantor:
A rebase on latest master should fix the build errors. I just merged a fix for the CI.

Bastian-Krause (approved):
I've tested this, works like a charm, thanks!

I've pushed a fixup adding `SiSPMPowerDriver.get()` to get the current state of the socket (allowing e.g. `labgrid-client pw get`). While at it, I added an empty line before and after the SiSPMPowerDriver class definition.

If you are okay with the fixup, please squash it. Then we should be good to merge, finally :)

mbgg:
> I've tested this, works like a charm, thanks!
> 
> I've pushed a fixup adding `SiSPMPowerDriver.get()` to get the current state of the socket (allowing e.g. `labgrid-client pw get`). While at it, I added an empty line before and after the SiSPMPowerDriver class definition.
> 
> If you are okay with the fixup, please squash it. Then we should be good to merge, finally :)

sorry I accidentally did a git push -f overwriting the changes you proposed. Unfortunately I can't see the code you changed now, could you push you changes again, please. Sorry for the inconvenience.

Bastian-Krause:
Here you go.

mbgg:
> Here you go.

Thaks a lot. Changes look good to me. I applied them and pushed again.

mbgg:
It's a year now since I created this PR. Is there anything else you need from me to get this accepted?

Emantor:
> It's a year now since I created this PR. Is there anything else you need from me to get this accepted?

No, I don't think so. Thanks for the contribution.

PR #552, opened 2020-01-06
RFC: Add BackgroundProcessProtocol, let SSHDriver/ShellDriver implement it 

Bastian-Krause:
**Description**
The _CommandProtocol_ allows executing commands on the target. `CommandProtocol.run()` blocks until the process terminated and stdout/stderr/exit code are returned. Some use cases require non-blocking communication with the process because something else needs to be done concurrently (think of a listening _netcat_ on the target and another _netcat_ on the testing host connecting to it). Maybe even multiple commands need to be run concurrently (if the underlying communication mechanism allows that).

The _BackgroundProcessProtocol_ allows exactly that. It defines a method to start a command in background, a method to read stdout/stderr from that background process and a method to check if the background process terminated (and if so with which exit code). The abstract _BackgroundProcessProtocol_ provides an example how to use it. Users of this protocol have to implement the timeout handling themselves. In order to simplify it `labgrid.util.timeout` can now be used as a context manager (see example).

This PR implements this protocol for the _SSHDriver_ and the _ShellDriver_. These driver probably benefit most from this feature.

**Checklist**
- [x] Documentation for the feature (in the BackgroundProcessProtocol)
- [ ] Tests for the feature 
- [x] PR has been tested

### DIFF
--- a/labgrid/driver/shelldriver.py
+++ b/labgrid/driver/shelldriver.py
@@ -11,17 +11,35 @@ from pexpect import TIMEOUT
 import xmodem
 
 from ..factory import target_factory
-from ..protocol import CommandProtocol, ConsoleProtocol, FileTransferProtocol
+from ..protocol import (CommandProtocol, ConsoleProtocol, FileTransferProtocol,
+                        BackgroundProcessProtocol)
 from ..step import step
 from ..util import gen_marker
+from ..util.timeout import Timeout
 from .commandmixin import CommandMixin
 from .common import Driver
 from .exception import ExecutionError
 
 
+@attr.s(eq=False)
+class ShellProcess:
+    """Represents a shell background process"""
+    command = attr.ib(validator=attr.validators.instance_of(str))
+    marker = attr.ib(validator=attr.validators.instance_of(str))
+    exitcode = attr.ib(
+        default=None,
+        validator=attr.validators.optional(attr.validators.instance_of(int))
+    )
+
+    @property
+    def running(self):
+        return self.exitcode is None
+
+
 @target_factory.reg_driver
 @attr.s(eq=False)
-class ShellDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
+class ShellDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol,
+                  BackgroundProcessProtocol):
     """ShellDriver - Driver to execute commands on the shell
     ShellDriver binds on top of a ConsoleProtocol.
 
@@ -55,6 +73,8 @@ class ShellDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
         self._xmodem_cached_rx_cmd = ""
         self._xmodem_cached_sx_cmd = ""
 
+        self.background_process = None
+
     def on_activate(self):
         if self._status == 0:
             self._await_login()
@@ -71,7 +91,17 @@ class ShellDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
         self._run("dmesg -n 1")
 
     def on_deactivate(self):
+        if self.background_process.running:
+            self.console.sendcontrol('c')
+            self.console.expect(self.prompt)
+            self.background_process.exitcode = 130
+
         self._status = 0
+
+    @Driver.check_active
+    @step(args=['cmd'], result=True)
+    def run(self, cmd, timeout=30.0, codec="utf-8", decodeerrors="strict"):
+        return self._run(cmd, timeout=timeout, codec=codec, decodeerrors=decodeerrors)
 
     def _run(self, cmd, *, timeout=30.0, codec="utf-8", decodeerrors="strict"):
         """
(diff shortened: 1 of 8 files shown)


labgrid/driver/shelldriver.py
```diff
@@ -71,39 +91,151 @@ def on_activate(self):
...
+        if not timeout:
+            expect.append(TIMEOUT)
+
+        index, before, match, _ = self.console.expect(expect, timeout=timeout)
+
+        if index == 0:
+            self.background_process.exitcode = int(match.group(2))
+            stdout = match.group(1)
+        elif index == 1:
+            stdout = self.console._expect.read(size=len(before))
```
Bastian-Krause: Is there a better way to retrieve output until EOF without using the private `_expect.read()` here?

labgrid/util/timeout.py
```diff
@@ -23,3 +24,10 @@ def remaining(self):
     @property
     def expired(self):
         return self._deadline <= time.monotonic()
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, *args):
+        if self.expired:
+            raise TIMEOUT("Timeout of {} seconds exceeded".format(self.timeout))
```
Bastian-Krause: Is `pexpect.TIMEOUT` the correct exception to raise here?

labgrid/driver/shelldriver.py
```diff
@@ -71,39 +91,151 @@ def on_activate(self):
...
         cmd - cmd to run on the shell
         """
-        # FIXME: Handle pexpect Timeout
+        timeout_ = None if timeout is None else Timeout(float(timeout))
+
+        process = self._run_as_background_process(cmd)
+        read_timeout = timeout_.remaining if timeout_ else None
+        stdout, stderr = self._read_from_background_process(process, timeout=read_timeout)
+
+        if timeout_ and timeout_.expired:
```
jremmet: Maybe this is a race condition? Timeout can expire short after _read_from_background_process successfully read till exitcode. 
```suggestion
        if timeout_ and timeout_.expired and self.background_process.running:
```
Do we need to cleanup the process in case of a timeout here?

Bastian-Krause: You are right, this situation can happen. Here in `ShellDriver._run()` your solution is correct. But what about the usage example of the _BackgroundProcessProtocol_ [here](https://github.com/labgrid-project/labgrid/pull/552/files#diff-8032c949da39d97978728cb39890d762R12)? Should we live this corner case there?

Bastian-Krause: Regarding clean up: yes, we need to clean up in that case, that's missing. I guess we should move the background process clean up lines from `ShellDriver.on_deactivate()` into a separate function. Users of the BackgroundProcessProtocol need to do this themselves too, so we should add something like `BackgroundProcessProtocol.clean_up_background_process()`. I am still not sure how to expose all this is in a user-friendly way. Suggestions welcome :)

labgrid/driver/sshdriver.py
```diff
@@ -118,10 +126,46 @@ def _run(self, cmd, codec="utf-8", decodeerrors="strict", timeout=None): # pylin
...
+        return self._run(cmd, codec=codec, decodeerrors=decodeerrors)
+
+    def _run(self, cmd, codec="utf-8", decodeerrors="strict", timeout=None): # pylint: disable=unused-argument
+        timeout_ = None if timeout is None else Timeout(float(timeout))
+
+        process = self._run_as_background_process(cmd)
+        read_timeout = timeout_.remaining if timeout_ else None
+        stdout, stderr = self._read_from_background_process(process, timeout=read_timeout)
+
+        if timeout_ and timeout_.expired:
```
jremmet: Also check if process is meanwhile finished here like in the shell driver
```suggestion
        if timeout_ and timeout_.expired and process.poll():
```

labgrid/driver/shelldriver.py
jremmet: My usecase is to start a not stopping command (like iperf -s) on the target, send some data over other interface and than force stop the command on the target and get the output so far.
So I like the idea of `BackgroundProcessProtocol.clean_up_background_process()`, maybe we call it `BackgroundProcessProtocol.stop_background_process()`

jremmet: for the usage example of the BackgroundProcessProtocol: whats about adding the timout to `BackgroundProcessProtocol.run_as_background_process`? `BackgroundProcessProtocol.poll_background_process()` could than raise TIMEOUT if needed

something like
```
process = run_as_background_process('dd if=/dev/random bs=1M count=100', timeout=300.0)
try:
    while driver.poll_background_process(process) is None:
       stdout, stderr = driver.read_from_background_process(process, timeout=1.0)
       # do something else
    exitcode = driver.poll_background_process(process)
  except TIMEOUT:
      driver.stop_background_process(process)
```
`BackgroundProcessProtocol.read_from_background_process()` blocks here and servers as sleep.

labgrid/protocol/backgroundprocessprotocol.py
```diff
@@ -0,0 +1,67 @@
...
+    driver.
+
+    Example usage:
+
+    >>> from labgrid.util.timeout import Timeout
+    >>> with Timeout(300.0) as timeout:
+    >>>     process = driver.run_as_background_process('dd if=/dev/random bs=1M count=100')
+    >>>     while driver.poll_background_process(process) is None and not timeout.expired:
+    >>>         time.sleep(min(timeout.remaining, 1.0))
+    >>>         stdout, stderr = driver.read_from_background_process(process, timeout=timeout.remaining)
```
jremmet: Here we don't want to block for the remaining time
```suggestion
    >>>         stdout, stderr = driver.read_from_background_process(process)
```

Bastian-Krause:
Thanks for your input!

An internal discussion lead to the following:
- the _BackgroundProcessProtocol_ will be merged into the _CommandProtocol_, all mentions of "background" will be removed
- instead of having read/poll/stop methods, separate _Popen_-like (per-driver) process handle objects with these methods (as well as `expect()`) will be introduced (defined by an abstract process handle in `labgrid.protocol.commandprotocol`)
- `read(timeout=)` will be `read(wait=)` to minimize confusion of fatal timeouts and non-fatal waits
- the read method will try to match everything but the marker, then it will match the marker (as opposed to matching the marker, reading "before" and then `ShellDriver.console._expect.read(size=len(before))`)
- the read method will return bytes, no VT100 substitution, no decoding, no newline translation (we cannot be sure that we read complete sequences, so leave this to the user)
- `CommandProtocol.run()` will perform these clean ups as before
- when instantiating process handle the driver (to allow calls to `ShellDriver.expect()` and access of other driver instance variables) and a timeout will be passed to it
- the process handle will be usable as a context manager
- if available [coreutils' timeout](https://www.gnu.org/software/coreutils/timeout) will be used to make things easier
- the SSHDriver will use pexpect.spawn() instead of subprocess.Popen()

jremmet:
> the read method will try to match everything but the marker, then it will match the marker (as opposed to matching the marker, reading "before" and then ShellDriver.console._expect.read(size=len(before)))

I tried to work on this issue. But due to the [non greedy](https://pexpect.readthedocs.io/en/stable/overview.html?highlight=non%20greedy#beware-of-and-at-the-end-of-patterns) behaviour in expect and that a negative lookahead will block till the marker is read, I run into issues. Especially if read with wait=0.
My solution still uses read and also check if "before" contains the start of a marker.
[WIP/shelldriver_popen](https://github.com/jremmet/labgrid/tree/WIP/shelldriver_popen)

dnltz:
Hey @Bastian-Krause, is this draft still planned or dropped? We see a need for this and would like to remove our workaround for this.

Bastian-Krause:
@dnltz I worked for a customer on this. Unfortunately the concrete use case vanished into thin air, so the work stalled here. If you are interested to work on this (see current state of discussions above), please do so. If you want us to work on this, please contact our sales team.

Bastian-Krause:
Superseded by #835.

PR #556, opened 2020-01-07
driver: networkusbstoragedriver: fix ExecutionErrors for unavailable storage

Bastian-Krause:
**Description**
NetworkUSBStorageDriver does not have a `storage_path` member. Instead `storage.path` is meant here. Fix that.

**Checklist**
- [ ] Tests for the feature 
- [x] PR has been tested

Fixes #363

### DIFF
--- a/labgrid/driver/networkusbstoragedriver.py
+++ b/labgrid/driver/networkusbstoragedriver.py
@@ -52,7 +52,7 @@ class NetworkUSBStorageDriver(Driver):
     def write_image(self, filename=None, mode=Mode.DD):
         if not self.storage.path:
             raise ExecutionError(
-                "{} is not available".format(self.storage_path)
+                "{} is not available".format(self.storage.path)
             )
         if filename is None and self.image is not None:
             filename = self.target.env.config.get_image_path(self.image)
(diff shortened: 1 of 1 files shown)


Bastian-Krause:
Don't merge yet, the if condition does not make sense yet.

labgrid/driver/networkusbstoragedriver.py
```diff
@@ -52,7 +52,7 @@ def on_deactivate(self):
     def write_image(self, filename=None, mode=Mode.DD):
         if not self.storage.path:
```
Bastian-Krause: @Emantor Is this correct? If so, we cannot use the non-existent path in the Exception message below. Or should this be `self.storage.avail`?

labgrid/driver/networkusbstoragedriver.py
```diff
@@ -101,7 +101,7 @@ def write_image(self, filename=None, mode=Mode.DD):
     def get_size(self):
         if not self.storage.path:
```
Bastian-Krause: Same as above.

labgrid/driver/networkusbstoragedriver.py
Emantor: It is correct. This will print `None is not available` in the error case. I think its best to remove the variable from the error and just return a generic `device not available`

Bastian-Krause:
Replaced by #573.

PR #557, opened 2020-01-07
contrib/systemd: add systemd services

sephalon:
**Description**

Adds systemd services for coordinator and exporter.

While the coordinator can be run under a dynamic user, the exporter needs a dedicated system group: Users, that are a member of the `labgrid` group are then able to store `ManagedFile`s in `/var/cache/labgrid`. Therefore also adds a sysusers.d file that creates the `labgrid` user and group. Further hardening (e.g. dropping privileges) of the services has been omitted; however, it is questionable whether this is needed in a lab environment anyway.

One can set the coordinator URL the exporter should connect to by overriding the exporter service file; i.e. one would execute `systemctl edit labgrid-exporter.service` and add the following snippet:

    [Service]
    Environment="LG_CROSSBAR=ws://<your-host>:<your-port>/ws"

Tested on Ubuntu 18.04 LTS.

Documentation is yet to be written. The paths in the service files conform to what you would see in packaged installation (I'm working on integrating these changes into the Debian package right now); we could also use pseudo paths (and patch them during packaging) if you prefer that.

**Checklist**
- [ ] Tests for the feature 
- [x] PR has been tested

### DIFF
--- a/contrib/systemd/labgrid-coordinator.service
+++ b/contrib/systemd/labgrid-coordinator.service
@@ -1,0 +1,13 @@
+[Unit]
+Description=Labgrid Coordinator
+After=network.target
+
+[Service]
+ExecStart=/path/to/labgrid-coordinator/venv/bin/crossbar start --logformat=syslogd --cbdir /var/lib/labgrid-coordinator --config /etc/labgrid/coordinator.yaml
+ExecStop=/usr/bin/labgrid-coordinator stop --cbdir /var/lib/labgrid-coordinator
+Restart=on-abort
+DynamicUser=yes
+StateDirectory=labgrid-coordinator
+
+[Install]
+WantedBy=multi-user.target
--- a/contrib/systemd/labgrid-exporter.service
+++ b/contrib/systemd/labgrid-exporter.service
@@ -1,0 +1,16 @@
+[Unit]
+Description=Labgrid Exporter
+After=network.target
+
+[Service]
+ExecStart=/path/to/labgrid/venv/bin/labgrid-exporter /etc/labgrid/exporter.yaml
+Restart=on-abort
+User=labgrid
+Group=labgrid
+# Adjust to your distribution (most often "dialout" or "tty")
+SupplementaryGroups=dialout
+CacheDirectory=labgrid
+CacheDirectoryMode=1775
+
+[Install]
+WantedBy=multi-user.target
--- a/contrib/systemd/tmpfiles.d/labgrid.conf
+++ b/contrib/systemd/tmpfiles.d/labgrid.conf
@@ -3,4 +3,4 @@
 # it is advised to at least change the group from the default for your
 # environment
 # Path			Mode	UID	GID	Age Argument
-d /var/cache/labgrid	1775	root	users	2d
+d /var/cache/labgrid	1775	labgrid	labgrid	2d
--- a/doc/getting_started.rst
+++ b/doc/getting_started.rst
@@ -176,6 +176,8 @@ same machine. The client is used to access functionality provided by an
 exporter. Over the course of this tutorial we will set up a coordinator and
 exporter, and learn how to access the exporter via the client.
 
+.. _remote-getting-started-coordinator:
+
 Coordinator
 ~~~~~~~~~~~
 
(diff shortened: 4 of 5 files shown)


Emantor:
I am in favor of merging the unit files, however I am againt changing the current `tmpdiles.d` example. Its an example and each distribution needs to adjust it for itself, when adding the necessary packaging.

contrib/systemd/labgrid-exporter.service
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description=Labgrid Exporter
+After=network.target
+
+[Service]
+ExecStart=/usr/bin/labgrid-exporter /etc/labgrid/exporter.yaml
+Restart=on-abort
+User=labgrid
+Group=labgrid
+SupplementaryGroups=dialout
```
jluebbe: The group for which serial ports are available depends on the distribution.

contrib/systemd/labgrid-coordinator.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Labgrid Coordinator
+After=network.target
+
+[Service]
+ExecStart=/usr/bin/labgrid-coordinator start --logformat=syslogd --cbdir /var/lib/labgrid-coordinator --config /etc/labgrid/coordinator.yaml
```
jluebbe: There is no standard `labgrid-coordinator` wrapper.

We use:
```
[Unit]
Description=labgrid coordinator

[Service]
User=labgrid
WorkingDirectory=/srv/labgrid/coordinator/workdir
Environment="PYTHONUNBUFFERED=1"
ExecStart=/usr/ptx-venvs/labgrid-staging/bin/crossbar start --loglevel=debug --logformat=syslogd
Restart=always

[Install]
WantedBy=multi-user.target
```

Using the stop action is better, though.

jluebbe: The PYTHONUNBUFFERED is useful to have immediate output in the journal.

contrib/systemd/labgrid-exporter.service
sephalon: Correct, I added a comment that this needs to be adjusted to the distribution. Is it ok?

contrib/systemd/labgrid-coordinator.service
sephalon: > There is no standard labgrid-coordinator wrapper.

Right, this only existed in my Debian package. I have added one now in `setup.py`.

> The PYTHONUNBUFFERED is useful to have immediate output in the journal.

Ok, however I would leave enabling this to the user (via a override file, see above) in case one needs it for debugging.

sephalon:
> I am in favor of merging the unit files, however I am against changing the current tmpdiles.d example. Its an example and each distribution needs to adjust it for itself, when adding the necessary packaging.

Did you see that I also added a sysusers.d file that will automatically create a `labgrid` user and group? Thanks to systemd [1], all this can be done now in a fairly distribution-independent way. I think we should try to establish good default settings upstream; changes can still be made by the respective package authors.

[1] "Thanks" might be a bit controversial, I know ;-)

contrib/systemd/tmpfiles.d/labgrid.conf
```diff
@@ -1,6 +1,3 @@
 # Labgrid saves files uploaded to the exporter in /var/cache
-# This configuration file creates the directory and is meant as a starting point
-# it is advised to at least change the group from the default for your
-# environment
```
Emantor: Please don't remove this part. This can also be incorrect, since the users of the exporter now need to be part of the labgrid group, where an existing environment may reuse existing group memberships.

sephalon: Ok, re-added.

sephalon:
I have added some documentation on how to use the systemd files; please review as well.

sephalon:
Any more comments on this? I would like to continue with the Debian package for which this PR is a prerequisite for.

Emantor:
This needs feedback from @jluebbe.

sephalon:
@jluebbe Would you please review this?

setup.py
```diff
@@ -63,6 +63,7 @@
         ],
         'console_scripts': [
             'labgrid-client = labgrid.remote.client:main',
+            'labgrid-coordinator = crossbar:run',
```
Emantor, line 66: Please remove this, not every labgrid installation will have crossbar available, since this is only required on the coordinator. If crossbar isn't available this falsely implies that the coordinator can be used.

doc/getting_started.rst
```diff
@@ -327,6 +327,73 @@ See :ref:`remote-usage` for some more advanced features.
...
+
+   .. code-block:: console
+
+      # systemctl daemon-reload
+
+#. Start the coordinator:
+
+   .. code-block:: console
+
+      # systemctl start labgrid-coordinator
```
Emantor, line 380: Likewise, please remove these instructions. Crossbar should be installed and used by creating a separate virtualenv with crossbar in it and without the labgrid wrapper.

contrib/systemd/labgrid-coordinator.service
Emantor: This line should have an example on howto use crossbar from a separate virtualenv.

Emantor (changes requested):

setup.py
sephalon: Done, removed.

doc/getting_started.rst
sephalon: I marked this step as optional and tried to make it more clear that the coordinator should be in a separate virtual environment from the exporter (and client).

Emantor (approved):

Emantor:
please squash the fixup, that should also fix the DCO error.

sephalon:
Done, squashed and rebased.

jluebbe (approved):

Emantor (approved):

PR #558, opened 2020-01-10
driver/sshdriver: add keepalive

Emantor:
**Description**
Add a keepalive function to the SSHDriver, similar to the utils ssh
implementation.

**Checklist**
- [ ] Tests for the feature 
- [x] PR has been tested

### DIFF
--- a/labgrid/driver/sshdriver.py
+++ b/labgrid/driver/sshdriver.py
@@ -28,6 +28,7 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
     def __attrs_post_init__(self):
         super().__attrs_post_init__()
         self.logger = logging.getLogger("{}({})".format(self, self.target))
+        self._keepalive = None
 
     def on_activate(self):
         self.ssh_prefix = "-o LogLevel=ERROR"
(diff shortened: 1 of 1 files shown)


labgrid/driver/sshdriver.py
```diff
@@ -44,8 +44,11 @@ def on_activate(self):
         self.ssh_prefix += " -o ControlPath={}".format(
             self.control
         ) if self.control else ""
+        self._keepalive = None
```
jluebbe: This needs to be set in `__attrs_post_init__` as well.

labgrid/driver/sshdriver.py
```diff
@@ -118,6 +121,7 @@ def _run(self, cmd, codec="utf-8", decodeerrors="strict", timeout=None): # pylin
         returns:
         (stdout, stderr, returncode)
         """
+        assert self._check_keepalive()
```
jluebbe: This should raise an ExecutionError as it can be triggered even by valid tests.

labgrid/driver/sshdriver.py
```diff
@@ -44,8 +44,11 @@ def on_activate(self):
         self.ssh_prefix += " -o ControlPath={}".format(
             self.control
         ) if self.control else ""
+        self._keepalive = None
+        self._start_keepalive();
 
     def on_deactivate(self):
+        self._stop_keepalive()
```
jluebbe: This needs a try/finally, so the master is always cleaned up.

Emantor:
Tested this manually by activating the driver and waiting for a sufficiently long time, the driver was still responsive and could run commands.

PR #560, opened 2020-01-14
exporter broken tracking

jluebbe:
**Description**
When the exporter is unable to start or stop subprocesses for resources, it's better to handle that explicitly rather than wait for errors when using them.

**Checklist**
- [x] Tests for the feature 
- [x] PR has been tested

### DIFF
--- a/labgrid/remote/exporter.py
+++ b/labgrid/remote/exporter.py
@@ -27,6 +27,14 @@ except pkg_resources.DistributionNotFound:
 exports = {}
 reexec = False
 
+class ExporterError(Exception):
+    pass
+
+
+class BrokenResourceError(ExporterError):
+    pass
+
+
 @attr.s(eq=False)
 class ResourceExport(ResourceEntry):
     """Represents a local resource exported via a specific protocol.
@@ -48,6 +56,26 @@ class ResourceExport(ResourceEntry):
         for key in self.local_params:
             del self.params[key]
         self.start_params = None
+        self._broken = None
+
+    # if something criticial failed for an export, we can mark it as
+    # permanently broken
+    @property
+    def broken(self):
+        return self._broken
+
+    @broken.setter
+    def broken(self, reason):
+        assert self._broken is None
+        assert type(reason) == str
+        assert reason
+        self._broken = reason
+        # By setting the acquired field, we block places from using this
+        # resource. For now, when trying to acquire a place with a match for
+        # this resource, we get 'resource is already in used by <broken>',
+        # instead of an unspecific error.
+        self.data['acquired'] = '<broken>'
+        self.logger.error("marked as broken: %s", reason)
 
     def _get_start_params(self):  # pylint: disable=no-self-use
         return {}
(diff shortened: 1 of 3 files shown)


labgrid/remote/exporter.py
```diff
@@ -48,6 +67,22 @@ def __attrs_post_init__(self):
         for key in self.local_params:
             del self.params[key]
         self.start_params = None
+        self._broken = None
+
+    # is something criticial failed for an export, we can mark it as
```
Bastian-Krause:
```suggestion
    # if something criticial failed for an export, we can mark it as
```

labgrid/remote/exporter.py
```diff
@@ -48,6 +67,22 @@ def __attrs_post_init__(self):
...
+    # permanently broken
+    @property
+    def broken(self):
+        return self._broken
+
+    @broken.setter
+    def broken(self, reason):
+        assert self._broken is None
+        assert type(reason) == str
+        assert len(reason) > 0
```
Bastian-Krause:
```suggestion
        assert reason
```
Should be enough at this point.

labgrid/remote/exporter.py
```diff
@@ -48,6 +67,22 @@ def __attrs_post_init__(self):
...
+        self._broken = reason
+        self.data['acquired'] = '<broken>'
```
Bastian-Krause: I guess this needs a little documentation somewhere.

labgrid/remote/exporter.py
```diff
@@ -90,38 +139,48 @@ def poll(self):
...
+            params['extra']['broken'] = self.broken
         if self.params != params:
             self.data['params'].update(params)  # pylint: disable=unsubscriptable-object
             dirty = True
 
         return dirty
 
     def acquire(self, *args, **kwargs):  # pylint: disable=arguments-differ
+        if self.broken:
+            raise BrokenResourceError("cannot acquire broken resource (original reason): %s", self.broken)
```
Bastian-Krause: Why do you use the % style formatting here?

labgrid/remote/exporter.py
```diff
@@ -90,38 +139,48 @@ def poll(self):
...
         super().acquire(*args, **kwargs)
         self.poll()
 
     def release(self, *args, **kwargs):  # pylint: disable=arguments-differ
+        if self.broken:
+            raise BrokenResourceError("cannot release broken resource (original reason): %s", self.broken)
```
Bastian-Krause: And here?

tests/conftest.py
```diff
@@ -117,7 +117,11 @@ def exporter(tmpdir, crossbar):
         """
     Testport:
         NetworkSerialPort:
-          {host: 'localhost', port: 4000}
+          host: 'localhost'
+          port: 4000
+    Broken:
+        RawSerialPort:
+          port: 'none'
```
Bastian-Krause: Just to make sure I understand this: This lets `assert start_params['path'].startswith('/dev/')` raise an AssertionError. This is then caught and `broken` is set to 'start failed'. Correct?

jluebbe: Yes, this was the easiest way to support a test for this code.

labgrid/remote/exporter.py
jluebbe: I've added a comment here.

labgrid/remote/exporter.py
```diff
@@ -35,6 +35,17 @@ class BrokenResourceError(ExporterError):
...
+def log_subprocess_kernel_stack(logger, child):
+    if child.poll() is not None:  # nothing to check if no longer running
+        return
+    try:
+        with open('/proc/{}/stack'.format(child.pid), 'r') as f:
+            stack = f.read()
+            stack = stack.strip()
+    except PermissionError:
+        return
+    logger.info("current kernel stack of {} is:\n".format(child.args)+stack)
```
Bastian-Krause: Nit: % style formatting should be used in logging functions.

jluebbe: Yes, but this is very far from performance critical code. ;)

Bastian-Krause: Agreed, but this will be another pylint warning.

labgrid/remote/exporter.py
```diff
@@ -27,6 +27,25 @@
...
+def log_subprocess_kernel_stack(logger, child):
+    if child.poll() is not None:  # nothing to check if no longer running
+        return
+    try:
+        with open('/proc/{}/stack'.format(child.pid), 'r') as f:
+            stack = f.read()
+            stack = stack.strip()
+    except PermissionError:
+        return
+    logger.info("current kernel stack of %s is:\n%s" % (child.args, stack))
```
Bastian-Krause:
```suggestion
    logger.info("current kernel stack of %s is:\n%s", child.args, stack)
```

jluebbe: Thanks, updated again. :man_facepalming:

PR #564, opened 2020-01-27
Various smaller fixes and cleanups

Emantor:
**Description**
Add some crossbar documentation, fix SSH timeout handling and remove `hawkbit` and `usbstick`.

**Checklist**
- [x] PR has been tested


CHANGES.rst
```diff
@@ -32,6 +32,7 @@ New Features in 0.3.0
 - The binding dictionary can now supports type name strings in addition to the
   types themselves, avoiding the need to import a specific protocol or driver
   in some cases.
+- The ``HawkbitTestClient`` and ``USBStick`` classes have been removed
```
Bastian-Krause: This is a breaking change, right? Better put in the corresponding section.

Emantor:
I would not introduce this in this PR, since you have sent one which does the mentioned change. Are you otherwise fine with this PR @Bastian-Krause?

Bastian-Krause:
Oh right, I did that in #574. That means I can replace 
`def _run(self, command: str, *, timeout: int = None):`
with
`def _run(self, command: str, *, timeout: int = 30):`
in that PR, right?

Emantor:
Yes.

> Oh right, I did that in #574. That means I can replace
> `def _run(self, command: str, *, timeout: int = None):`
> with
> `def _run(self, command: str, *, timeout: int = 30):`
> in that PR, right?

Yes.

Bastian-Krause:
Done.

PR #566, opened 2020-01-27
crossbar: move labgrid specific files to general workdir

krevsbech:
**Description**
Change the crossbar workdir configuration, for labgrid, from being
the parent dir to the actual workdir of crossbar. This implies
that all files related to crossbar are in the same dir.

- [x] PR has been tested

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -45,6 +45,10 @@ Breaking changes in 0.3.0
   setup the cleanup job.
 - ``@attr.s(cmp=False)`` is deprecated and all classes have been moved to
   ``@attr.s(eq=False)``, this release requires attrs version 19.2.0
+- Coordinator work dir is now set to the same dir as the crossbar configuration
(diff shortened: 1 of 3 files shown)


CHANGES.rst
```diff
@@ -45,6 +45,8 @@ Breaking changes in 0.3.0
   setup the cleanup job.
 - ``@attr.s(cmp=False)`` is deprecated and all classes have been moved to
   ``@attr.s(eq=False)``, this release requires attrs version 19.2.0
+- Coordinator now saves the labgrid specific files like ``places.yaml`` and
+  ``resources.yaml`` in its working-dir, Previously it would use ``..``.
```
jluebbe: This is not correct. The coordinator always saves these files in its working dir. The change is that the crossbar sets the coordinator working dir to the config directory, instead of one level up.

krevsbech: Point taken. Thanks.

PR #567, opened 2020-01-30
Improve process output logging, adjust client log levels

Bastian-Krause:
**Description**
#469 introduced ProcessWrapper, acting "like `subprocess.check_output`, but takes callbacks to send the output of the command to". This PR aims to improve the user experience of process output:

Output is now split on "\r" instead of "\r\n". This allows processing output of commands rewriting lines. It is currently used for "labgrid-client write-image" which again uses "dd status=progress".

Command output is in many cases not recognizable by the user without knowing the actual command. This gets even worse when multiple processes emit output. This is addressed by printing the command as well as its PID when started and prepend the PID to every log line of output.

To make use of this ProcessWrapper's logging callback is enabled for labgrid-client. Command output that is understandable without further context can now be printed with `ProcessWrapper.check_output(command, print_on_silent_log=True)`. This is enabled examplarily for NetworkUSBStorageDriver.

**Checklist**
- [ ] Tests for the feature 
- [x] PR has been tested

### DIFF
--- a/labgrid/util/helper.py
+++ b/labgrid/util/helper.py
@@ -60,7 +60,7 @@ class ProcessWrapper:
                 res.extend(parts)
                 for part in parts:
                     for callback in self.callbacks:
-                        callback(part)
+                        callback(part, process)
             process.poll()
             if process.returncode is not None:
                 break
@@ -73,7 +73,7 @@ class ProcessWrapper:
             if buf[-1] != b'\n':
                 buf += b'\n'
             for callback in self.callbacks:
-                callback(buf)
+                callback(buf, process)
         if process.returncode != 0:
             raise subprocess.CalledProcessError(process.returncode,
                                                 command,
@@ -93,7 +93,7 @@ class ProcessWrapper:
         self.callbacks.remove(callback)
 
     @staticmethod
-    def log_callback(message):
+    def log_callback(message, process):
         """Logs process output message along with its pid."""
         import logging
         logger = logging.getLogger("Process")
@@ -102,7 +102,7 @@ class ProcessWrapper:
             logger.info(message)
 
     @staticmethod
-    def print_callback(message):
+    def print_callback(message, _):
         """Prints process output message."""
         message = message.decode(encoding="utf-8", errors="replace")
         print("\r{}".format(message), end='')
--- a/labgrid/util/helper.py
+++ b/labgrid/util/helper.py
@@ -107,9 +107,17 @@ class ProcessWrapper:
         Loglevel is logging.INFO."""
         self.register(ProcessWrapper.log_callback)
 
+    def disable_logging(self):
+        """Disables process output logging."""
+        self.unregister(ProcessWrapper.log_callback)
+
     def enable_print(self):
         """Enables process output to print."""
         self.register(ProcessWrapper.print_callback)
+
+    def disable_print(self):
+        """Disables process output printing."""
+        self.unregister(ProcessWrapper.print_callback)
 
 
 processwrapper = ProcessWrapper()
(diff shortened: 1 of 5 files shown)


labgrid/util/helper.py
```diff
@@ -89,20 +89,26 @@ def unregister(self, callback):
         assert callback in self.callbacks
         self.callbacks.remove(callback)
 
+    @staticmethod
+    def log_callback(message, process):
```
Emantor: This already introduces the process argument, should be moved to the third commit.

Bastian-Krause: You're absolutely right, I must've messed up the commits here.

Bastian-Krause: Rebased the PR and fixed this.

labgrid/util/helper.py
```diff
@@ -94,12 +94,15 @@ def log_callback(message):
...
+        message = message.decode(encoding="utf-8", errors="replace").strip("\n")
+        if message:
+            logger.info(message)
 
     @staticmethod
     def print_callback(message):
         """Prints process output message."""
-        print(message.decode(encoding="utf-8", errors="replace"))
+        message = message.decode(encoding="utf-8", errors="replace")
+        print("\r{}".format(message), end='')
```
jluebbe: Won't this hide other output with doesn't end in a `\n`?

labgrid/driver/networkusbstoragedriver.py
```diff
@@ -93,7 +93,8 @@ def write_image(self, filename=None, mode=Mode.DD, partition=None, skip=0, seek=
                 "bs={}".format(block_size),
                 "skip={}".format(skip),
                 "seek={}".format(seek),
-                "conv=fdatasync"
+                "conv=fdatasync",
+                "oflag=dsync"
```
jluebbe: What's the performance impact when combined with the bs=512 change that is already merged?

labgrid/util/helper.py
```diff
@@ -106,9 +106,17 @@ def enable_logging(self):
         Loglevel is logging.INFO."""
         self.register(ProcessWrapper.log_callback)
 
+    def disable_logging(self):
+        """Disables process output logging."""
+        self.unregister(ProcessWrapper.log_callback)
```
jluebbe: We may need an assert here to detect unbalanced enable/disable?

labgrid/util/helper.py
Bastian-Krause: Correct, I've pushed a fixup that appends `\n` to `buf` in case it's not already there.

labgrid/driver/networkusbstoragedriver.py
Bastian-Krause: It's pretty bad, dd'ing 380 MB takes 3741s (with `oflag=dsync`) vs. 118.966s (without). Should we drop `oflag=dsync` `if skip or seek`? I personally use skip and seek for bootloader replacement only (while keeping the MBR).

labgrid/util/helper.py
Bastian-Krause: The register() and unregister() methods already have these asserts.

Bastian-Krause:
@jluebbe Could you please comment on the open points above?

Bastian-Krause:
Ping?

labgrid/driver/networkusbstoragedriver.py
Bastian-Krause: I'll drop the oflag=dsync patch.

Bastian-Krause:
Force-pushed:
- rebased on current master
- dropped the dd oflag=dsync commit

Emantor (approved):

PR #572, opened 2020-02-03
target: improve exception messages

Bastian-Krause:
**Description**
When getting named resources and drivers fails because the requested name is unavailable the error messages do not say which name was not found. Add the name in these cases. While at it use named args for extensive format strings and remove redundancies in exception messages.

**Checklist**
- [ ] PR has been tested (parts of it)


Bastian-Krause:
Force-push: fixed tests that depended on the previous exception messages.

PR #573, opened 2020-02-03
driver: networkusbstorage: make decorator check that resources are bound

Bastian-Krause:
**Description**
Instead of explicitly checking if self.storage.path is set use the `@Driver.check_active` decorator. In order to make this work for USBMassStorage resources overwrite its avail attribute.

**Checklist**
- [x] PR has been tested

Fixes #363
Replaces #556

### DIFF
--- a/labgrid/driver/networkusbstoragedriver.py
+++ b/labgrid/driver/networkusbstoragedriver.py
@@ -46,12 +46,9 @@ class NetworkUSBStorageDriver(Driver):
     def on_deactivate(self):
         pass
 
+    @Driver.check_active
     @step(args=['filename'])
     def write_image(self, filename=None, mode=Mode.DD):
-        if not self.storage.path:
-            raise ExecutionError(
-                "{} is not available".format(self.storage_path)
-            )
         if filename is None and self.image is not None:
             filename = self.target.env.config.get_image_path(self.image)
         assert filename, "write_image requires a filename"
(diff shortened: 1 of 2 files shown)


Bastian-Krause:
Note: we discussed this internally. While @Emantor's argument is correct (there is no guarantee that resources used by the driver are available when a driver method is called, it is guaranteed only right after driver activation) this applies to all drivers and resources. Deactivating drivers as a result of its resources being unavailable would be nice to have. This is not what this PR is about though.

jluebbe:
> Note: we discussed this internally. While @Emantor's argument is correct (there is no guarantee that resources used by the driver are available when a driver method is called, it is guaranteed only right after driver activation) this applies to all drivers and resources. Deactivating drivers as a result of its resources being unavailable would be nice to have. This is not what this PR is about though.

As soon as the target's update_resources() method is called, it will deactivate drivers which are bound to now unavailable resources.

Bastian-Krause:
Quick note: before this PR was merged it was possible to use the `NetworkUSBStorageDriver`'s `write_image()` and `get_size()` methods without activating the driver. This was never intended, always activate a driver before using it.

With this PR being merged an exception is raised:
```
labgrid.binding.StateError: NetworkUSBStorageDriver.write_image can not be called (NetworkUSBStorageDriver(target=Target(name='mytarget', env=Environment(config_file='config.yaml')), name=None, state=<BindingState.bound: 1>, image=None) is in state bound)
```

PR #574, opened 2020-02-05
Update protocols, make drivers follow protocols

Bastian-Krause:
**Description**
Make method args of `CommandProtocol`, `FileTransferProtocol` consistent/reflect driver reality. Now let drivers implementing protocols follow them strictly.

This PR includes converting `run()`'s/`run_check()`'s `codec`/`decodeerrors` kwargs to driver attributes.

**Checklist**
- [ ] PR has been tested (parts of it)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -56,6 +56,9 @@ Breaking changes in 0.3.0
   renamed to `USBStorageDriver`.
   A deprecated `NetworkUSBStorageDriver` exists temporarily for compatibility
   reasons.
+- ``codec`` and ``decodeerrors`` are no longer ``run()``/``run_check()``
+  kwargs, but driver (``codec``) and `CommandMixin` (``decodeerrors``)
+  attributes.
 
 Release 0.2.0 (released Jan 4, 2019)
 ------------------------------------
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -817,6 +817,7 @@ Arguments:
     to pass before sending a newline to device.
   - console_ready (regex): optional, pattern used by the kernel to inform
     the user that a console can be activated by pressing enter.
+  - codec (str, default="utf-8"): optional, encoding used to decode output bytes
 
 .. _conf-sshdriver:
 
@@ -849,6 +850,7 @@ Arguments:
     (only used if password is not set)
   - stderr_merge (bool): set to True to make `run()` return stderr merged with
       stdout, and an empty list as second element.
+  - codec (str, default="utf-8"): optional, encoding used to decode output bytes
 
 UBootDriver
 ~~~~~~~~~~~
@@ -880,6 +882,7 @@ Arguments:
   - bootstring (str): optional, regex to match on Linux Kernel boot
   - login_timeout (int): optional, timeout for login prompt detection in
     seconds (default 60)
+  - codec (str, default="utf-8"): optional, encoding used to decode output bytes
 
 SmallUBootDriver
 ~~~~~~~~~~~~~~~~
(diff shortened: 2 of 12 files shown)


Bastian-Krause:
Force-push:
- mentioned `codec` and `decodeerrors` in CHANGES.rst's "Breaking" section
- make `SSHDriver.run()` pass timeout argument, set default timeout to 30 as specified by CommandProtocol

labgrid/driver/sshdriver.py
```diff
@@ -103,24 +105,24 @@ def _check_master(self):
         return self._start_own_master()
 
     @Driver.check_active
-    @step(args=['cmd'], result=True)
-    def run(self, cmd, codec="utf-8", decodeerrors="strict", timeout=None): # pylint: disable=unused-argument
-        return self._run(cmd, codec=codec, decodeerrors=decodeerrors)
+    @step(args=['command'], result=True)
+    def run(self, command: str, *, timeout: int = 30):
```
jluebbe: Why do we want to have a default timeout for the SSHDriver? Compared to the serial based drivers, SSH easily detects a crashed target, so introducing such a short timeout might be surprising.

Bastian-Krause: I thought it would be a good thing to have consistent timeouts over all drivers implementing the `CommandProtocol`. I understand now that timeouts are a way to detect a crashed target in case of serial communication. In that case I will remove the default timeout from the `CommandProtocol` and set it to None for the `SSHDriver`, alright?

Bastian-Krause:
Force-push:
- rebased on current master
- remove default timeout of `CommandProtocol` `run()` and `run_check()`
- set default timeout of `SSHDriver` `run()` and `run_check()` to `None`
- moved `decodeerrors` to `CommandMixin`

Note: the new process handling API being discussed in #552 will return bytes instead of strings. There is no need to decode for (sub)string search or regex matching when this is used.

labgrid/driver/sshdriver.py
Bastian-Krause: Implemented as suggested, see below.

Bastian-Krause:
Force-push: rebased on master

Emantor:
Rebased on current master to fix merge conflict in `sshdriver.py`.

Emantor:
@Bastian-Krause please recheck the rebase was done correctly.

Bastian-Krause:
@Emantor Rebase looks good.

Rebased it once again on current master.

Bastian-Krause:
Ping?

CHANGES.rst
```diff
@@ -56,6 +56,9 @@ Breaking changes in 0.3.0
   renamed to `USBStorageDriver`.
   A deprecated `NetworkUSBStorageDriver` exists temporarily for compatibility
   reasons.
+- ``codec`` and ``decodeerrors`` are no longer ``run()``/``run_check()``
+  kwargs, but driver (``codec``) and `CommandMixin` (``decodeerrors``)
+  attributes.
```
jluebbe, line 61: Why is codec a driver attribute and decodeerrors on CommandMixin?

This should also describe how a user should update their code and configuration. Is the encoding changeable at runtime?

doc/configuration.rst
```diff
@@ -817,6 +817,7 @@ Arguments:
     to pass before sending a newline to device.
   - console_ready (regex): optional, pattern used by the kernel to inform
     the user that a console can be activated by pressing enter.
+  - codec (str, default="utf-8"): optional, encoding used to decode output bytes
```
jluebbe, line 820: Is "output" the bytes sent or received by labgrid? Shouldn't the encoding apply to both directions?

labgrid/driver/commandmixin.py
```diff
@@ -13,6 +13,7 @@ class CommandMixin:
 
     def __attrs_post_init__(self):
         super().__attrs_post_init__()
+        self.decodeerrors = 'strict'
```
jluebbe, line 16: This is not an attrs attribute, so it can't be specified during construction. What is the reason for the difference?

PR #576, opened 2020-02-10
Getting started cleanup

TheMeaningfulEngineer:
**This is still a Draft PR. Will polish the history before the final PR is done. Currently hoping to get higher level feedback**

**Description**

An iteration of the `getting started` page. The goal is to provide a step by step introduction to labgrid concepts through copy pastable examples.

Only makes changes to the documentation.

The documentation can be built successfully from this branch.

**Currently the main RFC is:**

* Why separate Resources from Drivers in a seemingly arbitrary way?
* Resources don't work well with the analogy of "global variables" due to their coupling to certain drivers. What would be a mental model to describe a Resource then. 
* Why use Resources vs "Driver properties" (entries living under the driver in the yaml)?  

**Checklist**
- [ ] Tests for the feature 

- [ ] PR has been tested


doc/getting_started.rst
```diff
@@ -14,155 +14,186 @@ these usually are:
...
-          $ pip uninstall pyserial
-          $ pip install https://github.com/labgrid-project/pyserial/archive/v3.4.0.1.zip#egg=pyserial
-
-   This pyserial version has two fixes for an Issue we found with Serial over IP
-   multiplexers. Additionally it reduces the Serial over IP traffic considerably
-   since the port is not reconfigured when labgrid changes the timeout (which is
-   done inside the library a lot).
+    git clone https://github.com/labgrid-project/labgrid
+    cd labgrid && pip install -r requirements.txt
+    python3 setup.py install
```
leiflm, line 34: pip actually supports installing from a git repository, right away:
```sh
pip install git+https://github.com/labgrid-project/labgrid/@master#egg=labgrid
```
However, I think that `master` should likely replaced with Sphinx' [`release`](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-release) variable. Which, in turn, should be set to the git ref (`v0.2.0`, `master`, ...) .

jluebbe: @leiflm I thought it wouldn't use the requirements.txt from the repository in that case (which is the main reason we currently suggest to do it this way).

leiflm: > I thougt it wouldn't use the `requirements.txt` from the repository in that case.

@jluebbe you are, most likely, correct. Why wouldn't you change the requirements in `setup.py` accordingly, though?

Anyway, pip also supports to specify the `requirements.txt` using a URI:
```sh
pip install -r https://raw.githubusercontent.com/labgrid-project/labgrid/master/requirements.txt
```

Emantor: The Python packaging User Guide does not recommend pinning dependencies in `setup.py`, see [here](https://packaging.python.org/discussions/install-requires-vs-requirements/).

leiflm: Thanks for the insight on why `setup.py` is not adjusted. I didn't see beforehand, that the current `requirements.txt` actually pinned all versions of the required packages. Reading the docs, I assumed it would just be a bit more restrictive about package versions used (i.e. `A>=1,<2`).

I assume that nobody has a setup at hand to verify which packages actually need to be kept pinned down? I guess that would be an issue of its own :)

Emantor:
Closing this, the documentation has been refactored since this was initially opened.

PR #579, opened 2020-02-18
binding: improve StateError message

Bastian-Krause:
**Description**
Currently the error message does not tell a user (without detailed knowledge of driver binding states) why the method called cannot be called at this time:

```
Driver.method can not be called (Driver(target=Target(name='name', env=Environment(config_file='conf.yaml')), name=None, state=<BindingState.bound: 1>) is in state bound)
```
The exception is raised only if the state is not active, so make the error message say this:
```
Driver(target=Target(name='name', env=Environment(config_file='x86-esr.yaml')), name=None, state=<BindingState.bound: 1>) has not been activated, Driver.method cannot be called in state "bound"
```

**Checklist**
- [x] PR has been tested

### DIFF
--- a/labgrid/binding.py
+++ b/labgrid/binding.py
@@ -94,8 +94,8 @@ class BindingMixin:
         def wrapper(self, *_args, **_kwargs):
             if self.state is not BindingState.active:
                 raise StateError(
-                    "{} can not be called ({} is in state {})".format(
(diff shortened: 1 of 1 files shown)


Emantor:
I had the same thought when you posted into the issue.

PR #580, opened 2020-02-19
remote/client: add name argument for power

strumtrar:
**Description**

Add an optional name argument for the power command to support
places with multiple power drivers.

One or multiple names can be supplied.

Tested locally on a place with two NetworkPowerPorts and with labgrid-pytest.

Signed-off-by: Steffen Trumtrar <s.trumtrar@pengutronix.de>

### DIFF
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -691,42 +691,48 @@ class ClientSession(ApplicationSession):
     def power(self):
         place = self.get_acquired_place()
         action = self.args.action
+        names = self.args.name
         delay = self.args.delay
         target = self._get_target(place)
         from ..driver.powerdriver import NetworkPowerDriver, PDUDaemonDriver, USBPowerDriver
         from ..resource.power import NetworkPowerPort, PDUDaemonPort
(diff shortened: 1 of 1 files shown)


jluebbe:
As discussed internally, this needs to keep the current default and handle finding all power resources as well.

avoine:
I was working on the Sentry agent (https://github.com/labgrid-project/labgrid/pull/595) and realized I need this too. We are doing power supply redundancy tests so we need to power off only one of the port but also all of them by default. Ping me if you want me to give it a shot for a solution.

avoine:
A solution could be to use an "aggregating driver" and changing the method by adding :

  ```
    def power(self):
    [...]
      try:
            drv = target.get_driver(PowerProtocol)
        except NoDriverFoundError:
        for name in names:
                try:
                    for resource in target.resources:
       [...]
```

like for the bootstrap method.

Also, maybe this could be generalized for example for flashing multiple banks with the FlashromDriver, flipping multiple GPIO, etc.

Emantor:
I already have a WIP branch and will probably post it next week.

Emantor:
I'll reopen with my changes.

PR #586, opened 2020-03-12
driver/power/gude24.py: Raise execution error if device is blocked

jgoetzinger:
**Description**
When the Gude Expert Power Control 8080 receives a request it blocks requests from other IP addresses for 120 seconds. Therefor when user A creates a request to the web interface user B with a different IP can not make a request for two minutes. The return message user B will get is:
```html
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN">
<html><head>
<meta http-equiv="Content-Type" content="text/html; charset=iso-8859-1">
<meta http-equiv="cache-control" content="no-cache">
<link rel="stylesheet" type="text/css" href="style.css">
<title>Device blocked</title>
</head><body><h2 align="center"><font color="red">Device is blocked by another user with IP xyz</font></h2>
<hr size=1 noshade><p align="center">[<a href="/">back</A>]
</body></html> 
```
The previous implementation of power_set would return ok, even if the device is blocked and the command has not been executed. This change raises an execution error to show that the device is locked and the command has not been executed.

**Checklist**
- [ ] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/labgrid/driver/power/gude24.py
+++ b/labgrid/driver/power/gude24.py
@@ -8,6 +8,8 @@ Driver has been tested with:
 * Gude Expert Power Control 8080
 """
 
+import re
+
 import requests
 
 from ..exception import ExecutionError
@@ -67,6 +69,10 @@ def get_state(request, index):
                 return True
 
             raise ExecutionError("failed to parse the port status")
+
+    m = re.search("Device is blocked by another user with IP ([^\s<]+)", request.text)
+    if m:
+        raise ExecutionError("device is blocked by another user with IP {}".format(m.group(1)))
 
     # if we got this far, something is wrong with the website
     raise ExecutionError("failed to determine status of power port {}".format(index))
--- a/labgrid/driver/power/gude24.py
+++ b/labgrid/driver/power/gude24.py
@@ -1,43 +1,65 @@
+"""
+This driver implements a power port for Gude Power Switches with up to
+24 ports.
+These switches differ in their API to the previous 8-port switches for set-
+and get-commands.
+
+Driver has been tested with:
+* Gude Expert Power Control 8080
+"""
+
 import requests
 
 from ..exception import ExecutionError
 
 
-# This driver implements a power port for Gude Power Switches with up to
-# 24 ports.
-# These switches differ in their API to the previous 8-port switches for set-
-# and get-commands.
-#
-# Driver has been tested with:
-# * Gude Expert Power Control 8080
-
 PORT = 80
 
 def power_set(host, port, index, value):
-    # The gude web-interface uses different pages for the three groups of
-    # switches. The web-interface always uses the 'correct' page to set a
-    # value. But commands for all pages are accepted on all pages.
+    """
+    The gude web-interface uses different pages for the three groups of
+    switches. The web-interface always uses the 'correct' page to set a
+    value. But commands for all pages are accepted on all pages.
+    """
     index = int(index)
     assert 1 <= index <= 24
     # access the web interface...
     value = 1 if value else 0
-    r = requests.get(
+    response = requests.get(
         "http://{}:{}/ov.html?cmd=1&p={}&s={}".format(host, port, index, value)
     )
-    r.raise_for_status()
+
+    # Check, that the port is in the desired state
+    state = get_state(response, index)
+    if state != value:
+        raise ExecutionError("failed to set port {} to status {}".format(index, value))
 
 
 def power_get(host, port, index):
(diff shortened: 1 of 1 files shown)


labgrid/driver/power/gude24.py
```diff
@@ -25,6 +25,8 @@ def power_set(host, port, index, value):
         "http://{}:{}/ov.html?cmd=1&p={}&s={}".format(host, port, index, value)
     )
     r.raise_for_status()
+    if "Device is blocked by another user" in r.text:
+        raise ExecutionError("Could not set power, device is blocked by other user.")
```
ukleinek: I wonder if instead of grepping for a specific problem it would be wiser to (at least try) to grep for any error and use its text in the Exception to be raised. This also had the advantage that the IP address is included in the message.

Having said that, using a method that allows error handling in a more reliable way would be great. There is http://wiki.gude.info/EPC_HTTP_Interface which documents a json interface. This isn't however available on the network power ports I have access to. :-|

jluebbe: A better approach would be to use PDUDaemon after https://github.com/pdudaemon/pdudaemon/pull/78 or something similar is merged. The requests are serialized and come from the same IP.
As a workaround, we have a socat forward on the exporter for our lab.

jluebbe:
A better solution for the underlying problem would be to use PDUDaemon after https://github.com/pdudaemon/pdudaemon/pull/78 or something similar is merged. The requests are serialized and come from the same IP.
As a workaround, we have a socat forward on the exporter for our lab.

Better error handling in this driver would be useful nevertheless.

labgrid/driver/power/gude24.py
jgoetzinger: I agree that just filtering for this one exception is not sufficient. I made some bigger changes to the driver, that allow to check if the set method changed the value correctly without code duplication. With this implementation the whole response will be printed if the state of the ports can not be checked.

jgoetzinger:
What do you think about the second approach I made for better error handling without searching for a specific error message?

labgrid/driver/power/gude24.py
```diff
@@ -1,49 +1,73 @@
...
             if line.find(",0") > 0:
                 return False
             if line.find(",1") > 0:
                 return True
 
             raise ExecutionError("failed to parse the port status")
+
     # if we got this far, something is wrong with the website
-    raise ExecutionError("failed to find the port")
+    print(request.text)
```
Emantor, line 72: leftover debug print.

jgoetzinger: Still it might be good to get the response of the request to see why it failed. In our case, request.text contains the reason (device is locked) as well as the ip of the user that locked the device.

jluebbe:
I've squashed your two commits, as the second solution looks much nicer. I've also done some minor style changes.

In addition, I added an untested commit which tries to pull the error message from the response via an regex. Does this work for you?

jgoetzinger:
One closing bracket was missing in your commit. After fixing this, it works as expected.

jluebbe:
> One closing bracket was missing in your commit. After fixing this, it works as expected.

Ah, yes. :) Do you see the conflicting IP in the exception message?

jgoetzinger:
Yes, here is the output after I locked the web interface from a different machine:
```shell
  File "/opt/venvs/labgrid/lib/python3.7/site-packages/labgrid/remote/client.py", line 1563, in main
    args.func(session)
  File "/opt/venvs/labgrid/lib/python3.7/site-packages/labgrid/remote/client.py", line 722, in power
    res = getattr(drv, action)()
  File "/opt/venvs/labgrid/lib/python3.7/site-packages/labgrid/binding.py", line 100, in wrapper
  File "/opt/venvs/labgrid/lib/python3.7/site-packages/labgrid/driver/powerdriver.py", line 143, in get
    return self.backend.power_get(self._host, self._port, self.port.index)
  File "/opt/venvs/labgrid/lib/python3.7/site-packages/labgrid/driver/power/gude24.py", line 47, in power_get
    state = get_state(response, index)
  File "/opt/venvs/labgrid/lib/python3.7/site-packages/labgrid/driver/power/gude24.py", line 74, in get_state
    raise ExecutionError("device is blocked by another user with IP {}".format(m.group(1)))
labgrid.driver.exception.ExecutionError: device is blocked by another user with IP 172.252.216.190
```

jluebbe (dismissed):

Emantor (approved):
Re-approving because of rebase.

PR #589, opened 2020-03-24
sshdriver: ssh_prefix to list and adjust users

Emantor:
**Description**
Convert ssh_prefix to a list and adjust all users. This was a string
previously, which led to get() and put() being broken on master, since
the string contained spaces. Convert it to a list, adjust the callers
and add new tests to test the SSHDriver against localhost. We also add a
new pytest option to supply the username, to make this work on travis.

Fixes: 1d7b97b838f7 (driver/sshdriver: Allow whitespaces in filenames Prevent filenames with whitespaces from beeing split.)

**Checklist**
- [x] Tests for the feature 
- [x] PR has been tested

Thanks @krevsbech for the report on IRC.

### DIFF
--- a/labgrid/driver/sshdriver.py
+++ b/labgrid/driver/sshdriver.py
@@ -31,19 +31,20 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
         self._keepalive = None
 
     def on_activate(self):
-        self.ssh_prefix = "-o LogLevel=ERROR"
+        self.ssh_prefix = ["-o", "LogLevel=ERROR"]
         if self.keyfile:
             keyfile_path = self.keyfile
             if self.target.env:
                 keyfile_path = self.target.env.config.resolve_path(self.keyfile)
-            self.ssh_prefix += " -i {}".format(keyfile_path)
-        self.ssh_prefix += " -o PasswordAuthentication=no" if (
-            not self.networkservice.password) else ""
+            self.ssh_prefix += ["-i", keyfile_path ]
+        if not self.networkservice.password:
+            self.ssh_prefix += ["-o", "PasswordAuthentication=no"]
+
         self.control = self._check_master()
-        self.ssh_prefix += " -F /dev/null"
-        self.ssh_prefix += " -o ControlPath={}".format(
-            self.control
-        ) if self.control else ""
+        self.ssh_prefix += ["-F", "/dev/null"]
+        if self.control:
+            self.ssh_prefix += ["-o", "ControlPath={}".format(self.control)]
+
         self._keepalive = None
         self._start_keepalive();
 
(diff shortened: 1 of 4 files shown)


labgrid/driver/sshdriver.py
```diff
@@ -164,15 +166,13 @@ def get_status(self):
...
-            "scp",
-            self.ssh_prefix,
-            "-P", str(self.networkservice.port),
-            filename,
-            "{user}@{host}:{remotepath}".format(
-                user=self.networkservice.username,
-                host=self.networkservice.address,
-                remotepath=remotepath)
+        transfer_cmd = ["scp"] + self.ssh_prefix + \
+            ["-P", str(self.networkservice.port),
```
jluebbe: This works as well:
```
>>> l = [1,2,3]
>>> print([0, *l, 4])
[0, 1, 2, 3, 4]
```
So use
```suggestion
        transfer_cmd = [
            "scp",
            *self.ssh_prefix,
            "-P", str(self.networkservice.port),
```
and avoid the reformatting.

labgrid/driver/sshdriver.py
```diff
@@ -191,15 +191,13 @@ def put(self, filename, remotepath=''):
...
-            "scp",
-            self.ssh_prefix,
-            "-P", str(self.networkservice.port),
-            "{user}@{host}:{filename}".format(
-                user=self.networkservice.username,
-                host=self.networkservice.address,
-                filename=filename),
-            destination
+        transfer_cmd = ["scp"] + self.ssh_prefix + \
+            ["-P", str(self.networkservice.port),
```
jluebbe: As above, this was intentionally split to multiple lines by parameter.

labgrid/driver/sshdriver.py
```diff
@@ -236,7 +234,7 @@ def _cleanup_own_master(self):
 
     def _start_keepalive(self):
         """Starts a keepalive connection via the own or external master."""
-        args = ["ssh"] + self.ssh_prefix.split() + ["cat"]
+        args = ["ssh"] + self.ssh_prefix + ["cat"]
```
jluebbe: Use `*self.ssh_prefix`.

tests/conftest.py
```diff
@@ -146,13 +146,17 @@ def pytest_addoption(parser):
                      help="Run sigrok usb tests with fx2lafw device (0925:3881)")
     parser.addoption("--local-sshmanager", action="store_true",
                      help="Run SSHManager tests against localhost")
+    parser.addoption("--ssh-username", default=None,
+                     help="SSH username to use for SSHService testing")
```
jluebbe: There is no `SSHService` (it should really be called that, but it's `NetworkService`). Below as well.

tests/test_sshdriver.py
```diff
@@ -6,49 +6,88 @@
...
+    s = target.get_driver("SSHDriver")
+    return s
+
+def test_create_fail_missing_resource(target):
+    with pytest.raises(NoResourceFoundError):
         SSHDriver(target, "ssh")
-        s = target.get_driver("SSHDriver")
-        return s
-
-class TestSSHDriver:
```
jluebbe: Getting rid of the test classes should be mentioned in the commit message.

Emantor:
Even if I re-request a review, this still shows up as "Changes Requested" in the PR overview :disappointed:

Emantor:
I added another commit which disables the `UserKnownHostsFile` and `StrictHostKeyChecking` inside the prefix. This way its also picked up by all other commands.

PR #594, opened 2020-03-30
USBSerialPorts: Add baudrate

mbgg:
Up to now we only support 115200 baud for the serial console.
Add a new option to allow a different baud rate to be used.

Signed-off-by: Matthias Brugger <matthias.bgg@gmail.com>

Documentation for the feature already exists but it wasn't actually implemented.
I tested this with a MediaTek board that has a baudrate of 921600 on it's serial console.

**Description**

**Checklist**
- [ ] Tests for the feature 

- [ ] PR has been tested

### DIFF
--- a/labgrid/remote/exporter.py
+++ b/labgrid/remote/exporter.py
@@ -202,6 +202,7 @@ class SerialPortExport(ResourceExport):
         return {
             'host': self.host,
             'port': self.port,
+            'speed': self.local.speed,
             'extra': {
                 'path': self.local.port,
             }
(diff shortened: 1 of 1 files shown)


labgrid/resource/udev.py
```diff
@@ -224,6 +224,13 @@ def read_attr(self, attribute):
 @target_factory.reg_resource
 @attr.s(eq=False)
 class USBSerialPort(USBResource, SerialPort):
+    """This resource describes a USB serial port.
+
+    Args:
+        speed (int): baudrate
+    """
+    speed = attr.ib(default=115200, validator=attr.validators.instance_of(int))
+
```
Emantor: This shouldn't be required, the `SerialPort` class already contains the speed attribute.

mbgg: You are right. I fixed this and pushed again (I know, github workflow is about adding fix patches on top of the pull request. sorry for that).

Emantor: IMO it's fine to just force-push, since github now includes a diff between the force-pushes.

Emantor:
I tested this locally, works as expected.

Emantor:
@mbgg I force pushed to update the branch, but now the DCO is incorrect, since your commit and author differed.

mbgg:
@Emantor hope that is fixed now. Too many developement machines with too many different gitconfigs...

PR #595, opened 2020-04-01
driver/power: add support for Sentry PDU

avoine:
add support for Sentry PDU

### DIFF
--- a/labgrid/driver/power/sentry.py
+++ b/labgrid/driver/power/sentry.py
@@ -1,0 +1,94 @@
+from ..exception import ExecutionError
+from ...util.helper import processwrapper
+
+INDEX_TO_OID = {
+    1: "1.1.1",
+    2: "1.1.2",
+    3: "1.1.3",
+    4: "1.1.4",
+    5: "1.1.5",
+    6: "1.1.6",
+    7: "1.1.7",
+    8: "1.1.8",
+    9: "1.2.1",
+    10: "1.2.2",
+    11: "1.2.3",
+    12: "1.2.4",
+    13: "1.2.5",
+    14: "1.2.6",
+    15: "1.2.7",
+    16: "1.2.8",
+    17: "1.3.1",
+    18: "1.3.2",
+    19: "1.3.3",
+    20: "1.3.4",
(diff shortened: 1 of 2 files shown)


labgrid/driver/power/sentry.py
```diff
@@ -0,0 +1,94 @@
+from ..exception import ExecutionError
```
Emantor, line 1: Please add a short comment on the top with which device this connector was tested.

labgrid/driver/power/sentry.py
```diff
@@ -0,0 +1,94 @@
...
+    47: "1.6.7",
+    48: "1.6.8",
+}
+
+
+BASE_STATUS_OID = ".1.3.6.1.4.1.1718.3.2.3.1.10"
+BASE_CTRL_OID = ".1.3.6.1.4.1.1718.3.2.3.1.11"
+
+def _snmp_get(host, oid):
+    out = processwrapper.check_output(
```
Emantor, line 60: W already support `pysnmp` for SNMP and also include `pysnmp-mib` in `snmp-requirements.txt`. Please switch to using pysnmp directly instead of using an external program, i.e.
```suggestion
    import pysnmp.hlapi
```

and so forth here and in `_snmp_set`.

jluebbe: Hmm, we already have the calls to the snmpset tool in `apc.py`, so I wouldn't mind doing the same here. Especially when this is only temporary before moving to pdudaemon.

Emantor: I didn't check this, in this case its fine to not convert to the python module.

avoine:
> We recommend to implement power switchting within [pdudaemon](https://github.com/pdudaemon/pdudaemon,) however due to the unmerged [pdudaemon/pdudaemon#78](https://github.com/pdudaemon/pdudaemon/pull/78), labgrid and pdudaemon are not fully compatible yet, I'd be okay to merge this for now. I'd suggest to add support to pdudaemon as well, eventually we want to fully switch over.

ok, I'll create a PR to add Sentry's support to pdudaemon.

PR #596, opened 2020-04-02
labgrid/remote: remove explicit path from ser2net

Emantor:
**Description**
Remove the explicit path from ser2net. Setting PATH correctly is for the
user to decide, and this allows overriding it externally with a locally
compiled version by modifying PATH.

Reasonably sure this breaks debian packages which run labgrid as a non-root user. We could modify the service units for this.
On the other hand this will be useful to have to test newer ser2net versions locally without updating the system.
@jluebbe what do you think?
**Checklist**
- [x] Feature has been tested.

### DIFF
--- a/labgrid/remote/exporter.py
+++ b/labgrid/remote/exporter.py
@@ -8,7 +8,9 @@ import sys
 import os
 import time
 import traceback
+import shutil
 import subprocess
+import warnings
 from socket import gethostname, getfqdn
 import attr
 from autobahn.asyncio.wamp import ApplicationRunner, ApplicationSession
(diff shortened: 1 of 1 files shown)


jluebbe:
Hmm, we could use `shutil.which` to do the lookup ourselves and fall back to `/usr/bin/ser2net` with a (one-time) warning when not found?

Emantor:
> Hmm, we could use `shutil.which` to do the lookup ourselves and fall back to `/usr/bin/ser2net` with a (one-time) warning when not found?

Sounds good, implemented.

Emantor:
Hold up, this doesn't work.

Issue #597, opened 2020-04-08
stdout and stderr not available in the exceptions from ssh driver 


TheMeaningfulEngineer:
If the ssh subprocess throws an error for whatever the reason (the example will show a permission error), the text of the error isn't accessible in the `ExecutionError` that gets raised.

```
from labgrid import Target
from labgrid.resource import NetworkService
from labgrid.driver import SSHDriver

lg_target = Target("Device")
lg_network_service = NetworkService(lg_target,
                                    name='network_service',
                                    address='192.168.3.11',
                                    username='root')

lg_ssh_driver = SSHDriver(lg_target, name=None)
try:
    lg_target.activate(lg_ssh_driver)
except Exception as e:
    print(e.msg)
    print(e.stdout)
    print(e.stderr)
```

**Current behavior**

Notice that the `root@192.168.3.11: Permission denied (publickey).` isn't a result of the print statements but it gets printed from within the library.

```
root@192.168.3.11: Permission denied (publickey).
failed to connect to 192.168.3.11 with ['ssh', '-f', '-o', 'LogLevel=ERROR', '-o', 'PasswordAuthentication=no', '-x', '-o', 'ConnectTimeout=30', '-o', 'ControlPersist=300', '-o', 'UserKnownHostsFile=/dev/null', '-o', 'StrictHostKeyChecking=no', '-o', 'ServerAliveInterval=15', '-MN', '-S', '/tmp/labgrid-ssh-tmp-njzanbrm/control-192.168.3.11', '-p', '22', 'root@192.168.3.11'] and 255
None
None
```

**Expected behavior**

```
failed to connect to 192.168.3.11 with ['ssh', '-f', '-o', 'LogLevel=ERROR', '-o', 'PasswordAuthentication=no', '-x', '-o', 'ConnectTimeout=30', '-o', 'ControlPersist=300', '-o', 'UserKnownHostsFile=/dev/null', '-o', 'StrictHostKeyChecking=no', '-o', 'ServerAliveInterval=15', '-MN', '-S', '/tmp/labgrid-ssh-tmp-njzanbrm/control-192.168.3.11', '-p', '22', 'root@192.168.3.11'] and 255
None
root@192.168.3.11: Permission denied (publickey).
```

TheMeaningfulEngineer:
The reason for this is likely the way how the `stdout` and `stderr` are piped.
[...]

PR #598, opened 2020-04-08
sshdriver: Add stdout and stderr to the subprocess exception

TheMeaningfulEngineer:
Closes https://github.com/labgrid-project/labgrid/issues/597

Running the example from the issue:

```
from labgrid import Target
from labgrid.resource import NetworkService
from labgrid.driver import SSHDriver

lg_target = Target("Device")
lg_network_service = NetworkService(lg_target,
                                    name='network_service',
                                    address='192.168.3.15',
                                    username='root')

lg_ssh_driver = SSHDriver(lg_target, name=None)
try:
    lg_target.activate(lg_ssh_driver)
except Exception as e:
    print(e.msg)
    print(e.stdout)
    print(e.stderr)
```

```
$ python labgrid_issue.py 
Failed to connect to 192.168.3.15 with ['ssh', '-f', '-o', 'LogLevel=ERROR', '-o', 'PasswordAuthentication=no', '-x', '-o', 'ConnectTimeout=30', '-o', 'ControlPersist=300', '-o', 'UserKnownHostsFile=/dev/null', '-o', 'StrictHostKeyChecking=no', '-o', 'ServerAliveInterval=15', '-MN', '-S', '/tmp/labgrid-ssh-tmp-6nhhwiwh/control-192.168.3.15', '-p', '22', 'root@192.168.3.15'] and 255
['']
['ssh: connect to host 192.168.3.15 port 22: No route to host\r\n']
```

The `stdout` and `stderr` are printed as lists because that seems to be the requirement of the [ExecutionError arguments](https://github.com/labgrid-project/labgrid/blob/b0cc54dc2e2c9cf94c9c98e22f00a150ce475597/labgrid/driver/exception.py#L9).

**Description**

**Checklist**
- [ ] Tests for the feature 

- [ ] PR has been tested

### DIFF
--- a/labgrid/driver/sshdriver.py
+++ b/labgrid/driver/sshdriver.py
@@ -56,13 +56,16 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
 
     def _start_own_master(self):
         """Starts a controlmaster connection in a temporary directory."""
+
+        timeout = 30
+
         self.tmpdir = tempfile.mkdtemp(prefix='labgrid-ssh-tmp-')
         control = os.path.join(
             self.tmpdir, 'control-{}'.format(self.networkservice.address)
         )
         # use sshpass if we have a password
         args = ["sshpass", "-e"] if self.networkservice.password else []
-        args += ["ssh", "-f", *self.ssh_prefix, "-x", "-o", "ConnectTimeout=30",
+        args += ["ssh", "-f", *self.ssh_prefix, "-x", "-o", "ConnectTimeout={}".format(timeout),
                  "-o", "ControlPersist=300", "-o",
                  "UserKnownHostsFile=/dev/null", "-o", "StrictHostKeyChecking=no",
                  "-o", "ServerAliveInterval=15", "-MN", "-S", control, "-p",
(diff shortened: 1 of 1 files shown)


Emantor:
It looks like python 3.5 does not support `capture_output=true`, also the connection to localhost times out, the `process.run` never finishes. The exception handling is also broken for the `TimeoutExpired` case, there is no `returncode` attribute inside of this exception.

TheMeaningfulEngineer:
Thanks for the feedback @Emantor.
Do you perhaps have a code snippet for running just the ssh driver tests locally?
The turnaround will be much quicker.

Emantor:
`pytest -k sshdriver --ssh-username <your username> tests`
This filters the tests using `-k` and runs only those for the SSHDriver. `--ssh-username` enables the SSHDriver tests against localhost, so you'll need to ensure that the supplied user can connect to localhost without a password.

labgrid/driver/sshdriver.py
```diff
@@ -72,19 +72,32 @@ def _start_own_master(self):
...
                 raise ExecutionError(
-                    "failed to connect to {} with {} and {}".
-                    format(self.networkservice.address, args, self.process.wait())
+                    "Failed to connect to {} with {} and {}".
+                    format(self.networkservice.address, args, return_value),
+                    stdout=stdout,
+                    stderr=stderr
                 )
         except subprocess.TimeoutExpired:
+            stdout = self.process.stdout.readlines()
```
TheMeaningfulEngineer, line 93: This will block until the process either closes or writes a line. It defeats the purpose of the upper timeout. Must change this or remove it.

(the original motivation was to at least try to give some info on what the subprocess printed out before we killed it with the timeout from `self.process.wait`

TheMeaningfulEngineer: There seems to not be a quick fix for this. The [solution I've encountered](https://stackoverflow.com/a/4896288/1214214) basically suggest implementing an async read in a separated thread with a queue. 

Dropping the attempt to show std* data on a forced timeout.

Emantor:
Please squash down the second commit into the first.

Emantor:
There are now two `Signed-off-by` tags in your commit, while not wrong, you probably want to remove the first one :grin:

PR #603, opened 2020-05-12
support binding outbound SSH connections to specific interfaces, add support for scp, rsync and sshfs to labgrid-client

jluebbe:
**Description**
When the target is only reachable via link-local addresses or overlapping network ranges, we need to make sure to connect from a specific interface on the exporter. This is possible using the SO_BINDTODEVICE socket option (which requires root privileges, so we need a sudo helper script) and some additional SSH tunneling.

As SSH is now more widely usable, also implement the low-hanging fruits in labgrid-client and SSHDriver (scp, rsync and sshfs).

**Checklist**
- [ ] Tests for the feature 
- [x] PR has been tested

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -10,6 +10,7 @@ Package: labgrid
 Architecture: any
 Pre-Depends: dpkg (>= 1.16.1), python3, ${misc:Pre-Depends}
 Depends: ${python3:Depends}, ${misc:Depends}, ${shlibs:Depends}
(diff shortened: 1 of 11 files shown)


jluebbe:
> Nit: In commit `helpers: add labgrid-bound-connect`, `:` is used instead of `.` to end the sentence.

I've fixed that and rebased on master.

PR #605, opened 2020-05-14
driver: sshdriver: avoid password authentication fallback

jremmet:
I saw issues with the control socket dropping me to interactive shell to
enter password:

INFO: Socket already closed
Control socket connect(/tmp/labgrid-ssh-tmp-ik7fhcku/control-192.168.0.10): Connection refused

To avoid this fallback don't allow password authentication after the
control socket runs.

**Description**

**Checklist**
- [ ] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/labgrid/driver/sshdriver.py
+++ b/labgrid/driver/sshdriver.py
@@ -105,6 +105,9 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
 
         self.logger.debug('Connected to %s', self.networkservice.address)
 
+        # We don't want to silently fall back to password authentication from now
+        if self.networkservice.password:
+            self.ssh_prefix += ["-o", "PasswordAuthentication=no"]
         return control
 
     def _check_master(self):


jluebbe:
"INFO: Socket already closed" comes from _cleanup_own_master(), called from on_deactivat(). After that, on_activate() would need to run before anything should try to use the control socket again.

As we want all connections to go via the master, we should never allow fallback to creating new standalone connections if the master is gone. So, that option should have no effect...

The _check_keepalive() function should ensure that we sill have a a working connection to the target. How do you get into the situation of not being able to connect to the control socket?

jremmet:
The setup is small tests which makes a transition to shell, request ssh via get_driver("SSHDriver") and use it's put to place a small file on target. On power down deactivate_all_drivers() is called.
Sometime the test asks for password authentication.
"INFO: Socket already closed" may come after that.
Adding the _check_keepalive() check to "put" didn't show the problem 
I'll debug it an return with more info.

PR #606, opened 2020-05-19
YKUSH use cli and make available for remote access

mbgg:
**Description**

This pull requests has two important changes.
1) It changes the backend to access the YKUSH. The ykush python implementation is in a really bad state. The "official" tool [1] isn't maintained. The pypip implementation [2] lacks basic functionality like for example connect to several Yepkits via the serial number.
Let's just use the ykush command line tool, which get's updated and has all the functionality we want.

2) Make YKUSH accessible through remote access. With this patch we can export the access to the YKUSH and toggle the power via the labgrid-client interface.
    
[1] https://github.com/Yepkit/pykush
[2] https://github.com/alvarop/pykush/

**Checklist**
- [ ] Tests for the feature 

- [ ] PR has been tested


labgrid/driver/powerdriver.py
```diff
@@ -182,26 +182,35 @@ def get(self):
...
-        self.pykush = self.pykush_mod.YKUSH(serial=self.port.serial)
+        if self.target.env:
+            self.tool = self.target.env.config.get_tool('ykush') or 'ykushcmd'
+        else:
+            self.tool = 'ykushcmd'
+
+    def _get_ykushcmd_prefix(self):
+        prefix = self.port.command_prefix+[
+            self.tool,
+            "-s", str(self.port.serial),
```
jluebbe, line 198: Does the ykushcmd tool talk to all connected devices to find the one with the correct serial or does it use the USB descriptor information? Otherwise, one could get problems with concurrent accesses (which probably also exist already with pykush).

labgrid/remote/exporter.py
```diff
@@ -341,6 +341,26 @@ def _get_params(self):
...
+
+    def _get_params(self):
+        """Helper function to return parameters"""
+        return {
+            'host': self.host,
+            'busnum': self.local.busnum,
+            'devnum': self.local.devnum,
+            'path': self.local.path,
+            'vendor_id': self.local.vendor_id,
+            'model_id': self.local.model_id,
```
jluebbe, line 359: It seems, that most of these are not used by the driver. It would be better to tell ykushcmd the busnum/devnum or path, so it wouldn't need to scan. Maybe that's supported?

labgrid/resource/udev.py
```diff
@@ -466,6 +466,27 @@ def __attrs_post_init__(self):
...
+    index = attr.ib(default=-1, validator=attr.validators.instance_of(int),
+                    converter=int)
+
+    def filter_match(self, device):
+        match = (device.properties.get('ID_VENDOR_ID'), device.properties.get('ID_MODEL_ID'))
+
+        if match not in [("04d8", "f2f7")]:
+             return False
+
+        return super().filter_match(device)
```
jluebbe, line 488: Without additional matche filters, this would match all ykush devices, even ones with different serials. So one would need to specify both serial and path in the resource configuration, which seems redundant.

labgrid/resource/remote.py
```diff
@@ -207,6 +207,17 @@ def __attrs_post_init__(self):
         super().__attrs_post_init__()
 
 
+@target_factory.reg_resource
+@attr.s(eq=False)
+class NetworkYKUSHPowerPort(RemoteUSBResource):
+    """The NetworkYKUSHPowerPort describes a remotely YEPKIT YKUSH switchable USB hub"""
+    serial = attr.ib(default=None, validator=attr.validators.instance_of(str))
+    index = attr.ib(default=None, validator=attr.validators.instance_of(int))
```
jluebbe, line 215: These should not have defaults. The current attrs library supports adding keyword-only arguments, so the default=None workaround shouldn't be necessary any more.

labgrid/resource/udev.py
```diff
@@ -475,8 +475,8 @@ class YKUSHPowerPort(USBResource):
         serial (str): serial of the YKUSH device
         index (int): port index
     """
-    serial = attr.ib(default=None, validator=attr.validators.instance_of(str))
-    index = attr.ib(default=None, validator=attr.validators.instance_of(int),
+    serial = attr.ib(default="dummy", validator=attr.validators.instance_of(str))
+    index = attr.ib(default=-1, validator=attr.validators.instance_of(int),
```
jluebbe, line 479: Adding suggest support this way only seems to work, because the serial is not checked during match. We should either have matching based on the path and autodetect the serial or the other way around, not require both to be specified manually. That would likely require some more intelligence in suggesst.

jluebbe:
In the medium term, I'd like to replace labgrid's powerdrivers with pdudaemon, which also has support for YKUSH (at least one of the versions). @mattface indicated that the approach to simplify it and support blocking requests (https://github.com/pdudaemon/pdudaemon/pull/78) can continue.

mbgg:
> In the medium term, I'd like to replace labgrid's powerdrivers with pdudaemon, which also has support for YKUSH (at least one of the versions). @mattface indicated that the approach to simplify it and support blocking requests ([pdudaemon/pdudaemon#78](https://github.com/pdudaemon/pdudaemon/pull/78)) can continue.

Does this mean you want me to abandon this pull request? I'd prefer to go on with this to get something working in the short term, then we can move that to pdudaemon in the future.

jluebbe:
> Does this mean you want me to abandon this pull request? I'd prefer to go on with this to get something working in the short term, then we can move that to pdudaemon in the future.

Besides the missing maintenance, what are the issues you have with https://github.com/Yepkit/pykush/? Just the missing remote support? I took a short look at the ykushcmd implementation and it seemed to have a lot of complexity for something that should be relatively simple. ;)

If you can solve the inconsistency between matching on serial number vs. matching on the bus topology, I'd merge the PR, as it would allow us to access them remotely.

mbgg:
> > Does this mean you want me to abandon this pull request? I'd prefer to go on with this to get something working in the short term, then we can move that to pdudaemon in the future.
> 
> Besides the missing maintenance, what are the issues you have with https://github.com/Yepkit/pykush/? Just the missing remote support? I took a short look at the ykushcmd implementation and it seemed to have a lot of complexity for something that should be relatively simple. ;)
> 

The problem with pykush is that the pip version has changed and doesn't work with the labgrid implementation anymore. So you would need to install pykush from https://github.com/Yepkit/pykush instead from pip.

> If you can solve the inconsistency between matching on serial number vs. matching on the bus topology, I'd merge the PR, as it would allow us to access them remotely. I'd suspect though, that finishing that pdudaemon PR may be easier to do.

Could we use

jluebbe:
> > > Does this mean you want me to abandon this pull request? I'd prefer to go on with this to get something working in the short term, then we can move that to pdudaemon in the future.
> > 
> > 
> > Besides the missing maintenance, what are the issues you have with https://github.com/Yepkit/pykush/? Just the missing remote support? I took a short look at the ykushcmd implementation and it seemed to have a lot of complexity for something that should be relatively simple. ;)
> 
> The problem with pykush is that the pip version has changed and doesn't work with the labgrid implementation anymore. So you would need to install pykush from https://github.com/Yepkit/pykush instead from pip.

I've only ever tried it with the version from git, which is also referred to in the source. We could add a git entry for it in a requirements.txt.

> > If you can solve the inconsistency between matching on serial number vs. matching on the bus topology, I'd merge the PR, as it would allow us to access them remotely. I'd suspect though, that finishing that pdudaemon PR may be easier to do.
> 
> Could we use ID_SERIAL_SHORT from udevadm info for matching in udev?
> 
> It shows me:
> E: ID_SERIAL_SHORT=YKC6071

I don't have a ykush available for testing. Is that the same serial as used with the command line tool?

mbgg:
> > > > Does this mean you want me to abandon this pull request? I'd prefer to go on with this to get something working in the short term, then we can move that to pdudaemon in the future.
> > > 
> > > 
> > > Besides the missing maintenance, what are the issues you have with https://github.com/Yepkit/pykush/? Just the missing remote support? I took a short look at the ykushcmd implementation and it seemed to have a lot of complexity for something that should be relatively simple. ;)
> > 
> > 
> > The problem with pykush is that the pip version has changed and doesn't work with the labgrid implementation anymore. So you would need to install pykush from https://github.com/Yepkit/pykush instead from pip.
> 
> I've only ever tried it with the version from git, which is also referred to in the source. We could add a git entry for it in a requirements.txt.
> 

That would work as well. I can find out how to do that.

> > > If you can solve the inconsistency between matching on serial number vs. matching on the bus topology, I'd merge the PR, as it would allow us to access them remotely. I'd suspect though, that finishing that pdudaemon PR may be easier to do.
> > 
> > 
> > Could we use ID_SERIAL_SHORT from udevadm info for matching in udev?
> > It shows me:
> > E: ID_SERIAL_SHORT=YKC6071
> 
> I don't have a ykush available for testing. Is that the same serial as used with the command line tool?

Yes:
```
# ykushcmd -l

Attached YKUSH Boards

---------------------
YKUSH device found with Serial Number: YKC6071
  Manufacturer: Yepkit Lda.
  Product:      YKUSH
```

I'm not sure how I'll have to implement this. Using the existing ID_VENDOR_ID and ID_MODEL_ID and add ID_SERIAL_SHORT or just use ID_SERIAL_SHORT?

labgrid/resource/udev.py
mbgg: Ok, I'll drop suggest support for now. We can add it later in a seperate pull request.

mbgg:
Ok the pull request now add the pykush dependency to the requierements file and matches against the serial number of the YKUSH.

@jluebbe please review again :)

labgrid/resource/udev.py
mbgg: ok, now it matches against the serial only which is discoverable.

Emantor:
I have pushed a commit onto your branch which should fix the travis failure.

mbgg:
> I have pushed a commit onto your branch which should fix the travis failure.

Thanks! This fixed the issue in travis but not yet in docker/build.sh

mbgg:
@jluebbe any comments on this? It's been a while since I pushed to code for review :)

.travis.yml
```diff
@@ -6,6 +6,9 @@ addons:
       libow-dev
       openssh-server
       openssh-client
+      libudev-dev
+      libusb-1.0-0-dev
+      libfox-1.6-dev
```
Emantor, line 11: We have since switched to github workflows, since travis and opensource apparently don't agree any longer.

requirements.txt
```diff
@@ -12,3 +12,4 @@ PyYAML==5.3.1
 ansicolors==1.1.8
 pyusb==1.1.0
 six>=1.13.0
+https://github.com/Yepkit/pykush/archive/master.zip#egg=pykush
```
krevsbech, line 15: As I read this you are not using a tag from the project but resent master. In order to maintain CI stability and to avoid sudden regressions due to dependencies breaking something I believe you must point to a certain version.

docker/base/Dockerfile
```diff
@@ -8,7 +8,7 @@ COPY ./ /opt/labgrid/
 
 RUN set -e ;\
     apt update -q=2 ;\
-    apt install -q=2 --yes --no-install-recommends python3 python3-dev python3-pip python3-setuptools python3-wheel git build-essential libsnappy-dev ;\
+    apt install -q=2 --yes --no-install-recommends python3 python3-dev python3-pip python3-setuptools python3-wheel git build-essential libsnappy-dev libusb-1.0-0-dev libudev-dev ;\
```
krevsbech, line 11: Consider moving the lib dependency to the images where it is needed.
As I understand your contribution this is relevant for client and exporter ?

Bastian-Krause:
@mbgg I guess most of the issues this PR tried to solve are now solved by #1143. Is there something left? If not, can we close this?

PR #615, opened 2020-06-04
sshdriver: add background method

tiljeset:
Add method for running commands in the background.

Signed-off-by: Morten Hein Tiljeset <mhti@triax.com>

**Description**

Add support for running tasks in the background using the SSHDriver. This is an
automated testing equivalent of running the command in a separate shell.

I use this for running a command which streams data to my DUT. I've tested
locally that this both starts the application and cleans up correctly after the
fact, and that it does not interfere with running other ssh commands.

**Checklist**
- [x] Tests for the feature
- [x] PR has been tested

### DIFF
--- a/labgrid/driver/sshdriver.py
+++ b/labgrid/driver/sshdriver.py
@@ -170,6 +170,57 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
             stderr.pop()
         return (stdout, stderr, sub.returncode)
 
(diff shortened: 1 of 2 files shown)


jluebbe:
There is a pending PR (#552), which has a similar goal. There are still some unresolved questions there, and it's not clear how best to proceed.

Emantor:
The pending PR is relevant here, we will need a solutions for all drivers that implement the same protocol.

PR #624, opened 2020-06-18
smallubootdriver and getting_started fixes

ynezz:
* smallubootdriver: fix missing init_commands argument support
* doc/getting_started: fix missing python3-setuptools package

- [x] PR has been tested on TP-Link Archer C7v5 device

### DIFF
--- a/labgrid/driver/smallubootdriver.py
+++ b/labgrid/driver/smallubootdriver.py
@@ -58,6 +58,8 @@ class SmallUBootDriver(UBootDriver):
 
         # wait until UBoot has reached it's prompt
         self.console.expect(self.prompt)
+        for command in self.init_commands:  #pylint: disable=not-an-iterable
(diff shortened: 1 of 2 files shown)


labgrid/driver/smallubootdriver.py
```diff
@@ -58,6 +58,8 @@ def _await_prompt(self):
 
         # wait until UBoot has reached it's prompt
         self.console.expect(self.prompt)
+        for command in self.init_commands:  #pylint: disable=not-an-iterable
+            self._run(command)
```
Emantor, line 62: `self._run` is correct here instead of `self._run_check`, since the `SmallUBootDriver` returns `0` for every command anyway.

ynezz: Yep, it doesn't work with `_run_check`.

Issue #565, opened 2020-01-27
ser2net >= 4.0 breaks configline option


Emantor:
Ser2net switched to a YAML format in the 4.0 release and marked the `-C` configuration file for old style configuration as unsupported. The same was done for `-c`, which is the configuration line option we are using for exporters.

Emantor:
Pasting my mail conversation with the ser2net maintainer on this topic.

On Mon, Jan 27, 2020 at 02:37:05PM +0100, Rouven Czerwinski wrote:
> Hello Corey,
> 
> we are users of the configline option within ser2net. We use ser2net as
> an ondemand process to spawn when we want to expose a local serial port
> over the network. We spawn individual ser2net processes for each serial
> port, this way we can disable access to the port when a user no longer
> needs it.
> The code within our hardware control library can be found at:
> https://bit.ly/2GxqGnj
> Since the configline options is effectively deprecated, we would like
> to know if there are alternatives within the ser2net/gensio ecosystem.
> Thank you for your work on ser2net and have a nice day.

I have realized that would be a problem, but I'm not sure what to do
about it at the moment.

The gensiot program could probably do what you need, and it would be
simpler than ser2net.  You could run something like:

  gensiot -a -i 'serialdev,/dev/ttyS1,9600N81,local' 'telnet(rfc2217),tcp,3000'

And you get basically the same thing as a since ser2net connection.  I
think all the remote serial control stuff will work.

You don't get the ability to have a control port, and gensiot will terminate
when the remote end disconnects, so you will have to respawn it each
time.  If you want the authentication capabilities, it's not going to be
as flexible as ser2net (but encryption should be easy).

-corey

PR #628, opened 2020-08-04
remote/exporter: Use YAML syntax for ser2net if version > 4.0.0

cbs101:
See:
https://github.com/labgrid-project/labgrid/issues/565

Verified to work when using ser2net version 2.10.0 and version 4.1.8.
Test performed: 
- Start exporter
- Check which port is chosen
- Connect using telnet to the ip / port
Verify that commands be send, and output seen.

### DIFF
--- a/labgrid/remote/exporter.py
+++ b/labgrid/remote/exporter.py
@@ -230,7 +230,7 @@ class SerialPortExport(ResourceExport):
                 '-d',
                 '-n',
                 '-Y', 'connection: &con01#  accepter: telnet(rfc2217,mode=server),{}'.format(self.port),
-                '-Y', '  connector: serialdev,{},115200n81,local'.format(start_params['path']
+                '-Y', '  connector: serialdev(nouucplock=true),{},115200n81,local'.format(start_params['path']
                 ),
             ])
         else:
@@ -238,6 +238,7 @@ class SerialPortExport(ResourceExport):
                 self.ser2net_bin,
                 '-d',
                 '-n',
+                '-u',
                 '-C',
                 '{}:telnet:0:{}:115200 NONE 8DATABITS 1STOPBIT LOCAL'.format(
                     self.port, start_params['path']
--- a/labgrid/remote/exporter.py
+++ b/labgrid/remote/exporter.py
@@ -220,15 +220,29 @@ class SerialPortExport(ResourceExport):
         assert self.child is None
         assert start_params['path'].startswith('/dev/')
         self.port = get_free_port()
-        self.child = subprocess.Popen([
-            self.ser2net_bin,
-            '-d',
-            '-n',
-            '-C',
-            '{}:telnet:0:{}:{} 8DATABITS NONE 1STOPBIT LOCAL'.format(
-                self.port, start_params['path'], self.local.speed
-            ),
-        ])
+
+        # Ser2net has switched to using YAML format at version 4.0.0.
+        _, _, version = str(subprocess.check_output([self.ser2net_bin,'-v'])).split(' ')
+        major_version = version.split('.')[0]
+        if int(major_version) >= 4:
+            self.child = subprocess.Popen([
+                self.ser2net_bin,
+                '-d',
+                '-n',
+                '-Y', 'connection: &con01#  accepter: telnet(rfc2217,mode=server),{}'.format(self.port),
+                '-Y', '  connector: serialdev,{},115200n81,local'.format(start_params['path']
+                ),
+            ])
+        else:
+            self.child = subprocess.Popen([
+                self.ser2net_bin,
+                '-d',
+                '-n',
+                '-C',
+                '{}:telnet:0:{}:115200 NONE 8DATABITS 1STOPBIT LOCAL'.format(
+                    self.port, start_params['path']
+                ),
+            ])
         try:
             self.child.wait(timeout=0.5)
             raise ExporterError("ser2net for {} exited immediately".format(start_params['path']))


labgrid/remote/exporter.py
```diff
@@ -220,15 +220,30 @@ def _start(self, start_params):
...
-            '-d',
-            '-n',
-            '-C',
-            '{}:telnet:0:{}:{} 8DATABITS NONE 1STOPBIT LOCAL'.format(
-                self.port, start_params['path'], self.local.speed
-            ),
-        ])
+
+        # Ser2net has switched to using YAML format at version 4.0.0.
+        _, _, version = str(subprocess.check_output(['ser2net','-v'])).split(' ')
```
Emantor, line 225: Use `self.ser2net_bin` instead of hard coding `'ser2net'`, this is resolved during init.

labgrid/remote/exporter.py
```diff
@@ -220,15 +220,30 @@ def _start(self, start_params):
...
+                '-n',
+                '-Y',
+                'connection: &con01#  accepter: telnet, {}'.format(self.port),
+                '-Y',
+                '  connector: serialdev,{},115200n81,local'.format(start_params['path']
+                ),
+            ])
+        else:
+            self.child = subprocess.Popen([
+                '/usr/sbin/ser2net',
```
Emantor, line 239: same as above, please use `self.ser2net_bin`.

labgrid/remote/exporter.py
```diff
@@ -220,15 +220,30 @@ def _start(self, start_params):
...
-            '-n',
-            '-C',
-            '{}:telnet:0:{}:{} 8DATABITS NONE 1STOPBIT LOCAL'.format(
-                self.port, start_params['path'], self.local.speed
-            ),
-        ])
+
+        # Ser2net has switched to using YAML format at version 4.0.0.
+        _, _, version = str(subprocess.check_output(['ser2net','-v'])).split(' ')
+        if version > "4.0.0":
```
Emantor, line 226: we should probably parse out the major version and do the check against integers. This breaks if another `.0` is added:
```
python 3.8.3 (default, May 13 2020, 19:59:26)
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> "4.0.0.0" > "4.0.0"
True
```

labgrid/remote/exporter.py
cbs101: Will do. Thanks for the comment!

labgrid/remote/exporter.py
cbs101: Will do. Thanks for the comment!

labgrid/remote/exporter.py
cbs101: Will do

cbs101:
do _NOT_ merge. 

The change does indeed cause ser2net to properly start a telnet connection, that can be accessed using telnet.
But:
Finally had some time to try the full monty (coordinater, client, exporter) with newest labgrid. Using ser2net of 4.1.8 does not work, due to something that fails to be setup. 
Error seen after acquiring the resource, and wanting to do a simple shell test

self = Serial<id=0x7fa0bf247c18, open=False>(port='rfc2217://<ip erased>:<port erased>?ign_set_control', baudrate=115200, bytesize=8, parity='N', stopbits=1, timeout=None, xonxoff=False, rtscts=False, dsrdtr=False)

    def _reconfigure_port(self):
        """Set communication parameters on opened port."""
        if self._socket is None:
            raise SerialException("Can only operate on open ports")
    
        # if self._timeout != 0 and self._interCharTimeout is not None:
            # XXX
    
        if self._write_timeout is not None:
            raise NotImplementedError('write_timeout is currently not supported')
            # XXX
    
        # Setup the connection
        # to get good performance, all parameter changes are sent first...
        if not 0 < self._baudrate < 2 ** 32:
            raise ValueError("invalid baudrate: {!r}".format(self._baudrate))
        self._rfc2217_port_settings['baudrate'].set(struct.pack(b'!I', self._baudrate))
        self._rfc2217_port_settings['datasize'].set(struct.pack(b'!B', self._bytesize))
        self._rfc2217_port_settings['parity'].set(struct.pack(b'!B', RFC2217_PARITY_MAP[self._parity]))
        self._rfc2217_port_settings['stopsize'].set(struct.pack(b'!B', RFC2217_STOPBIT_MAP[self._stopbits]))
    
        # and now wait until parameters are active
        items = self._rfc2217_port_settings.values()
        if self.logger:
            self.logger.debug("Negotiating settings: {}".format(items))
        timeout = Timeout(self._network_timeout)
        while not timeout.expired():
            time.sleep(0.05)    # prevent 100% CPU load
            if sum(o.active for o in items) == len(items):
                break
        else:
>           raise SerialException("Remote does not accept parameter change (RFC2217): {!r}".format(items))
E           serial.serialutil.SerialException: Remote does not accept parameter change (RFC2217): dict_values([baudrate:REQUESTED, datasize:REQUESTED, parity:REQUESTED, stopsize:REQUESTED])

/usr/local/lib/python3.7/dist-packages/serial/rfc2217.py:534: SerialException

Needs more investigations to why it is happening, and exactly what is going. Hopefully i will soon have more time to look into it, but unfortunately it wont happen now.

Leaving the PR open, in case someone has time to dig into what it is that fails to be setup.

Emantor:
By chance I'm just working on it. What is missing in this PR is that it only sets up a telnet connection, not a RFC2217 connection. For this `telnet(rfc2217)` is required. Also the `tcp` line can be dropped. I'll push an update soon.

Emantor:
I have pushed a new version, please test, this works on my local machine.

cbs101:
@Emantor : 
Getting further, but still having issues.

Error:
/usr/local/lib/python3.7/dist-packages/labgrid-0.2.1.dev477+gd1df2cd.d20200813-py3.7.egg/labgrid/target.py:425: in activate                                                                                                                                                                                                                                     
    client.on_activate()                                                                                                                                                                                                                              
/usr/local/lib/python3.7/dist-packages/labgrid-0.2.1.dev477+gd1df2cd.d20200813-py3.7.egg/labgrid/driver/serialdriver.py:65: in on_activate                                                                                                                                                                                                                                   
    self.open()                                                                                                                                                                                                                                       
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _                                   
                                                                                                                                                                                                                                                      
self = SerialDriver(target=Target(name='main', env=Environment(config_file='configuration/rpi2/remote.yaml')), name=None, state=<BindingState.bound: 1>, txdelay=0.0)                                                                                 
                                                                                                                                                                                                                                                      
    def open(self):                                                                                                                                                                                                                                   
        """Opens the serialport, does nothing if it is already closed"""                                                                                                                                                                              
        if not self.status:                                                                                                                                                                                                                           
            try:                                                                                                                                                                                                                                      
                self.serial.open()                                                                                                                                                                                                                    
            except serial.SerialException as e:                                                                                                                                                                                                       
                raise serial.SerialException(                                                                                                                                                                                                         
>                   "Could not open serial port {}".format(self.serial.port)) from e                                                                                                                                                                  
E               serial.serialutil.SerialException: Could not open serial port rfc2217://192.168.17.154:37035?ign_set_control                                                                                                                          
                                                                                                                                                                                                                                                      
/usr/local/lib/python3.7/dist-packages/labgrid-0.2.1.dev477+gd1df2cd.d20200813-py3.7.egg/labgrid/driver/serialdriver.py:100: SerialException

Emantor:
This looks like a permission issue, do you have the correct permissions to open the serial port as the exporter user?

cbs101:
I got both ser2net 2.10.0 and ser2net 4.1.8 on my target. Switching between them without touching the rest of the system.
When using ser2net 2.10.0 it does work, whereas with 4.1.8 it does not. So don't assume it's related to permission.
Unsure whether related to network.

Emantor:
Hm, I'm using ser2net 4.2.0 locally. Maybe there was a change in between minor revisions?

cbs101:
Found 4.2 in master meta-openembedded, so build that on put on target
I see the same problem as when using 4.1.8.

Hoping to get some time to look a bit more in depth to why it is happening.

Emantor:
This might also depend on the version of gensio, the underlying stream IO library, running 2.1.3 here.

cbs101:
Very true. I'm using 1.5.3.
I'll try to update and re-test.

Thanks a lot for the assistance! It is highly appreciated!

Emantor:
I'd like to get this in, we can address further ser2net issues in a separate PR.

Emantor (dismissed):

Emantor (approved):

PR #635, opened 2020-09-08
ExternalPowerDriver: split into Port and Driver

leiflm:
To support the export of an ExternalPowerDriver as a resource,
split the declaration into Driver and Port.

Signed-off-by: Leif Middelschulte <leif.middelschulte@klsmartin.com>

**Description**
Allow `ExternalPowerDriver` to be exported as a resource, like `PDUDaemonPort`.

**Checklist**
- [x] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -61,7 +61,7 @@ Breaking changes in 0.3.0
   A deprecated `NetworkUSBStorageDriver` exists temporarily for compatibility
   reasons.
 - `ExternalPowerDriver` is now split into a Port and a Driver, like PDUDaemon.
-  Configuration files need to be updated to declare the `cmd_*` as a resource.
+  Configuration files need to be updated to declare the ``cmd_*`` as a resource.
 
 Release 0.2.0 (released Jan 4, 2019)
 ------------------------------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -36,6 +36,8 @@ New Features in 0.3.0
   ``--skip``, ``--seek``.
 - UBootDriver now allows overriding of default boot command (``run bootcmd``)
   via new ``boot_command`` argument.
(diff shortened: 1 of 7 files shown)


leiflm:
I'm not sure whether this should be a resource or rather be implemented using the `Agent` mechanism.

jluebbe:
The ExternalPowerDriver was mainly intended as a quick alternative for local setups, not for use with exporters. The resource doesn't really correspond to an interface.
I'd very much prefer if we could use the PDUDaemon support for that (see https://github.com/pdudaemon/pdudaemon/pull/78). There shouldn't too much work left to get that merged, but I've been unable to find some spare cycles for that. @mattface seems busy as well. :)

jluebbe:
This would break existing users of the ExternalPowerDriver, right?

leiflm:
> This would break existing users of the ExternalPowerDriver, right?

Yes, it would. I [mentioned](https://github.com/labgrid-project/labgrid/pull/635/files#diff-db23dcd814354c954091a9b90dbfd92aR63) it in the *Breaking changes in 0.3.0*-section accordingly.

Emantor:
What kind of resource am I locking in this case? I'd much rather add proper support via a power driver than making the ExternalPowerDriver remote capable.

leiflm:
> What kind of resource am I locking in this case? 

You'd lock a power supply that is linked to a place. But this is not really about locking, but about the process of announcement, acquiring and using a *Place*.

> I'd much rather add proper support via a power driver

[Done](https://github.com/pdudaemon/pdudaemon/pull/85), a while ago.

> than making the ExternalPowerDriver remote capable.

Fair enough. This was just about declaring obscure/ad-hoc setups (devices and their resources) via a single `labgrid-exporter`.

So if you feel uncomfortable with this change, close it.

PR #642, opened 2020-10-09
Update pytest and pytest-* to current versions

bgermann:
Pytest 6 was released recently with a lot of improvements and bug fixes.
Use the latest pytest version and the latest pytest plugin versions which
might have changes to be compatible with the 6.x versions.

**Description**

**Checklist**

- [x] PR has been tested

### DIFF
--- a/labgrid/pytestplugin/fixtures.py
+++ b/labgrid/pytestplugin/fixtures.py
@@ -50,7 +50,7 @@ def pytest_addoption(parser):
 
 
 @pytest.fixture(scope="session")
-def env(request):
+def env(request, record_testsuite_property):
     """Return the environment configured in the supplied configuration file.
     It contains the targets contained in the configuration file.
     """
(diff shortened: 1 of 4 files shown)


Emantor:
From the coverage diff it looks like the `junitxml` usage we currently use is no longer valid. This should be fixed before moving to pytest 6.x, since this outputs a jenkins compatible XML.

bgermann:
I think this should be fixed.

labgrid/pytestplugin/fixtures.py
```diff
@@ -60,49 +60,47 @@ def env(request):
         pytest.skip("missing environment config (use --lg-env)")
 
     if request.config.pluginmanager.hasplugin('junitxml'):
```
Emantor, line 62: According to the source code at [pytest](https://docs.pytest.org/en/stable/_modules/_pytest/junitxml.html#record_testsuite_property), this should no longer be necessary, the fixture will return a no-op function in this case.

Emantor (changes requested):
This looks good, thank you!
With the removal mentioned in the comments and a bump of the minimal pytest version in `setup.py` this should be good to go.

Emantor (approved):

PR #646, opened 2020-10-20
serialdriver: Prevent pyserial version check fails

bgermann:
The version might contain 'b' in a component, e.g., with the newest beta
version 3.5b0. This makes the int() conversion fail. Switch to packaging.version comparison.

**Description**

**Checklist**
- [x] PR has been tested

### DIFF
--- a/labgrid/driver/serialdriver.py
+++ b/labgrid/driver/serialdriver.py
@@ -2,6 +2,7 @@ import logging
 import warnings
 
 import attr
+from packaging import version
 from pexpect import TIMEOUT
 import serial
 import serial.rfc2217
@@ -22,11 +23,11 @@ class SerialDriver(ConsoleExpectMixin, Driver, ConsoleProtocol):
     """
     # pyserial 3.2.1 does not support RFC2217 under Python 3
     # https://github.com/pyserial/pyserial/pull/183
-    if tuple(int(x) for x in serial.__version__.split('.')) <= (3, 2, 1):
+    if version.parse(serial.__version__) <= version.Version('3.2.1'):
         bindings = {"port": "SerialPort", }
     else:
         bindings = {"port": {"SerialPort", "NetworkSerialPort"}, }
-    if tuple(int(x) for x in serial.__version__.split('.')) < (3, 4, 0, 1):
+    if version.parse(serial.__version__) != version.Version('3.4.0.1'):
         message = ("The installed pyserial version does not contain important RFC2217 fixes.\n"
                    "You can install the labgrid fork via:\n"
                    "pip uninstall pyserial\n"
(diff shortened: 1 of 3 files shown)


labgrid/driver/serialdriver.py
```diff
@@ -22,11 +22,11 @@ class SerialDriver(ConsoleExpectMixin, Driver, ConsoleProtocol):
     """
     # pyserial 3.2.1 does not support RFC2217 under Python 3
     # https://github.com/pyserial/pyserial/pull/183
-    if tuple(int(x) for x in serial.__version__.split('.')) <= (3, 2, 1):
+    if tuple(x for x in serial.__version__.split('.')) <= ('3', '2', '1'):
         bindings = {"port": "SerialPort", }
     else:
         bindings = {"port": {"SerialPort", "NetworkSerialPort"}, }
-    if tuple(int(x) for x in serial.__version__.split('.')) < (3, 4, 0, 1):
+    if tuple(x for x in serial.__version__.split('.')) < ('3', '4', '0', '1'):
```
Emantor, line 29:
```suggestion
    if version(serial.__version__) != version('3.4.0.1'):
```

labgrid/driver/serialdriver.py
```diff
@@ -22,11 +22,11 @@ class SerialDriver(ConsoleExpectMixin, Driver, ConsoleProtocol):
     """
     # pyserial 3.2.1 does not support RFC2217 under Python 3
     # https://github.com/pyserial/pyserial/pull/183
-    if tuple(int(x) for x in serial.__version__.split('.')) <= (3, 2, 1):
+    if tuple(x for x in serial.__version__.split('.')) <= ('3', '2', '1'):
```
Emantor, line 25: We should switch this over to use `version` from the `packaging` module. We require setuptools to run setup.py anyway and it includes the packaging module.
```suggestion
    if version(serial.__version__) <= version('3.2.1'):
```

Emantor (changes requested):
Thanks, just some improvements and a fix for the version comparison (we want users to use our `3.4.0.1` version, but `3.5.0b` is "bigger"). Also the `from packaging import version` line needs to be added at the top.

labgrid/driver/serialdriver.py
bgermann: setuptools is a buildtime dependency but this makes packaging a runtime dependency, so I specified it as one.

Emantor (approved):

labgrid/driver/serialdriver.py
Emantor: Yes, I agree its better to be explicit here. Considering that we advocate for an installation from git we require setuptools anyway, but I hope to eventually make the pypi installation route for pytest usage available again.

Issue #620, opened 2020-06-12
Wrong SSH forward for proxied labgrid-client when connecting to isolated exporter


Bastian-Krause:
Trying to connect to a NetworkSerialPort exported by an isolated exporter (`labgrid-exporter -i`) via proxied labgrid-client.
I've added a `print(complete_cmd)` statement in [SSHConnection._run_socket_command()](https://github.com/labgrid-project/labgrid/blob/d11a5ecc86decd74d3ff86b6fb8b730c9ea4a962/labgrid/util/ssh.py#L169) right before `subprocess.check_call()` to help debugging the forwards.
```
$ labgrid-client -vvd -x ws://labgrid/ws -P proxyhost con
  DEBUG: Creating SSHConnection for proxyhost
  DEBUG: Found existing control socket
   INFO: Using existing SSH connection to proxyhost
  DEBUG: Started keepalive for proxyhost
['ssh', '-x', '-o', 'LogLevel=ERROR', '-o', 'PasswordAuthentication=no', '-O', 'forward', '-L34565:labgrid:20408', 'proxyhost']
  DEBUG: expanded remote resources for place simulator: [NetworkPowerPort(target=Target(name='simulator', env=None), name='NetworkPowerPort', state=<BindingState.bound: 1>, avail=True, model='gude', host='gude7', index='5'), NetworkSerialPort(target=Target(name='simulator', env=None), name='USBSerialPort', state=<BindingState.bound: 1>, avail=True, host='exporter', port=48395, speed=115200, protocol='rfc2217'), ...]
['ssh', '-x', '-o', 'LogLevel=ERROR', '-o', 'PasswordAuthentication=no', '-O', 'forward', '-L37649:exporter:48395', 'proxyhost']
  DEBUG: Creating SSHConnection for exporter
  DEBUG: Found existing control socket
   INFO: Using existing SSH connection to exporter
  DEBUG: Started keepalive for exporter
['ssh', '-x', '-o', 'LogLevel=ERROR', '-o', 'PasswordAuthentication=no', '-O', 'forward', '-L59887:localhost:37649', 'exporter']
connecting to NetworkSerialPort(target=Target(name='simulator', env=None), name='USBSerialPort', state=<BindingState.bound: 1>, avail=True, host='exporter', port=48395, speed=115200, protocol='rfc2217') calling microcom -s 115200 -t localhost:59887
connected to ::1 (port 59887)
Escape character: Ctrl-\
Type the escape character to get to the prompt.
[...]

PR #647, opened 2020-10-26
util/proxy: if LG_PROXY set, use as only forward

Emantor:
**Description**
If the LG_PROXY Variable is set, we expect to be able to reach all of
the infrastructure using this variable. Do not request a second forward
even if the resource is from an isolated exporter.

**Checklist**
- [ ] Tests for the feature 
- [ ] PR has been tested

Fixes https://github.com/labgrid-project/labgrid/issues/620

### DIFF
--- a/labgrid/util/proxy.py
+++ b/labgrid/util/proxy.py
@@ -51,6 +51,7 @@ class ProxyManager:
         if cls._force_proxy:
             port = sshmanager.request_forward(cls._force_proxy, host, port)
             host = 'localhost'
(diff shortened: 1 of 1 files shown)


Emantor:
@jluebbe mentioned that another way is to require that users have setup correct `.ssh/config` entries for each isolated exporter, in this case an isolated exporter would supersede a setup `LG_PROXY` environment variable. This needs to be documented and than enforced in the proxymanager.

PR #650, opened 2020-10-28
driver/usbstoragedriver: log what actually gets written

agners:
Add log entires just before actual writing is executed. This is useful
when checking logs to figure out what exaclty got written.

Signed-off-by: Stefan Agner <stefan@agner.ch>

**Checklist**
- [ ] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/labgrid/driver/usbstoragedriver.py
+++ b/labgrid/driver/usbstoragedriver.py
@@ -82,13 +82,16 @@ class USBStorageDriver(Driver):
             raise ExecutionError("Timeout while waiting for medium")
 
         partition = "" if partition is None else partition
+        remote_path = mf.get_remote_path()
+        target = "{}{}".format(self.storage.path, partition)
 
         if mode == Mode.DD:
+            self.logger.info('Writing %s to %s using dd.', remote_path, target)
             block_size = '512' if skip or seek else '4M'
             args = [
                 "dd",
-                "if={}".format(mf.get_remote_path()),
-                "of={}{}".format(self.storage.path, partition),
+                "if={}".format(remote_path),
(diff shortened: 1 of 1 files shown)


labgrid/driver/usbstoragedriver.py
```diff
@@ -82,13 +82,16 @@ def write_image(self, filename=None, mode=Mode.DD, partition=None, skip=0, seek=
             raise ExecutionError("Timeout while waiting for medium")
 
         partition = "" if partition is None else partition
+        remote_path = mf.get_remote_path()
+        target = "{}{}".format(self.storage.path, partition)
 
         if mode == Mode.DD:
+            self.logger.info('Writing %s to %s using dd.', remote_path, target)
```
Emantor, line 89: This is useful if the path is on the local computer, but causes confusion if the file is copied over the network.
The solution here is to use the information already contained in the processwrapper below:
The processwrapper has a step decorator, but currently we have no infrastructure to turn Step Events into logging events. I have a work in progress branch which will make many similar issues visible in the logging output.

agners: We could also just log the file name?

By logging on processwrapper level I guess you would log the whole command executed? But then it would also show the local path when working over network, wouln't it?

IMHO that the local path is shown is also not that big of deal. Could be even helpful for debugging.

Emantor: Since I want the used loggers for the new Logging Reporter to be configurable anyway, this logging info won't hurt.

Emantor (approved):

Emantor:
Thanks!

PR #652, opened 2020-10-29
binding: check resources in check_active

Emantor:
**Description**
Ensure that not only the driver has been activated but that all bound
suppliers are available as well. This guards against cases where the
driver was activated at some point, but the resource is no longer
available when driver functions are used at a later point.

**Checklist**
- [x] Tests for the feature 
- [x] PR has been tested


agners (approved):
LGTM

Emantor:
This is essentially useless until we also call `update_resources()` on the target to poll the current status.

jluebbe:
And calling `update_resources()` every `check_active()` would be much too expensive.

PR #653, opened 2020-10-29
util/helper: use low level read for PTY

agners:
Executing a very shored lived process (e.g. a cat of a /sys file) using
check_output sometimes does not return the output the process returned.
Debugging into it showed that the process indeed returned the expected
value, but the PTY file descriptor also returned a EIO error:
  read(17, "131072\r\n", 4096)            = 8
  read(17, 0x55bd3aff9750, 4096)          = -1 EIO (Input/output error)

The PTY file descriptor is passed to os.fdopen() which returns a
io.BufferedReader instance. It turns out that io.BufferedReader is not
suitable to handle PTY file descriptors. It handles the EIO error as an
actual error and swallows the previously read bytes. The Python Issue
5380 has more details: https://bugs.python.org/issue5380.

Use low-level I/O operation so we can properly handle -EIO.

Signed-off-by: Stefan Agner <stefan@agner.ch>

**Checklist**
- [ ] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/labgrid/util/helper.py
+++ b/labgrid/util/helper.py
@@ -41,19 +41,19 @@ class ProcessWrapper:
         # close sfd so we notice when the child is gone
         os.close(sfd)
         # get a file object from the fd
-        omfd = os.fdopen(mfd, 'rb')
         buf = b""
         while True:
             try:
-                raw = omfd.read(4096)
+                raw = os.read(mfd, 4096)
+            except BlockingIOError as ex:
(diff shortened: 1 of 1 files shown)


Emantor:
Nice find. A test for this would be great, however this may be hard to reproduce in a test.

Emantor (approved):
Tested and works for me.

PR #656, opened 2020-10-30
CI Support for python 3.9

Emantor:
**Description**
Add CI for python 3.9 and adjust the classifiers in setup.py

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,5 +52,7 @@ matrix:
           fi
     - stage: optional
       python: "nightly"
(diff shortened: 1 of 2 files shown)


jluebbe (approved):

Emantor:
Looks like numpy is not python 3.9 ready yet, but used by crossbar.

Emantor:
Moved `python3.9` to optional. I'd keep the classifier since library usage is not affected.

jluebbe (approved):

PR #661, opened 2020-11-04
remote/client: remove unused variables from _get_ssh()

Bastian-Krause:
**Description**
These are leftovers from the method prior to 77ee439
("driver/sshdriver: move interactive SSH access to the SSHDriver").

The instanciation of NetworkService is still needed to bind a
NetworkService with an IP of an EthernetPort resource to the target.

**Checklist**
- [x] PR has been tested

### DIFF
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -966,19 +966,15 @@ class ClientSession(ApplicationSession):
     def _get_ssh(self):
         place = self.get_acquired_place()
         target = self._get_target(place)
-        env = os.environ.copy()
 
         from ..resource import NetworkService
         try:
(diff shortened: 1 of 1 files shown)


Emantor:
AFAICS this code tries to resolve the IP address of the place by using an `EthernetPort` which is exported via SNMP and translating it to a `NetworkService` for the `SSHDriver` to use. Or am I reading the code wrong?

Bastian-Krause:
You are right, the instanciation of NetworkService is still needed to bind a
NetworkService with an IP of an EthernetPort resource to the target. I've adjusted the PR accordingly.

Emantor (approved):

Issue #660, opened 2020-11-04
Various client sub commands return 0 on command error


Bastian-Krause:
The return code of the executed commands `ssh`, `scp`, `rsync`, `telnet`, `gst-launch-1.0` is lost:
```
bst@erbse$ labgrid-client -c env.yaml ssh false
Selected role main from configuration file
Shared connection to 192.168.10.75 closed.
connection lost (remote exit code 1)
bst@erbse$ echo $?
0
```

```
bst@erbse$ labgrid-client -c conf.yaml scp foo :bar
Selected role main from configuration file
   INFO: Connected to 192.168.23.39
   INFO: Running command: ['scp', '-o', 'ControlPath=/tmp/labgrid-ssh-tmp-ag484isf/control-192.168.23.39', 'foo', '_:bar']
foo: No such file or directory
bst@erbse$ echo $?
0
```

```
bst@erbse$ labgrid-client -c conf.yaml rsync foo :bar
Selected role main from configuration file
   INFO: Connected to 192.168.23.39
   INFO: Running command: ['rsync', '-v', '--rsh=ssh -o ControlPath=/tmp/labgrid-ssh-tmp-2vhi3t7a/control-192.168.23.39', '-rlpt', '--one-file-system', '--progress', 'foo', '_:bar']
sh: rsync: not found
rsync: connection unexpectedly closed (0 bytes received so far) [sender]
rsync error: remote command not found (code 127) at io.c(235) [sender=3.1.3]
bst@erbse$ echo $?
0
```

```
bst@erbse$ labgrid-client -c conf.yaml telnet
Selected role main from configuration file
Trying 192.168.23.39...
telnet: Unable to connect to remote host: Connection refused
connection lost
bst@erbse$ echo $?
0
```

```
bst@erbse$ labgrid-client video
WARNING: erroneous pipeline: no element "vaapipostproc"
bst@erbse$ echo $?
0
```

```
bst@erbse$ labgrid-client audio
WARNING: erroneous pipeline: no element "alsasrc"
bst@erbse$ echo $?
0
```

Emantor:
Please try #662.

Bastian-Krause:
Every subcommand method that does not raise an exception on error is affected by this. Other than scp/rsync, at least console.

Bastian-Krause:
Added ssh remote execution case.

Bastian-Krause:
See discussion in #662 for a way forward to fix this.

PR #662, opened 2020-11-04
remote/client: return code for sshfs, rsync & scp

Emantor:
**Description**
Correctly return the returncode from sshfs, rsync & scp calls.

**Checklist**
- [ ] PR has been tested

Fixes https://github.com/labgrid-project/labgrid/issues/660

### DIFF
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -43,6 +43,10 @@ class ServerError(Error):
     pass
 
 
+class InteractiveCommandError(Error):
+    pass
+
+
 class ClientSession(ApplicationSession):
     """The ClientSession encapsulates all the actions a Client can Invoke on
     the coordinator."""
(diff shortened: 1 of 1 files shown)


Bastian-Krause (changes requested):
This does nothing.

The subcommand methods are called in [labgrid/remote/client.py:1602](https://github.com/labgrid-project/labgrid/blob/1216385455c0e324cd07f3ad481aa1bc33432ee6/labgrid/remote/client.py#L1602) and in [labgrid/remote/client.py:1604](https://github.com/labgrid-project/labgrid/blob/1216385455c0e324cd07f3ad481aa1bc33432ee6/labgrid/remote/client.py#L1604). Their return value is lost there.

I had a quick look through the affected methods, most do not return anything at all, some return True/False. We could set `exitcode` if it is an integer or False there. But I am not sure that this is the desired way though. Maybe better raise an exception in the subcommand methods directly and rely on the client's main() function to set `exitcode` (like most of the other methods do?

[](url)

Emantor:
> I had a quick look through the affected methods, most do not return anything at all, some return True/False. We could set `exitcode` if it is an integer or False there. But I am not sure that this is the desired way though. Maybe better raise an exception in the subcommand methods directly and rely on the client's main() function to set `exitcode` (like most of the other methods do)?

Yes, raising exception lets us output specific error messages for the user for each exception, which improves the usability of the client command.

Bastian-Krause:
@Emantor Why did you close this? The latest state of the patches look good to me. I have a WIP branch on top of this to fix all issues from #660.

PR #663, opened 2020-11-08
docs: improve library usage docs

rohieb:
The concept of "activation" is mentioned a few times in the docs, but it is explained nowhere. Try to change that by distilling the code from `labgrid.target.Target.activate()` into an explanation with an example. Also document how to create multiple resources of the same type and bind them to drivers through the API. While at it, add a few more links to the API doc, and throw in some smaller drive-by improvements to the docs.

### DIFF
--- a/doc/usage.rst
+++ b/doc/usage.rst
@@ -229,8 +229,8 @@ Active drivers can be accessed by class (any :any:`Driver <labgrid.driver>` or
   FakeConsoleDriver(target=Target(name='main', …), name='console', …)
 
 After you are done with the target, optionally call the cleanup method on your
-target. While labgrid registers an atexit handler to cleanup targets, this has
-the advantage that exceptions can be handled by your application:::
+target. While labgrid registers an ``atexit`` handler to cleanup targets, this has
+the advantage that exceptions can be handled by your application::
 
   >>> try:
   >>>     target.cleanup()
--- a/doc/usage.rst
+++ b/doc/usage.rst
@@ -204,7 +204,7 @@ Then, a :any:`Driver` needs to be created on the `Target`::
   >>> sd = SerialDriver(t, name=None)
 
 
-As the `SerialDriver` declares a binding to a SerialPort, the target binds it
+As the :any:`SerialDriver` declares a binding to a :any:`SerialPort`, the target binds it
 to the resource created above::
 
   >>> sd.port
@@ -217,8 +217,8 @@ Before the driver can be used, it needs to be activated::
   >>> t.activate(sd)
   >>> sd.write(b'test')
 
-Active drivers can be accessed by class (any `Driver` or `Protocol`) using some
-syntactic sugar::
+Active drivers can be accessed by class (any :any:`Driver <labgrid.driver>` or
+:any:`Protocol <labgrid.protocol>`) using some syntactic sugar::
 
   >>> target = Target('main')
   >>> console = FakeConsoleDriver(target, 'console')
--- a/doc/usage.rst
+++ b/doc/usage.rst
@@ -171,8 +171,6 @@ Library
 labgrid can be used directly as a Python library, without the infrastructure
 provided by the pytest plugin.
 
-Creating and Configuring Targets
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 The labgrid library provides two ways to configure targets with resources and
 drivers: either create the :any:`Target` directly or use :any:`Environment` to
 load a configuration file.
@@ -182,7 +180,7 @@ load a configuration file.
    targets using the python atexit module.
 
 Targets
-^^^^^^^
+~~~~~~~
 At the lower level, a :any:`Target` can be created directly::
 
   >>> from labgrid import Target
@@ -238,7 +236,7 @@ the advantage that exceptions can be handled by your application::
   >>>     <your code here>
 
 Environments
-^^^^^^^^^^^^
+~~~~~~~~~~~~
 In practice, is is often useful to separate the `Target` configuration from the
 code which needs to control the board (such as a test case or installation
 script).
--- a/doc/usage.rst
+++ b/doc/usage.rst
@@ -181,6 +181,13 @@ load a configuration file.
 
 Targets
 ~~~~~~~
+
+.. note::
+   In most cases it is easier to :ref:`use a complete environment from a YAML
+   file <usage_environments>` instead of manually creating and activating objects.
+   Nevertheless, we explain this in the following to clarify the underlying concepts,
+   and how to work with targets on a lower level, e.g. in strategies.
+
 At the lower level, a :any:`Target` can be created directly::
 
   >>> from labgrid import Target
@@ -284,6 +291,8 @@ the advantage that exceptions can be handled by your application::
   >>>     target.cleanup()
   >>> except Exception as e:
   >>>     <your code here>
+
+.. _usage_environments:
 
 Environments
 ~~~~~~~~~~~~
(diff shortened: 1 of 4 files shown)


rohieb:
What's still unclear to me is when a driver needs to be deactivated. Some strategies do this, but I cannot see why.

Emantor:
> What's still unclear to me is when a driver needs to be deactivated. Some strategies do this, but I cannot see why.

Drivers need to be deactivated in the following cases:
- One of their suppliers is required by another driver which is to be activated, this logic can be found in `target.py`:
https://github.com/labgrid-project/labgrid/blob/0c5a45daf973805d703a2c61aee8789d83358a85/labgrid/target.py#L424-L428
This allows i.e. the `ShellDriver` to be activated even if the `BareboxDriver` is currently activated. Labgrid will deactivate the `BareboxDriver` in this case.
- Some drivers setup persistent objects which are only valid for as long as the target is booted, i.e. the `SSHDriver` will fail if it's not deactivated and reactivated when the target is rebooted, since the control socket master connection is gone.


Drivers additionally require to be activated for most user accessible functions, to prevent usage errors, this is done by the `@driver.check_active` decorator:
https://github.com/labgrid-project/labgrid/blob/0c5a45daf973805d703a2c61aee8789d83358a85/labgrid/driver/shelldriver.py#L379-L382

rohieb:
OK, I've added a short section on driver deactivation now. I think the last commit explaining the `set_binding_map` hack is not really necessary if we plan to extend the Driver constructor with a binding map.

doc/usage.rst
```diff
@@ -204,21 +207,34 @@ Then, a :any:`Driver` needs to be created on the `Target`::
...
+Before the driver can be used, it needs to be activated.
+On driver activation, labgrid makes sure that all resources needed by the
+driver are available::
 
   >>> t.activate(sd)
   >>> sd.write(b'test')
 
-Active drivers can be accessed by class (any `Driver` or `Protocol`) using some
-syntactic sugar::
+If a resource is not available (or not available after a certain timeout,
```
TheMeaningfulEngineer, line 227: Hi, am also trying to figure out the concepts and terminology behind labgrid. :)

As far as I understood, the word "resource" perhaps isn't the best choice of words for this context.
It seems there are two things "resources" might try to address:

* the ["Resource"](https://labgrid.readthedocs.io/en/latest/overview.html#resources) as introduced by labgrid.
* the "actual resource" to which the above is a proxy. I.e. `/dev/ttyUSB0`

From what I've been seeing it's not possible to even create a driver if the correct "Resource" isn't present on the target. This happens a step before activation and is a requirement of the binding phase. It would perhaps be better to use another term for the "actual resource" to describe what might happen in the activation phase.

In this example where I'm mostly trying to describe the implicit assumptions in the code I describe the "actual resource" as an "object below the driver which a driver consumes". It's not great wording either :) 

```
username = "pi"
password = "raspberry"                                                                                                                                                                    
                                                                                                                                                                                                         
rpi = Target("RPi")                                                                                                                                                                                                
                                                                                                                                                                                                           
network_service = NetworkService(rpi,                                                                                                                                                                      
                                 address="192.168.18.210",                                                                                                                                                 
                                 name="rpi_network",                                                                                                                                                       
                                 username=username,                                                                                                                             
                                 password=password)
raw_serial_port = RawSerialPort(rpi, None, port="/dev/ttyUSB0", speed=115200)                                                                                                                                      
# At this point both resources belong to the target. This happened implicitly.                                                                                                                                     
 
                                                                                                                                                                                                                   
ssh_driver = SSHDriver(rpi, name=None)                                                                                                                                                                             
serial_driver = SerialDriver(rpi, name=None)                                                                                                                                                                       
prompt_regex_pattern = f'{username}' + r'@\w+:[^ ]+ '                                                                                                                                                              
login_regex_pattern = ' login: '
shell_driver = ShellDriver(rpi, name=None,                                                                                                                                                                         
                           prompt=prompt_regex_pattern,                                                                                                                                                            
                           login_prompt=login_regex_pattern,                                                                                                                                                       
                           username=username,                                                                                                                                                                      
                           password=password)                                                                                                                                                                      
                                                                                                                                                                                                                   
# By belonging to the same target, the drivers automatically bounded to the resources and other drivers that fit their requirements.                                                                               
# The SSHDriver bounded to NetworkService
# The ShellDriver bounded to SerialDriver whicb bounded to RawSerialPort                                                                                                                                  
                                                                                                                                                                                                                   
rpi.activate(ssh_driver)
rpi.activate(shell_driver)                                                                                                                                                                                         
                                                                                                                                                                                                                   
# When activated the drivers consume the object below it (either the resource or the other driver).                                                                                                                
# After that they are ready to be used
```

rohieb: We could use "hardware resource" and "Resource object", would that be better?

doc/usage.rst
```diff
@@ -204,21 +207,34 @@ Then, a :any:`Driver` needs to be created on the `Target`::
...
   RawSerialPort(target=Target(name='example', env=None), name=None, state=<BindingState.bound: 1>, avail=True, port='/dev/ttyUSB0', speed=115200)
   >>> sd.port is rsp
   True
 
-Before the driver can be used, it needs to be activated::
+Driver Activation
+^^^^^^^^^^^^^^^^^
+Before the driver can be used, it needs to be activated.
+On driver activation, labgrid makes sure that all resources needed by the
+driver are available::
```
TheMeaningfulEngineer, line 222: > makes sure that all resources needed by the driver are available

What are you trying to say here?
Because a few sentences below there is refutal of `labgrid's ability to making sure...`.
FWIW here is a suggestion :)

```suggestion
Before the bounded driver can be used, it needs to be activated.
An activation is an act of the driver consuming the underlying resource on a system level. 

If we use a car analogy, binding is the process of screwing the car together from parts and activation is igniting the engine.

I.e. If a `SerialDriver` is bound to a `RawSerialPort` representing `/dev/ttyUSB0`. An external application could like `minicom` could consume the `/dev/ttyUSB0`. Once the driver is activated, that is no longer possible. ::
```

doc/usage.rst
TheMeaningfulEngineer: Don't have a clear one for that. :/
While writing the comment above a `resource on the level of the OS` seemed like a good candidate.

doc/usage.rst
rohieb: I like tha analogy!

doc/usage.rst
```diff
@@ -217,9 +217,12 @@ to the resource created above::
...
 ^^^^^^^^^^^^^^^^^
-Before the driver can be used, it needs to be activated.
-On driver activation, labgrid makes sure that all resources needed by the
-driver are available::
+Before the bounded driver can be used, it needs to be activated.
+An activation is an act of the driver consuming the underlying resource on the level of the OS. 
+
+If we use a car analogy, binding is the process of screwing the car together from parts and activation is igniting the engine.
+
+I.e. If a `SerialDriver` is bound to a `RawSerialPort` representing `/dev/ttyUSB0`. An external application could like `minicom` could consume the `/dev/ttyUSB0`. Once the driver is activated, that is no longer possible. ::
```
Emantor, line 225: `could like` reads wrong in the second sentence.

rohieb: Yes, I plan to polish the PR a bit more, also with the other suggestions above.

doc/usage.rst
rohieb: OK, I think I've got all of this in the current version now. Sometimes I just used "hardware" when it is clear what is meant.

Emantor (approved):

PR #664, opened 2020-11-10
driver/sshdriver: force no config when no hostname given

Bastian-Krause:
**Description**
The scp/rsync/sshfs methods use the SSH control master. In these cases no valid hostname is required. Certain SSH config option combinations break this behavior, e.g.:

```
Host *
    CanonicalDomains          example.com
    CanonicalizePermittedCNAMEs *.example:*.example.com
```

This leads to:

`Could not resolve hostname _: Name or service not known`

or

`Could not resolve hostname : Name or service not known`

Fix this by forcing no config file for these commands.

**Checklist**
- [x] PR has been tested

### DIFF
--- a/labgrid/driver/sshdriver.py
+++ b/labgrid/driver/sshdriver.py
@@ -232,6 +232,7 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
             dst = '_' + dst
 
         complete_cmd = ["scp",
+                "-F", "none",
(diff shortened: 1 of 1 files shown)


jluebbe:
According to ssh(1), `-F "none"` is the intended way to disable reading of config files.

Bastian-Krause:
Force-pushed "none" instead of "/dev/null". Also added another commit to do the same in on_activate().

jluebbe (approved):
Looks fine to me.

Emantor (approved):

PR #668, opened 2020-11-12
driver: add eaton epdu network power driver

b2vn:
The NetworkPowerDriver for the pdu uses PySNMP.

Signed-off-by: Nikolaj Rahbek <nikolaj@b2vn.org>

**Description**

Added support for Eaton ePDU via SMNP. The resource is implements the NetworkPowerPort, and therefore works like other power drivers, e.g. the apc power driver.

An usage example has been added under `examples/power/`, and includes both an environment yaml and a pytest example.

The driver has been tested on Eaton eMMA10.

**Checklist**
- [x] Documentation for the feature - as a usage example
- [x] Tests for the feature 
- [x] PR has been tested

### DIFF
--- a/examples/power/power_example.py
+++ b/examples/power/power_example.py
@@ -1,0 +1,18 @@
+import pytest
+
+
+@pytest.fixture()
+def pdu(target):
+    return target.get_driver('NetworkPowerDriver')
+
+
+def test_something(pdu):
+    pdu.on()
+    # Note that the pdu might not have switched on when requesting it's
+    # status immediately
+    assert pdu.get() is True
+
+    # ... do something
+
+    pdu.off()
+    assert pdu.get() is False
--- a/labgrid/driver/power/eaton.py
+++ b/labgrid/driver/power/eaton.py
@@ -1,0 +1,66 @@
+from pysnmp import hlapi
+from ..exception import ExecutionError
+
+
+OID = ".1.3.6.1.4.1.534.6.6.7.6.6.1"
+NUMBER_OF_OUTLETS = 16
+
+
+class _Snmp:
+    """A class that helps wrap pysnmp"""
+    def __init__(self, host, community, port=161):
+        if port is None:
+            port = 161
+
+        self.engine = hlapi.SnmpEngine()
+        self.transport = hlapi.UdpTransportTarget((host, port))
+        self.community = hlapi.CommunityData(community, mpModel=0)
+        self.context = hlapi.ContextData()
+
+    def get(self, oid):
+        g = hlapi.getCmd(self.engine, self.community, self.transport,
+            self.context, hlapi.ObjectType(hlapi.ObjectIdentity(oid)),
+            lookupMib=False)
+
+        error_indication, error_status, _, res = next(g)
+        if error_indication or error_status:
+            raise ExecutionError("Failed to get SNMP value")
+        return res[0][1]
+
+    def set(self, oid, value):
+        identify = hlapi.ObjectType(hlapi.ObjectIdentity(oid),
+                   hlapi.Integer(value))
+        g = hlapi.setCmd(self.engine, self.community, self.transport,
+            self.context, identify, lookupMib=False)
+        next(g)
+
+
+def power_set(host, port, index, value):
+    assert 1 <= int(index) <= NUMBER_OF_OUTLETS
+
+    _snmp = _Snmp(host, 'public', port=port)
+    cmd_id = 4 if int(value) else 3
+    outlet_control_oid = "{}.{}.0.{}".format(OID, cmd_id, index)
+
+    _snmp.set(outlet_control_oid, 1)
+
+
+def power_get(host, port, index):
+    assert 1 <= int(index) <= NUMBER_OF_OUTLETS
+
+    _snmp = _Snmp(host, 'public', port=port)
+    output_status_oid = "{}.2.0.{}".format(OID, index)
+
+    value = _snmp.get(output_status_oid)
+
+    if value == 1:  # On
+        return True
+    if value == 0:  # Off
+        return False
+
+    if value == 3:  # Pending on - treat as on
+        return True
+    if value == 2:  # Pending off - treat as off
+        return False
+
+    raise ExecutionError("failed to get SNMP value")
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,6 +3,7 @@ Release 0.5.0 (unreleased)
 
 New Features in 0.5.0
 ~~~~~~~~~~~~~~~~~~~~~
+- Support for Eaton ePDU added, and can be used as a NetworkPowerPort.
 
 Bug fixes in 0.5.0
 ~~~~~~~~~~~~~~~~~~
--- a/examples/power/env.yaml
+++ b/examples/power/env.yaml
@@ -1,0 +1,10 @@
+targets:
+  main:
+    resources:
+      NetworkPowerPort:
+        model: eaton
+        host: 1.2.3.4
+        index: 1
+    drivers:
+      NetworkPowerDriver:
+        delay: 0.5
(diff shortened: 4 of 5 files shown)


labgrid/driver/power/eaton.py
```diff
@@ -0,0 +1,65 @@
...
+
+    if(value == 1):  # On
+        return True
+    if(value == 0):  # Off
+        return False
+
+    if(value == 3):  # Pending on - treat as on
+        return True
+    if(value == 2):  # Pending off - treat as off
+        return False
```
Bastian-Krause, line 63: Unnecessary parens. Could be simplified:
```suggestion
    if value in (1, 3):  # on or pending on
        return True
    if value in (0, 2):  # off or pending off
        return False
```

Bastian-Krause (changes requested):

labgrid/driver/power/eaton.py
```diff
@@ -0,0 +1,65 @@
...
+        identify = hlapi.ObjectType(hlapi.ObjectIdentity(oid),
+                   hlapi.Integer(value))
+        g = hlapi.setCmd(self.engine, self.community, self.transport,
+            self.context, identify, lookupMib=False)
+        next(g)
+
+
+def power_set(host, port, index, value):
+    assert port is None
+    index = int(index)
```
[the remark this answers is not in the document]
b2vn: Kind of... the json have index as a string, e.g. "1". But the cast can be integrated in the `assert` on the next line.

labgrid/driver/power/eaton.py
b2vn: I decided against that solution. Writing it this was, it is explicit the an assumption was made.

labgrid/driver/power/eaton.py
```diff
@@ -0,0 +1,65 @@
...
+        self.community = hlapi.CommunityData(community, mpModel=0)
+        self.context = hlapi.ContextData()
+
+    def get(self, oid):
+        g = hlapi.getCmd(self.engine, self.community, self.transport,
+            self.context, hlapi.ObjectType(hlapi.ObjectIdentity(oid)),
+            lookupMib=False)
+
+        _,_,_, res = next(g)
+        return res[0][1]
```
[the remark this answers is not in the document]
b2vn: Good point. The example follows (some of)  the examples in their documentation. I have added some validation so at least it throws a more meaningful exception if this fails.

b2vn:
@Bastian-Krause will you take another look. Review comments implemented.

Emantor:
Please rebase on master, this should switch the CI over to Github Actions.

labgrid/driver/power/eaton.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+OID = ".1.3.6.1.4.1.534.6.6.7.6.6.1"
+NUMBER_OF_OUTLETS = 16
+
+
+class _Snmp:
+    """A class that helps wrap pysnmp"""
+    def __init__(self, host, community, port=161):
+        if(port is None):
```
Bastian-Krause, line 12: Please drop the parens.

labgrid/driver/power/eaton.py
```diff
@@ -0,0 +1,66 @@
...
+        self.community = hlapi.CommunityData(community, mpModel=0)
+        self.context = hlapi.ContextData()
+
+    def get(self, oid):
+        g = hlapi.getCmd(self.engine, self.community, self.transport,
+            self.context, hlapi.ObjectType(hlapi.ObjectIdentity(oid)),
+            lookupMib=False)
+
+        error_indication, error_status, _, res = next(g)
+        if(error_indication or error_status):
```
Bastian-Krause, line 26: Please drop the parens.

labgrid/driver/power/eaton.py
```diff
@@ -0,0 +1,65 @@
...
+
+        _,_,_, res = next(g)
+        return res[0][1]
+
+    def set(self, oid, value):
+        identify = hlapi.ObjectType(hlapi.ObjectIdentity(oid),
+                   hlapi.Integer(value))
+        g = hlapi.setCmd(self.engine, self.community, self.transport,
+            self.context, identify, lookupMib=False)
+        next(g)
```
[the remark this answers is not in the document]
Bastian-Krause: Will this fail now in case the set operation did not work? Am I missing something here?

labgrid/driver/power/eaton.py
```diff
@@ -0,0 +1,66 @@
...
+
+    if(value == 1):  # On
+        return True
+    if(value == 0):  # Off
+        return False
+
+    if(value == 3):  # Pending on - treat as on
+        return True
+    if(value == 2):  # Pending off - treat as off
+        return False
```
Bastian-Krause, line 64: Please drop the parens.

Bastian-Krause (changes requested):

labgrid/driver/power/eaton.py
b2vn: The setCmd usage is done as in the example in pysnmp documentation. But you have a point. According to the documentation setCmd raise on error, but it can return an error code. I have added the same check on the return code as in the get() method.

labgrid/driver/power/eaton.py
Bastian-Krause: The parens are still there.

labgrid/driver/power/eaton.py
Bastian-Krause: The parens are still there.

examples/power/power_example.py
```diff
@@ -0,0 +1,18 @@
...
+@pytest.fixture()
+def pdu(target):
+    return target.get_driver('NetworkPowerDriver')
+
+
+def test_something(pdu):
+    pdu.on()
+    # Note that the pdu might not have switched on when requesting it's
+    # status immediately
+    assert pdu.get() == True
```
Bastian-Krause, line 13: Maybe use `is` instead of `==` here?

examples/power/power_example.py
```diff
@@ -0,0 +1,18 @@
...
+
+    # ... do something
+
+    pdu.off()
+    assert pdu.get() == False
```
Bastian-Krause, line 18: Maybe use `is` instead of `==` here?

Bastian-Krause (changes requested):

labgrid/driver/power/eaton.py
```diff
@@ -0,0 +1,66 @@
+from pysnmp import hlapi
```
Emantor, line 1: This pysnmp dependency needs to be added to one of the requirements files.
However, I'd much rather have this switched over to the snmpwalk binary, instead of using the python module.

Emantor (changes requested):

labgrid/driver/power/eaton.py
jluebbe: @Emantor You probably mean `snmpget`/`snmpset` as used in https://github.com/labgrid-project/labgrid/blob/master/labgrid/driver/power/sentry.py.
If the driver works well as is and because we already have https://github.com/labgrid-project/labgrid/blob/master/snmp-requirements.txt, I wouldn't insist on switching to calling the external binaries.
@b2vn Did you test it with the version in the `snmp-requirements.txt`?

Emantor: Ah, nevermind. I was under the impression that we did not have a pysnmp dependency already, but this was added with EthernetPort support (I always forget about that). In this case this is fine and should work.

Emantor (dismissed):

Bastian-Krause:
Sorry for letting you wait so long. I've rebased this on current master and moved the `CHANGES.rst` entry to the upcoming 0.5.0 release section.

Bastian-Krause (approved):

PR #669, opened 2020-11-13
SSHDriver cleanup

Emantor:
**Description**
- Remove support for user control master connections
- Document the `cleanup` function for targets in the library case
- Minor adjustments to `argument` printing

**Checklist**
- [x] PR has been tested

### DIFF
--- a/doc/usage.rst
+++ b/doc/usage.rst
@@ -177,6 +177,10 @@ The labgrid library provides two ways to configure targets with resources and
 drivers: either create the :any:`Target` directly or use :any:`Environment` to
 load a configuration file.
 
+.. note::
+   On exit of your script/application, labgrid will call ``cleanup()`` on the
+   targets using the python atexit module.
+
 Targets
 ^^^^^^^
 At the lower level, a :any:`Target` can be created directly::
@@ -223,6 +227,15 @@ syntactic sugar::
   FakeConsoleDriver(target=Target(name='main', …), name='console', …)
   >>> target[FakeConsoleDriver, 'console']
   FakeConsoleDriver(target=Target(name='main', …), name='console', …)
+
+After you are done with the target, optionally call the cleanup method on your
+target. While labgrid registers an atexit handler to cleanup targets, this has
+the advantage that exceptions can be handled by your application:::
+
+  >>> try:
+  >>>     target.cleanup()
+  >>> except Exception as e:
+  >>>     <your code here>
 
 Environments
 ^^^^^^^^^^^^
--- a/labgrid/target.py
+++ b/labgrid/target.py
@@ -1,4 +1,5 @@
 import abc
+import atexit
 import logging
 from time import monotonic, sleep
 from collections import Counter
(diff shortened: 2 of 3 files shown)


doc/usage.rst
```diff
@@ -177,6 +177,12 @@ The labgrid library provides two ways to configure targets with resources and
 drivers: either create the :any:`Target` directly or use :any:`Environment` to
 load a configuration file.
 
+.. warning::
+   After your script/application is done using labgrid as a library, ensure to
+   call the ``cleanup()`` function all created targets. This will deactivate all
```
jluebbe, line 182: Maybe it would be better to use the [atexit module](https://docs.python.org/3/library/atexit.html) in the Target to trigger this automatically? We'd need to make sure to handle any exceptions gracefully.

doc/usage.rst
```diff
@@ -224,6 +230,12 @@ syntactic sugar::
   >>> target[FakeConsoleDriver, 'console']
   FakeConsoleDriver(target=Target(name='main', …), name='console', …)
 
+After you are done with the target, call the cleanup function to ensure
+drivers are deactivated and processes are stopped:::
+
+  >>> target.clean()
```
jluebbe, line 236: The function name is cleanup(). If we go the atexit approach, this would be optional but could catch some errors earlier.

jluebbe (changes requested):

doc/usage.rst
Emantor: Neat, I'll add that for the target. Handling this should IMO be done by catching the exception, mentioning that an error happened and directing the user to call `cleanup()` himself to get the full exception and investigate the problem.

doc/usage.rst
Emantor: I'll reword this to conform to this use case.

doc/usage.rst
```diff
@@ -224,6 +228,15 @@ syntactic sugar::
   >>> target[FakeConsoleDriver, 'console']
   FakeConsoleDriver(target=Target(name='main', …), name='console', …)
 
+After you are done with the target, optionally call the cleanup function on your
```
Bastian-Krause, line 231:
```suggestion
After you are done with the target, optionally call the cleanup method on your
```

doc/usage.rst
```diff
@@ -224,6 +228,15 @@ syntactic sugar::
...
+target. While labgrid registers an atexit handler to cleanup targets, this has
+the advantage that exceptions can be handled by your application:::
+
+  >>> try:
+  >>>     target.cleanup()
+  >>> except Exception as e:
+  >>> <your code here>
```
Bastian-Krause, line 238:
```suggestion
  >>>     <your code here>
```

labgrid/target.py
```diff
@@ -466,6 +468,14 @@ def deactivate_all_drivers(self):
         for drv in reversed(self.drivers):
             self.deactivate(drv)
 
+    def _atexit_cleanup(self):
+        try:
+            self.cleanup()
+        except Exception as e:
+            print("An exception occured during cleanup, call the cleanup() "
+                  "function on targets yourself to explictly handle exceptions.")
```
Bastian-Krause, line 476:
```suggestion
                  "method on targets yourself to handle exceptions explicitly.")
```

Bastian-Krause (changes requested):

jluebbe (approved):

PR #670, opened 2020-11-20
remote/config: allow use of environment variables in exporter config

esben:
**Description**

This extends the jinja2 templating support in exporter configuration file with
support for using OS environment variables, allowing for more flexibility when
using templated configurations for multiple exporters.

Tested locally by using a templated configuration with the group name
specified using an OS environment variable.

**Checklist**
- [ ] Tests for the feature 
- [x] PR has been tested

### DIFF
--- a/labgrid/remote/config.py
+++ b/labgrid/remote/config.py
@@ -2,6 +2,7 @@ from pprint import pprint
 
 import attr
 import jinja2
+import os
 
 from ..util.yaml import load
 from ..exceptions import NoConfigFoundError
@@ -23,7 +24,7 @@ class ResourceConfig:
             raise NoConfigFoundError(
                 "{} could not be found".format(self.filename)
             )
-        rendered = template.render()
+        rendered = template.render(env=os.environ)
         pprint(('rendered', rendered))
         self.data = load(rendered)
         pprint(('loaded', self.data))
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -61,6 +61,8 @@ New Features in 0.3.0
 - ``labgrid-client`` now respects the ``LG_HOSTNAME`` and ``LG_USERNAME``
   environment variables to set the hostname and username when accessing
   resources.
+- Exporter configuration file ``exporter.yaml`` now allows use of environment
+  variables.
 
 Breaking changes in 0.3.0
 ~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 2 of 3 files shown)


esben:
I don't see where to add tests for the exporter.yaml format.  If it is needed, where should I implement it?

Emantor:
Out of curiosity: how are you using this feature? LGTM.

Emantor (approved):

esben:
I am using this to build in exporter configs into a container for balenaOS, using the (dynamic) device name in the exporter config, so that the name of the resources can be managed simply by naming the devices in balenaCloud.

jluebbe:
The full resource name already includes the exporter's hostname. Wouldn't that be enough for this use-case? How does `labgrid-client resources` look in your case?

labgrid/remote/config.py
```diff
@@ -23,7 +23,8 @@ def __attrs_post_init__(self):
             raise NoConfigFoundError(
                 "{} could not be found".format(self.filename)
             )
-        rendered = template.render()
+        import os
```
Bastian-Krause, line 26: Please move this import to the top of the file.

esben: > The full resource name already includes the exporter's hostname. Wouldn't that be enough for this use-case? How does `labgrid-client resources` look in your case?

Yes, you are of-course right.  No point in having the same string appear twice in the resource name.

I still believe the change is a worthy improvement.

esben:
> The full resource name already includes the exporter's hostname. Wouldn't that be enough for this use-case? How does `labgrid-client resources` look in your case?

Yes, you are of-course right.  No point in having the same string appear twice in the resource name.

I still believe the change is a worthy improvement.

Emantor:
> > The full resource name already includes the exporter's hostname. Wouldn't that be enough for this use-case? How does `labgrid-client resources` look in your case?
> 
> Yes, you are of-course right. No point in having the same string appear twice in the resource name.
> 
> I still believe the change is a worthy improvement.

Pardon me asking, but I'd like a usecase besides the hostname inclusion before we merge this. Since the exporter configuration supports YAML templating already, I don't really see a reason to include environment variables. Thanks.

esben:
> Pardon me asking, but I'd like a usecase besides the hostname inclusion before we merge this. Since the exporter configuration supports YAML templating already, I don't really see a reason to include environment variables. Thanks.

Well, this change is not an alternative to YAML templating.  It simply allows using environment variables in the existing Jinja2 template support. And the use-case is that systems (such as balenaCloud) that allows setting device specific environment variables, would clearly benefit from this by allowing a YAML template to be configurable from the existing environment variable configuration support.

I don't think I am adding something dramatically new, or taking the project in a new direction.  I am simply proposing a small improvement to the existing template support, in a way that is AFAIK pretty standard when using Jinja2 templating.

Emantor:
Merged manually. Thanks!

PR #673, opened 2020-11-23
smallubootdriver fixes

ynezz:
- [x] PR has been tested

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -977,7 +977,7 @@ Arguments:
     prompt
   - boot_expression (str): optional, regex to match the U-Boot start string
     defaults to "U-Boot 20\d+"
-  - login_timeout (str): optional, timeout for the password/login detection
+  - login_timeout (int): optional, timeout for the password/login detection
 
 BareboxDriver
 ~~~~~~~~~~~~~
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -975,6 +975,7 @@ Arguments:
   - prompt (regex): U-Boot prompt to match
   - init_commands (tuple): tuple of commands to execute after matching the
     prompt
+  - boot_secret (str): optional, secret used to unlock prompt
   - boot_expression (str): optional, regex to match the U-Boot start string
     defaults to "U-Boot 20\d+"
   - login_timeout (int): optional, timeout for the password/login detection
--- a/labgrid/driver/smallubootdriver.py
+++ b/labgrid/driver/smallubootdriver.py
@@ -39,10 +39,14 @@ class SmallUBootDriver(UBootDriver):
     This driver was created especially for the following devices:
 
     - TP-Link WR841 v11
+
+    Args:
+        boot_secret (str): optional, secret used to unlock prompt
+        login_timeout (int): optional, timeout for login prompt detection,
     """
 
     boot_secret = attr.ib(default="a", validator=attr.validators.instance_of(str))
-    login_timeout = attr.ib(default=60.0, validator=attr.validators.instance_of(float))
+    login_timeout = attr.ib(default=60, validator=attr.validators.instance_of(int))
 
     @step()
     def _await_prompt(self):


Emantor:
PR looks good.
On a side note, why do you use the `init_commands` here? Just to retrieve the IP address and setup U-Boot? That may be better handled by calling `run()` yourself and configuring the timeout for each individual command.

ynezz:
> On a side note, why do you use the `init_commands` here? Just to retrieve the IP address and setup U-Boot? That may be better handled by calling `run()` yourself and configuring the timeout for each individual command.

```yaml
      SmallUBootDriver:
        prompt: 'GRX500 #'
        boot_secret: ''
        boot_expression: 'Hit any key to stop autoboot'
        init_command_timeout: 60
        init_commands:
          - setenv bootdelay 0
          - setenv serverip 192.168.1.2
          - setenv ipaddr 192.168.1.1
          - setenv bootargs debug earlyprintk console=ttyLTQ0,115200
          - setenv fullimage NETGEAR_RAX40-squashfs-fullimage.img
          - run update_fullimage
          - run run_bootcore
          - ubi read $(loadaddr) $(kernel_vol)
          - run flashargs addmisc
      UBootStrategy: {}
```

Having all the device specific bits in the YAML file allows me to have single universal Python code.

Bastian-Krause:
This is not how the environment yaml is meant to be used. Your commands from `init_commands` belong in a custom Strategy.

I know that creating a custom strategy might be daunting, especially when you're duplicating most parts of an already existing Strategy. I'd like to change this by turning the Strategies shipped with labgrid (in `labgrid/strategy/`) into GraphStrategies (or into a new type of less complex linear Strategies with the same "one state per method" approach). This makes it possible to use inheritance. With this your custom Strategy could be as simple as:

```python
import attr
from labgrid.factory import target_factory
from labgrid.strategy import GraphStrategy, UBootStrategy

@target_factory.reg_driver
@attr.s(eq=False)
class RAX40Strategy(UBootStrategy):
    @GraphStrategy.depends("off")
    def state_uboot(self):
        super().state_uboot()
        self.uboot.run("setenv bootdelay 0")
        self.uboot.run("setenv serverip 192.168.1.2")
        self.uboot.run("setenv ipaddr 192.168.1.1")
        self.uboot.run("setenv bootargs debug earlyprintk console=ttyLTQ0,115200")
        self.uboot.run("setenv fullimage NETGEAR_RAX40-squashfs-fullimage.img")
        self.uboot.run("run update_fullimage")
        self.uboot.run("run run_bootcore")
        self.uboot.run("ubi read $(loadaddr) $(kernel_vol)")
        self.uboot.run("run flashargs addmisc")
```

ynezz:
Thanks for reviewing my PR.

> This is not how the environment yaml is meant to be used.

You know, vendors are crazy, so one never knows what is going to solve with the next device. Anyway, how the YAML environment should be used? You allow overriding of `login_timeout`, so why not other fixed timeouts? How is that `init_commands` timeout so special?

> With this your custom Strategy could be as simple as:

I'm probably using labgrid in quite different way then it was designed for, I agree. But that's the beauty of the FOSS, right? I really like the "everything device specific defined in YAML" approach, because YAML is more accessible then Python for newcomers, it's single point of truth etc.

My solution allows me to have everything in single YAML file. I've currently nothing device specific in the [Python code](https://gitlab.com/prpl-foundation/prplwrt/prplwrt/-/blob/prplwrt/.gitlab/scripts/testbed-device.py) (well, there are currently OS specific and testbed/network setup specific parts, but I plan to move those somehow into YAML file one day as well).

Your solution would need extra class for every device, people would need to edit/understand YAML _and_ Python as well.

Emantor:
So what you really want is per target options, see https://github.com/labgrid-project/labgrid/pull/490. With these you can add generic per target configuration in the YAML file. These could than be handled in your python files generically and with error handling, which we don't provide for the `init_commands`.
I'll get this into Labgrid on friday.

Bastian-Krause:
> Thanks for reviewing my PR.
> 
> > This is not how the environment yaml is meant to be used.
> 
> You know, vendors are crazy, so one never knows what is going to solve with the next device. Anyway, how the YAML environment should be used? You allow overriding of `login_timeout`, so why not other fixed timeouts? How is that `init_commands` timeout so special?

Oh, sorry, I did not look at your changes carefully enough. I thought you added `init_commands`, but that's already part of the SmallUBootDriver. I don't have a strong opinion on the timeout though.

> 
> > With this your custom Strategy could be as simple as:
> 
> I'm probably using labgrid in quite different way then it was designed for, I agree. But that's the beauty of the FOSS, right? I really like the "everything device specific defined in YAML" approach, because YAML is more accessible then Python for newcomers, it's single point of truth etc.

I do not agree with that. The Strategies shipped with labgrid cover only very basic use cases. Just look at the variety of drivers in labgrid vs. drivers used in these basic Strategies. As soon as you really want to interact with the board beyond booting, you need to dive into Python and labgrid.

> My solution allows me to have everything device specific in single YAML file. I've currently nothing device specific in the [Python code](https://gitlab.com/prpl-foundation/prplwrt/prplwrt/-/blob/prplwrt/.gitlab/scripts/testbed-device.py) (well, there are currently OS specific and testbed/network setup specific parts, but I plan to move those somehow into YAML file one day as well).

Having no device-specific Python code is very unusual. Forcing everything into yaml configs is IMHO not intended.

> Your solution would need extra class for every device, people would need to edit/understand YAML _and_ Python as well.

Yes, that's expected.

ynezz:
So I've removed the `init_command_timeout` feature and left only the smallubootdriver related fixes.

PR #675, opened 2020-11-24
Add sdwire driver

JPEWdev:
Adds a driver for the Tizen SDWire based device SD Card multiplexer.

This device is very similar to the already existing SDMux device (which
is distinct from the Tizen SDMux device on which SDWire is based).
The primary differences are:
 1) SDWire is controlled with the `sd-mux-ctrl` tool
 2) The serial number for SDWire devices is tied to the control chip
    (an FTDI-based GPIO solution).

See: https://wiki.tizen.org/SDWire

**Description**

**Checklist**
- [ ] Tests for the feature 

- [ ] PR has been tested

### DIFF
--- a/labgrid/driver/usbsdwiredriver.py
+++ b/labgrid/driver/usbsdwiredriver.py
@@ -1,0 +1,40 @@
+# pylint: disable=no-member
+import attr
+
+from .common import Driver
+from ..factory import target_factory
+from ..step import step
+from .exception import ExecutionError
+from ..util.helper import processwrapper
+
+@target_factory.reg_driver
+@attr.s(eq=False)
+class USBSDWireDriver(Driver):
+    """The USBSDWireDriver uses the sd-mux-ctrl tool to control SDWire hardware
+
+    Args:
+        bindings (dict): driver to use with usbsdmux
+    """
+    bindings = {
+        "mux": {"USBSDWireDevice", "NetworkUSBSDWireDevice"},
+    }
+
+    def __attrs_post_init__(self):
+        super().__attrs_post_init__()
+        if self.target.env:
+            self.tool = self.target.env.config.get_tool('sd-mux-ctrl') or 'sd-mux-ctrl'
+        else:
+            self.tool = 'sd-mux-ctrl'
+
+    @Driver.check_active
+    @step(title='sdmux_set', args=['mode'])
+    def set_mode(self, mode):
+        if not mode.lower() in ['dut', 'host']:
+            raise ExecutionError("Setting mode '%s' not supported by USBSDWireDriver" % mode)
+        cmd = self.mux.command_prefix + [
+            self.tool,
+            '--dut' if mode.lower() == "dut" else "--ts",
+            "-e",
+            self.mux.control_serial,
+        ]
+        processwrapper.check_output(cmd)
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -927,7 +927,8 @@ class ClientSession(ApplicationSession):
         action = self.args.action
         target = self._get_target(place)
         from ..driver.usbsdmuxdriver import USBSDMuxDriver
-        from ..resource.remote import NetworkUSBSDMuxDevice
+        from ..driver.usbsdwiredriver import USBSDWireDriver
+        from ..resource.remote import NetworkUSBSDMuxDevice, NetworkUSBSDWireDevice
         drv = None
         for resource in target.resources:
             if isinstance(resource, NetworkUSBSDMuxDevice):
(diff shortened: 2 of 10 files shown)


Emantor (approved):
Overall this looks okay. Letting this sit till next week for @Bastian-Krause & @jluebbe to take a look.

labgrid/resource/udev.py
```diff
@@ -407,6 +407,50 @@ def path(self):
...
+    def poll(self):
+        super().poll()
+        if self.device is None:
+            self.disk_path = None
+        else:
+            if not self.avail:
+                for child in self.device.parent.children:
+                    if child.subsystem == 'block' and child.device_type == 'disk':
+                        self.disk_path = child.device_node
+                self.control_serial = self.device.get('ID_SERIAL_SHORT')
```
agners, line 448: This leads to a warning here:
```
test_install_image.py::test_flash_image
  /usr/lib/python3.8/_collections_abc.py:660: DeprecationWarning: Will be removed in 1.0. Access properties with Device.properties.
    return self[key]

-- Docs: https://docs.pytest.org/en/stable/warnings.html
```

This should help:

```suggestion
                self.control_serial = self.device.properties.get('ID_SERIAL_SHORT')
```

agners:
I actually have a similar patch locally which I wanted to polish a bit for submission. The current state relies on two discrete drivers: `USBSDWireDriver` and `USBMassStorage` for mass storage (since it is possible to set a serial also on the USB2640. The combined approach seems smarter though.

Other than the small nit, the code looks fine for me and also tested the driver on my end:
Reviewed-by: Stefan Agner <stefan@agner.ch>
Tested-by: Stefan Agner <stefan@agner.ch>

Thanks @JPEWdev !

labgrid/resource/udev.py
```diff
@@ -407,6 +407,50 @@ def path(self):
...
+    def avail(self):
+        return bool(self.disk_path and self.control_serial)
+
+    # Forbid the USBResource super class to set the avail property
+    @avail.setter
+    def avail(self, prop):
+        pass
+
+    # Overwrite the poll function. Only mark the SDWire as available if both
+    # paths are available.
```
Bastian-Krause, line 438: @Emantor Does this fit here? I know it is copied from USBSDWireDevice. Maybe I misunderstand something, but I think it does not fit there either.

labgrid/resource/udev.py
```diff
@@ -407,6 +407,50 @@ def path(self):
...
+    def poll(self):
+        super().poll()
+        if self.device is None:
+            self.disk_path = None
```
Bastian-Krause, line 442: Why is `self.control_serial` not set to `None` here?

labgrid/remote/client.py
```diff
@@ -936,9 +938,17 @@ def sd_mux(self):
                 except NoDriverFoundError:
                     drv = USBSDMuxDriver(target, name=None)
                 break
+            elif isinstance(resource, NetworkUSBSDWireDevice):
```
Bastian-Krause, line 941: No need for an `elif` here.
```suggestion
            if isinstance(resource, NetworkUSBSDWireDevice):
```

labgrid/remote/client.py
```diff
@@ -1104,13 +1114,13 @@ def write_image(self):
         place = self.get_acquired_place()
         target = self._get_target(place)
         drv = None
-        from ..resource.remote import NetworkUSBMassStorage, NetworkUSBSDMuxDevice
+        from ..resource.remote import NetworkUSBMassStorage, NetworkUSBSDMuxDevice, NetworkUSBSDWireDevice
```
Bastian-Krause, line 1117: Line longer than 100 characters.

Bastian-Krause (changes requested):

labgrid/resource/udev.py
Emantor: I think it applies here too. We don't want to support partial availability for these devices, since we'll get errors in the USBMassStorageDriver otherwise.

labgrid/resource/udev.py
Emantor: Yes, we should unset these here.

labgrid/resource/udev.py
```diff
@@ -407,6 +407,50 @@ def path(self):
...
 
+@target_factory.reg_resource
+@attr.s(eq=False)
+class USBSDWireDevice(USBResource):
+    """The USBSDWireDevice describes an attached SDWire device,
+    it is identified via USB using udev
+    """
+
+    disk_path = attr.ib(default=None)
+    control_serial = attr.ib(default=None)
```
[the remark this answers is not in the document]
Emantor: No, added for the USBSDWire Device.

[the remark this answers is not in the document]
Bastian-Krause: For the record: in my comment above I wanted to clarify that the `USBSDMuxDevice` (not the `USBSDWireDevice`, copy pasta..) does not use a validator either.

labgrid/resource/udev.py
Bastian-Krause: The comment itself is correct, I am unsure about the position though, it applies to the `avail` property above, doesn't it?

labgrid/driver/usbsdwiredriver.py
```diff
@@ -0,0 +1,41 @@
...
+        if not mode.lower() in ['dut', 'host']:
+            raise ExecutionError("Setting mode '%s' not supported by USBSDWireDriver" % mode)
+        cmd = self.mux.command_prefix + [
+            self.tool,
+            '--dut' if mode.lower() == "dut" else "--ts",
+            "-e",
+            self.mux.control_serial,
+        ]
+        processwrapper.check_output(cmd)
+
```
Bastian-Krause, line 41: Traling newline.

labgrid/resource/udev.py
Emantor: It does, but was copied from USBSDMux. I am fine with merging this and fixing this later.

Bastian-Krause (approved):

PR #678, opened 2020-11-27
remote/client: try protocols before driver

Emantor:
For the power and digital_io command, try to get a bound driver via
protocol first. This way already instanced drivers from a passed in
environment file can be chosen, even if they are not part of upstream
labgrid, or require a driver as a supplier.

- [x] PR has been tested

### DIFF
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -696,26 +696,30 @@ class ClientSession(ApplicationSession):
         from ..driver.powerdriver import NetworkPowerDriver, PDUDaemonDriver, USBPowerDriver
         from ..resource.power import NetworkPowerPort, PDUDaemonPort
         from ..resource.remote import NetworkUSBPowerPort
+
         drv = None
-        for resource in target.resources:
-            if isinstance(resource, NetworkPowerPort):
-                try:
-                    drv = target.get_driver(NetworkPowerDriver)
-                except NoDriverFoundError:
-                    drv = NetworkPowerDriver(target, name=None, delay=delay)
-                break
-            elif isinstance(resource, NetworkUSBPowerPort):
-                try:
-                    drv = target.get_driver(USBPowerDriver)
-                except NoDriverFoundError:
-                    drv = USBPowerDriver(target, name=None, delay=delay)
-                break
-            elif isinstance(resource, PDUDaemonPort):
-                try:
-                    drv = target.get_driver(PDUDaemonDriver)
-                except NoDriverFoundError:
-                    drv = PDUDaemonDriver(target, name=None, delay=int(delay))
-                break
+        try:
+            drv = target.get_driver("PowerProtocol")
+        except NoDriverFoundError:
+            for resource in target.resources:
+                if isinstance(resource, NetworkPowerPort):
+                    try:
+                        drv = target.get_driver(NetworkPowerDriver)
+                    except NoDriverFoundError:
+                        drv = NetworkPowerDriver(target, name=None, delay=delay)
+                        break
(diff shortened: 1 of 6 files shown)


labgrid/remote/client.py
```diff
@@ -693,29 +693,34 @@ def power(self):
...
-                    drv = USBPowerDriver(target, name=None, delay=delay)
-                break
-            elif isinstance(resource, PDUDaemonPort):
-                try:
-                    drv = target.get_driver(PDUDaemonDriver)
-                except NoDriverFoundError:
-                    drv = PDUDaemonDriver(target, name=None, delay=int(delay))
-                break
+        try:
+            target.get_driver(PowerProtocol)
```
jluebbe, line 703: Why don't you use the name as a string to avoid the import?

labgrid/remote/client.py
```diff
@@ -743,43 +748,48 @@ def digital_io(self):
...
-                break
-            elif isinstance(resource, NetworkLXAIOBusPIO):
-                try:
-                    drv = target.get_driver(LXAIOBusPIODriver, name=name)
-                except NoDriverFoundError:
-                    target.set_binding_map({"pio": name})
-                    drv = LXAIOBusPIODriver(target, name=name)
-                break
+        try:
+            target.get_driver(DigitalOutputProtocol)
```
jluebbe, line 755: Here as well.

labgrid/remote/client.py
Emantor: Because at this point, no class has been registered in case no environment file was given on the command line. I don't remember why we only allow already bound drivers to be resolved by string.

labgrid/remote/client.py
```diff
@@ -696,26 +696,30 @@ def power(self):
...
-                    drv = USBPowerDriver(target, name=None, delay=delay)
-                break
-            elif isinstance(resource, PDUDaemonPort):
-                try:
-                    drv = target.get_driver(PDUDaemonDriver)
-                except NoDriverFoundError:
-                    drv = PDUDaemonDriver(target, name=None, delay=int(delay))
-                break
+        try:
+            target.get_driver("PowerProtocol")
```
SmithChart, line 702: I assume this should be `` drv = target.get_driver("PowerProtocol")``.
Otherwise you will end up without a Driver, even if a protocol is present.

labgrid/remote/client.py
```diff
@@ -743,43 +747,47 @@ def digital_io(self):
...
-                break
-            elif isinstance(resource, NetworkLXAIOBusPIO):
-                try:
-                    drv = target.get_driver(LXAIOBusPIODriver, name=name)
-                except NoDriverFoundError:
-                    target.set_binding_map({"pio": name})
-                    drv = LXAIOBusPIODriver(target, name=name)
-                break
+        try:
+            target.get_driver("DigitalOutputProtocol")
```
SmithChart, line 753: Same here. Add a ``drv =``.

SmithChart (changes requested):
Tested the changes on the client on my setup with the PowerProtocol on top of a GPIO.
Your PR does not work:
```
(venv) chris@demopi:~/demo $ labgrid-client -c motorbone.yaml power on
Selected role main and place motorbone from configuration file
labgrid-client: error: target has no compatible resource available
```

With the changes suggested it works. Please double check :)

labgrid/remote/client.py
Emantor: indeed. fixed.

labgrid/remote/client.py
Emantor: indeed. fixed.

PR #681, opened 2020-12-02
Add libsnappy dev and document requirement for crossbar

Emantor:
**Description**
The new pip resolver correctly resolves that crossbar requires autobahn with compression. Document it and add it to the workflow.

With this the CI builds again: https://github.com/Emantor/labgrid/actions/runs/395994168

**Checklist**
- [x] PR has been tested


Bastian-Krause (approved):
Thanks!

PR #685, opened 2020-12-03
util/helper: allow multiple callbacks

jremmet:
when pytest plugin pytester runs a testdir based test enable_logging is
called more than once. This triggers the assertion in register.
So remove it an take multiple entries of the callback

Signed-off-by: Jan Remmet <j.remmet@phytec.de>

Fixes #684

PR #686, opened 2020-12-04
util/helper: multiple registration of callbacks

Emantor:
**Description**
The pytester pytest plugin will setup multiple pytest sessions and calls
the pytest_configure multiple times. This leads to multiple registration
of the logging callback, which triggers the assert that the callback can
only be registered once. Remove the assert and replace it with a check
to register the callback only once.

**Checklist**
- [x] PR has been tested

Fixes https://github.com/labgrid-project/labgrid/issues/684

### DIFF
--- a/labgrid/util/helper.py
+++ b/labgrid/util/helper.py
@@ -101,12 +101,14 @@ class ProcessWrapper:
 
     def register(self, callback):
         """Register a callback with the ProcessWrapper"""
-        assert callback not in self.callbacks
+        if callback in self.callbacks:
+            return
         self.callbacks.append(callback)
 
     def unregister(self, callback):
         """Unregister a callback with the ProcessWrapper"""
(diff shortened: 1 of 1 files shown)


Bastian-Krause:
@jremmet mentioned in #685 that `ProcessWrapper.unregister()` is also called multiple times. How does this work with `assert callback in self.callbacks` in `ProcessWrapper.unregister()`?

Emantor:
@Bastian-Krause fixed by doing the same for unregister.

jremmet:
For my pytester /  testdir tests this works and processwrapper logging INFO also looks fine. :+1:

Emantor:
> For my pytester / testdir tests this works and processwrapper logging INFO also looks fine. +1

Yep, thanks for the minimal testcase in your issue, it was easy to verify that the fix works :+1:

PR #689, opened 2020-12-07
usbstoragedriver: Try blockdev command to get disk size

JPEWdev:
Attempt to use the blockdev command to determine the size of a block
device. This command has the advantage that it directly calls
ioctl(BLKGETSIZE64) on the device, meaning it will work as long as the
user has read permission on the actual block device (e.g. /dev/sdb)
instead of requiring permissions to read the sysfs file.

Signed-off-by: Joshua Watt <JPEWhacker@gmail.com>

**Description**

**Checklist**
- [ ] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/labgrid/driver/usbstoragedriver.py
+++ b/labgrid/driver/usbstoragedriver.py
@@ -76,13 +76,9 @@ class USBStorageDriver(Driver):
         # wait for medium
         timeout = Timeout(10.0)
         while not timeout.expired:
-            try:
-                if self.get_size() > 0:
-                    break
-                time.sleep(0.5)
-            except ValueError:
-                # when the medium gets ready the sysfs attribute is empty for a short time span
-                continue
+            if self.get_size() > 0:
+                break
+            time.sleep(0.5)
         else:
             raise ExecutionError("Timeout while waiting for medium")
 
@@ -147,9 +143,11 @@ class USBStorageDriver(Driver):
     @Driver.check_active
     @step(result=True)
     def get_size(self):
-        args = ["cat", "/sys/class/block/{}/size".format(self.storage.path[5:])]
-        size = subprocess.check_output(self.storage.command_prefix + args)
-        return int(size)*512
+        args = ["blockdev", "--getsize64", self.storage.path]
+        p = subprocess.run(self.storage.command_prefix + args, stdout=subprocess.PIPE)
+        if p.returncode != 0:
+            return 0
+        return int(p.stdout.decode('utf-8'))
 
 
 @target_factory.reg_driver


Emantor:
`blockdev` is contained in `essential` for debian and part of the `base` metapackage for archlinux (together with cat from moreutils), so IMO its safe to assume that a device has blockdev and I'd be in favor of replacing `cat` wholesale with `blockdev`. cc @jluebbe & @Bastian-Krause for their thoughts on this.

labgrid/driver/usbstoragedriver.py
```diff
@@ -122,9 +122,9 @@ def write_image(self, filename=None, mode=Mode.DD, partition=None, skip=0, seek=
     @Driver.check_active
     @step(result=True)
     def get_size(self):
-        args = ["cat", "/sys/class/block/{}/size".format(self.storage.path[5:])]
+        args = ["/sbin/blockdev", "--getsize64", self.storage.path]
```
Emantor, line 125: This shouldn't be `/sbin/blockdev`, it should rather be `blockdev`. `PATH` needs to be setup correctly in the user environment, not all distributions have `blockdev` in sbin.

Emantor: blockdev also behaves differently, instead of returning `0` it will return `blockdev: cannot open /dev/sdb: No medium found` and exit with `1`, this handling needs to be adapted here.

Bastian-Krause: You should also drop the `try .. except` from `write_image()`.

Bastian-Krause:
In Debian `blockdev` resides in `/sbin/` which is not part of an unpriviliged user's `PATH`. Calling `/sbin/blockdev --getsize64 /dev/sdd` works without additional privileges though.

JPEWdev:
> In Debian `blockdev` resides in `/sbin/` which is not part of an unpriviliged user's `PATH`. Calling `/sbin/blockdev --getsize64 /dev/sdd` works without additional privileges though.

Ya, I originally had that for that reason... @Emantor any ideas?

Emantor:
Yeah, we need to check whether blockdev is in `PATH` and if it isn't, try `/sbin/<binary>` before failing. So:
- this should use the processwrapper
- the processwrapper should check the binary availability before running the command (this allows us better user errors either way, i.e. by notifying the user if a command is not available)

jluebbe:
As I see it, you need sysfs access to run a labgrid-exporter in any case, as udev won't work otherwise. So I'd much prefer to keep this as-is, since the sysfs API for this is identical on every system.

JPEWdev:
> As I see it, you need sysfs access to run a labgrid-exporter in any case, as udev won't work otherwise. So I'd much prefer to keep this as-is, since the sysfs API for this is identical on every system.

I'm specifically running labgrid-exporter in a container and passing through the host udev. I'd rather not bind mount in /sys/ as well if it can be avoided

jluebbe:
> > As I see it, you need sysfs access to run a labgrid-exporter in any case, as udev won't work otherwise. So I'd much prefer to keep this as-is, since the sysfs API for this is identical on every system.
> 
> I'm specifically running labgrid-exporter in a container and passing through the host udev. I'd rather not bind mount in /sys/ as well if it can be avoided

I see (read-only) access to /sys as a standard kernel interface, and so I wouldn't want to add workarounds for this case. If /dev is already mounted into the container, I don't see the downside of a read-only /sys mount.

Emantor:
Read only access to `/sys` is the way to go.

PR #690, opened 2020-12-07
client: Add LG_CLAIM_HOSTNAME/LG_CLAIM_USER

JPEWdev:
Adds environment variables LG_CLAIM_HOSTNAME/LG_CLAIM_USER to control
the hostname and user that used to make a claim on a place. This is
useful in container based build pipelines where multiple pipeline stages
may need to use a claim, but the hostname and user name running in each
container differs (e.g. Tekton in Kubernetes).

Signed-off-by: Joshua Watt <JPEWhacker@gmail.com>

**Description**

**Checklist**
- [ ] Tests for the feature 

- [ ] PR has been tested

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -58,6 +58,9 @@ New Features in 0.3.0
 - Add power driver to support GEMBIRD SiS-PM implementing SiSPMPowerDriver.
 - A cleanup of the cleanup functions was performed, labgrid should now clean up
   after itself and throws an error if the user needs to handle it himself.
+- ``labgrid-client`` now respects the ``LG_HOSTNAME`` and ``LG_USERNAME``
+  environment variables to set the hostname and username when accessing
+  resources.
 
 Breaking changes in 0.3.0
 ~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -46,6 +46,12 @@ class ClientSession(ApplicationSession):
     """The ClientSession encapsulates all the actions a Client can Invoke on
     the coordinator."""
 
+    def gethostname(self):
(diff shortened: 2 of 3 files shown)


Emantor:
My guess is that this is used to make the locking semantics static across multiple containers & users, i.e. your pipeline will always lock the place with something like `tekton/labgrid-test`.

JPEWdev:
> My guess is that this is used to make the locking semantics static across multiple containers & users, i.e. your pipeline will always lock the place with something like `tekton/labgrid-test`.

Correct. Specifically, I capture the (unique) name of the container/pipeline used to make the reservation, and the use that name in all subsequent containers. This makes it easy to trace a reservation back to a specific pipeline in the event something goes wrong and it gets left dangling (which is pretty rare because I have a finalize task that releases any reservations when the pipeline exits).

Emantor (approved):

jluebbe:
Is there a specific reason why you called it "claim"? That's not a term we've used in labgrid so far. Why not just "LG_HOSTNAME" and "LG_USERNAME"?

JPEWdev:
> Is there a specific reason why you called it "claim"? That's not a term we've used in labgrid so far. Why not just "LG_HOSTNAME" and "LG_USERNAME"?

It made sense to me.... but I don't really care. If you want it changed, let me know.

Emantor (changes requested):
This needs to be added to the `labgrid-client` man page. I also prefer the either `LG_USER/HOSTNAME` or `LG_ACQUIRE_USER/HOSTNAME`, since we use `acquire` & `release` as the new naming semantics.

jluebbe:
Please just use LG_USERNAME/LG_HOSTNAME. :) They are also used for reservations, and we also have the lock alias for acquire.

Emantor (approved):

PR #695, opened 2020-12-15
graphstrategy: document errors and public methods

Bastian-Krause:
**Description**
High-level documentation is already available in doc/development.rst. Having doc strings on the errors and public methods still makes sense, since strategies that want to use these features must inherit from the GraphStrategy and must know how the public methods are intended to work.

**Checklist**
- [ ] PR has been tested

### DIFF
--- a/labgrid/strategy/graphstrategy.py
+++ b/labgrid/strategy/graphstrategy.py
@@ -13,14 +13,17 @@ __all__ = [
 
 
 class GraphStrategyError(StrategyError):
+    """Generic GraphStrategy error"""
     pass
 
 
 class InvalidGraphStrategyError(GraphStrategyError):
+    """GraphStrategy error raised during initialization of broken strategies"""
     pass
 
 
 class GraphStrategyRuntimeError(GraphStrategyError):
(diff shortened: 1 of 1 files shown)


Bastian-Krause:
Note that this is introduces a change in behavior. But in my opinion the previous behavior was not expected.

Emantor (approved):

Bastian-Krause:
This PR originally included:
> For certain development use cases Strategies inheriting from GraphStrategy might effectively skip execution of certain states by
> returning early if a certain parameter is given to the strategy. If this parameter is set, the developer knows that the board is already in a certain state.
This additional state information must be reset in its extended invalidate() method in order to handle error conditions correctly.
>
>The GraphStrategy calls invalidate() in its initialization phase to avoid code duplication. This causes the additional state information passed to such a strategy to be lost immediately during initialization.
>
> To avoid this, move the code resetting the state to a dedicated, private method and call this method during initialization and on invalidate(). This causes code in extended invalidate() methods to be executed on error and on explicit calls to invalidate() only, but not during initialization.

I dropped this part in favor of implementing the `force()` method (introduced by #622) of my custom strategy.

PR #698, opened 2021-01-11
mqtt: support for tasmota power ports

Emantor:
**Description**
Support for MQTT Tasmota Devices.

**Checklist**
- [x] Tests for the feature 
- [x] PR has been tested


labgrid/driver/mqtt.py
```diff
@@ -0,0 +1,81 @@
...
+
+@target_factory.reg_driver
+@attr.s(eq=False)
+class TasmotaPowerDriver(Driver, PowerProtocol):
+    bindings = {
+            "power": {"TasmotaPowerPort", "NetworkTasmotaPowerPort"}
+    }
+    delay = attr.ib(default=2.0, validator=attr.validators.instance_of(float))
+    _client = attr.ib(default=mqtt.Client(), validator=attr.validators.instance_of(mqtt.Client))
+    _status_lock = attr.ib(default=threading.Lock())
```
jluebbe, line 27: Either initialize in `__attrs_post_inst__` or use `factory=`.

labgrid/resource/lxaiobus.py
```diff
@@ -10,13 +10,14 @@
...
+
+    _last = attr.ib(default=0.0, validator=attr.validators.instance_of(float))
+
     def __attrs_post_init__(self):
         super().__attrs_post_init__()
 
         self.log = logging.getLogger('LXAIOBusNodeManager')
 
-        self._last = 0.0
-
```
jluebbe, line 19: These changes are unrelated.

labgrid/resource/mqtt.py
```diff
@@ -0,0 +1,81 @@
...
+from ..factory import target_factory
+from ..util import Timeout
+
+@attr.s(eq=False)
+class MQTTManager(ResourceManager):
+    _available = attr.ib(default=attr.Factory(list), validator=attr.validators.instance_of(list))
+    _avail_lock = attr.ib(default=threading.Lock())
+    _clients = attr.ib(default=attr.Factory(dict), validator=attr.validators.instance_of(dict))
+    _topics = attr.ib(default=attr.Factory(list), validator=attr.validators.instance_of(list))
+    _topic_lock = attr.ib(default=threading.Lock())
```
jluebbe, line 19: You could probably use `factory=` for these. I'm not sure if maybe type annotations are a better solution for internal attributes.

labgrid/resource/mqtt.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def _create_mqtt_connection(self, host):
+        client = mqtt.Client()
+        client.connect(host)
+        client.on_connect = self._on_connect
+        client.on_message = self._on_message
+        client.loop_start()
+        return client
+
+    def on_resource_added(self, resource):
```
jluebbe, line 37: Should this be private?

labgrid/driver/mqtt.py
```diff
@@ -0,0 +1,81 @@
...
+        with self._status_lock:
+            self._status = status
+
+    def _on_connect(self, client, userdata, flags, rc):
+        client.subscribe(self.power.status_topic)
+
+    @Driver.check_active
+    @step()
+    def on(self):
+        self._client.publish(self.power.power_topic, payload="ON")
```
jluebbe, line 56: Do we get feedback if this worked? When does this return?

labgrid/driver/mqtt.py
```diff
@@ -0,0 +1,81 @@
...
+    def _get_status(self):
+        with self._status_lock:
+            status = self._status
+        return status
+
+    @Driver.check_active
+    @step()
+    def get(self):
+        self._client.publish(self.power.power_topic)
+        while self._status is None:
```
jluebbe, line 79: Use util.timeout here.

labgrid/remote/exporter.py
```diff
@@ -535,6 +535,23 @@ def _get_params(self):
 
 exports["LXAIOBusPIO"] = LXAIOBusNodeExport
 
+@attr.s(eq=False)
+class TasmotaPowerPortExport(ResourceExport):
+    """ResourceExport for LXAIOBusNode devices accessed via the HTTP API"""
```
jluebbe, line 540: The docstring is wrong.

labgrid/resource/mqtt.py
```diff
@@ -0,0 +1,81 @@
+import queue
+import logging
+from time import monotonic
+
+import attr
+import paho.mqtt.client as mqtt
```
jluebbe, line 6: This should also be imported only when needed.

labgrid/resource/mqtt.py
```diff
@@ -0,0 +1,81 @@
...
+    _topics = attr.ib(default=attr.Factory(list), validator=attr.validators.instance_of(list))
+    _topic_lock = attr.ib(default=threading.Lock())
+    _last = attr.ib(default=0.0, validator=attr.validators.instance_of(float))
+
+    def __attrs_post_init__(self):
+        super().__attrs_post_init__()
+        self.log = logging.getLogger('MQTTManager')
+
+    def _on_connect(self, client, userdata, flags, rc):
+        client.subscribe("#")
```
jluebbe, line 27: Is this a wildcard subscription? Shouldn't we subscribe to a more specific path?

labgrid/resource/mqtt.py
```diff
@@ -0,0 +1,81 @@
...
+        host = resource.host
+        if host not in self._clients:
+            self._clients[host] = self._create_mqtt_connection(host)
+
+    def _on_message(self, client, userdata, msg):
+        payload = msg.payload.decode('utf-8')
+        topic = msg.topic
+        if payload == "Online":
+            with self._avail_lock:
+                self._available.append(topic)
```
jluebbe, line 47: `_available` is a list but used like a set?

labgrid/resource/mqtt.py
```diff
@@ -0,0 +1,81 @@
...
+        super().__attrs_post_init__()
+
+
+@target_factory.reg_resource
+@attr.s(eq=False)
+class TasmotaPowerPort(MQTTResource):
+    power_topic = attr.ib(default=None,
+                         validator=attr.validators.instance_of(str))
+    status_topic = attr.ib(default=None,
+                         validator=attr.validators.instance_of(str))
```
jluebbe, line 81: Could we derive all these topic from a common prefix?

jluebbe (changes requested):
We should add at least add a simple test which imports this, but better connect to a MQTT broker.

Maybe we don't need the split between NetworkTasmotaPowerPort and TasmotaPowerPort. This looks similar to the NetworkPowerPort and NetworkSerialPort, which are exported unmodified for external resources.

labgrid/driver/mqtt.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+@target_factory.reg_driver
+@attr.s(eq=False)
+class TasmotaPowerDriver(Driver, PowerProtocol):
+    bindings = {
+            "power": {"TasmotaPowerPort", "NetworkTasmotaPowerPort"}
+    }
+    delay = attr.ib(default=2.0, validator=attr.validators.instance_of(float))
+    _client = attr.ib(default=mqtt.Client(), validator=attr.validators.instance_of(mqtt.Client))
```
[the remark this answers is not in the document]
Emantor: Since this is an `attr.ib` my expectation was that every class will get its own client.

labgrid/driver/mqtt.py
Emantor: `factory` is syntactic sugar for this same construct, but  I don't mind switching.

labgrid/resource/mqtt.py
Emantor: Same as above, but I can switch this to use `factory`.

labgrid/resource/mqtt.py
Emantor: This is called by the `ResourceManager` class, changing this will require changing all other classes using `ResourceManager` as well.

labgrid/driver/mqtt.py
Emantor: Yes, we should handle the returned class and wait for the topic to be delivered.

labgrid/resource/mqtt.py
Emantor: In theory yes, in practice we have already changed single topics inside our own infrastructure, which will make derivation from a simple token impossible. I implemented it this way at first, but everybody seems prone to change the topics on their own, so I'd rather  provide each topic separately.

labgrid/resource/mqtt.py
Emantor: We could subscribe only to the topics for whom we have registered resources.

labgrid/driver/mqtt.py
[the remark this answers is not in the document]
jluebbe: No, with default= you set a single default value:
```
>>> import attr
>>> @attr.s()
... class Foo:
...   bar = attr.ib(default=dict())
... 
>>> a = Foo()
>>> b = Foo()
>>> a.bar is b.bar
True
>>> a.bar['this'] = 'that'
>>> b.bar
{'this': 'that'}
```

labgrid/driver/mqtt.py
jluebbe: Factory expects a function that should be called to produce the per-instance value.

labgrid/resource/mqtt.py
jluebbe: Ah, yes.

labgrid/resource/mqtt.py
jluebbe: Could you give an example of these topics?

labgrid/driver/mqtt.py
```diff
@@ -0,0 +1,81 @@
...
+from ..protocol import PowerProtocol
+from ..step import step
+from ..util import Timeout
+
+
+@target_factory.reg_driver
+@attr.s(eq=False)
+class TasmotaPowerDriver(Driver, PowerProtocol):
+    bindings = {
+            "power": {"TasmotaPowerPort", "NetworkTasmotaPowerPort"}
```
Emantor, line 23: We talked about this in a short call and agreed that the `NetworkTasmotaPortPort` does not make much sense. The exporter should export a plan `TasmotaPowerPort` instead, which gets instanciated again on the client side.

labgrid/resource/mqtt.py
```diff
@@ -0,0 +1,81 @@
...
+import logging
+from time import monotonic
+
+import attr
+import paho.mqtt.client as mqtt
+import threading
+
+from .common import ManagedResource, ResourceManager, Resource
+from ..factory import target_factory
+from ..util import Timeout
```
Bastian-Krause, line 11: Unused import.

labgrid/resource/mqtt.py
```diff
@@ -0,0 +1,81 @@
+import queue
+import logging
+from time import monotonic
+
+import attr
+import paho.mqtt.client as mqtt
+import threading
```
Bastian-Krause, line 7: Wrong import order, standard library first, then third-party, then local imports.

labgrid/resource/mqtt.py
```diff
@@ -0,0 +1,81 @@
...
+
+from .common import ManagedResource, ResourceManager, Resource
```
Bastian-Krause, line 9: Unused import: `Resource` is not used anywhere.

Bastian-Krause (dismissed):
Some nitpicks :)

labgrid/resource/mqtt.py
Emantor: We could combine `status_topic` and `power_topic`, I'd like to keep the `avail_topic` apart since this is part of the MQTTResource and should work with other MQTT devices as well.

labgrid/driver/mqtt.py
```diff
@@ -0,0 +1,95 @@
...
+    @Driver.check_active
+    @step()
+    def on(self):
+        self._publish(self.power.power_topic, "ON")
+
+
+    @Driver.check_active
+    @step()
+    def off(self):
+        self._publish(self.power.power_topic, "OFF")
```
Emantor, line 71: We should wait for the message to be published and check that the status gets set via the get_status function.

labgrid/driver/mqtt.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def on_deactivate(self):
+        self._client.loop_stop()
+
+    def _on_message(self, client, userdata, msg):
+        if msg.payload == b'ON':
+            status = True
+        elif msg.payload == b'OFF':
+            status = False
+        with self._status_lock:
```
Emantor, line 48: We don't need a `threading.Lock` for single python assignements or checks, so we can drop the lock from the driver here.

Emantor:
I have adressed all review comments, the PR should be good to go.

labgrid/driver/mqtt.py
```diff
@@ -0,0 +1,97 @@
...
+
+    @Driver.check_active
+    @step()
+    def get(self):
+        self._client.publish(self.power.power_topic)
+        timeout = Timeout(3.0)
+        while self._status is None:
+            time.sleep(0.1)
+            if timeout.expired:
+                raise MQTTError("Could not get initial status")
```
jluebbe, line 96: Why is this the initial status?

labgrid/resource/lxaiobus.py
```diff
@@ -10,6 +10,7 @@
 
 @attr.s(eq=False)
 class LXAIOBusNodeManager(ResourceManager):
+
```
jluebbe, line 13: Spurious change.

labgrid/remote/exporter.py
```diff
@@ -617,7 +617,6 @@ def _get_params(self):
 
 exports["LXAIOBusPIO"] = LXAIOBusNodeExport
 
-
```
jluebbe, line 620: Spurious change.

labgrid/resource/mqtt.py
```diff
@@ -0,0 +1,76 @@
...
+    def _on_message(self, client, userdata, msg):
+        payload = msg.payload.decode('utf-8')
+        topic = msg.topic
+        if payload == "Online":
+            with self._avail_lock:
+                self._available.add(topic)
+        if payload == "Offline":
+            if topic in self._available:
+                with self._avail_lock:
+                    self._available.remove(topic)
```
jluebbe, line 46: Use `.discard(topic)`.

jluebbe (changes requested):

labgrid/driver/mqtt.py
Emantor: Since the initial status is unknown when running in the `labgrid-client` cli we need to request the topic and wait for the request to hit the message handler, therefore the initial state is `None`.

labgrid/resource/mqtt.py
jluebbe: Ah, sorry, i should have been more explicit: with `.remove()`, the initial check is no longer needed.

Emantor: I guess you meant with `.discard()` the check for the object to be in the set is not required. I have adjusted the PR accordingly.

jluebbe (approved):

PR #699, opened 2021-01-11
Fix Network sysfs GPIO

JPEWdev:
**Description**
Fix accessing sysfs GPIOs over a network

**Checklist**
- [x] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/labgrid/util/agents/sysfsgpio.py
+++ b/labgrid/util/agents/sysfsgpio.py
@@ -9,7 +9,6 @@ import os
 
 class GpioDigitalOutput:
     _gpio_sysfs_path_prefix = '/sys/class/gpio'
-    _buffered_file_access = False
 
     @staticmethod
     def _assert_gpio_line_is_exported(index):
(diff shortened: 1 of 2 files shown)


labgrid/util/agents/sysfsgpio.py
```diff
@@ -9,17 +9,15 @@
...
     @staticmethod
     def _assert_gpio_line_is_exported(index):
         gpio_sysfs_path = os.path.join(GpioDigitalOutput._gpio_sysfs_path_prefix,
                                        'gpio{0}'.format(index))
         if not os.path.exists(gpio_sysfs_path):
             export_sysfs_path = os.path.join(GpioDigitalOutput._gpio_sysfs_path_prefix, 'export')
-            with open(export_sysfs_path, mode='r+',
-                      buffering=GpioDigitalOutput._buffered_file_access) as export:
-                export.write(str(index))
+            with open(export_sysfs_path, mode='wb') as export:
```
Emantor, line 19: Do we want to turn of buffering with `buffering=0` here as well?

JPEWdev: We could if you really wanted to, but since the file is only open transiently I doubt the buffer mode matters.

Emantor: > We could if you really wanted to, but since the file is only open transiently I doubt the buffer mode matters.

You're right, this is just used to check/set the exports anyway. LGTM

Emantor (approved):

PR #700, opened 2021-01-14
resource: DeditecRelais8: add ID_MODEL

hallerjulian:
Newer (?) Deditec 8 way relais board have a device individual and unique
serial number. In order to make those work with labgrid, the
ID_SERIAL_SHORT must be user definable.
To not break existing setups, the fallback value will be the already
existing default.

### DIFF
--- a/labgrid/resource/udev.py
+++ b/labgrid/resource/udev.py
@@ -577,8 +577,8 @@ class DeditecRelais8(USBResource):
 
     def __attrs_post_init__(self):
         self.match['ID_VENDOR'] = 'DEDITEC'
-        # the serial is the same for all boards with the same model
-        self.match['ID_SERIAL_SHORT'] = 'DT000014'
+        # Deditec does not set proper model/vendor IDs, so we use ID_MODEL instead
+        self.match['ID_MODEL'] = 'DEDITEC_USB-OPT_REL-8'
         super().__attrs_post_init__()
 
     @property


Emantor:
Do you have multiple devices on hand? Do they follow the same naming scheme i.e. `DTxxxxxx`? This ID might have been specific to the board I was using. Even then I'd be more in favor of removing the whole line.

jluebbe:
I think the problem with identifying the Deditec devices was that they don't have proper vendor/model IDs. Without the ID_SERIAL_SHORT check, we'd match any Deditec device, irrespective of whether the driver would be compatible.

Could you provide the output of `udevadm info /dev/…` of your device for future reference?

hallerjulian:
I have currently only 2 devices on hand:
```
ID_SERIAL_SHORT=DT01404JHG9L
ID_SERIAL_SHORT=DT014054AYS8
```

I attached the complete udevadm output of both devices.

[udevadm_deditec1.txt](https://github.com/labgrid-project/labgrid/files/5819609/udevadm_deditec1.txt)
[udevadm_deditec2.txt](https://github.com/labgrid-project/labgrid/files/5819611/udevadm_deditec2.txt)

hallerjulian:
To continue with the patch, would you be okay with removing both the hard-coded 'ID_SERIAL_SHORT' and 'ID_VENDOR' matches? Or would you more like to keep the 'ID_VENDOR' match (with the risk the the user unintentionally matches any Deditec device)?

Emantor:
I'd prefer removing the `ID_SERIAL_SHORT` and using `ID_MODEL=DEDITEC_USB-OPT_REL-8` instead, however I currently don't have access to the deditec device we tested with. Maybe using `ID_MODEL_ID` would be even more preferable.

hallerjulian:
I updated the commit to now use ID_MODEL. ID_MODEL_ID and ID_VENDOR_ID cannot be used, the identify the device as FTDI chip.

PR #702, opened 2021-01-15
labgrid-bound-connect: Delete IP cache

JPEWdev:
Delete the ARP or NDP cache before attempting to connect to a service.
For devices that are assigned an ephemeral MAC address on boot instead
of using a fixed one, these caches may be out of date which results in
the connection failing.

**Checklist**
- [ ] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/helpers/labgrid-bound-connect
+++ b/helpers/labgrid-bound-connect
@@ -7,6 +7,7 @@ import argparse
 import os
 import sys
 import ipaddress
+import subprocess
 
 
 def main(ifname, address, port):
@@ -33,6 +34,12 @@ def main(ifname, address, port):
         prefix = "TCP6:[{}]:{}".format(address, port)
     else:
         raise RuntimeError("Invalid IP version '{}'".format(address.version))
+
+    # Delete the IP lookup cache for the address in case it is stale
+    subprocess.run(["ip", "neigh", "del", str(address), "dev", ifname], 
+        stdout=suprocess.DEVNULL, 
+        stderr=subprocess.DEVNULL
+    )
 
     args.append(','.join([
         prefix,


jluebbe:
Hmm, this is for the case where you have the reuse the same IP with a different MAC, right?

How sure are you that this doesn't cause problems with other use cases, maybe when a connection to the same IP but a different port is already active?

JPEWdev:
> Hmm, this is for the case where you have the reuse the same IP with a different MAC, right?

Correct. Specifically, the DUT I'm testing doesn't have persistent MAC addresses, and the exporter gets confused about where it should send packets.

> 
> How sure are you that this doesn't cause problems with other use cases, maybe when a connection to the same IP but a different port is already active?

This is for the mapping of IP address -> MAC address, so deleting the cache won't affect any current connection. It might slightly delay the next packet when the kernel sees that it no longer has an IP -> MAC address mapping for it and does a query to discover it.

I can't think of any reasonable case where deleting the cache here would cause a problem that wouldn't eventually become apparent when the cache naturally times out and the kernel has to refresh it.

jluebbe:
> > Hmm, this is for the case where you have the reuse the same IP with a different MAC, right?
> 
> Correct. Specifically, the DUT I'm testing doesn't have persistent MAC addresses, and the exporter gets confused about where it should send packets.

OK.

> > How sure are you that this doesn't cause problems with other use cases, maybe when a connection to the same IP but a different port is already active?
> 
> This is for the mapping of IP address -> MAC address, so deleting the cache won't affect any current connection. It might slightly delay the next packet when the kernel sees that it no longer has an IP -> MAC address mapping for it and does a query to discover it.
> 
> I can't think of any reasonable case where deleting the cache here would cause a problem that wouldn't eventually become apparent when the cache naturally times out and the kernel has to refresh it.

In the normal case, an entry will become stale before it is removed from the neighbor cache, so the kernel can refresh it in the background. Maybe it would be better to reduce base_reachable_time{,_ms} for the interface where systems without persistent MACs are connected? That would allow the kernel to handle reprobing automatically.

helpers/labgrid-bound-connect
```diff
@@ -34,6 +35,9 @@ def main(ifname, address, port):
     else:
         raise RuntimeError("Invalid IP version '{}'".format(address.version))
 
+    # Delete the IP lookup cache for the address in case it is stale
+    subprocess.run(["ip", "neigh", "del", str(address), "dev", ifname])
```
jluebbe, line 39: We need to make sure that no output is produced on stdout, as that would be indistinguishable from data sent be the remote side.

JPEWdev:
> > > Hmm, this is for the case where you have the reuse the same IP with a different MAC, right?
> > 
> > 
> > Correct. Specifically, the DUT I'm testing doesn't have persistent MAC addresses, and the exporter gets confused about where it should send packets.
> 
> OK.
> 
> > > How sure are you that this doesn't cause problems with other use cases, maybe when a connection to the same IP but a different port is already active?
> > 
> > 
> > This is for the mapping of IP address -> MAC address, so deleting the cache won't affect any current connection. It might slightly delay the next packet when the kernel sees that it no longer has an IP -> MAC address mapping for it and does a query to discover it.
> > I can't think of any reasonable case where deleting the cache here would cause a problem that wouldn't eventually become apparent when the cache naturally times out and the kernel has to refresh it.
> 
> In the normal case, an entry will become stale before it is removed from the neighbor cache, so the kernel can refresh it in the background. Maybe it would be better to reduce base_reachable_time{,_ms} for the interface where systems without persistent MACs are connected? That would allow the kernel to handle reprobing automatically.

Personally, I think that's a bad user experience. It would be much better if labgrid worked "out of the box" for this situation. In particular, this type of thing is pretty specific to on-device automated testing; under normal circumstances you aren't going to have a network where a device's MAC address is changing but it's IP address is the same; it's rare enough that manual correction (or simply waiting a little longer is fine). I think users might have a hard time realizing that's even the problem and that they need to adjust the setting. I it took me a while to figure it out because when I ssh'd into the exporter to investigate or ran the tests locally, sufficient time had passed that everything worked. It was only when doing automated CI where tests are run in succession that it had a problem. It seems like having labgrid clear out the cache to make sure it's connecting to the correct MAC which has very little downside, (even if the code is a bit annoying) can prevent a lot of user annoyance/frustration.

Emantor:
I agree here, it's way harder for users to figure out what happens, even if we add documentation for this. We are even restraining this to only one address and this specific interface, so I am fine with this change.

jluebbe (approved):

PR #704, opened 2021-01-18
config/environment/target: move get_target_features() from Env to Config, add Target helpers

Bastian-Krause:
**Description**
As discussed in #490 move `get_target_features()` from `Environment` to `Config`. Also add helpers for target options and features to `Target`.

The pytest plugin uses the now deprecated `Environment.get_target_features()`. I wonder how we can change that since the target seems not known at that time in `pytest_configure()`/`pytest_collection_modifyitems()`.

### DIFF
--- a/labgrid/target.py
+++ b/labgrid/target.py
@@ -485,3 +485,7 @@ class Target:
     def get_target_features(self):
         """Helper to retrieve all features of this target"""
         return self.env.config.get_target_features(self.name)
+
+    def get_target_option(self, name, default=None):
+        """Helper to retrieve target option of this target"""
+        return self.env.config.get_target_option(self.name, name, default=default)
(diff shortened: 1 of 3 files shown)


Emantor (approved):

Emantor:
Guess I was to fast with the review. 

> The pytest plugin uses the now deprecated `Environment.get_target_features()`. I wonder how we can change that since the target seems not known at that time in `pytest_configure()`/`pytest_collection_modifyitems()`.

Hm. We should probably hand over the config instead of the environment.

labgrid/target.py
```diff
@@ -481,3 +481,11 @@ def cleanup(self):
         self.deactivate_all_drivers()
         for res in reversed(self.resources):
             self.deactivate(res)
+
+    def get_target_features(self):
+        """Helper to retrieve all features of this target"""
+        return self.env.config.get_target_features(self.name)
```
jluebbe, line 487: Hmm, maybe it would make more sense to pass the features and options to make_target() and store them as a copy in each target? That would avoid following the path upwards from target to env to config...

PR #706, opened 2021-01-20
Xilinx JTAG support

sephalon:
**Description**

Provides support for accessing Xilinx devices via JTAG:
- `XilinxUSBJTAG` resource describing a Xilinx-compatible USB JTAG adapter
- `VivadoHWServerExport` resource export, spawning a Vivado hardware server for each `XilinxUSBJTAG` resource
- `XSDBDriver`, which runs Xilinx System Debugger (XSDB) Tcl commands via the Vivado hardware server
- `xlx` client subcommands, showcasing how to program a bitstream using the `XSDBDriver`

The list of supported JTAG adapters is by no means exhaustive, but is limited by our lab inventory we were able to test with.

In our lab we have seen more complicated scenarios, in which one needs to start/stop Vivado hardware server(s), sometimes even of different versions, during a test run. Right now, such use cases are not supported by the current architecture, but we could provide an alternative `VivadoHWServerDriver` that dynamically spawns a Vivado hardware server on the target via SSH. The `VivadoHWServerExport` must then be deactivated, likely with a flag in the respective `XilinxUSBJTAG` resource.

**Checklist**
- [ ] Tests for the feature 
- [x] PR has been tested

### DIFF
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -1271,10 +1271,10 @@ class ClientSession(ApplicationSession):
 
     def _check_xlx_env(self):
         if not "XILINX_VIVADO" in os.environ:
-            print("xlx subcommands must be invoked from within a Vivado environment", file=sys.stderr)
+            print("xsdb must be invoked from within a Vivado environment", file=sys.stderr)
             exit(1)
 
-    def _get_xlx(self, name):
+    def _get_fpga(self, name):
         place = self.get_acquired_place()
         target = self._get_target(place)
         from ..driver.xsdbdriver import XSDBDriver
@@ -1293,17 +1293,17 @@ class ClientSession(ApplicationSession):
         target.activate(drv)
         return drv
 
-    def xlx_run_xsdb(self):
+    def fpga_run_xsdb(self):
         self._check_xlx_env()
-        drv = self._get_xlx(self.args.resource)
+        drv = self._get_fpga(self.args.resource)
 
         processwrapper.enable_print()
         drv.run([self.args.tcl_cmds])
         processwrapper.disable_print()
 
-    def xlx_program_bitstream(self):
+    def fpga_program_bitstream(self):
         self._check_xlx_env()
-        drv = self._get_xlx(self.args.resource)
+        drv = self._get_fpga(self.args.resource)
 
         processwrapper.enable_print()
         drv.program_bitstream(self.args.bitstream)
@@ -1784,23 +1784,23 @@ def main():
     tmc_subparser.add_argument('action', choices=['info', 'values'])
     tmc_subparser.set_defaults(func=ClientSession.tmc_channel)
 
-    subparser = subparsers.add_parser('xlx', help="connect to a Xilinx Vivado hardware server")
+    subparser = subparsers.add_parser('fpga', help="interact with FPGA")
     subparser.set_defaults(func=lambda _: subparser.print_help())
-    xlx_subparsers = subparser.add_subparsers(
+    fpga_subparsers = subparser.add_subparsers(
         dest='subcommand',
         title='available subcommands',
         metavar="SUBCOMMAND",
     )
 
-    xlx_subparser = xlx_subparsers.add_parser('xsdb', help="run XSDB")
-    xlx_subparser.add_argument('-r,', '--resource', help="resource name")
-    xlx_subparser.add_argument('tcl_cmds', help="Tcl commands")
-    xlx_subparser.set_defaults(func=ClientSession.xlx_run_xsdb)
+    fpga_subparser = fpga_subparsers.add_parser('xsdb', help="run XSDB")
+    fpga_subparser.add_argument('-r,', '--resource', help="resource name")
+    fpga_subparser.add_argument('tcl_cmds', help="Tcl commands")
+    fpga_subparser.set_defaults(func=ClientSession.fpga_run_xsdb)
 
-    xlx_subparser = xlx_subparsers.add_parser('program-bitstream', help="program bitstream")
-    xlx_subparser.add_argument('-r,', '--resource', help="resource name")
-    xlx_subparser.add_argument('bitstream', type=pathlib.PurePath, help="bistream file")
-    xlx_subparser.set_defaults(func=ClientSession.xlx_program_bitstream)
+    fpga_subparser = fpga_subparsers.add_parser('program-bitstream', help="program bitstream")
+    fpga_subparser.add_argument('-r,', '--resource', help="resource name")
+    fpga_subparser.add_argument('bitstream', type=pathlib.PurePath, help="bitstream file")
+    fpga_subparser.set_defaults(func=ClientSession.fpga_program_bitstream)
 
     subparser = subparsers.add_parser('write-image', help="write an image onto mass storage")
     subparser.add_argument('-w', '--wait', type=float, default=10.0)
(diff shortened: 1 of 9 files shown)


sephalon:
I somehow missed that resource exports are already startet/stopped when a place is acquired/released; therefore I revised my initial statement accordingly.

In the meanwhile, I pushed some more commits which are partially still WiP; please see the respective commit messages and documentation changes.

labgrid/remote/client.py
```diff
@@ -1531,6 +1572,24 @@ def main():
     tmc_subparser.add_argument('action', choices=['info', 'values'])
     tmc_subparser.set_defaults(func=ClientSession.tmc_channel)
 
+    subparser = subparsers.add_parser('xlx', help="connect to a Xilinx Vivado hardware server")
```
Emantor, line 1575: In the long run we'll want to support tools like [openfpgaloader](https://github.com/trabucayre/openFPGALoader), this should rather be called `fpga`.

labgrid/remote/client.py
```diff
@@ -1531,6 +1572,24 @@ def main():
...
+    subparser.set_defaults(func=lambda _: subparser.print_help())
+    xlx_subparsers = subparser.add_subparsers(
+        dest='subcommand',
+        title='available subcommands',
+        metavar="SUBCOMMAND",
+    )
+
+    xlx_subparser = xlx_subparsers.add_parser('xsdb', help="run XSDB")
```
Emantor, line 1583: I guess this exports a gdb compatible interface, so something not xilinx specific like `gdb` would be better.

labgrid/remote/exporter.py
```diff
@@ -273,6 +274,79 @@ def _stop(self, start_params):
...
+            self.local.hw_server_cmd,
+            '-e', "'set jtag-port-filter {}'".format(self._get_serial()),
+            '-s{}'.format(self.local.agent_url),
+            '-p{}'.format(self.local.gdb_port),
+        ]
+        if self.local.log_level:
+            args.append('-l{}'.format(','.join(self.local.log_level)))
+        args.extend(self.local.extra_args)
+        # TODO: Add timeout
+        self.child = subprocess.Popen(' '.join(args), preexec_fn=os.setsid,
```
Emantor, line 332: Please use `start_new_session=True` instead of `preexec_fn=os.setsid`.

labgrid/remote/client.py
sephalon: Making the interface vendor-neutral sounds like a good idea; the `xlx` subparser was more like a first try.

I have the following two high-level `fpga` subcommands in mind:
- `program-bitstream`: Program bitstream via `xsdb` (Xilinx), `quartus_pgm` (Intel) or `openFPGALoader` (vendor-neutral) into volatile memory
- `program-flash`: Program data (bitstream and software for SoCs) via `program_flash` (Xilinx), `quartus_hps` (Intel) or `openFPGALoader` (vendor-neutral) into non-volatile memory

In additon, there could be vendor-specific, lower level commands like `xsdb` for the Xilinx System Debugger Tcl-based shell.

I only have first-hand experience with the Xilinx toolchain; so feel free to fill in the gaps here (especially on the Lattice side).

labgrid/remote/client.py
sephalon: This is a proprietary, Tcl-based shell; see also my other comment about the client interface architecture.

labgrid/remote/exporter.py
sephalon: Done

labgrid/remote/client.py
sephalon: I have implemented the interface as proposed.

sephalon:
We have had this feature under testing for a while now and came up with a few bug fixes as well as a slight refactoring of the XSDB interface. From our side, this feature is done; I will squash the commit history a bit more after your feedback.

sessl3r:
I'm the new guy @missinglinkelectronics having an eye on our labgrid PRs and trying to step in for @sephalon here.
What is the status on this PR in your opinion? We are using this feature internally since we opened the PR without any issues. Are you interested in it? Shall we change something more?
I see you want some focus on openfpgaloader. This is a good alternative but in many cases will not be sufficient in all cases (for neither of the fpga vendors we know).
Please let me known how we should proceed here.

labgrid/resource/udev.py
```diff
@@ -354,6 +361,31 @@ def filter_match(self, device):
...
+        self.match['DEVTYPE'] = 'usb_device'
+        super().__attrs_post_init__()
+
+    def filter_match(self, device):
+        match = (device.properties.get('ID_VENDOR_ID'), device.properties.get('ID_MODEL_ID'))
+
+        if match not in [("0403", "6010"),  # Trenz Electronic TE0790-03
+                         ("0403", "6014"),  # Digilent JTAG-SMT2/JTAG-HS3
+                         ("03fd", "0013"),  # Xilinx Platform Cable USB
+                         ("03fd", "0008")]: # Xilinx Platform Cable USB II
```
ozan956, line 384: The match list is Xilinx/Digilent/Trenz only. 

There's overlap with the existing USBDebugger resource (e.g. 0403:6010, 0403:6014 are matched by both), and USBDebugger has since grown a broader adapter table. 

Worth deciding whether XSDBDriver should just bind the existing USBDebugger/NetworkUSBDebugger (like OpenOCDDriver does) rather than introduce a parallel Xilinx-only resource, it helps to avoid two resources fighting over the same FT2232 devices.

ozan956 (commented):
Thanks for this.

A few things:

API is quite thin for real FPGA-SoC bring-up. run() + program_bitstream() covers "program the PL," but the common flows need more like:
- program PL and download a kernel/elf to a soft or hard core and run it
- Zynq/ZynqMP bring-up with rst -system + ps7_init/psu_init before dow.

Would you be open to a couple of higher-level methods (e.g. load_fabric, load_elf) so this covers Microblaze and Zynq bring-up, not just bitstream load?

File staging support also needed. program_bitstream/run reference paths that must already exist on the host running xsdb. For a remote exporter that means the bitstream/elf has to be put there before.  ManagedFile(path, self.interface).sync_to_resource() (as OpenOCDDriver do) would let callers pass a local path and have it copied to the exporter automatically.

ozan956:
Hey, @sephalon @sessl3r,

Thanks for opening/following up on this. Are you still interested in that?

I also would like to see this to be merged.

@Emantor what is your current thoughts on the PR?

sessl3r:
Hey @ozan956,
internally we use this feature (and a similar one for QuartusPGM) productive, but we haven't updated this PR.

1. udev.py: we completely removed the matching and accept just everything now - as the user already selected a correct device
2. API for PS: what we implemented is switching bootmode, everything else is up to the user for now (you can append to the driver). Downloading elf baremetal would be a point which might make sense for us internally. Downloading other things in our systems is most of the times done via sd-mux or the buildsystems

Have a look here for our current state: https://github.com/missinglinkelectronics/labgrid/blob/v24.0.0%2Bmle/labgrid/driver/xsdbdriver.py

ozan956:
> Hey @ozan956, internally we use this feature (and a similar one for QuartusPGM) productive, but we haven't updated this PR.
> 
> Have a look here for our current state: https://github.com/missinglinkelectronics/labgrid/blob/v24.0.0%2Bmle/labgrid/driver/xsdbdriver.py

Thanks @sephalon, I will take a closer look. It is great to hear that this is already functional and working well on your side :smile: 

I would also be happy to see this PR rebased onto the current master so we can move it forward.

PR #709, opened 2021-01-22
Test suite fixes

Emantor:
**Description**

Various fixes in the test suite.

**Checklist**
- [x] PR has been tested


tests/test_crossbar.py
```diff
@@ -6,18 +6,19 @@
 
 import pytest
 import pexpect
-import psutil
```
jluebbe, line 9: Why not use https://docs.pytest.org/en/latest/skipping.html#skipping-on-a-missing-import-dependency?

tests/test_pyvisa.py
```diff
@@ -1,6 +1,12 @@
+from importlib.util import find_spec
+
 from labgrid.resource.pyvisa import PyVISADevice
 from labgrid.driver.pyvisadriver import PyVISADriver
 
+import pytest
+
+pytestmark = pytest.mark.skipif(not find_spec("pyvisa"),
+                              reason="crossbar required")
```
jluebbe, line 9: Here `pytest.importorskip` should work as well.

tests/test_crossbar.py
Emantor: Because I didn't know this exists.

PR #712, opened 2021-01-29
target: duplicate bindings under differing names

Emantor:
**Description**

The duplicate binding mechanism forbid the duplicate binding of the same
supplier:
```
  class DriverWithMultiA(Driver):
      bindings = {
        "res1": ResourceA,
	"res2": ResourceA,
      }
```
Is not allowed since the same resource is bound twice. We do howver have
Driver implementing two protocols, like the QEMUDriver:
```
  class QEMUDriver(ConsoleExpectMixin, Driver, PowerProtocol, ConsoleProtocol):
```
Here we do want to allow the binding of the QEMUDriver under differing
Names, namely binding it as a PowerProtocol and ConsoleProtocol.
Rework the code to keep a set of (requirement, supplier) to check
against and prevent binding against the same class with a duplicate
binding error, however allow it if the supplier was found under a
different requirement.

**Checklist**
- [x] Tests for the feature 
- [ ] PR has been tested
- [x] Test for use with Optional Bindings, i.e. ({A, None} vs. {B, None})
- [x] readjust error messages

### DIFF
--- a/labgrid/target.py
+++ b/labgrid/target.py
@@ -293,6 +293,7 @@ class Target:
 
         # locate suppliers
         bound_suppliers = []
+        bound_req_pairs = set()
         for name, requirements in client.bindings.items():
             explicit = False
             if isinstance(requirements, Driver.NamedBinding):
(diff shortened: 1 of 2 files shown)


jluebbe (changes requested):
As discussed internally, this needs a test that optional bindings are handled correctly. Also the commit message should use the protocol-based example.

jluebbe (approved):

PR #715, opened 2021-02-02
driver/sshdriver: add retry for fast failures

### LINKED ISSUE #641: SSHDriver not working reliably but sometimes asking for users password
Hi there,

just using labgrid (v0.2.1 on Ubuntu 18.04) within pytest. I defined some basic local.yaml and some simple test which just executes the hostname command on a remote and checks if the hostname is correct (minimized example for this issue). This is working but not always. When I execute the labgrid-pytest command multiple times it sometimes asks for the password and sometimes it just works as expected and passes.

```
local.yaml
targets:
  c1:
    resources:
      NetworkService:
        address: 'pc1'
        username: 'user'
        password: 'pass'
    drivers:
      SSHDriver: {}
```

```
test_hostname.py
import pytest
from labgrid.protocol import CommandProtocol

@pytest.fixture(scope="module")
def client1(env):
  t = env.get_target("c1")
  return t.get_driver(CommandProtocol)

def test_hostname(client1):
  assert client1.run("echo $(hostname)") == (["pc1"], [], 0)
```

Commandline:
`$ labgrid-client --lg-env=local.yaml`

Result:
Sometimes it passes as one would expect.
Sometimes the following is printed:

```
[...]
collected 2 items

test_hostname.py user@pc1 password:
```

Here I can enter nothing and ssh asks me again. If the correct password is entered it is also not working interestingly. However it should never ask me, and of course not randomly.

With -vvv some more info is given:
```
client1 = SSHDriver(target=Target(name='client1', env=Environment(config_file='local.yaml')), name=None, state=<BindingState.active: 2>, keyfile='', stderr_merge=False)

    def test_hostname(client1):
>       assert client1.run("echo $(hostname)") == (["pc1"], [], 0)
E    AssertionError: assert ([],\n ['Control socket connect(/tmp/labgrid-ssh-tmp-w_vfgo72/control-pc1): '\n  'Connection refused\r',\n  'Permission denied, please try again.\r',\n  'Permission denied, please try again.\r',\n  'user@pc1: Permission denied (publickey,password).\r'],\n 255) == (['pc1'], [], 0)
E      At index 0 diff: [] != ['pc1']
E      Full diff:
E        (
E      +  ['pc1'],
E         [],
E      +  0,
E      -  ['Control socket connect(/tmp/labgrid-ssh-tmp-w_vfgo72/control-pc1): '
E      -   'Connection refused\r',
E      -   'Permission denied, please try again.\r',
E      -   'Permission denied, please try again.\r',
E      -   'user@pc1: Permission denied (publickey,password).\r'],
E      -  255,
E        )
```

This output is gained if I hit enter three times without giving a password. So this output sounds like no password was found in the environment.

Already tried some things which seem to change how often the password request happens (just an feeling, not really tested it long enough):
* Initally the scope of the fixture was session, tested module and also none==function
* Check if normal ssh to same pc with same credentials is reliable (yes it is)
* Executed in a clean session (no tmux, special bash configs etc.)

esben:
**Description**

Implement retry loop around start of ssh controlmaster connection,
avoiding "connection refused" errors caused by targets starting sshd after network interface is brought up.

Fixes #641

### DIFF
--- a/labgrid/driver/sshdriver.py
+++ b/labgrid/driver/sshdriver.py
@@ -5,6 +5,7 @@ import os
 import shutil
 import subprocess
 import tempfile
+import time
 
 import attr
 
@@ -15,6 +16,7 @@ from .common import Driver
 from ..step import step
 from .exception import ExecutionError
 from ..util.proxy import proxymanager
+from ..util.timeout import Timeout
 
 
 @target_factory.reg_driver
@@ -58,7 +60,23 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
     def _start_own_master(self):
         """Starts a controlmaster connection in a temporary directory."""
 
-        timeout = 30
+        timeout = Timeout(30.0)
+
+        # Retry start of controlmaster, to allow handle failures such as
+        # connection refused during target startup
+        connect_timeout = round(timeout.remaining)
+        while True:
+            if connect_timeout == 0:
+                raise Exception("Timeout while waiting for ssh connection")
+            try:
+                return self._start_own_master_once(connect_timeout)
+            except ExecutionError as e:
+                if timeout.expired:
+                    raise e
+                time.sleep(0.5)
+                connect_timeout = round(timeout.remaining)
+
+    def _start_own_master_once(self, timeout):
 
         self.tmpdir = tempfile.mkdtemp(prefix='labgrid-ssh-tmp-')
         control = os.path.join(


labgrid/driver/sshdriver.py
```diff
@@ -60,6 +61,19 @@ def _start_own_master(self):
 
         timeout = 30
 
+        # Retry start of controlmaster, to allow handle failures such as
+        # connection refused during target startup
+        retry_deadline = time.time() + timeout
```
jluebbe, line 66: Please use labgrid's Timeout helper class, which also correctly uses CLOCK_MONOTONIC.

labgrid/driver/sshdriver.py
```diff
@@ -60,6 +61,19 @@ def _start_own_master(self):
...
+        while True:
+            try:
+                return self._start_own_master_once(timeout)
+            except ExecutionError as e:
```
jluebbe, line 70: This should include a short sleep (maybe 0.5 seconds) to avoid degrading into a busy poll.

jluebbe (changes requested):

labgrid/driver/sshdriver.py
esben: Done.

labgrid/driver/sshdriver.py
esben: Done.

jluebbe (approved):
Looks fine to me. @Emantor?

Emantor (approved):

PR #716, opened 2021-02-02
Labgrid __version__ and client version support

### LINKED ISSUE #714: add a way to find out the installed version
This was already wished for in #585, but not addressed there, and today I had the same use case. Especially when installed from a source tree, the git revision would be nice to know too. A simple `labgrid-client version` could do the trick, and I think the version is available in the setuptools metadata…?

Emantor:
**Description**
PEP 396 compliant `__version__` and output in the labgrid-client command.

**Checklist**
- [ ] Tests for the feature 
- [x] PR has been tested

Fixes #714

### DIFF
--- a/labgrid/__init__.py
+++ b/labgrid/__init__.py
@@ -5,3 +5,7 @@ from .exceptions import NoConfigFoundError
 from .factory import target_factory
 from .step import step, steps
 from .stepreporter import StepReporter
+
+from importlib.metadata import version
+
+__version__ = version("labgrid")
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -26,6 +26,7 @@ from .. import Target, target_factory
 from ..util.proxy import proxymanager
 from ..util.helper import processwrapper
 from ..driver import Mode
+from .. import __version__ as lg_version
 
 txaio.use_asyncio()
 txaio.config.loop = asyncio.get_event_loop()
@@ -1208,6 +1209,9 @@ class ClientSession(ApplicationSession):
             print("Reservation '{}':".format(res.token))
             res.show(level=1)
 
+    async def print_version(self):
+        print("Labgrid version", lg_version)
+
 
 def start_session(url, realm, extra):
     from autobahn.asyncio.wamp import ApplicationRunner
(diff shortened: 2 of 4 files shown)


rohieb:
Hmm…
```
$ labgrid-client version
Could not connect to coordinator: [Errno 111] Connect call failed ('127.0.0.1', 20408)
```

Is this supposed to happen? I would expect that it also works without a coordinator.

labgrid/__init__.py
```diff
@@ -5,3 +5,7 @@
 from .factory import target_factory
 from .step import step, steps
 from .stepreporter import StepReporter
+
+from importlib.metadata import version
```
Bastian-Krause, line 9: Wrong import order: standard imports first, then third-party libraries, then local imports.

labgrid/remote/client.py
```diff
@@ -1208,6 +1209,9 @@ async def print_reservations(self):
             print("Reservation '{}':".format(res.token))
             res.show(level=1)
 
+    async def print_version(self):
+        print("Labgrid version", lg_version)
```
Bastian-Krause, line 1213: Just an idea: we could additionally print the exporter version here, right? AFAICS there is no "coordinator version" exposed, is it?

Bastian-Krause (changes requested):

labgrid/remote/client.py
jluebbe: Currently, the client doesn't directly access the exporter.

labgrid/__init__.py
jluebbe: Also, importlib.metadata is not available in 3.6 and marked as provisional:
https://docs.python.org/3/library/importlib.metadata.html

labgrid/remote/client.py
Bastian-Krause: Too bad, okay.

labgrid/__init__.py
Emantor: The alternative is too require setuptools at runtime.

Emantor:
I did some testing with `time` and `labgrid-client`, `pkg_resources` adds a measurable 20% to startup time, while `importlib_metadata` is barely noticable. I am still not entirely sure how useful this is since we require a running coordinator anyway. If you want to fetch the currently available version, you can also execute `python -c "import labgrid; print(\"labgrid version\", labgrid.__version__)"`, which does not require any further support from labgrid.

Emantor:
I am closing this, IMO we don't get much benefit from this. If you are curious about your currently installed version, `pip freeze` will always have the correct answer. For development environments the solution presented here is even more confusing, since the version will be the one which was created when `pip install -e .` was called and not the currently checked out commit.

PR #717, opened 2021-02-02
driver/power: add support for Siglent SPD3000X series power supplies

esben:
**Description**

Add support for using Siglent SPD3000X series power supplies, such as SPD3303X and SPD3303X-E with NetworkPowerDriver.

**Checklist**
- [x] Tests for the feature 
- [x] PR has been tested


requirements.txt
```diff
@@ -12,3 +12,4 @@ PyYAML==5.3.1
 ansicolors==1.1.8
 pyusb==1.1.0
 six>=1.13.0
+python-vxi11==0.9
```
jluebbe, line 15: Hmm, this way, it will be installed for every labgrid user, although probably only a few will use it.

Maybe wrap the import in a `try:`/`except:` and tell the user to install the module? You could document the requirement in the NetworkPowerDriver section in doc/configuration.rst, as well.

labgrid/driver/power/siglent.py
```diff
@@ -0,0 +1,21 @@
+""" tested with Siglent SPD3303X-E, and should be compatible with all SPD3000X series modules"""
+
+import vxi11
+
+
+def power_set(host, index, value):
+    index = int(index)
+    assert 1 <= index <= 2
+    value = "ON" if value else "OFF"
+    psu = vxi11.Instrument(host)
```
jluebbe, line 10: Is this using a fixed TCP port? If so, you should set `PORT` in the backend module, as then labgrid can do SSH port forwarding. (see netio_kshell.py)

requirements.txt
esben: True.

But it is a pure-python module, 2144 lines of code (including comments and blank lines), and no dependencies.
Are you sure it is worth making it more difficult to use to avoid pulling it in?

It is worth noticing that the protocol is not Siglent specific, but could be used to extend support to many other VXI-11 compatible instruments.

If you still think that I should wrap the import, where do you want that done?  In siglent.py or i powerdriver.py?

labgrid/driver/power/siglent.py
esben: It is a bit more complicated than that.  VXI-11 protocol specifes both a TCP and a UDP port.

jluebbe: Hmm, can it work with either? How would it select TCP vs. UDP?

requirements.txt
jluebbe: In siglent.py. For special cases like this, we also have {modbus,onewire,snmp,xena}-requirements.txt. So maybe a vxi-requirements? And also add it to dev-requirements so that you get all of them for development.

esben: Ok.

But shouldn't it suffice to move the dependency to vxi11-requirements.txt?
The siglent.py module is only going to be imported when a `NetworkPowerPort` is configured with `modul: siglent`.
And cannot come up with something much more sensible than just raising an exception, which the import does anyway.

If you take the `onewire` module example, it simply calls
`
    self._module = import_module('onewire')
`

which end up throwing an exception, just as the import in siglent.py will.

labgrid/driver/power/siglent.py
esben: I don't know, I just used it, and it worked :)

esben: Actually, looking at the python-vxi11 code, it seems like UDP is only used for VXI-11 device discovery, which doesn't seem to work with the Siglent device I have here anyway.

But, the TCP communication uses port 111 (RPC/portmap), which returns another port, which is then used for the remaining communication.  In my case it was 9009, but I don't know if that is guaranteed to be a fixed port number.

jluebbe: Oh, OK. Why do people use portmapper...

The leave out the PORT entry and we'll have to figure out forwarding later, if needed.

requirements.txt
jluebbe: You're right. That's fine.

labgrid/driver/power/siglent.py
```diff
@@ -0,0 +1,21 @@
+""" tested with Siglent SPD3303X-E, and should be compatible with all SPD3000X series modules"""
```
jluebbe, line 1: Maybe this should be called siglent_spd3000x.py to avoid confusion? You could also mention the "model" name in the changelog.

jluebbe (changes requested):
Also, your Signed-off-by is missing.

labgrid/driver/power/siglent.py
esben: I did consider that, but that would mean that `siglent_spd3000x` would be encoded in configuration files.
But as other Siglent PSU's are likely to be easy to support with the same code, such as SPD1000X series, I think it makes most sense to use a more generic name.
Only change I needed for other Siglent VXI-11 supplies would most likely be to adapt the check for index values, and probably also the decoding of status bits.

jluebbe: Maybe siglent_spd or _vxi11 instead?

esben: I would go for siglent_vxi11 then.  And I should use `_` and not `-` ?

jluebbe: `-` is not a valid character in python module names.

If you're confident that siglent won't switch, I'm fine with that. @Emantor?

esben: VXI-11 is based on HPIB/GPIB, which was defined in the late 1960's.  Instrument control is not an area with much new-shiny in the protocol development :grin:

esben: @jluebbe Is it okay to resolve this as it is now?

jluebbe (approved):

jluebbe:
The CI issue is unrelated and was fixed with #721.

PR #720, opened 2021-02-03
Fix DuplicateBinding Error for optional bindings of the same protcol

Emantor:
**Description**

Allow optional bindings to the same protocol if both are not available. We simply ignore this case since we don't care about a duplicate binding than.

**Checklist**

- [x] Tests for the feature 
- [x] PR has been tested

### DIFF
--- a/labgrid/target.py
+++ b/labgrid/target.py
@@ -347,7 +347,7 @@ class Target:
                 raise NoSupplierFoundError("conflicting suppliers matching {} found in target {}".format(requirements, self))  # pylint: disable=line-too-long
             else:
                 supplier = suppliers[0]
-            if (requirement, supplier) in bound_req_pairs:
+            if supplier is not None and (requirement, supplier) in bound_req_pairs:
                 raise BindingError(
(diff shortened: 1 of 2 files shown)


labgrid/target.py
```diff
@@ -349,8 +349,8 @@ def bind_driver(self, client):
                 supplier = suppliers[0]
             if supplier is not None and (requirement, supplier) in bound_req_pairs:
                 raise BindingError(
-                    "duplicate bindings of {} for {} found in target {}".format(
-                        supplier, requirement, self)
+                    "duplicate bindings of {} to {} for {} found in target {}".format(
+                        supplier, supplier_name, requirement, self)
```
jluebbe, line 353: This should be `name` instead of `supplier_name`, as we're interested in the name of the binding here.

jluebbe (changes requested):

labgrid/target.py
Emantor: Right, fixed.

jluebbe (approved):

PR #723, opened 2021-02-04
driver/power: add support for REST based relays controlling power

esben:
**Description**

Add a new driver to support relay based power control where relay is controlled with a REST API with a simple GET/PUT request to the same URL.

This is different from simplerest driver, as it relies on REST interface using
GET to query status, and PUT to change status, where as simplerest driver uses
GET for both purposes.

These two models are therefore inherently incompatible, and supplements each
other.

**Checklist**
- [x] Tests for the feature 
- [x] PR has been tested

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,6 +15,7 @@ New Features in 0.4.0
   supports 8 or 256 colors, and defaults to the respective dark variant.
   The 256-color schemes now use purple instead of green for the ``run`` lines to
   make them easier distinguishable from pytest's "PASSED" output.
+- Network controlled relay providing GET/PUT based REST API
 
 Breaking changes in 0.4.0
 ~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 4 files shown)


Emantor:
We already have the `simplerest` model for this which coincidentally implements almost the same functions.

jluebbe:
An alternative might be to add this to the existing simplerest driver via a new attribute on the resource... But as almost every line would be different, I'm not sure that would be better.

We need some documentation on when to choose either, though. Otherwise it would be too confusing.

esben:
@jluebbe @Emantor Do you have somewhere in mind where this kind of documentation should be placed?  I didn't find any place where the existing `NetworkPowerDriver` models where documented, other than in the source files themselves.
Would it be sufficient if I try to cleanup the documentation in the top of labgrid/driver/power/rest.py ?

jluebbe:
I've added some documentation on the available backends to the NetworkPowerDriver section.

jluebbe (approved):

PR #727, opened 2021-02-11
ubootdriver: allow overriding of await boot timeout

ynezz:
**Description**

Current 30 second timeout used by default in `await_boot()` might not be
enough in some corner cases where we for example boot large initramfs
images over network, which take some time to load, decompress and boot:

```
 await_boot state='start'
    expect state='start', args={'pattern': 'Linux version \\d'}
    expect state='stop', result=(0, b"bootm 0x83000000 ... b'Linux version 4'), duration=42.025
```

So make it possible to override the default timeout via new `boot_timeout` variable.

**Checklist**
- [x] PR has been tested on ipq40xx based device

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,8 @@ New Features in 0.4.0
 
 - The `NetworkPowerDriver` now additionally supports:
   - Siglent SPD3000X series power supplies
+- UBootDriver now allows overriding of currently fixed await boot timeout
+  via new ``boot_timeout`` argument.
 
 Breaking changes in 0.4.0
 ~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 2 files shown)


Emantor:
I have nothing against this change, just wondering why it takes the device 42 seconds to get to the kernel version string. Decompression without caches?

ynezz:
> Just wondering why it takes the device 42 seconds to get to the kernel version string. Decompression without caches?

You know, it's just another downstream U-Boot on [GL.iNet GL-B1300](https://openwrt.org/toh/gl.inet/gl.inet_gl-b1300) device with IPQ4028 SoC and I don't have a clue, why it's happening. That initramfs image has 15MB compressed and it simply takes ages to boot even if I do it manually.

Emantor (approved):

PR #728, opened 2021-02-11
docker: base: fix Python cryptography build failure

ynezz:
It seems like recent Python cryptography package versions need Rust
compiler present for OpenSSL wrapper, which leads to following error:

```
 generating cffi module 'build/temp.linux-x86_64-3.7/_openssl.c'
 running build_rust
 ...
 error: Can not find Rust compiler
```

There is following note regarding this issue in the documentation:

> If you are on RHEL/CentOS/Fedora/Debian/Ubuntu or another distribution
 derived from the preceding list, then you should upgrade pip (in a
 virtual environment!) and attempt to install cryptography again before
 trying to install the Rust toolchain. These platforms will receive a
 binary wheel and require no compiler if you have an updated pip!

So fix this by recommended pip update.

References: https://cryptography.io/en/latest/installation.html#rust

### DIFF
--- a/dockerfiles/base/Dockerfile
+++ b/dockerfiles/base/Dockerfile
@@ -9,6 +9,7 @@ COPY ./ /opt/labgrid/
 RUN set -e ;\
     apt update -q=2 ;\
     apt install -q=2 --yes --no-install-recommends python3 python3-dev python3-pip python3-setuptools python3-wheel git build-essential libsnappy-dev ;\
+    pip3 install -U pip;\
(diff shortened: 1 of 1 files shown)


ynezz:
The failure is visible for example here https://github.com/labgrid-project/labgrid/pull/727/checks?check_run_id=1879138064#step:3:779

Emantor (approved):

krevsbech:
Just saw your fix now. Thanks for fixing it!

PR #730, opened 2021-02-14
ColoredStepReporter: be compatible with 8-color terminals

rohieb:
The default color schemes use 256-color escape sequences, which are not fully supported by all terminals (e.g. the Linux virtual console).

First refactor the exiting code a bit.

Then add two more dark and light color schemes that only use the original 8 colors. Then try to detect if the terminal in use understands the enhanced escape codes, and in that case use the 256-color dark scheme by default.

As a side effect, the two new color schemes also work well in terminals with "solarized" color schemes.

**Checklist**
- [x] Tests for the feature 
- [x] PR has been tested

### DIFF
--- a/labgrid/pytestplugin/reporter.py
+++ b/labgrid/pytestplugin/reporter.py
@@ -129,7 +129,7 @@ class StepReporter:
 class ColoredStepReporter(StepReporter):
     EVENT_COLORS_DARK = {
         'expect$': 8, # dark gray (a lot of output, blend into background)
-        'run': 10, # green
+        'run': 129, # light purple
         'state_': 51, # light blue
         'transition$': 45, # blue
         'cycle$|on$|off$': 246, # light gray
@@ -137,7 +137,7 @@ class ColoredStepReporter(StepReporter):
 
     EVENT_COLORS_LIGHT = {
         'expect$': 250, # light gray (a lot of output, blend into background)
-        'run': 10, # green
+        'run': 93, # purple
         'state_': 51, # light blue
         'transition$': 45, # blue
         'cycle$|on$|off$': 8, # dark gray
--- a/labgrid/pytestplugin/reporter.py
+++ b/labgrid/pytestplugin/reporter.py
@@ -128,6 +128,22 @@ class StepReporter:
 
 class ColoredStepReporter(StepReporter):
     EVENT_COLORS_DARK = {
+        'expect$': 'blue', # (a lot of output, blend into background)
+        'run': 'magenta',
+        'state_': 'cyan',
+        'transition$': 'yellow',
+        'cycle$|on$|off$': 'white',
+    }
+
+    EVENT_COLORS_LIGHT = {
+        'expect$': 'white', # (a lot of output, blend into background)
+        'run': 'magenta',
+        'state_': 'cyan',
+        'transition$': 'yellow',
+        'cycle$|on$|off$': 'blue',
+    }
+
+    EVENT_COLORS_DARK_256COLOR = {
         'expect$': 8, # dark gray (a lot of output, blend into background)
         'run': 129, # light purple
         'state_': 51, # light blue
@@ -135,7 +151,7 @@ class ColoredStepReporter(StepReporter):
         'cycle$|on$|off$': 246, # light gray
     }
 
-    EVENT_COLORS_LIGHT = {
+    EVENT_COLORS_LIGHT_256COLOR = {
         'expect$': 250, # light gray (a lot of output, blend into background)
         'run': 93, # purple
         'state_': 51, # light blue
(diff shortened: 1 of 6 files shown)


rohieb:
Hmmm, on second thought, the green color for the "run" lines is really similar to the green "PASSED" output of pytest. So I think green and red should not be used at all in the ColoredStepReporter. Should I remap the "run" lines to something else? Magenta seems to work well.

rohieb:
> So I think green and red should not be used at all in the ColoredStepReporter. Should I remap the "run" lines to something else? Magenta seems to work well.

Did that now, and also rebased to master to get rid of the pthonyn-cryptography issue.

labgrid/pytestplugin/reporter.py
```diff
@@ -128,29 +128,57 @@ def pytest_runtest_logreport(self, report):
...
     def __init__(self, terminalreporter, *, rewrite=False):
         super().__init__(terminalreporter, rewrite=rewrite)
 
-        scheme = os.environ.get('LG_COLOR_SCHEME', 'dark')
-        if scheme == 'light':
-            self.color_scheme = ColoredStepReporter.EVENT_COLORS_LIGHT
-        else:
-            self.color_scheme = ColoredStepReporter.EVENT_COLORS_DARK
+        default_scheme = 'dark'
+        if os.environ.get('TERM', '').endswith("-256color"):
```
Emantor, line 173: This is hugely inaccurate, please use something like the `infocmp` command to retrieve the supported colors.

rohieb: Alright, `curses.tigetnum()` seems to do the trick.

labgrid/pytestplugin/reporter.py
```diff
@@ -128,29 +129,58 @@ def pytest_runtest_logreport(self, report):
...
+        'run': 'magenta',
+        'state_': 'cyan',
+        'transition$': 'yellow',
+        'cycle$|on$|off$': 'white',
+    }
+
+    EVENT_COLORS_LIGHT = {
+        'expect$': 'white', # (a lot of output, blend into background)
+        'run': 'magenta',
+        'state_': 'cyan',
```
rohieb, line 142: BTW, I could not find anything that uses `state_`…

rohieb:
Apparently the curses module is too smart/lazy to re-read the terminal database more than once, so there are no tests for the part where the default color scheme is determined.

labgrid/pytestplugin/reporter.py
```diff
@@ -3,6 +3,7 @@
 import os
 import re
 import colors
+import curses
```
Emantor, line 6: This new dependency needs to be described in setup.py and the requirements files as well. We should only import this if the `ColoredStepReporter`  is used and raise an appropriate error message if curses is not available, with a fallback to the old behaviour of assuming 256 colors.

rohieb: `curses` is part of the Python standard library.

Emantor (approved):

labgrid/pytestplugin/reporter.py
rohieb: Nevertheless, it might be disabled when compiling Python, so I pushed a fix for that nevertheless.

labgrid/pytestplugin/reporter.py
Bastian-Krause: The GraphStrategy's state methods use that.

labgrid/pytestplugin/reporter.py
```diff
@@ -128,29 +128,62 @@ def pytest_runtest_logreport(self, report):
...
+            default_scheme = 'dark'
+            curses.setupterm()
+            if curses.tigetnum("colors") >= 256:
+                default_scheme = 'dark-256color'
+        except ModuleNotFoundError:
+            default_scheme = 'dark-256color'
+
+        scheme = os.environ.get('LG_COLOR_SCHEME', default_scheme)
+        if not scheme in ColoredStepReporter.EVENT_COLOR_SCHEMES.keys():
+            logging.warning("Color scheme '{}' unknown".format(scheme))
```
Bastian-Krause, line 183: Please use lazy % formatting in logging.

rohieb: Hmm. Is there any rules when to use which? Both variants seem to be all over the code.

Emantor: Yes, logging functions always use the `%` formatting syntax, this is even a linting errors. Other strings in labgrid usually use `"".format()`, but can be replaced with f-strings in many cases. This is now possible since we dropped python 3.5.

rohieb: Ahh, you were talking about something like `logging.warning("%s", "foobar")`, but I thought of Python's string interpolation like `"%s" % "foobar"`. Yes, the former makes more sense here since the API directly supports it.

rohieb: Fixed now.

rohieb:
I also added a snippet for CHANGES.md, and another test to get two lines more coverage… :)

labgrid/pytestplugin/reporter.py
```diff
@@ -128,29 +128,62 @@ def pytest_runtest_logreport(self, report):
...
+            import curses
+            default_scheme = 'dark'
+            curses.setupterm()
+            if curses.tigetnum("colors") >= 256:
+                default_scheme = 'dark-256color'
+        except ModuleNotFoundError:
+            default_scheme = 'dark-256color'
+
+        scheme = os.environ.get('LG_COLOR_SCHEME', default_scheme)
+        if not scheme in ColoredStepReporter.EVENT_COLOR_SCHEMES.keys():
```
Bastian-Krause, line 182:
```suggestion
        if scheme not in ColoredStepReporter.EVENT_COLOR_SCHEMES.keys():
```
Personally, I like this better, but it's up to you.

tests/test_stepreporter.py
```diff
@@ -72,3 +76,41 @@ def test_step_reporter(power_env, pw_cycle_test, color_scheme, pytest_extra_para
...
+
+    with open("/dev/null") as f:
+        csr = ColoredStepReporter(MockTerminalReporter(f))
+    assert csr.color_scheme == ColoredStepReporter.EVENT_COLOR_SCHEMES[result_color_scheme]
+
+    caplog_warnings = [x.message for x in caplog.records if x.levelno == logging.WARNING]
+    if warning_msg:
+        assert warning_msg in caplog_warnings
+    else:
+        assert [] == caplog_warnings
```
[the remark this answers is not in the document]
rohieb: Internalised behaviour from my C coding days… I like to do it this way more than `var == constant` because it errors out when I accidentally forget one `=` and it becomes an assignment. Your suggestion looks better, so I used it.

labgrid/pytestplugin/reporter.py
rohieb: Yep, looks better.

PR #733, opened 2021-02-16
Docker build updates

JPEWdev:
Updates the docker image build to do a multistage build for each client. This allows sharing the labgrid-base image as an temporary intermediate image without needing to officially publish it, which in turn makes derived images and building for multiple architectures eaiser.

Also updates a few of the packages included in the images

**Checklist**
- [ ] Tests for the feature 

- [x] PR has been tested


dockerfiles/Dockerfile
```diff
@@ -0,0 +1,119 @@
...
+
+VOLUME /opt/crossbar
+
+EXPOSE 20408
+
+CMD ["crossbar", "start", "--config", "/opt/labgrid/.crossbar/config.yaml"]
+
+#
+# ser2net
+#
```
krevsbech, line 54: I see the point as the distro release is quite old. However for maintainability would it be possible to inherit from another distro having a newer release?

krevsbech (commented):
Thanks for your contribution, it is a good idea to use staging. Besides for the ser2net thing I have no comments

dockerfiles/Dockerfile
JPEWdev: I need the latest release of ser2net (4.3.3) to fix a bug. 

If we want to stick with Debian, I think the only option would be to move to _testing_ (e.g. "bullseye").

krevsbech: I see no strong reason to stick to a certain distro if the image works with latest debian base, alpine, fine by me. 
But I see a burden of maintaining packages and dependencies on our self. So would it be possible to try with a distro?

JPEWdev: @krevsbech I checked Alpine and Fedora, but unfortunately it looks like no one has a version of ser2net newer that 3.5. I'm not quite sure why that is but regardless we would have to compile it manually with one of them also. I think that makes the only remaining option to switch to bullseye. Would you like to try that or compile it locally?

krevsbech: I see, then I agree with you that this is the best solution. I will go ahead and approve your contribution.

krevsbech (approved):

krevsbech (commented):
-

dockerfiles/Dockerfile
```diff
@@ -0,0 +1,122 @@
...
+        libssl-dev \
+        pkg-config \
+        file \
+        libyaml-dev \
+    && \
+    apt clean && \
+    rm -rf /var/lib/apt/lists/*
+
+RUN mkdir -p /src
+
```
krevsbech, line 76: Please add comment why we do add packages by our self and that this can be removed once the base image bump the package.

.github/workflows/docker.yml
```diff
@@ -16,6 +16,7 @@ jobs:
         password: ${{ secrets.DOCKERHUB_TOKEN }}
     - name: Build docker image
       run: |
+        pip3 install --user setuptools
```
krevsbech, line 19: Is this the reason why the build fails

krevsbech (changes requested):
Just minors, please add comment for the ser2net stuff and ensure that the pipeline does not fail

JPEWdev:
@krevsbech OK, I had a few problems getting the docker CI job working, sorry for the noise. It should be ready now.

Thanks!

krevsbech (approved):

Issue #697, opened 2020-12-17
sshdriver: sshpass: fails when connecting through remote exporter


alegsan:
Hi,

our exporter configuration looks like that:
```
750-8216:                                                                           
  USBSerialPortService:                                                             
    cls: USBSerialPort                                                              
    match:                                                                          
      '@ID_SERIAL_SHORT': '1284744278'                                              
                                                                                    
  USBSerialPortSerial:                                                              
    cls: USBSerialPort                                                              
    match:                                                                          
      '@sys_name': '3-1.3.2.4.3:1.0'                                                
                                                                                    
  USBSDMuxDevice:                                                                   
    match:                                                                          
      ID_SERIAL_SHORT: '000000000563'                                               
                                                                                    
  NetworkService:                                                                   
    address: 192.168.1.157                                                          
    username: root                                                                  
    password: root                                                                                                
```

When we want to open a ssh session to the `NetworkService` from a remote client we get the exception:
```
(labgrid-venv) alegsan@nas:~/workspace/git/labgrid$ labgrid-client -vvv -P lc013508 -p 750-8216 ssh
  DEBUG: Creating SSHConnection for lc013508
  DEBUG: No existing control socket found
  DEBUG: ControlSocket: /tmp/labgrid-connection-vcga0abn/control-lc013508
  DEBUG: Master Start command: ssh -x -o LogLevel=ERROR -o PasswordAuthentication=no -n -MN -o ConnectTimeout=30 -o ControlPersist=300 -o ControlMaster=yes -o ControlPath=/tmp/labgrid-connection-vcga0abn/control-lc013508 -o StrictHostKeyChecking=yes lc013508
  DEBUG: Connected to lc013508
   INFO: Created new SSH connection to lc013508
  DEBUG: Started keepalive for lc013508
  DEBUG: Running control command: ssh -x -o LogLevel=ERROR -o PasswordAuthentication=no -o ControlMaster=no -o ControlPath=/tmp/labgrid-connection-vcga0abn/control-lc013508 -O forward -L52732:127.0.0.1:20408 lc013508
[...]

PR #737, opened 2021-02-22
driver/sshdriver: use SSH_ASKPASS instead of sshpass

bitshifter1:
**Description**

Fixes bug #697 when connecting to a place via ssh proxy. I have tried to use both, subprocess/Popen/write and pexpect/spawn/expect/sendline to write the password to stdin, but as sshpass, it only works when connecting without proxy and fails when connecting via ssh proxycommand. Then I found the builtin ssh option SSH_ASKPASS, which is the only left way to pass the password from a script, or a graphical user interface to the agent. It may not be the most elegant solution, but it works for us and it is robust.

**Checklist**
- [ ] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/labgrid/driver/sshdriver.py
+++ b/labgrid/driver/sshdriver.py
@@ -2,6 +2,7 @@
 """The SSHDriver uses SSH as a transport to implement CommandProtocol and FileTransferProtocol"""
 import logging
 import os
+import stat
 import shutil
 import subprocess
 import tempfile
@@ -82,10 +83,8 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
         control = os.path.join(
             self.tmpdir, 'control-{}'.format(self.networkservice.address)
         )
-        # use sshpass if we have a password
-        args = ["sshpass", "-e"] if self.networkservice.password else []
 
-        args += ["ssh", "-f", *self.ssh_prefix, "-x", "-o", "ConnectTimeout={}".format(timeout),
+        args = ["ssh", "-f", *self.ssh_prefix, "-x", "-o", "ConnectTimeout={}".format(timeout),
                  "-o", "ControlPersist=300", "-o",
                  "UserKnownHostsFile=/dev/null", "-o", "StrictHostKeyChecking=no",
                  "-o", "ServerAliveInterval=15", "-MN", "-S", control.replace('%', '%%'), "-p",
@@ -111,12 +110,21 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
             ]
 
         env = os.environ.copy()
+        pass_file = ''
         if self.networkservice.password:
-            env['SSHPASS'] = self.networkservice.password
+            fd, pass_file = tempfile.mkstemp()
+            os.fchmod(fd, stat.S_IRWXU)
+            #with openssh>=8.4 SSH_ASKPASS_REQUIRE can be used to force SSH_ASK_PASS
+            #openssh<8.4 requires the DISPLAY var and a detached process with start_new_session=True
+            env = {'SSH_ASKPASS': pass_file, 'DISPLAY':'', 'SSH_ASKPASS_REQUIRE':'force'}
+            with open(fd, 'w') as f:
+                f.write("#!/bin/sh\necho " + self.networkservice.password)
+
         self.process = subprocess.Popen(args, env=env,
                                         stdout=subprocess.PIPE,
                                         stderr=subprocess.STDOUT,
-                                        stdin=subprocess.DEVNULL)
+                                        stdin=subprocess.DEVNULL,
+                                        start_new_session=True)
 
         try:
             subprocess_timeout = timeout + 5
@@ -147,6 +155,9 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
                 "Subprocess timed out [{}s] while executing {}".
                 format(subprocess_timeout, args),
             )
+        finally:
+            if self.networkservice.password and os.path.exists(pass_file):
+                os.remove(pass_file)
 
         if not os.path.exists(control):
             raise ExecutionError(
--- a/labgrid/driver/sshdriver.py
+++ b/labgrid/driver/sshdriver.py
@@ -399,7 +399,7 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
 
     def _start_keepalive(self):
         """Starts a keepalive connection via the own or external master."""
-        args = ["ssh", *self.ssh_prefix, "cat"]
+        args = ["ssh", *self.ssh_prefix, self.networkservice.address, "cat"]
 
         assert self._keepalive is None
         self._keepalive = subprocess.Popen(
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -359,7 +359,7 @@ A NetworkService describes a remote SSH connection.
 The example describes a remote SSH connection to the computer `example.computer`
 with the username `root`.
 Set the optional password password property to make SSH login with a password
-instead of the key file (needs sshpass to be installed)
+instead of the key file.
 
 When used with ``labgrid-exporter``, the address can contain a device scope
 suffix (such as ``%eth1``), which is especially useful with overlapping address
--- a/dockerfiles/Dockerfile
+++ b/dockerfiles/Dockerfile
@@ -26,7 +26,7 @@ RUN set -e ;\
     pip3 install --no-cache-dir -r requirements.txt ;\
     SETUPTOOLS_SCM_PRETEND_VERSION="$VERSION" python3 setup.py install ;\
     apt update -q=2 ;\
-    apt install -q=2 --yes --no-install-recommends microcom openssh-client sshpass rsync jq qemu-system; \
+    apt install -q=2 --yes --no-install-recommends microcom openssh-client rsync jq qemu-system; \
     apt clean ;\
     rm -rf /var/lib/apt/lists/*
 
--- a/tests/test_docker.py
+++ b/tests/test_docker.py
@@ -12,10 +12,7 @@ from labgrid.exceptions import NoResourceFoundError
 
 
 def check_external_progs_present():
-    """Determine if host machine has sshpass and a usable docker daemon"""
-    from shutil import which
-    if not which('sshpass'):
-        return False
+    """Determine if host machine has a usable docker daemon"""
     try:
         import docker
         dock = docker.from_env()


labgrid/driver/sshdriver.py
```diff
@@ -111,12 +110,19 @@ def _start_own_master_once(self, timeout):
             ]
 
         env = os.environ.copy()
+        pass_file = ''
         if self.networkservice.password:
-            env['SSHPASS'] = self.networkservice.password
+            fd, pass_file = tempfile.mkstemp()
+            os.fchmod(fd, stat.S_IRWXU)
+            env = {'SSH_ASKPASS': pass_file, 'DISPLAY':'0:0'}
```
jluebbe, line 117: Newer SSH versions also have SSH_ASKPASS_REQUIRE, which we could set to `force`. Maybe we could set DISPLAY to a clearly invalid value without breaking this mechanism?

labgrid/driver/sshdriver.py
```diff
@@ -111,12 +110,19 @@ def _start_own_master_once(self, timeout):
...
+            with open(fd, 'w') as f:
+                f.write("echo " + self.networkservice.password)
```
jluebbe, line 119: This needs `#!/bin/sh\n`.

labgrid/driver/sshdriver.py
```diff
@@ -111,12 +110,19 @@ def _start_own_master_once(self, timeout):
...
+
         self.process = subprocess.Popen(args, env=env,
                                         stdout=subprocess.PIPE,
                                         stderr=subprocess.STDOUT,
-                                        stdin=subprocess.DEVNULL)
+                                        stdin=subprocess.DEVNULL,
+                                        preexec_fn=os.setsid)
```
jluebbe, line 125: Why do we need setsid here? Doesn't this risk leaving a process behind?

labgrid/driver/sshdriver.py
```diff
@@ -148,6 +154,9 @@ def _start_own_master_once(self, timeout):
                 format(subprocess_timeout, args),
             )
 
+        if self.networkservice.password and os.path.exists(pass_file):
+            os.remove(pass_file)
```
jluebbe, line 158: This should be in a try:/finally: block.

jluebbe (changes requested):
Thanks, this is a very interesting approach! I've got a few comments.

labgrid/driver/sshdriver.py
jluebbe: Hmm, this could also be used to do something like `echo $LG_SSH_PASSWORD`, right?

labgrid/driver/sshdriver.py
bitshifter1: I have tried openssh 8.4p1 nightly and with SSH_ASKPASS_REQUIRE=force it works without DISPLAY variable and without the preexec_fn=os.setsid.

labgrid/driver/sshdriver.py
bitshifter1: I believe in the pre 8.4 versions they do an extra check that the ssh call is detached from tty.
I'm not sure, I see four ssh processes in htop (even in the new 8.4 version with SSH_ASKPASS_REQUIRE=force and without setsid). How does it look like when you use sshpass and proxy?
Untitled

labgrid/driver/sshdriver.py
bitshifter1: I tried to clean up earlier, directly after Popen, but then ssh complains about the missing file. I'll do a finally after the big try.

bitshifter1:
On 22.02.21 17:20, Jan Lübbe wrote:
> *@jluebbe* commented on this pull request.
> 
> ------------------------------------------------------------------------
> 
> In labgrid/driver/sshdriver.py 
> <https://github.com/labgrid-project/labgrid/pull/737#discussion_r580384893>:
> 
>>          if self.networkservice.password:
> -            env['SSHPASS'] = self.networkservice.password
> +            fd, pass_file = tempfile.mkstemp()
> +            os.fchmod(fd, stat.S_IRWXU)
> +            env = {'SSH_ASKPASS': pass_file, 'DISPLAY':'0:0'}
> +            with open(fd, 'w') as f:
> +                f.write("echo " + self.networkservice.password)
> 
> Hmm, this could also be used to do something like |echo 
> $LG_SSH_PASSWORD|, right?
Hm, seems not to work, even with the 8.4 version and 
SSH_ASKPASS_REQUIRE=force. I thinks that's why you have to pass an env 
vector to Popen :(

labgrid/driver/sshdriver.py
```diff
@@ -82,10 +83,8 @@ def _start_own_master_once(self, timeout):
...
             self.tmpdir, 'control-{}'.format(self.networkservice.address)
         )
-        # use sshpass if we have a password
-        args = ["sshpass", "-e"] if self.networkservice.password else []
 
-        args += ["ssh", "-f", *self.ssh_prefix, "-x", "-o", "ConnectTimeout={}".format(timeout),
+        args = ["ssh", "-f", *self.ssh_prefix, "-x", "-o", "ConnectTimeout={}".format(timeout),
                  "-o", "ControlPersist=300", "-o",
                  "UserKnownHostsFile=/dev/null", "-o", "StrictHostKeyChecking=no",
                  "-o", "ServerAliveInterval=15", "-MN", "-S", control.replace('%', '%%'), "-p",
```
bitshifter1, line 90: Can we remove the ServerAliveInterval argument and leave it for the ~/.ssh/config? Because of the missing
ServerAliveCountMax value our session logges out after 60s.

Emantor: This line is missing a `["-F", "none"]`to ignore the user configuration file. ServerAliveInterval is correct here, since the keepalive started in `on_activate` should keep the socket open as long as your device is reachable. This breaks if you restart the target unexpectedly without deactivating the SSHDriver first and is a usage error.

jluebbe: We explicitly add `-F none` to `self.ssh_prefix` after starting the master. Are you sure that was not intentional and a change wouldn't break user configs (such as per device keys)?

jluebbe:
> > Hmm, this could also be used to do something like |echo $LG_SSH_PASSWORD|, right?

> Hm, seems not to work, even with the 8.4 version and SSH_ASKPASS_REQUIRE=force. I thinks that's why you have to pass an env vector to Popen :(

Yes, and we may also pass the env variable to the target, which would be surprising. (I removed the comment when I realized that ;)

labgrid/driver/sshdriver.py
Emantor: You mean the usecase where you have target keys configured in you SSH config? AFAICS this is a mistake and keys should be directly configured in the YAML. I am pretty sure this is a leftover when we removed the support for user supplied masters altogether and this didn't work for several years. I don't think we should support per user keys configured in the SSH config and instead want them to be defined explicitly in the YAML file.

jluebbe: OK. But then we should move the addition of `-F none` up instead of adding it to individual calls.

bitshifter1: Ok, I found the _start_keepalive function, but I had to add the address to the args in order for the "cat" command to be executed correctly. Unfortunately, I am still being logged out after 60s.

    def _start_keepalive(self):
             args = ["ssh", *self.ssh_prefix, self.networkservice.address, "cat"]
`

bitshifter1: > This line is missing a `["-F", "none"]`to ignore the user configuration file. ServerAliveInterval is correct here, since the keepalive started in `on_activate` should keep the socket open as long as your device is reachable. This breaks if you restart the target unexpectedly without deactivating the SSHDriver first and is a usage error.

@Emantor the start_keepalive keeps the socket open, but not the labgrid-client connection, right? I think the ServerAliveInterval could be removed, or increased, to get a higher connection timeout for the client.

bitshifter1, line 90: I just realized it, in line 90, instead of "-S", shouldn't it be "-o", "ControlPath=" + control.replace('%', '%%'), because a new socket needs to be created. The -S is for reusing, isn't it? I think the ControlMaster variable needs to be also set properly. Oh man, I am such a noob with openssh :)

bitshifter1:
Do I need to do anything else for this to progress? Please let me know I am still new to github :relaxed:

labgrid/driver/sshdriver.py
Emantor: To me the description in `man ssh` and `man ssh_config` read like `-S` and `-o ControlPath=` are equivalent. The only way to find out for sure is testing.

bitshifter1: @Emantor ok, never mind. We are using a dropbear server, which doesn't seem to work very well together with openssh ServerAliveInterval.

jluebbe: Is that a known issue for dropbear? I'd prefer to keep the ServerAlive, as otherwise we'd need to implement "ping" ourselves via the cat...

bitshifter1: It looks like the openssh ServerAliveInterval is just not supported by dropbear. I found a dropbear server option that allows to increase the idle timeout, but was not able to keep the connection open permanently. I am fine with how it works right now, as long as the connections stays open for longer than a minute.

krevsbech (commented):
If this approach is used we need to remove sshpass from the client docker (dockerfiles/Dockerfile) as it would not be a dependency anymore?(url)

bitshifter1:
> If this approach is used we need to remove sshpass from the client docker (dockerfiles/Dockerfile) as it would not be a dependency anymore?
@krevsbech thanks, removed it from the Dockerfile and the test_docker script.

labgrid/driver/sshdriver.py
Emantor: We could add a driver option to support dropbear. From reading the dropbear changelog, it simply does not respond to the server keepalive messages, thus the client thinks the server is dead and terminates the connection, AFAICS. So for dropbear we would remove the `ServerAliveInterval` and maybe enable the TCP keepalive only.

bitshifter1: Sounds like a good idea :smile: I can add that in another patch.

bitshifter1:
Guys, I made one more change to the SSH_ASKPASS fix. I now set the SSH_ASKPASS_REQUIRE=force, because that is required for openssh-client 8.4 to work when no Display is set. I thought I already tested with version 8.4 last time, but I guess I messed up when I tried to make install the self compiled client into the system. I now tested with a docker image ubuntu 21.04 which comes with version 8.4p1

labgrid/driver/sshdriver.py
bitshifter1: > We could add a driver option to support dropbear. From reading the dropbear changelog, it simply does not respond to the server keepalive messages, thus the client thinks the server is dead and terminates the connection, AFAICS. So for dropbear we would remove the `ServerAliveInterval` and maybe enable the TCP keepalive only.

@Emantor I tested by adding a server_type attribute to distinguish between openssh and dropbear and I played with removing the ServerAliveInterval and adding TCPKeepAlive (yes and no). I even removed all occurrences of ControlPersist=300 and also tried by setting yes and no. However without the ServerAliveIntervalI I am being disconnected after exactly 5 minutes. This is regardless of using a proxy or not. I guess the openssh client, dropbear server combo needs some more testing.

The client disconnection log looks like this. The server says the the client exited correctly.

```
...
debug2: channel 4: rcvd eof
debug2: channel 4: output open -> drain
debug2: channel 4: obuf empty
debug2: channel 4: chan_shutdown_write (i0 o1 sock -1 wfd 12 efd 13 [write])
debug2: channel 4: output drain -> closed
debug3: receive packet: type 98
debug1: client_input_channel_req: channel 4 rtype exit-status reply 0
debug3: mux_exit_message: channel 4: exit message, exitval 0
debug3: receive packet: type 97
debug2: channel 4: rcvd close
debug2: channel 4: chan_shutdown_read (i0 o3 sock -1 wfd 11 efd 13 [write])
debug2: channel 4: input open -> closed
debug3: channel 4: will not send data after close
debug2: channel 4: send close
debug3: send packet: type 97
debug2: channel 4: is dead
debug2: channel 4: gc: notify user
debug3: mux_master_session_cleanup_cb: entering for channel 4
debug2: channel 3: rcvd close
debug2: channel 3: output open -> drain
debug2: channel 3: chan_shutdown_read (i0 o1 sock 10 wfd 10 efd -1 [closed])
debug2: channel 3: input open -> closed
debug2: channel 4: gc: user detached
debug2: channel 4: is dead
debug2: channel 4: garbage collecting
debug1: channel 4: free: client-session, nchannels 5
debug3: channel 4: status: The following connections are open:
  #1 mux-control (t16 r2 i0/0 o0/0 e[closed]/0 fd 6/6/-1 sock 6 cc -1)
  #2 client-session (t4 r0 i0/0 o0/0 e[write]/0 fd 7/8/9 sock -1 cc 1)
  #3 mux-control (t16 nr0 i3/0 o1/16 e[closed]/0 fd 10/10/-1 sock 10 cc -1)
  #4 client-session (t4 r1 i3/0 o3/0 e[write]/0 fd -1/-1/13 sock -1 cc -1)

debug2: channel 3: obuf empty
debug2: channel 3: chan_shutdown_write (i3 o1 sock 10 wfd 10 efd -1 [closed])
debug2: channel 3: output drain -> closed
debug2: channel 3: is dead (local)
debug2: channel 3: gc: notify user
debug3: mux_master_control_cleanup_cb: entering for channel 3
debug2: channel 3: gc: user detached
debug2: channel 3: is dead (local)
debug2: channel 3: garbage collecting
debug1: channel 3: free: mux-control, nchannels 4
debug3: channel 3: status: The following connections are open:
  #1 mux-control (t16 r2 i0/0 o0/0 e[closed]/0 fd 6/6/-1 sock 6 cc -1)
  #2 client-session (t4 r0 i0/0 o0/0 e[write]/0 fd 7/8/9 sock -1 cc 1)
  #3 mux-control (t16 nr0 i3/0 o3/0 e[closed]/0 fd 10/10/-1 sock 10 cc -1)

debug3: send packet: type 1
client_loop: send disconnect: Broken pipe
```

jluebbe (changes requested):
The commit looks fine. We should document that `start_new_session=True` is only for OpenSSH < 8.4, which doesn't have `SSH_ASKPASS_REQUIRE` yet.
If keepalive doesn't work with Dropbear, that should be fixed there instead of a adding manual workarounds in labgrid.

bitshifter1:
> The commit looks fine. We should document that `start_new_session=True` is only for OpenSSH < 8.4, which doesn't have `SSH_ASKPASS_REQUIRE` yet.

Updated the comment

> If keepalive doesn't work with Dropbear, that should be fixed there instead of a adding manual workarounds in labgrid.

I agree :grin:

jluebbe (approved):

Emantor (approved):

PR #739, opened 2021-02-25
requirements.txt: move pyvisa requirements to general requirements

krevsbech:
**Description**
As pyvisa has no host OS dependencies there is no good reason for
managing the dependencies for this in it's own file

Tested locally and by rebuilding docker images

**Checklist**
- [x] PR has been tested

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -13,5 +13,4 @@ psutil==5.6.6
 -r xena-requirements.txt
 -r graph-requirements.txt
 -r docker-requirements.txt
--r pyvisa-requirements.txt
 -r vxi11-requirements.txt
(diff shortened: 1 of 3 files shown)


jluebbe:
According to https://github.com/pyvisa/pyvisa#requirements it needs the VISA library to be installed. It has no python dependencies, though. Is this widely use enough to install it by default?

krevsbech:
> According to https://github.com/pyvisa/pyvisa#requirements it needs the VISA library to be installed. It has no python dependencies, though. Is this widely use enough to install it by default?

I am not the one to judge if it is used in general or not. I will let you decide. If you think so, I will put a phrase in the docs that one need the VISA lib, if not I will just close the PR.

PR #740, opened 2021-02-26
Dockerfile: Add dev-requirements dependencies to client image

krevsbech:
**Description**
Add dev-requirements and lib onewire to the labgrid-client  image 

- [x] PR has been tested

### DIFF
--- a/dockerfiles/Dockerfile
+++ b/dockerfiles/Dockerfile
@@ -21,14 +21,14 @@ FROM labgrid-base AS labgrid-client
 ARG VERSION
 
 RUN set -e ;\
+    apt update -q=2 ;\
+    apt install -q=2 --yes --no-install-recommends microcom openssh-client sshpass rsync jq qemu-system libow-dev ;\
+    apt clean ;\
+    rm -rf /var/lib/apt/lists/* ;\
     cd /opt/labgrid ;\
     pip3 install yq ;\
-    pip3 install --no-cache-dir -r requirements.txt ;\
-    SETUPTOOLS_SCM_PRETEND_VERSION="$VERSION" python3 setup.py install ;\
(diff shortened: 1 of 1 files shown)


jluebbe (approved):

jluebbe:
With dev-requirements, it will also pull in the full crossbar, yapf and the documentation tools, which should not be needed for a client only container. Is the additional size so small that this doesn't hurt?

Otherwise we could add a "full-client-requirements.txt" and use that for the container.

krevsbech:
> With dev-requirements, it will also pull in the full crossbar, yapf and the documentation tools, which should not be needed for a client only container. Is the additional size so small that this doesn't hurt?
> 
> Otherwise we could add a "full-client-requirements.txt" and use that for the container.

I see your proposal for at full-client-requirements.txt as a good solution, not so much because of disk space but it think it will make it easier to see what dependencies are required where. I will close this PR and make that in a new one, when time allows.
Anyone who disagree ?

Emantor:
> > With dev-requirements, it will also pull in the full crossbar, yapf and the documentation tools, which should not be needed for a client only container. Is the additional size so small that this doesn't hurt?
> > Otherwise we could add a "full-client-requirements.txt" and use that for the container.
> 
> I see your proposal for at full-client-requirements.txt as a good solution, not so much because of disk space but it think it will make it easier to see what dependencies are required where. I will close this PR and make that in a new one, when time allows.
> Anyone who disagree ?

Sounds good to me.

Bastian-Krause:
@krevsbech If this is still relevant for you, we could now move to the "dev" extra easily.

Bastian-Krause:
No feedback for two years, I'm closing this. Please update and reopen if it's still relevant.

PR #743, opened 2021-03-04
Bump crossbar & autobahn

Emantor:
**Description**

This should fix the CI failures due to incompatibilities with twisted.

**Checklist**
- [x] PR has been tested


Emantor:
Looks like we need to limit `twisted<21.2.0` for python 3.6 to work.

jluebbe (approved):

jluebbe (approved):

PR #751, opened 2021-03-20
sshdriver: Add port forwarding

JoshuaWatt:
**Description**

Adds a context manager function that will forward a port on the local
host to a port on the target

**Checklist**
- [x] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/labgrid/driver/sshdriver.py
+++ b/labgrid/driver/sshdriver.py
@@ -1,5 +1,6 @@
 # pylint: disable=no-member
 """The SSHDriver uses SSH as a transport to implement CommandProtocol and FileTransferProtocol"""
+import contextlib
 import logging
 import os
 import shutil
@@ -15,6 +16,7 @@ from .commandmixin import CommandMixin
 from .common import Driver
 from ..step import step
 from .exception import ExecutionError
+from ..util.helper import get_free_port
 from ..util.proxy import proxymanager
 from ..util.timeout import Timeout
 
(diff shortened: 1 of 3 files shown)


JoshuaWatt:
I'll happily add tests if someone knows of a good way of testing the port bindings

Emantor:
> I'll happily add tests if someone knows of a good way of testing the port bindings

We already have some tests for port forwarding in `test_util.py`, with `ssh_localhost` as a fixture you should be able to test the port forwards for the `SSHDriver` using the same mechanism.

labgrid/driver/sshdriver.py
```diff
@@ -222,6 +224,82 @@ def interact(self, cmd=None):
...
+        """
+        if not self._check_keepalive():
+            raise ExecutionError("Keepalive no longer running")
+
+        if localport is None:
+            localport = get_free_port()
+
+        complete_cmd = ["ssh",
+                "-L", "%d:localhost:%d" % (localport, remoteport),
+                "-o", "ControlPath={}".format(self.control.replace('%', '%%')),
```
jluebbe, line 253: Couldn't you use self.ssh_prefix here?

labgrid/driver/sshdriver.py
```diff
@@ -222,6 +224,82 @@ def interact(self, cmd=None):
...
+                "-N",
+                "_",
+        ]
+        self.logger.debug("Running command: %s", complete_cmd)
+        sub = subprocess.Popen(complete_cmd)
+        try:
+            yield localport
+        finally:
+            sub.kill()
+            sub.wait()
```
jluebbe, line 263: Hmm, we have a control master process anyway, so using `-O forward` should be possible. In that case, you wouldn't have an active process for the full duration (see `add_port_forward()` in `labgrid/util/ssh.py`). That would mean running the cleanup on shutdown, though.

labgrid/remote/client.py
```diff
@@ -1087,6 +1089,24 @@ def sshfs(self):
...
+                localport = stack.enter_context(drv.forward_local_port(remote, localport=local))
+                print("Forwarding local port %d to remote port %d" % (localport, remote))
+
+            for local, remote in self.args.remote:
+                localport = stack.enter_context(drv.forward_remote_port(remote, localport=local))
+                print("Forwarding remote port %d to local port %d" % (remote, localport))
+
+            try:
+                while True:
+                    signal.pause()
```
jluebbe, line 1106: This should mention that it's waiting for a CTRL-C.

labgrid/remote/client.py
```diff
@@ -1568,6 +1626,16 @@ def main():
...
     subparser.set_defaults(func=ClientSession.sshfs)
 
+    subparser = subparsers.add_parser('forward',
+                                      help="forward local port to remote target")
+    subparser.add_argument("--local", "-L", metavar="[LOCAL:]REMOTE",
+                           action=LocalPort,
+                           help="Forward local port LOCAL to remote port REMOTE. If LOCAL is unspecified, an arbitrary port will be chosen")
+    subparser.add_argument("--remote", "-R", metavar="REMOTE[:LOCAL]",
+                           action=RemotePort,
+                           help="Forward remote port REMOTE to local port LOCAL. If LOCAL is unspecified, an arbitrary port will be chosen")
```
jluebbe, line 1636: When is the arbitrary port feature useful for the command line client?

Also, I think that in the -R case, the LOCAL (target) needs to be explicit, otherwise you wouldn't know what it connects to (see my comment to `forward_remote_port()` in the driver).

labgrid/driver/sshdriver.py
```diff
@@ -222,6 +224,82 @@ def interact(self, cmd=None):
...
+
+    @Driver.check_active
+    @contextlib.contextmanager
+    def forward_remote_port(self, remoteport, localport=None):
```
jluebbe, line 267: Random local target port doesn't seem to be useful? And probing a random port on the remote isn't that straight forward either.

labgrid/remote/client.py
```diff
@@ -1647,8 +1715,8 @@ def main():
 
     # make any leftover arguments available for some commands
     args, leftover = parser.parse_known_args()
-    if args.command not in ['ssh', 'rsync']:
-        args = parser.parse_args()
+    if args.command not in ['ssh', 'rsync', 'forward']:
+        args = parser.parse_args(namespace=argparse.Namespace())
```
jluebbe, line 1719: If I read the docs correctly, this should already be the default?

jluebbe (changes requested):

labgrid/driver/sshdriver.py
JoshuaWatt: I tries this, but it doesn't work. Unless I'm misunderstanding the code, we have a control master process to the exporter, but not the target, so I don't _think_ that will work (someone please correct me if I'm wrong).

jluebbe (approved):
Looks good now, thanks for the test cases!

Emantor (approved):

PR #754, opened 2021-03-26
driver: quartushpsdriver: try reading JTAG ID from chain to identify errors

Bastian-Krause:
**Description**
PR #736's ccc17c86 ("driver: quartushpsdriver: retry jtagconfig until intact JTAG chain found") implemented retrying on a specific error message.

Every once in a while other errors occur. Since `jtagconfig` is part of the proprietary Quartus programmer suite, there is no reasonable way to tell intact JTAG chains apart from error messages other than trying to parse the JTAG ID.

An intact chain looks like this:

    4BA00477   SOCVHPS

..whereas errors look like that:

    Unable to lock chain - Chain in use - check  on 127.0.0.1

..or:

    Unable to read device chain - JTAG chain broken`

So try to find the JTAG ID of the first chain and make sure its a valid hexadecimal number. If that fails, log the message as a warning and try again for a limited time. We do not longer rely on a specific error message.

**Checklist**
- [ ] Tests for the feature 
- [x] PR has been tested

### DIFF
--- a/labgrid/driver/quartushpsdriver.py
+++ b/labgrid/driver/quartushpsdriver.py
@@ -2,6 +2,7 @@
 import subprocess
 import re
 import time
+import logging
 
 import attr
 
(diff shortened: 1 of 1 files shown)


jluebbe (approved):
If there's no better way, then this is probably as good as it gets. @Emantor ?

Emantor (approved):

PR #756, opened 2021-03-29
tests: fix warnings and atexit exceptions

Bastian-Krause:
**Description**
When running the test suite, various warnings and exceptions happen. Some occur because cleanup actions are called explicitly during tests leading to errors once the atexit registered cleanup methods run. Unregister the atexit cleanup methods for these tests explicitly. Another warning happens because not all of pytest's TerminalWriter features that we use are mocked. Extend `MockTW` accordingly.

**Checklist**
- [x] PR has been tested

Fixes #730
Fixes #687

### DIFF
--- a/tests/test_stepreporter.py
+++ b/tests/test_stepreporter.py
@@ -93,6 +93,8 @@ def test_step_reporter_color_scheme(caplog, curses_init, color_scheme, result_co
 
     class MockTW:
         _file = None
+        def line(self, line):
+            pass
     class MockTerminalReporter:
         def __init__(self, f):
             self._tw = MockTW()
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -1,6 +1,7 @@
 import os.path
 import subprocess
 import socket
+import atexit
 
 from shutil import which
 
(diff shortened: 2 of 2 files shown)


Emantor (dismissed):

rohieb (dismissed):
Hmm, I did not see such warnings for MockTW when I tested #730, but the StepReporter might have changed in the meantime.

Bastian-Krause:
> Hmm, I did not see such warnings for MockTW when I tested #730, but the StepReporter might have changed in the meantime.

The warning is there since your PR, see the [corresponding GitHub actions run](https://github.com/labgrid-project/labgrid/runs/1969842981?check_suite_focus=true#step:9:63).

rohieb:
> > Hmm, I did not see such warnings for MockTW when I tested #730, but the StepReporter might have changed in the meantime.
> 
> @rohieb The warning is there since your PR, see the [corresponding GitHub actions run](https://github.com/labgrid-project/labgrid/runs/1969842981?check_suite_focus=true#step:9:63).

Strange, I didn't see that on my machine when developing the test (maybe it was before I rebased onto then-current master). But it doesn't matter, thank you for fixing it.

jluebbe:
As the warning tells the user "call the cleanup() method on targets yourself to handle exceptions explictly", taking that approach shouldn't lead to further problems. The cleanup functions either need to unregister the atexit handler or be safe to call multiple times.

Bastian-Krause:
Dropped commit 
> tests/externalconsoledriver: unregister cleanup for explicit close call

in favor of #765.

Emantor (approved):

PR #759, opened 2021-03-29
driver/usbvideodriver: add basic MacroSilicon USB Video support

jluebbe:
This device only support JPEG.

**Checklist**
- [ ] Tests for the feature 
- [x] PR has been tested

### DIFF
--- a/labgrid/driver/usbvideodriver.py
+++ b/labgrid/driver/usbvideodriver.py
@@ -41,10 +41,11 @@ class USBVideoDriver(Driver, VideoProtocol):
 
     def get_pipeline(self):
         match = (self.video.vendor_id, self.video.model_id)
+        transport = " ! matroskamux streamable=true ! fdsink"
         if match == (0x046d, 0x082d):
-            return "v4l2src device={} ! {} ! h264parse ! fdsink"
+            return "v4l2src device={} ! {} ! h264parse" + transport
         if match == (0x046d, 0x0892):
-            return "v4l2src device={} ! {} ! decodebin ! vaapipostproc ! vaapih264enc ! h264parse ! fdsink"  # pylint: disable=line-too-long
+            return "v4l2src device={} ! {} ! decodebin ! vaapipostproc ! vaapih264enc ! h264parse" + transport  # pylint: disable=line-too-long
         raise InvalidConfigError("Unkown USB video device {:04x}:{:04x}".format(*match))
 
     @Driver.check_active
@@ -55,7 +56,7 @@ class USBVideoDriver(Driver, VideoProtocol):
         tx_cmd = self.video.command_prefix + ["gst-launch-1.0", "-q"]
         tx_cmd += pipeline.format(self.video.path, caps).split()
         rx_cmd = ["gst-launch-1.0"]
-        rx_cmd += "fdsrc ! h264parse ! avdec_h264 ! glimagesink sync=false".split()
+        rx_cmd += "playbin uri=fd://0".split()
 
         tx = subprocess.Popen(
             tx_cmd,
--- a/labgrid/driver/usbvideodriver.py
+++ b/labgrid/driver/usbvideodriver.py
@@ -45,24 +45,35 @@ class USBVideoDriver(Driver, VideoProtocol):
         raise InvalidConfigError("Unkown video format {} for device {:04x}:{:04x}".format(
             variant, self.video.vendor_id, self.video.model_id))
 
-    def get_pipeline(self):
+    def get_pipeline(self, path, caps, controls=None):
         match = (self.video.vendor_id, self.video.model_id)
-        transport = " ! matroskamux streamable=true ! fdsink"
         if match == (0x046d, 0x082d):
-            return "v4l2src device={} ! {} ! h264parse" + transport
-        if match == (0x046d, 0x0892):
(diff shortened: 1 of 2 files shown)


labgrid/driver/usbvideodriver.py
```diff
@@ -54,7 +62,8 @@ def stream(self, caps_hint=None):
         tx_cmd = self.video.command_prefix + ["gst-launch-1.0", "-q"]
         tx_cmd += pipeline.format(self.video.path, caps).split()
         rx_cmd = ["gst-launch-1.0"]
-        rx_cmd += "fdsrc ! h264parse ! avdec_h264 ! glimagesink sync=false".split()
+        #rx_cmd += "fdsrc ! h264parse ! avdec_h264 ! glimagesink sync=false".split()
+        rx_cmd += "fdsrc ! jpegparse ! jpegdec ! glimagesink sync=false".split()
```
jluebbe, line 66: This needs to be more flexible. Maybe return more information from _get_caps (encoder+decoder)?

Emantor: Yes, sgtm.

jluebbe: I've talked with @michaelolbrich about this, the simpler approach is to use mkv and let the client use playbin.

labgrid/driver/usbvideodriver.py
```diff
@@ -54,7 +63,7 @@ def stream(self, caps_hint=None):
         tx_cmd = self.video.command_prefix + ["gst-launch-1.0", "-q"]
         tx_cmd += pipeline.format(self.video.path, caps).split()
         rx_cmd = ["gst-launch-1.0"]
-        rx_cmd += "fdsrc ! h264parse ! avdec_h264 ! glimagesink sync=false".split()
+        rx_cmd += "playbin uri=fd://0 sync=false".split()
```
michaelolbrich, line 66: `playbin` has no `sync` property. And it should not be necessary here anyways.

jluebbe: OK. I've removed the sync=false.

jluebbe:
@Emantor @michaelolbrich I've also added support for passing v4l controls.

labgrid/driver/usbvideodriver.py
```diff
@@ -38,23 +44,36 @@ def select_caps(self, hint=None):
...
-            return "v4l2src device={} ! {} ! decodebin ! vaapipostproc ! vaapih264enc ! h264parse ! fdsink"  # pylint: disable=line-too-long
-        raise InvalidConfigError("Unkown USB video device {:04x}:{:04x}".format(*match))
+            controls = controls or "focus_auto=1"
+            inner = "h264parse"
+        elif match == (0x046d, 0x0892):
+            inner = "decodebin ! vaapipostproc ! vaapih264enc ! h264parse"
+        elif match == (0x534d, 0x2109):
+            inner = None  # just forward the jpeg frames
+        else:
+            raise InvalidConfigError("Unkown USB video device {:04x}:{:04x}".format(*match))
```
michaelolbrich, line 57: You could allow arbitrary cameras with something like this:
`v4l2src device=/dev/video0 ! 'image/jpeg;video/x-h264' ! parsebin ! matroskamux ! streamable=true ! fdsink`

Note that the order of the caps define the priority. So in this example, jpeg will be used if the camera supports both jpeg and h264.

michaelolbrich: And you need some caps here, otherwise some raw format will most likely be chosen.

jluebbe: We currently select the caps explicitly from low/mid/high settings.

michaelolbrich: You can use ranges there: `image/jpeg,width=[1,1280]height=[1,720],framerate=[1/1,15/1]`
From what I can tell, it tries to use the larges values that fit into the range.

jluebbe: OK. But for now, I'd prefer to use tested and explicit values there.

Emantor (dismissed):

Emantor:
LGTM, have not tested yet.

labgrid/driver/usbvideodriver.py
```diff
@@ -38,23 +44,36 @@ def select_caps(self, hint=None):
...
         match = (self.video.vendor_id, self.video.model_id)
         if match == (0x046d, 0x082d):
-            return "v4l2src device={} ! {} ! h264parse ! fdsink"
-        if match == (0x046d, 0x0892):
-            return "v4l2src device={} ! {} ! decodebin ! vaapipostproc ! vaapih264enc ! h264parse ! fdsink"  # pylint: disable=line-too-long
-        raise InvalidConfigError("Unkown USB video device {:04x}:{:04x}".format(*match))
+            controls = controls or "focus_auto=1"
+            inner = "h264parse"
+        elif match == (0x046d, 0x0892):
+            inner = "decodebin ! vaapipostproc ! vaapih264enc ! h264parse"
```
Emantor, line 53: This breaks this pipeline, however even on master this is currently displaying in green, do i'd be in favor of moving this to jpeg too and reworking the vaapi support eventually.

Emantor (approved):

labgrid/driver/usbvideodriver.py
jluebbe: @Emantor has switched this to jpeg.

PR #741, opened 2021-03-03
add NetworkManager remote control via agent

jluebbe:
**Description**
This adds support for controlling an exported NetworkInterface remotely using NetworkManager over an agent connection.

Via NetworkManager, we can configure and inspect Ethernet and WiFi (client&AP) interfaces. This is especially useful for testing WiFi interfaces on the DUT.

**Checklist**
- [x] Documentation for the feature
- [ ] Tests for the feature 
- [x] The arguments and description in doc/configuration.rst have been updated
- [ ] CHANGES.rst has been updated
- [x] PR has been tested

PR #762, opened 2021-03-30
TFTP/NFS/HTTP providers

jluebbe:
This builds on the network-agent PR #741.

**Description**
This supports "staging" files to a remote TFTP/NFS/HTTP server for
access by the target. Files below the 'internal' path need to be visible
to the target below the corresponding 'external' path.

The actual server needs to be set up outside of labgrid for now. Running
and configuring the respective server could be done later in the
exporter without changes to the client.

The files are synchronized to the exporter via the ManagedFile helper
and then symlinked to the actual provider path.

**Checklist**
- [ ] Tests for the feature 
- [x] PR has been tested


Emantor (approved):

doc/configuration.rst
```diff
@@ -797,6 +803,57 @@ Such device could be a signal generator.
...
+.. _HTTPProvider:
+
+TFTPProvider / NFSProvider / HTTPProvider
++++++++++++++++++++++++++++++++++++++++++
+
+.. code-block:: yaml
+
+   TFTPProvider:
+     internal: "/srv/tftp/board-23/"
+     external: "board-23/"
```
Emantor, line 833: Another example with the `HTTPProvider` is required here to understand that the URL should be part of the external string.

Emantor (approved):

doc/configuration.rst
jluebbe: I've added a HTTPProvider example.

PR #766, opened 2021-04-06
add support for Linux Automation USB-Mux

jluebbe:
This devices allows switching a USB device between the host and the DUT.

**Description**

**Checklist**
- [x] PR has been tested

### DIFF
--- a/labgrid/driver/lxausbmuxdriver.py
+++ b/labgrid/driver/lxausbmuxdriver.py
@@ -1,0 +1,49 @@
+# pylint: disable=no-member
+import attr
+
+from .common import Driver
+from ..factory import target_factory
+from ..step import step
+from .exception import ExecutionError
+from ..util.helper import processwrapper
+
+@target_factory.reg_driver
+@attr.s(eq=False)
+class LXAUSBMuxDriver(Driver):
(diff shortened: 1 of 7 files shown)


ejoerns (approved):
Tested this and works for me :+1:

Emantor (approved):

labgrid/driver/lxausbmuxdriver.py
```diff
@@ -0,0 +1,49 @@
...
+    def __attrs_post_init__(self):
+        super().__attrs_post_init__()
+        if self.target.env:
+            self.tool = self.target.env.config.get_tool('usbmuxctl') or 'usbmuxctl'
+        else:
+            self.tool = 'usbmuxctl'
+
+    @Driver.check_active
+    @step(title='usbmux_set', args=['links'])
+    def set_links(self, links):
```
SmithChart, line 29: These options do not reflect all options provided by the underlying tool: the `--no-id` option is missing.
With this option the user can switch the DUT port without changing the value of the ID pin. But to be honest: I am not sure if this feature will be used in practice.

jluebbe: We can still extend this later if needed.

PR #770, opened 2021-04-09
doc/usage: LG_PROXY: Add NetworkServices proxy note

tompreston:
NetworkServices specify their own proxies, which take precedence over
LG_PROXY. This isn't clear when specifying LG_PROXY, and I didn't know
both had to be configured in .ssh/config. I've added some documentation
showing how to debug this.

**Description**

**Checklist**
- N/A Tests for the feature 

- N/A The arguments and description in doc/configuration.rst have been updated

- N/A Add a section on how to use the feature to doc/usage.rst

- N/A Add a section on how to use the feature to doc/development.rst

- N/A PR has been tested

- N/A Man pages have been regenerated

### DIFF
--- a/doc/usage.rst
+++ b/doc/usage.rst
@@ -356,7 +356,10 @@ Takes effect only when used with ``--lg-colored-steps``.
 LG_PROXY
 ^^^^^^^^
 Specifies a SSH proxy host to be used for port forwards to access the
-coordinator and network resources.
+coordinator. Network resources made available by the exporter will prefer their
+own proxy, and only fallback to LG_PROXY.
+
+See also :ref:`overview-proxy-mechanism`.
 
 Simple Example
 ~~~~~~~~~~~~~~
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,7 @@ New Features in 0.4.0
   The 256-color schemes now use purple instead of green for the ``run`` lines to
   make them easier distinguishable from pytest's "PASSED" output.
 - Network controlled relay providing GET/PUT based REST API
+- Improved LG_PROXY documentation in docs/usage.rst.
 
 Bug fixes in 0.4.0
 ~~~~~~~~~~~~~~~~~~
--- a/doc/overview.rst
+++ b/doc/overview.rst
@@ -352,6 +352,8 @@ file names refer to a shared filesystem (such as NFS or SMB).
   For exporters which are not directly accessible via SSH, add the host to your
   .ssh/config file, with a ProxyCommand when need.
 
+.. _overview-proxy-mechanism:
+
 Proxy Mechanism
 ~~~~~~~~~~~~~~~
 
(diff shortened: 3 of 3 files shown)


tompreston:
I regenerated the man pages (I had to change the below on Fedora), but it didn't change any files in git.
```diff
diff --git a/man/Makefile b/man/Makefile
index 0543c82..8f3d181 100644
--- a/man/Makefile
+++ b/man/Makefile
@@ -2,7 +2,7 @@ PAGES := labgrid-client.1 labgrid-exporter.1 labgrid-device-config.1 labgrid-pyt
 COMPRESSED := labgrid-client.1.gz labgrid-exporter.1.gz labgrid-device-config.1.gz labgrid-pytest.1.gz
 
 %.1: %.rst
-       rst2man.py $< >$@.tmp
+       rst2man $< >$@.tmp
        mv -f $@.tmp $@
 
 %.1.gz : %.1
```

doc/usage.rst
```diff
@@ -358,6 +358,20 @@ LG_PROXY
 Specifies a SSH proxy host to be used for port forwards to access the
 coordinator and network resources.
 
+It is worth noting that each NetworkService will define its own proxy host in
+the `params.extra.proxy` variable, which will take precedence over `LG_PROXY`.
+The NetworkService `params.extra.proxy` will be used when `LG_PROXY` is set or
```
Emantor, line 363: This is not only true for the NetworkService, but for every resource which uses `get_host_and_port` from the `labgrid.util.proxy`. Labgrid always assumes that exporters are accessible via a configured SSH config. Indeed the isolated flag enforces that all connections to resources from the isolated exporter are tunneled through the exporter instead of the `LG_PROXY` host. I think the information is currently available in the docs, but not consolidated enough to understand the full picture easily.

tompreston: You're right, the `--isolator` mechanism and SSH config is explained in the [Proxy Mechanism section](https://github.com/labgrid-project/labgrid/blob/master/doc/overview.rst#proxy-mechanism).

I was expecting labgrid to use LG_PROXY for client-to-coordinator and client-to-exporter proxies. The current documentation suggests that LG_PROXY configures both, and that labgrid will only prefer the client-to-exporter proxy if the exporter is isolated.

In my case, an "isolated" exporter was not configured with `--isolated` and setting LG_PROXY selects the client-to-coordinator proxy (a nickname I have in .ssh/config), and activated the client-to-exporter proxy - a configuration I hadn't set up. I think the right configuration for my arrangement is:
1. `labgrid-exporter --isolated', with the exporter host in SSH config, as labgrid expects.
2. Configure client-to-coordinator with either LG_PROXY or LG_CROSSBAR, with coordinator host also in SSH config.

I've made another set of doc changes to make the current behaviour more clear.

Emantor:
> I regenerated the man pages (I had to change the below on Fedora), but it didn't change any files in git.
> 
> ```diff
> diff --git a/man/Makefile b/man/Makefile
> index 0543c82..8f3d181 100644
> --- a/man/Makefile
> +++ b/man/Makefile
> @@ -2,7 +2,7 @@ PAGES := labgrid-client.1 labgrid-exporter.1 labgrid-device-config.1 labgrid-pyt
>  COMPRESSED := labgrid-client.1.gz labgrid-exporter.1.gz labgrid-device-config.1.gz labgrid-pytest.1.gz
>  
>  %.1: %.rst
> -       rst2man.py $< >$@.tmp
> +       rst2man $< >$@.tmp
>         mv -f $@.tmp $@
>  
>  %.1.gz : %.1
> ```

The man pages only require regeneration if they have been changed, as this PR doesn't touch the man pages they don't need to be regenerated.

Emantor (approved):

jluebbe (approved):

jluebbe:
I've updated the commit message to match the changed content of this PR.

PR #772, opened 2021-04-13
exporter: Check /usr/sbin for ser2net

tompreston:
On Debian, the ser2net binary is installed at /usr/sbin/ser2net and it
can be run by any user. Since this isn't on the users PATH, we should
attempt to use it when failing to find a ser2net binary, before falling
back to /usr/bin/ser2net.

You can quickly see this like so:
```
$ podman run -it debian:bullseye
# apt update && apt install ser2net
# ls -l /usr/sbin/ser2net
-rwxr-xr-x. 1 root root 136696 Aug 24  2018 /usr/sbin/ser2net
```

**Checklist**
- N/A Documentation for the feature (not necessary, just searches in an additional place)
- N/A Tests for the feature (not necessary)

- N/A The arguments and description in doc/configuration.rst have been updated (not relevant)

- N/A Add a section on how to use the feature to doc/usage.rst (not relevant)

- N/A Add a section on how to use the feature to doc/development.rst (not relevant)

- [x] PR has been tested

- N/A Man pages have been regenerated (not relevant)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -17,6 +17,7 @@ New Features in 0.4.0
   make them easier distinguishable from pytest's "PASSED" output.
 - Network controlled relay providing GET/PUT based REST API
 - Improved LG_PROXY documentation in docs/usage.rst.
+- Exporter now checks /usr/sbin/ser2net for SerialPortExport
 
 Bug fixes in 0.4.0
 ~~~~~~~~~~~~~~~~~~
--- a/labgrid/remote/exporter.py
+++ b/labgrid/remote/exporter.py
@@ -6,6 +6,7 @@ import asyncio
 import logging
 import sys
 import os
+import os.path
 import time
 import traceback
 import shutil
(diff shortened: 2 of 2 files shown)


jluebbe (dismissed):

Emantor (approved):
Can't say I'm too happy to encode this into labgrid, but we do this in other places as well.

tompreston:
@Emantor How come? Have two fallbacks is a bit ugly - it's at `/usr/sbin/ser2net` in Ubuntu as well, and the source (this one at least) installs to sbin only.

What's the expected environment where we fallback to `/usr/bin`? Why not just fall back to `/usr/sbin/ser2net` and lose `/usr/bin/ser2net`?

Emantor:
> @Emantor How come? Have two fallbacks is a bit ugly - it's at `/usr/sbin/ser2net` in Ubuntu as well, and [the source](https://github.com/cminyard/ser2net/blob/master/Makefile.am#L1) (this one at least) installs to sbin only.
> 
> What's the expected environment where we fallback to `/usr/bin`? Why not just fall back to `/usr/sbin/ser2net` and lose `/usr/bin/ser2net`?

Arch Linux installs to [/usr/bin](https://github.com/archlinux/svntogit-community/blob/packages/ser2net/trunk/PKGBUILD). Either way, at least for the exporter it is recommended to setup `PATH` for the service in a way so that all binaries are found correctly using `which`. Thats how I solve the "where is the binary"-problem for NixOS as well, which has neither `/usr/bin/ser2net` nor `/usr/sbin/ser2net`.
I don't really understand why ser2net installs itself into sbin either way, ser2net is perfectly usable as a non root user as long as you have access to the serial port and use a network port >1024.

jluebbe (approved):

PR #773, opened 2021-04-13
exporter: Add --ser2net-port argument

tompreston:
Add an exporter argument to set the ser2net port. This is useful when
running the exporter in a container because you have to expose a known
port in the configuration.

The ser2net port defaults to any free port.

Running the exporter:
```
$ labgrid-exporter --ser2net-port=54321 exporter-conf/exporter.yaml
[snip]
2021-04-13T20:39:28 started ser2net for /dev/ttyUSB1 on port 54321
[snip]
```
**Description**

**Checklist**
- N/A Tests for the feature

- N/A The arguments and description in doc/configuration.rst have been updated

- N/A Add a section on how to use the feature to doc/usage.rst

- N/A Add a section on how to use the feature to doc/development.rst

- [x] PR has been tested

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,7 @@ New Features in 0.4.0
   The 256-color schemes now use purple instead of green for the ``run`` lines to
   make them easier distinguishable from pytest's "PASSED" output.
 - Network controlled relay providing GET/PUT based REST API
+- Add exporter argument --ser2net-port.
 
 Bug fixes in 0.4.0
 ~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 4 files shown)


Emantor:
This will only work if the exporter exposes exactly one serial port on the given container, but breaks if more serial ports are available.

tompreston:
Of course, I didn't think of that.

In addition, there are issues sharing the /dev/tty nodes since they change on input. It's better to use udev.

This was mentioned in IRC, but I'll mention it here too - it's probably best the exporter isn't run inside a container. I will switch to this workflow.

I'm happy for you to close this, many thanks.

Emantor:
> In addition, there are issues sharing the /dev/tty nodes since they change on input. It's better to use udev.
> 
> You mentioned this in IRC, but I'll mention it here too - it's probably best the exporter isn't run inside a container. I will switch to this workflow.

Yep, using the exporter inside a container also requires forwarding devices and events, we have no expertise and can't help with these setups. Using exporters without containers is much easier.

> I'm happy for you to close this, many thanks.

Sure.

PR #774, opened 2021-04-27
Debian package improvements

sephalon:
**Description**

- Package coordinator into separate virtual environment
- Split `labgrid` package into `base`, `coordinator`, `client` and `exporter`
- Package systemd files
- Package `labgrid-suggest`
- Move config manpage into section 5 and package manpages

Recently an independent [upstream Debian package](https://packages.debian.org/sid/labgrid) has been published. Ideally, we could put both packages onto the same basis in the near future; however I suppose the main issue is how to deal with [vendoring](https://lwn.net/Articles/842319/): Labgrid has quite a lot of dependencies and for our deployments, being able to build local packages from master using dh-virtualenv without having to worry too much about some Python dependency incompatibility has been quite useful so far. The upstream package is using distribution packages only however. I will notify the maintainer of this pull request; maybe he has some suggestions as well.

I think the package `labgrid-base` should be called `python3-labgrid` as the upstream package does; however dh-virtualenv seems to treat the `python3-` prefix specially and does not create a virtual environment then.

**Checklist**
- [ ] Tests for the feature 

- [ ] PR has been tested


jluebbe:
The `python3`-prefix is for python modules installed in the normal paths, so it doesn't apply here.
Splitting off the coordinator is fine (due to the additional dependancies), but the other dependencies are very similar, so I'd prefer a shared package.

Regarding vendoring: As I understand it, dh_virtualenv is not allowed in the Debian archive. The current Debian package is doesn't support labgrid-client or labgrid-exporter (due too old autobahn). @bgermann is preparing a new version with only labgrid-suggest:
https://salsa.debian.org/python-team/packages/labgrid/-/commit/53503ba1afef9a0c05aded6646cb0db49f2d6a2f

Using the same package name will probably confuse users, so maybe we should use a different one for the dh_virtualenv-based one and add `Conflicts:`?

sephalon:
> The `python3`-prefix is for python modules installed in the normal paths, so it doesn't apply here.

Yeah, the rename would have only made sense if we would have decided to drop dh-virtualenv. However that does not seem to be an option right now…

> Splitting off the coordinator is fine (due to the additional dependancies), but the other dependencies are very similar, so I'd prefer a shared package.

Ok, then I will merge `labgrid-base`, `labgrid-client` and `labgrid-exporter` back into a single `labgrid` package.

> Regarding vendoring: As I understand it, dh_virtualenv is not allowed in the Debian archive. The current Debian package is doesn't support labgrid-client or labgrid-exporter (due too old autobahn). @bgermann is preparing a new version with only labgrid-suggest:
> https://salsa.debian.org/python-team/packages/labgrid/-/commit/53503ba1afef9a0c05aded6646cb0db49f2d6a2f

It looks like going upstream is more of a long-term project then and not a solution for people wanting to use the newest and shiniest version anyway. What do you think about creating an official PPA for Labgrid?

(I am aware of the Docker containers, but they are rather inconvenient to use and deploy; I consider them more of a development/reference setup. AppImage/Flatpak/Snap might be an option too, but I have no experience with these tools.)

> Using the same package name will probably confuse users, so maybe we should use a different one for the dh_virtualenv-based one and add `Conflicts:`?

How about naming them `labgrid-venv` and `labgrid-coordinator-venv`?

sephalon:
I have merged client and exporter into a single package and added the `-venv` suffix as proposed in my previous comment. Please review.

PR #775, opened 2021-04-28
shelldriver: Handle intermingled output on login

JoshuaWatt:
Some devices may have chatty consoles during login;
specifically, if a user logs into the root serial console of a target
using systemd that has not finished booting, boot messages will begin to
appear after login. These messages can interrupt prompt detection and
cause test timeouts. To allow proper detection of the prompt:
1. Add an API to wait for the device to "settle", e.g.
   no output is seen on the console for a specified number of seconds and
   add an option for the shelldriver to utilize it
2. Move the password and post-login prompt detection into the same main
   main loop as the other prompt detection. This allows the same login and retry
   logic to be used which make it more robust against noise.

**Description**

**Checklist**
- [ ] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/labgrid/driver/consoleexpectmixin.py
+++ b/labgrid/driver/consoleexpectmixin.py
@@ -1,6 +1,7 @@
-from time import sleep
+import time
+import pexpect
 
-from ..util import PtxExpect
+from ..util import PtxExpect, Timeout
 from ..step import step
 from .common import Driver
 
@@ -33,7 +34,7 @@ class ConsoleExpectMixin:
                               len(data), data, self.txdelay)
             count = 0
             for i in range(len(data)):
-                sleep(self.txdelay)
+                time.sleep(self.txdelay)
                 count += self._write(data[i:i+1])
             return count
 
@@ -53,6 +54,16 @@ class ConsoleExpectMixin:
     def expect(self, pattern, timeout=-1):
         index = self._expect.expect(pattern, timeout=timeout)
         return index, self._expect.before, self._expect.match, self._expect.after
+
+    @Driver.check_active
+    @step(args=['quiet_time'])
+    def settle(self, quiet_time, timeout=120.0):
+        t = Timeout(timeout)
+        while not t.expired:
+            try:
+                self.read(timeout=quiet_time)
+            except pexpect.TIMEOUT:
+                break
 
     def resolve_conflicts(self, client):
         for other in self.clients:
--- a/labgrid/driver/shelldriver.py
+++ b/labgrid/driver/shelldriver.py
@@ -15,7 +15,7 @@ import xmodem
 from ..factory import target_factory
 from ..protocol import CommandProtocol, ConsoleProtocol, FileTransferProtocol
 from ..step import step
-from ..util import gen_marker
+from ..util import gen_marker, Timeout
 from .commandmixin import CommandMixin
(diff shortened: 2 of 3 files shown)


labgrid/driver/shelldriver.py
```diff
@@ -137,37 +163,32 @@ def _await_login(self):
...
-                    if self.password:
-                        self.console.sendline(self.password)
-                        remaining_time = (start + self.login_timeout) - time.time()
-                        self.console.expect(self.prompt, timeout=remaining_time)
-                    else:
-                        raise Exception("Password entry needed but no password set")
+
+                # Do not issue the check string until the password has been entered
+                check = not self.password
+                for index in poll_expect([self.prompt, "Password: "], check):
```
JoshuaWatt, line 169: I'm not sure if we really need a nested poll just to handle a password. We could also roll the password detection as just another case in the main poll loop.

Anyone have thoughts?

Emantor: We do this inner loop to decide if we actually need to enter the password after the login prompt. IMO this could be folded into the main loop.

labgrid/driver/consoleexpectmixin.py
```diff
@@ -54,6 +55,17 @@ def expect(self, pattern, timeout=-1):
...
         return index, self._expect.before, self._expect.match, self._expect.after
 
+    @Driver.check_active
+    @step(args=['quiet_time'])
+    def settle(self, quiet_time, timeout=None):
+        start = time.time()
+        try:
+            while self.read(timeout=quiet_time):
+                if timeout is not None and time.time() > start + timeout:
+                    raise pexpect.TIMEOUT("Timeout of {} seconds exceeded waiting for output to settle".format(timeout))
```
jluebbe, line 65: Use the `Timeout` helper class from labgrid.util, which also uses `time.monotonic` instead of `time.time()` (which is can have discontinuities).

labgrid/driver/shelldriver.py
```diff
@@ -42,6 +45,7 @@ class ShellDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
     login_timeout = attr.ib(default=60, validator=attr.validators.instance_of(int))
     console_ready = attr.ib(default="", validator=attr.validators.instance_of(str))
     await_login_timeout = attr.ib(default=2, validator=attr.validators.instance_of(int))
+    post_login_settle_time = attr.ib(default=-1, validator=attr.validators.instance_of(int))
```
jluebbe, line 48: Wouldn't `default=0` a more natural default to disable this behaviour?

labgrid/driver/shelldriver.py
```diff
@@ -111,6 +115,13 @@ def _await_login(self):
 
         start = time.time()
 
+        def remaining_time():
```
jluebbe, line 118: This can also be simplified by using the Timeout helper (although here we already incorrectly use `time.time()`).

jluebbe (changes requested):
The refactoring commit is fine.

labgrid/driver/shelldriver.py
jluebbe: This was resolved by @JoshuaWatt in a follow-up commit.

jluebbe (approved):

PR #778, opened 2021-04-30
driver/power/digipower: Make driver work

SmithChart:
**Description**

This changes basically make this driver work. I am not sure if it ever worked at all:
* Indexing of the `portstring` and `ports` -dicts has probably never worked
  the way it was done: `index` is explicitly casted to `int` but indices
  have been `str`.
* `response.content` is the bytes-response from the server.
  The way it was used before probably never worked the way it was done:
  A byte-array can not be splitted using a string.
  When using `response.text` decoding is done by `requests` what is
  probably the best idea here.

This problems showed up when I wanted to bring up a new lab in our infrastructure that is equipped with a Digipower power-switch.
I successfully tested the `get` and `set` methods against our device.

**Checklist**
- [ ] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/labgrid/driver/power/digipower.py
+++ b/labgrid/driver/power/digipower.py
@@ -12,18 +12,8 @@ def power_set(host, port, index, value):
         cgi = "offs.cgi"
 
     suffixstring = "0000000000000000"
-    portstring = {
-        1: "10000000",
-        2: "01000000",
-        3: "00100000",
-        4: "00010000",
-        5: "00001000",
-        6: "00000100",
-        7: "00000010",
-        8: "00000001",
-    }
     r = requests.get(
-        "http://{}:{}/{}?led={}".format(host, port, cgi, portstring[index] + suffixstring),
+        f"http://{host}:{port}/{cgi}?led={1 << 8 - index:08b}{suffixstring}",
         auth=("snmp", "1234"),
     )
     r.raise_for_status()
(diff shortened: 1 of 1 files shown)


Emantor (approved):

Emantor (approved):

labgrid/driver/power/digipower.py
```diff
@@ -13,14 +13,14 @@ def power_set(host, port, index, value):
...
-        "7": "00000010",
-        "8": "00000001",
+        1: "10000000",
+        2: "01000000",
+        3: "00100000",
+        4: "00010000",
+        5: "00001000",
+        6: "00000100",
+        7: "00000010",
+        8: "00000001",
```
Bastian-Krause, line 23: It's probably better to get rid of this altogether. What about `f'{1 << 8-index:08b}0000000000000000'` instead of `portstring[index] + suffixstring`?

labgrid/driver/power/digipower.py
```diff
@@ -38,13 +38,13 @@ def power_get(host, port, index):
...
-        "8": 17
+        1: 10,
+        2: 11,
+        3: 12,
+        4: 13,
+        5: 14,
+        6: 15,
+        7: 16,
+        8: 17
     }
```
Bastian-Krause, line 49: This is basically the same as above, but comma-separated. So something like this instead of states/ports/return?
```
state = r.text.split(',')[9+index]
return bool(int(state, 2))
```

Bastian-Krause (changes requested):

labgrid/driver/power/digipower.py
jluebbe: Maybe still keep portstring and suffixstring separate for reability. Otherwise the format is much nicer than a dict here.

Emantor (approved):

Bastian-Krause (approved):
Looks good, thanks!

jluebbe (approved):

PR #780, opened 2021-05-03
modbusdriver: retrieve exception information

Emantor:
We got a report of a user trying to access a modbus device with the
following error:
```
  File "/opt/venvs/labgrid/lib/python3.9/site-packages/labgrid/remote/client.py", line 1638, in main
    args.func(session)
  File "/opt/venvs/labgrid/lib/python3.9/site-packages/labgrid/remote/client.py", line 814, in digital_io
    drv.set(False)
  File "/opt/venvs/labgrid/lib/python3.9/site-packages/labgrid/binding.py", line 100, in wrapper
  File "/opt/venvs/labgrid/lib/python3.9/site-packages/labgrid/driver/modbusdriver.py", line 36, in set
    raise ExecutionError('Could not write coil (code={})'.format(error_code))
  labgrid.driver.exception.ExecutionError: Could not write coil (code=7)
```
The error code is 7, indicating that an exception happened, however
labgrid did not retrieve the exception code, so no further debugging was
possible. Unify the modbusdriver error handling and retrieve the
exception if it happened. Also ignore the ACKNOWLEDGE and NONE exceptions.

**Description**

- [ ] PR has been tested

### DIFF
--- a/labgrid/driver/modbusdriver.py
+++ b/labgrid/driver/modbusdriver.py
@@ -15,6 +15,7 @@ class ModbusCoilDriver(Driver, DigitalOutputProtocol):
     def __attrs_post_init__(self):
         super().__attrs_post_init__()
         self._module = import_module('pyModbusTCP.client')
+        self._consts = import_module('pyModbusTCP.constants')
         self.client = None
 
     def on_activate(self):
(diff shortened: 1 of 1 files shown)


labgrid/driver/modbusdriver.py
```diff
@@ -26,21 +26,31 @@ def on_activate(self):
     def on_deactivate(self):
         self.client = None
 
+    def handle_error(self, action):
+        error_code = self.client.last_error()
+        if error_code == 7:
+            exc = self.client.last_except()
+            if exc not in [0, 5]:
```
jluebbe, line 33: Some comments for the numeric values would be useful. Perhaps they are already available as Enums in pyModbusTCP?

Emantor: Done, We now additionally import the constants from pyModbusTCP and use instead of the magic values.

jluebbe (approved):
I've changed `handle_error` to a private function, as it should not be called from the outside.

jluebbe:
@Emantor, please merge if you're fine with my change.

PR #783, opened 2021-05-07
driver: shelldriver: add get_default_interface_device_name()/get_ip_interfaces()

Bastian-Krause:
**Description**
Based on labgridhelper.linux's `get_interface_ip()` function, introduce methods to retrieve the default route's interface device name as well as IP interfaces.

A `ShellDriver` bound on top of a `SerialDriver` is now able to retrieve the target's IP during runtime. That's useful for targets without a fixed IP. This could allow set up of automated SSH access in a future PR.

**Checklist**
- [ ] Tests for the feature 
- [x] PR has been tested

### DIFF
--- a/labgrid/driver/shelldriver.py
+++ b/labgrid/driver/shelldriver.py
@@ -536,3 +536,32 @@ class ShellDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
             IOError: if the provided localfile could not be found
         """
         return self._run_script_file(scriptfile, *args, timeout=timeout)
+
+    @Driver.check_active
+    def get_default_interface_device_name(self, version=4):
+        """ Retrieve the default route's interface device name.
+
+        Args:
+            version (int): IP version
+
+        Returns:
+            Name of the device of the default route
+
+        Raises:
+            ExecutionError: if no or multiple routes are set up
+        """
+        assert version in (4, 6)
+
+        regex = r"""default\s+via # leading strings
+                \s+\S+ # IP address
+                \s+dev\s+(\w+) # interface"""
+
+        default_route = self._run_check(f"ip -{version} route list default")
+        matches = re.findall(regex, "\n".join(default_route), re.X)
+
+        if not matches:
+            raise ExecutionError(f"No IPv{version} default route found")
+        if len(matches) > 1:
+            raise ExecutionError(f"Multiple IPv{version} default routes found")
+
+        return matches[0]
--- a/labgrid/driver/shelldriver.py
+++ b/labgrid/driver/shelldriver.py
@@ -6,6 +6,8 @@ import logging
 import re
 import time
 import shlex
+import ipaddress
+
 import attr
 from pexpect import TIMEOUT
 import xmodem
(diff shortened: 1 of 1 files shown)


labgrid/driver/shelldriver.py
```diff
@@ -565,3 +567,26 @@ def get_default_interface_device_name(self, version=4):
...
+        """
+        if device is None:
+            device = self.get_default_interface_device_name()
+
+        regex = r"""\d+: # leading number
+                \s+\w+ # interface name
+                \s+inet6?\s+(\S+) # IP address, prefix
+                .*global # global scope, not host scope"""
+
+        ip_show = self._run_check(f"ip -o addr show dev {device}")
```
JoshuaWatt, line 589: You can output directly parsable JSON with `ip -j addr`. I've done something similar to this before with https://github.com/JPEWdev/yocto-doom-demo/blob/main/ci/conftest.py#L29, but in that case I was assigning the address specified in the `NetworkResource` if it wasn't already assigned.

JoshuaWatt:
> **Description**
> Based on labgridhelper.linux's `get_interface_ip()` function, introduce methods to retrieve the default route's interface device name as well as IP interfaces.
> 
> A `ShellDriver` bound on top of a `SerialDriver` is now able to retrieve the target's IP during runtime. That's useful for targets without a fixed IP. This could allow set up of automated SSH access in a future PR.

This would be really useful for my use case, do you have an idea how it could be done?

> 
> **Checklist**
> 
>     * [ ]  Documentation for the feature
> 
>     * [ ]  Tests for the feature
> 
>     * [ ]  The arguments and description in doc/configuration.rst have been updated
> 
>     * [ ]  Add a section on how to use the feature to doc/usage.rst
> 
>     * [ ]  CHANGES.rst has been updated
> 
>     * [x]  PR has been tested

labgrid/driver/shelldriver.py
Bastian-Krause: Unfortunately busybox' `ip` does not support `-j`, so this is not an option.

Bastian-Krause:
> > A `ShellDriver` bound on top of a `SerialDriver` is now able to retrieve the target's IP during runtime. That's useful for targets without a fixed IP. This could allow set up of automated SSH access in a future PR.
> 
> This would be really useful for my use case, do you have an idea how it could be done?

Not sure. I guess a `NetworkService` should be instantiated for the target. Is this correct, @Emantor, @jluebbe? Where should that happen?

Emantor:
Usually this is done within the Strategy, i.e. you add an `ssh` state, boot the device to the shell, retrieve the ip and than add a NetworkService onto the target.

labgrid/driver/shelldriver.py
```diff
@@ -536,3 +538,55 @@ def run_script_file(self, scriptfile: str, *args, timeout: int = 60):
...
+        if not matches:
+            raise ExecutionError(f"No IPv{version} default route found")
+        if len(matches) > 1:
+            raise ExecutionError(f"Multiple IPv{version} default routes found")
+
+        return matches[0]
+
+    @Driver.check_active
+    def get_ip_interfaces(self, device=None):
+        """ Retrieves IP interfaces for given interface name.
```
jluebbe, line 573: I find the function name confusing. It returns the IP *addresses* of a single network *interface*. @Bastian-Krause probably chose `get_ip_interfaces` because the ipaddress modules has a class name `ipaddress.IPv4/6Interface`. This is a subclass of the corresponding `IPv4/6Address`, so it'd suggest to call it `get_ip_addresses`.

jluebbe (changes requested):

labgrid/driver/shelldriver.py
Bastian-Krause: Right, so should we still return IPv4Interface/IPv6Interface objects here or rather IPv4Address/IPv6Address?

jluebbe: Yes, the just contain additional information (the prefix length).

Bastian-Krause:
Force-pushed:
- renamed `get_ip_interfaces` -> `get_ip_addresses`
- improved docstring

jluebbe (approved):

PR #790, opened 2021-05-28
HIDRelay: Add serial number filtering

JoshuaWatt:
Adds support for checking the HID Relay serial number when scanning
devices. The serial number is fetched from the devices HID report.

**Description**

**Checklist**
- [ ] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -361,9 +361,11 @@ It currently supports the widely used "dcttech USBRelay".
    HIDRelay:
      index: 2
      invert: False
+     serial: 6QMBS
 
 - index (int): number of the relay to use (defaults to 1)
 - invert (bool): whether to invert the relay
+- serial (str): relay serial number (defaults to "")
 
 Used by:
   - `HIDRelayDriver`_
--- a/dockerfiles/Dockerfile
+++ b/dockerfiles/Dockerfile
@@ -115,6 +115,7 @@ RUN set -e ;\
     apt install -q=2 --yes --no-install-recommends \
         libyaml-0-2 \
         libsctp1 \
+        libusb-1.0-0 \
     ; \
     apt clean ;\
     rm -rf /var/lib/apt/lists/*
--- a/labgrid/resource/udev.py
+++ b/labgrid/resource/udev.py
@@ -7,6 +7,8 @@ from collections import OrderedDict
 
 import attr
 import pyudev
+import usb.core
+import usb.util
 
 from ..factory import target_factory
 from .common import ManagedResource, ResourceManager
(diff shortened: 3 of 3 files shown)


jluebbe:
We don't want to use libusb to open to devices during matching as this causes hard to debug issues if this happens while other instances are accessing those devices.

JoshuaWatt:
> We don't want to use libusb to open to devices during matching as this causes hard to debug issues if this happens while other instances are accessing those devices.

Any ideas then? You can't attach multiple HID relays to a single exporter otherwise, because the (stupidly) don't set the USB serial ID.

Is it OK that it is optional?

jluebbe:
For multiple relays, you'd have to use matching by path (ID_PATH).

Due to the difficulty of knowing whether a reported issue is be caused by a conflicting libusb access, I'm rather skeptical of any direct access to devices during matching. As other devices sometimes have the same issue with serial numbers that are only available in-band, I've looked at this before and not found a robust way to support this. So, currently resource implementations shouldn't do anything that might conflict with other processes. :/

In the exporter, "active checks" might be OK, as in that case no parallel access should be happening (the device would only be available after the matching process completes). It might be confusing to users though, if some match options are only available with the exporter.

JoshuaWatt:
> For multiple relays, you'd have to use matching by path (ID_PATH).
> 
> Due to the difficulty of knowing whether a reported issue is be caused by a conflicting libusb access, I'm rather skeptical of any direct access to devices during matching. As other devices sometimes have the same issue with serial numbers that are only available in-band, I've looked at this before and not found a robust way to support this. So, currently resource implementations shouldn't do anything that might conflict with other processes. :/
> 
> In the exporter, "active checks" might be OK, as in that case no parallel access should be happening (the device would only be available after the matching process completes). It might be confusing to users though, if some match options are only available with the exporter.

mm, what if the `HIDRelayDriver` sent the serial instead of the busnum and devnum to the `usb_hid_relay` agent and it can find the matching HID device? That might mean the serial should should mandatory, otherwise the exported busnum and devnum might be complete lies....

jluebbe:
> mm, what if the `HIDRelayDriver` sent the serial instead of the busnum and devnum to the `usb_hid_relay` agent and it can find the matching HID device? That might mean the serial should should mandatory, otherwise the exported busnum and devnum might be complete lies....

That would mean that the agent would have to open all HID relay devices, possibly conflicting with concurrent accesses.

JoshuaWatt:
Switched to having udev lookup the serial number using https://pypi.org/project/pyhid-usb-relay/ during an add event and adding a udev rule match in labgrid, which eliminates the racing

jluebbe:
That's a nice solution! Could you share your udev rule for others in the same situation?

jluebbe:
Ah, I found the example: https://github.com/JPEWdev/pyhid-usb-relay#fixing-serial-numbers

Issue #794, opened 2021-07-07
Storing passwords in a secure manner


initBasti:
Hello,

I just wanted to quickly check if I missed something. Within the getting started section, I set up an initial [local. yaml](https://labgrid.readthedocs.io/en/latest/getting_started.html#running-your-first-test) for the first test and I am expected to change the port, username, and password. I have searched through all search results for the word password in the documentation and within the GitHub issues, but I was not able to find any hint on how to securely manage passwords. I was expecting something like a keyring, gpg encrypted file, or some connection to a password manager, but it seems like I should use a plain text password in the `.yaml`.
Am I wrong or could somebody explain to me the current best practices for storing passwords for different boards?

Greetings,
Sebastian

jluebbe:
There is no support for encrypting passwords. The expectation is that images for testing would have well known passwords (and SSH keys and other credentials). labgrid itself also has no authentication for access to the coordinator, so it's intended to be used in a trusted environment.

initBasti:
Hey @jluebbe,

Thank you for your response, so in order to secure my test setup, I will have to make sure that the master device, which coordinates the work onto the devices under test, cannot be penetrated, right? I was just wondering because I would like to host a CI system on the master device and connect it to multiple test devices (for camera quality testing) and I thought that if an attacker is able to get into the master device he can then easily access all of the prepared test devices. Maybe I have a use case, which does not fit the design of labgrid.

Thank you.
Sebastian

jluebbe:
We're currently using a separate VLAN for the labs and only specific systems have access to that network (specific CI runners and developer hosts). That lab network also contains other network devices (power switches, serial servers, ...) that don't have any authentication, so it needs to be relatively protected in any case.

initBasti:
Thank you very much for giving me an example, I believe it would be awesome if you could link to a source describing how to properly set up a secure lab in the documentation or if the documentation would contain an example setup. Do you also think this would help or would that be above the scope of labgrid?

jluebbe:
I'd think that configuring the Linux hosts and routers in a test-lab setup would be out of scope for labgrid's documentation, as this is a complex topic and also applies to any lab network (with or without labgrid). If you'd like design support or review for your specific project setup, my employer (Pengutronix) can offer consulting.

initBasti:
Thank you for the hint but I am a poor developer using my own time and money to build a test lab for an open-source project, so this is out of my scope :) I'll try to figure it out and maybe I can contribute to your project as well.

[...]

PR #797, opened 2021-07-13
driver/power: Add backend driver for EG PMS2 LAN/WLAN

initBasti:
This driver adds support for the Energenie PMS2 LAN & WLAN programmable
power sockets. They are accessed over a web interface and require a
login before accessing the status of the socket or switching them on or off.
A password is required for the login in POST request and this driver
deliberately uses the default password, as the current state of labgrid
doesn't support password encryption, meaning the password would be
stored in plain text.

Signed-off-by: Sebastian Fricke <sebastian.fricke@posteo.net>

Small note about passwords:
From my discussion in this [issue](https://github.com/labgrid-project/labgrid/issues/794) I have gotten the notion, that security in labgrid is performed on the network level, therefore I have decided to keep this module as simple as possible by demanding the default password. If we want to allow the user to change the password, then I believe this patch will become a lot more complex as the NetworkPowerDriver does not support such an attribute and therefore there might be the need for an alternative approach.
I would like to hear your thoughts about this.

What do you use the feature for?
To work with my Energenie programmable power sockets within my personal lab.

How does labgrid benefit as a testing library from the feature?
A wider range of supported devices increases the usefulness of the library for a bigger audience. 

How did you verify the feature works?
I have tested it in my personal lab with the EG_PMS2_LAN device.

If hardware is needed for the feature, which hardware is supported, and which hardware did you test with?
Currently, I know that the device should work for both the EG_PMS2_LAN and EG_PMS2_WLAN devices as both are from the same series, but I only tested with EG_PMS2_LAN. The driver might also work for the EG_PM2 series but I have not tested it (the only difference between the PMS2 and PM2 seems to be surge protection).

**Checklist**
- [x] Tests for the feature 
- [ ] Add a section on how to use the feature to doc/usage.rst (should not be required for such a simple module)
- [ ] Add a section on how to use the feature to doc/development.rst (should not be required for such a simple module)
- [x] PR has been tested

### DIFF
--- a/labgrid/driver/power/eg_pms2_network.py
+++ b/labgrid/driver/power/eg_pms2_network.py
@@ -1,0 +1,77 @@
+import re
+import requests
+
+from ..exception import ExecutionError
+
+# This driver implementes a power port for the EG_PMS2_LAN & EG_PMS2_WLAN
+# devices, it works through a simple HTTP interface, that requires a login.
+# Driver has been tested with:
+# * EG_PMS2_LAN
+
+# The default HTTP port for usage via the SSH proxy.
+PORT = 80
+
+# The login was successful when we can locate a Log out button
+LOGIN_SUCCESS_REGEX = r'<div class="boxmenuitem"><a href="login\.html">Log Out</a></div>'
+# Search for a string similar to: "var sockstates = [1,1,1,1];" and create
+# a match group out of the numbers within the square brackets.
+SOCKSTATES_REGEX = r"var\s+sockstates\s+=\s+\[([01]),([01]),([01]),([01])\];"
+
+
+def login(base_url: str) -> None:
+    """
+    Use the default password 1, because labgrid doesn't support password
+    encryption, modifying the password doesn't secure the device as the
+    password would be stored as plain-text.
+    """
+    login_url = f"{base_url}/login.html"
+    try:
+        response = requests.post(login_url, data={'pw': 1})
+    except requests.exceptions.ConnectionError as err:
+        raise ExecutionError(
+            f"Device not found at {base_url} or the network interface of the "
+            "device is not active (press the reset button on the device)"
+        ) from err
+    if response.status_code != 200 or not re.search(LOGIN_SUCCESS_REGEX,
+                                                    response.text):
+        raise ExecutionError("Login to Energenie web interface failed")
+
+
+def logout(base_url: str) -> None:
+    """
+    After a successful login, the session is reserved for the IP address.
+    Logout explicitly to allow accessing the device from different hosts.
+    """
+    response = requests.get(f"{base_url}/login.html")
+    if response.status_code != 200:
+        raise ExecutionError("Logout from Energenie web interface failed")
+
+
+def power_set(host, port, index, value):
+    base_url = f"http://{host}:{port}"
+    index = int(index)
+    assert 1 <= index <= 4
+
+    value = 1 if value else 0
+    login(base_url=base_url)
+    response = requests.post(base_url, data={f'cte{index}': value})
+    response.raise_for_status()
+    logout(base_url=base_url)
+    if not response.status_code == 200:
+        raise ExecutionError(f"Switching socket at index {index} "
+                             f"{'on' if value else 'off'} failed")
+
+
+def power_get(host, port, index):
+    base_url = f"http://{host}:{port}"
+    index = int(index)
+    assert 1 <= index <= 4
+
+    # Fetch status
+    login(base_url=base_url)
+    response = requests.get(f"{base_url}/energenie.html")
+    response.raise_for_status()
+    logout(base_url=base_url)
+    match_group = re.search(SOCKSTATES_REGEX, response.text)
+    sockstates = [int(match_group.group(x)) for x in range(1, 5)]
+    return sockstates[index - 1] == 1
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,8 @@ New Features in 0.5.0
   considering only the closest marker.
 - The labgrid client SSH command is now able to instantiate the SSHDriver when
   there are multiple NetworkService resources available.
+- eg_pms2_network power port driver supports controlling the Energenie power
+  management series with devices like the EG_PMS2_LAN & EG_PMS2_WLAN
 
 Bug fixes in 0.5.0
 ~~~~~~~~~~~~~~~~~~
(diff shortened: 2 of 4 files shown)


Emantor (approved):
LGTM, you surely want to remove the `PORT` variable before merging.

initBasti:
@Emantor 
Sorry for the delay, had a busy week. I've removed the line in question, I hope that it is all fine now.
Thank you for the review.

labgrid/driver/power/eg_pms2_network.py
```diff
@@ -0,0 +1,73 @@
...
+            raise ExecutionError("Login to Energenie web interface failed")
+
+        result = func(host, port, index, value)
+
+        # Logout
+        response = requests.get(login_url)
+        if response.status_code != 200:
+            raise ExecutionError("Logout from Energenie web interface failed")
+        return result
+    return wrapper
```
jluebbe, line 47: Wouldn't two helper functions (login/logout) instead of the decorator be simpler?

initBasti: Hey Jan,

I guess they would be sufficient, I just applied my personal taste in this code and I am not mad if we go another route.
I just felt that...
```python
@login
def power_get(host, port, index):
    ...
    response = requests.get(status_url)
    ...
```
... concentrates more on the action that we are trying to do which is getting the status from the device. The login and logout felt more like a required extra, which can be abstracted out of the specific code into a decorator.
The alternative would be:
```python
def power_get(host, port, index):
    ...
    login(base_url=base_url)
    response = requests.get(status_url)
    logout(base_url=base_url)
    ...
```
Which from my point of view is a lot more verbose, so I prefer the decorator, but again no strong opinions, if simple helper functions fit more into the coding style of labgrid then I will change my code.

Greetings,
Sebastian

initBasti: @jluebbe 
Hey just wanted to check in, I would rebase this commit to the current tree, do you want me to refactor the code as well?

Greetings,
Sebastian

initBasti:
Hey @jluebbe @Emantor ,
I have refactored the decorator into two separate functions and documented why the logout functionality is useful. Additionally, I noticed that the change was declared for 3.0 which is already released so I changed it accordingly within the CHANGES.rst file.
I have tested this on my setup and it works without problems.
Greetings,
Sebastian

initBasti:
Hey @jluebbe  & @Emantor ,

just wanted to quickly check, is there something left to do for me?

Greetings,
Sebastian

Emantor (approved):

Emantor:
> Hey @jluebbe & @Emantor ,
> 
> just wanted to quickly check, is there something left to do for me?

AFAICS no, thanks for the ping.

labgrid/driver/power/eg_pms2_network.py
```diff
@@ -0,0 +1,74 @@
...
+import requests
+
+from ..exception import ExecutionError
+
+# This driver implementes a power port for the EG_PMS2_LAN & EG_PMS2_WLAN
+# devices, it works through a simple HTTP interface, that requires a login.
+# Driver has been tested with:
+# * EG_PMS2_LAN
+
+PORT = 80
```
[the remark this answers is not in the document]
jluebbe: This is needed for the SSH tunnel/proxy. I've made a commit to add it again.

jluebbe (approved):
Without the `PORT` variable, access via an SSH proxy will not work (see https://github.com/labgrid-project/labgrid/blob/master/labgrid/driver/powerdriver.py#L170). I've added a commit to fix that.

labgrid/driver/power/eg_pms2_network.py
```diff
@@ -0,0 +1,77 @@
...
+# devices, it works through a simple HTTP interface, that requires a login.
+# Driver has been tested with:
+# * EG_PMS2_LAN
+
+# The default HTTP port for usage via the SSH proxy.
+PORT = 80
+
+# The login was successful when we can locate a Log out button
+LOGIN_SUCCESS_REGEX = str("<div class=\"boxmenuitem\"><a href="
+                          "\"login\\.html\">Log Out</a></div>")
```
Bastian-Krause, line 16:
```suggestion
LOGIN_SUCCESS_REGEX = r'<div class="boxmenuitem"><a href="login\.html">Log Out</a></div>'
```

labgrid/driver/power/eg_pms2_network.py
```diff
@@ -0,0 +1,77 @@
...
+# Search for a string similar to: "var sockstates = [1,1,1,1];" and create
+# a match group out of the numbers within the square brackets.
+SOCKSTATES_REGEX = str("var\s+sockstates\s+=\s+"
+                       "\[([01]),([01]),([01]),([01])\];")
```
Bastian-Krause, line 20:
```suggestion
SOCKSTATES_REGEX = r"var\s+sockstates\s+=\s+\[([01]),([01]),([01]),([01])\];"
```

Bastian-Krause (changes requested):

Bastian-Krause (dismissed):
Looks good to me. I have not tested this, since I don't have the hardware needed.

labgrid/driver/power/eg_pms2_network.py
```diff
@@ -0,0 +1,75 @@
...
+    index = int(index)
+    assert 1 <= index <= 4
+
+    value = 1 if value else 0
+    login(base_url=base_url)
+    response = requests.post(base_url, data={f'cte{index}': value})
+    logout(base_url=base_url)
+    if not response.status_code == 200:
+        raise ExecutionError(f"Switching socket at index {index} "
+                             f"{'on' if value else 'off'} failed")
```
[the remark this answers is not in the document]
initBasti: I think in this case it might actually be more descriptive to add the custom message to highlight the state at which the error occurred (index and desired state of the socket).

labgrid/driver/power/eg_pms2_network.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+def power_get(host, port, index):
+    base_url = f"http://{host}:{port}"
+    index = int(index)
+    assert 1 <= index <= 4
+
+    # Fetch status
+    login(base_url=base_url)
+    response = requests.get(f"{base_url}/energenie.html")
```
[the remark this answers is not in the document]
Bastian-Krause: This is probably still a good idea, right?

Emantor (approved):
I rebased and added additional `raise_for_status()` calls.

initBasti:
@Emantor Thanks a lot! I was quite busy working myself into my new job, so I delayed finishing this patch... sorry for that!

PR #798, opened 2021-07-13
remote/exporter: info print ser2net command

Emantor:
**Description**
This is useful for debugging if ser2net is not starting correctly.

### DIFF
--- a/labgrid/remote/exporter.py
+++ b/labgrid/remote/exporter.py
@@ -230,16 +230,16 @@ class SerialPortExport(ResourceExport):
         _, _, version = str(subprocess.check_output([self.ser2net_bin,'-v'])).split(' ')
         major_version = version.split('.')[0]
         if int(major_version) >= 4:
-            self.child = subprocess.Popen([
+            cmd = [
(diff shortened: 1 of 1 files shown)


labgrid/remote/exporter.py
```diff
@@ -248,7 +250,9 @@ def _start(self, start_params):
                 '{}:telnet:0:{}:{} NONE 8DATABITS 1STOPBIT LOCAL'.format(
                     self.port, start_params['path'], self.local.speed
                 ),
-            ])
+            ]
+            self.logger.info("Starting ser2net with: {}".format(" ".join(cmd)))
+            self.child = subprocess.Popen(cmd)
```
jluebbe, line 255: Maybe pull both out of the if/else?

jluebbe (changes requested):

labgrid/remote/exporter.py
Emantor: Good point.

jluebbe (approved):

PR #799, opened 2021-07-15
WIP: yaml include

b2vn:
This is still early days for the functionality, but I would like to get feed back if the concept holds and should be part of labgrid in the future. 

**Description**

The functionality added it to allow to split the lg-env yaml into multiple files. The rationale for this is that we have multiple test sites that essential are the same with very minor differences, e.g. different ip addresses of resources. With this change it is possible to have a `common.yaml` (or any other name) that describes that describes all that is common and then implement the individual sites/envs in separate file that will include the common stuff.

Special attention: 

* I had to change the unittests for some of the existing tests. See ecd3f08e462f6e5b44d6b574c65f8b8d5677f8f9
* There is an issue in the current implementation of the yaml.py::load() method. The custom loader implemented does not work with anchors and references in the yaml. This is not originated to the include-feature, but becomes an issue for it. I do not know how to fix that problem correct, but this PR contains a work around. If the feature in this PR is welcomed into the master, I will create an issue on this. See the comments in `labgrid/util/yaml.py` in 2c0a9fce18ccbe7a1bb2218d365a092421ce7bfa
 
**Alternative solution**

The functionality covered in this PR can also be achieved with scripts that can "stitch" specialized yaml files. This solution would allow much more complex use cases that covered by this PR, e.g. storing sites and configuration in a database and be able to auto generate specific yaml files. However, it is assumed that the issue in this PR will cover far the most examples.

**Checklist**

- [ ] Tests for the feature 
- [x ] PR has been tested

### DIFF
--- a/examples/yamlinclude/README.md
+++ b/examples/yamlinclude/README.md
@@ -1,0 +1,99 @@
+# Splitting configuration into multiple yaml files
+
+**TODO**:*write a coherent prose*
+
+
+Essentially the files listed in `includes` are copy pasted into the "base" yaml with all the limitation it will cause.
+
+Limitations:
+
+* It is not allow to have multiple blocks with the same name. See the section 'Same block in multiple files'.
+* Anchors must be defined before they are referenced. See the section 'Defining anchors'
+
+## Same block in multiple files
+
+This is not possible due to limitation in YAML. In the following example the `defaults` block from `first.yaml` will be overwritten with the `defaults` block from `second.yaml`
+
+```yaml
+# filename: first.yaml
+defaults:
+    foo: "hello"
+
+includes:
+    - "second.yaml"
+```
+
+```yaml
+# filename: second.yaml
+defaults:
+    bar: "world"
+```
+
+The output produced by the **TODO***:name of this include functionality* will be
+
+```yaml
+# filename: first.yaml
+defaults:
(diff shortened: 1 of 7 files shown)


b2vn:
@Emantor, @Bastian-Krause and @jluebbe could you please have a look at this?

Emantor:
IMO we should use the existing [pyyaml-include](https://github.com/tanbro/pyyaml-include) module to implement this.

b2vn:
Oh, I did not know of pyyaml-include. Yes, that is obviously the right way to go. I will look into that. Thanks :)

Emantor:
> Oh, I did not know of pyyaml-include. Yes, that is obviously the right way to go. I will look into that. Thanks :)

Why do you need this in the first place? IMO this should all be covered by the remote infrastructure, i.e. by setting up an exporter and coordinator. For multiple boards you'll run out of ports on one machine eventually and growing to more machine will require the remote infrastructure anyway. The remote infrastrucures exporter is templated anyway, so you can export all configuration options at once. If your online anyway it may be easier to discuss this over [matrix](https://app.element.io/#/room/#labgrid:matrix.org).

jluebbe:
Closing this for now. Please reopen if you want to continue with this approach.

PR #806, opened 2021-07-26
driver: add modbus rtu driver

b2vn:
**Description**

Added a Modbus RTU driver to allow for labgrid tests to control modbus devices.

**Checklist**
- [X] Tests for the feature 
- [X] CHANGES.rst has been updated
- [X] PR has been tested

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -64,6 +64,42 @@ port 53867 and use a baud rate of 115200 with the RFC2217 protocol.
 Used by:
   - `SerialDriver`_
 
+ModbusRTU
++++++++++
+Describes the resource required to use the ModbusRTU driver.
+`Modbus RTU <https://en.wikipedia.org/wiki/Modbus>`_ is a communication
+protocol used to control many different kinds of electronic systems, such as
+thermostats, power plants, etc.
+Modbus is normally implemented on top of RS-485, though this is not strictly
+necessary, as long as the Modbus network only has one master (and up to 256
+slaves).
+
+The labgrid driver is implemented using the minimalmodbus Python library.
+The implementation only supports that labgrid will be the master on the Modbus
+network.
+For more information, see `minimalmodbus
+<https://minimalmodbus.readthedocs.io/en/stable/>`_.
+
+This resource and driver only supports local usage and will not work with an
+exporter.
+
+.. code-block:: yaml
+
+    ModbusRTU:
+      port: "/dev/ttyUSB0"
+      address: 16
+      speed: 115200
+      timeout: 0.25
+
+Arguments:
+  - port (str): tty the instrument is connected to, e.g. '/dev/ttyUSB0'
+  - address (int): slave address on the modbus, e.g. 16
+  - baudrate (bool): optional, default is 115200
+  - timeout (bool): optional, timeout in seconds. Default is 0.25 s
+
+Used by:
+  - `ModbusRTUDriver`_
+
 USBSerialPort
 +++++++++++++
 A USBSerialPort describes a serial port which is connected via USB and is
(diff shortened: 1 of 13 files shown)


b2vn:
@Emantor, @Bastian-Krause  would you take a look at this?

labgrid/resource/modbusrtu.py
```diff
@@ -0,0 +1,24 @@
...
+
+    Args:
+        port (str): tty the instrument is connected to, e.g. '/dev/ttyUSB0'
+        address (int): slave address on the modbus, e.g. 16
+        baudrate (bool): optional, default is 115200
+        timeout (bool): optional, timeout in seconds. Default is 0.25 s
+    """
+    port = attr.ib(validator=attr.validators.instance_of(str))
+    address = attr.ib(validator=attr.validators.instance_of(int))
+    speed = attr.ib(default=115200,
```
Emantor, line 21: The `SerialPort` baseclass already has port & speed, why is this redefined here?

labgrid/driver/modbusrtudriver.py
```diff
@@ -0,0 +1,64 @@
+from importlib import import_module
+import attr
+
+from ..factory import target_factory
+from .common import Driver
+
+
+@target_factory.reg_driver
+@attr.s(eq=False)
+class ModbusRTUDriver(Driver):
```
Emantor, line 10: What kind of device are you using this driver with? Rather than a blanked implementation where we basically pass through all the read/write functions, I'd rather add a specific device driver on top.

modbusrtu-requirements.txt
```diff
@@ -0,0 +1 @@
+minimalmodbus==1.0.2
```
Emantor, line 1: I guess that the `modbustcp` module we currently used can't be reused here, would be nice to have a one module solution to cover both cases. Maybe minimalmodbus supports tcp as well?

doc/configuration.rst
```diff
@@ -64,6 +64,27 @@ port 53867 and use a baud rate of 115200 with the RFC2217 protocol.
 Used by:
   - `SerialDriver`_
 
+ModbusRTU
++++++++++
+Descriped the resource required to use the Modbus RTU driver
```
Emantor, line 69: A bit more explanation would be helpful, maybe a link to a Modbus RTU protocol description. Is Modbus RTU always used over serial interfaces?

modbusrtu-requirements.txt
b2vn: Correct. Unfortunately modbus tcp and rtu are two quite different things. Also minimalmodbus does not support tcp.

labgrid/driver/modbusrtudriver.py
b2vn: I theory I agree. But I'm not sure which abstract driver it should implement. Modbus has a lot in common with the SerialDriver - how ever it does not make sense (IMO) to implement it and thereby give the user access to raw read() and write() methods, which would require the user to implement framing, CRC, ACK/NACK, timing, etc. that this driver is there to handle.

An other option would be to let this class inherit from something with eeprom as Modbus basically just allows you to read/write specific addresses on another device. However, I don't think any of the existing drivers offers that level of details.

b2vn:
@Emantor Could you have another look at this?

Emantor (approved):

jluebbe (dismissed):
I've fixed some typos in the docs an thing we can merge this as is.

@b2vn, if you have some time, you could add `@Driver.check_active` to the driver functions to avoid failing only when trying to access the instrument property.

Also, the checks in the resource `__attrs_post_init__` should be better handled via attrs.

Emantor (approved):

Emantor (approved):

jluebbe (approved):

jluebbe:
Although the GitHub UI doesn't show it, the CI run was successful: https://github.com/labgrid-project/labgrid/actions/runs/1506655629

PR #807, opened 2021-07-27
proxy: add ProxyError, check host and port

Emantor:
**Description**
Add a ProxyError and check that we have a valid host and port. If we
don't raise a ProxyError with a proper explanation.

**Checklist**
- [ ] Tests for the feature 
- [ ] PR has been tested

### DIFF
--- a/labgrid/util/proxy.py
+++ b/labgrid/util/proxy.py
@@ -9,6 +9,10 @@ from ..resource.common import Resource
 __all__ = ['proxymanager']
 
 
+class ProxyError(Exception):
+    pass
+
+
 class ProxyManager:
     """The ProxyManager class is only used inside labgrid.util.proxy (similar
     to a singleton), don't instanciate this class, use the exported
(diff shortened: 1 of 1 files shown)


labgrid/util/proxy.py
```diff
@@ -39,15 +43,23 @@ def get_host_and_port(cls, res, *, default_port=None, force_port=None):
...
             port = s.port
         else:
             port = getattr(res, 'port', None) or default_port
 
+        if host is None:
+            raise ProxyError(f"Invalid url: {url} does not contain a host")
+
+        if port is None:
+            raise ProxyError(f"Invalid url: {url} does not contain a port and no default set")
+
```
jluebbe, line 62: The input to this function is not an URL, so these messages are a bit confusing. The port can come from the arguments or from the resource. The hostname is always from the resource and can't be missing at this point (res.host should never be None, though).

jluebbe (changes requested):

labgrid/util/proxy.py
Emantor: Hm, I don't think we need the check here than, I'll remove it if that is fine with you.

jluebbe: Sounds fine.

jluebbe (approved):

PR #811, opened 2021-08-02
target: Quieten message when resource is unavailable

JoshuaWatt:
Having this message at info() level means it shows up on stdout when a
resource disappears during a test. This is unnecessary as any driver
using the resource will fail when an attempt is made to use it.

**Description**

**Checklist**
- [ ] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/labgrid/target.py
+++ b/labgrid/target.py
@@ -51,8 +51,15 @@ class Target:
         for resource in self.resources:
             resource.poll()
             if not resource.avail and resource.state is BindingState.active:
-                self.log.info("deactivating unavailable resource %s", resource.display_name)  # pylint: disable=line-too-long
-                self.deactivate(resource)
+                deactivated = self.deactivate(resource)
+                deactivated.remove(resource)
+                if deactivated:
+                    self.log.info("deactivating unavailable resource %s used by %s",
+                        resource.display_name,
+                        ", ".join(d.display_name for d in deactivated)
+                    )
+                else:
+                    self.log.debug("deactivating unavailable resource %s (unused)", resource.display_name)  # pylint: disable=line-too-long
 
     def await_resources(self, resources, timeout=None, avail=True):
         """
@@ -438,6 +445,8 @@ class Target:
         Recursively deactivate the client's clients and itself.
 
         This is needed to ensure that no client has an inactive supplier.
+
+        Returns the list of all objects that were deactivated
         """
         if isinstance(client, str):
             cls = target_factory.class_from_string(client)
@@ -446,7 +455,7 @@ class Target:
         assert client is not None
 
         if client.state is BindingState.bound:
-            return  # nothing to do
+            return [] # nothing to do
 
         if client.state is not BindingState.active:
             raise BindingError(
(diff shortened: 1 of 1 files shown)


Emantor (approved):

jluebbe:
I find it useful to see when a resource was deactivated, as normally you don't try to access unavailable resources. So this PR would leave the user with no information about forced state changes. Why do you think this information is only needed during development?

JoshuaWatt:
> I find it useful to see when a resource was deactivated, as normally you don't try to access unavailable resources. So this PR would leave the user with no information about forced state changes. Why do you think this information is only needed during development?

Per discussion on IRC, you'll get an error when you attempt to use the deactivated driver and it was thought this didn't need to be so chatty. More specifically, I have a device with a serial bootloader that comes and goes and this log message really clutters up stdout. If there's a way to make sure it's still logged somewhere else, that's fine with me.

jluebbe:
> > I find it useful to see when a resource was deactivated, as normally you don't try to access unavailable resources. So this PR would leave the user with no information about forced state changes. Why do you think this information is only needed during development?
> 
> Per discussion on IRC, you'll get an error when you attempt to use the deactivated driver and it was thought this didn't need to be so chatty. More specifically, I have a device with a serial bootloader that comes and goes and this log message really clutters up stdout. If there's a way to make sure it's still logged somewhere else, that's fine with me.

If you deactivate the resource before it disappears (perhaps before turning off power), you should not get this message. Would that help in your case?

JoshuaWatt:
> > > I find it useful to see when a resource was deactivated, as normally you don't try to access unavailable resources. So this PR would leave the user with no information about forced state changes. Why do you think this information is only needed during development?
> > 
> > 
> > Per discussion on IRC, you'll get an error when you attempt to use the deactivated driver and it was thought this didn't need to be so chatty. More specifically, I have a device with a serial bootloader that comes and goes and this log message really clutters up stdout. If there's a way to make sure it's still logged somewhere else, that's fine with me.
> 
> If you deactivate the resource before it disappears (perhaps before turning off power), you should not get this message. Would that help in your case?

I tried deactivating the driver, and that didn't remove the message. Having to poke into the driver bindings to deactivate the resource it's bound to feels messy to me?

jluebbe:
> I tried deactivating the driver, and that didn't remove the message. Having to poke into the driver bindings to deactivate the resource it's bound to feels messy to me?

Ah, you're right. Usually we only have direct access to the drivers in the strategy.

Perhaps it would be better to set this message to debug level and add a new message in Target.deactivate() while it's recursively deactivating drivers? Or would that still be too verbose to you? I fear that debugging issues with unexpectedly disappearing (USB) devices is almost impossible without some explicit notification.

JoshuaWatt:
> > I tried deactivating the driver, and that didn't remove the message. Having to poke into the driver bindings to deactivate the resource it's bound to feels messy to me?
> 
> Ah, you're right. Usually we only have direct access to the drivers in the strategy.
> 
> Perhaps it would be better to set this message to debug level and add a new message in Target.deactivate() while it's recursively deactivating drivers? Or would that still be too verbose to you? I fear that debugging issues with unexpectedly disappearing (USB) devices is almost impossible without some explicit notification.

@jluebbe Is this sort of what you had in mind?

Emantor (approved):

PR #817, opened 2021-08-23
strategy: wait for systemd operational state in shell states

Bastian-Krause:
**Description**
The system might not be in operational state once the login prompt is shown. Try to wait until the system is up and running for systemd-based systems. Use run() instead of run_check() for non-systemd based systems or if degraded/maintenance/stopping system states are reached.

This covers wider use-cases for the shipped strategies.

**Checklist**
- [ ] Tests for the feature 
- [.] PR has been tested

### DIFF
--- a/labgrid/strategy/bareboxstrategy.py
+++ b/labgrid/strategy/bareboxstrategy.py
@@ -56,6 +56,7 @@ class BareboxStrategy(Strategy):
             self.barebox.boot("")
             self.barebox.await_boot()
             self.target.activate(self.shell)
+            self.shell.run("systemctl is-system-running --wait")
         else:
             raise StrategyError(
                 "no transition found from {} to {}".
--- a/labgrid/strategy/shellstrategy.py
+++ b/labgrid/strategy/shellstrategy.py
@@ -46,6 +46,7 @@ class ShellStrategy(Strategy):
             self.target.activate(self.console)
             self.power.cycle()
             self.target.activate(self.shell)
+            self.shell.run("systemctl is-system-running --wait")
         else:
             raise StrategyError(
                 "no transition found from {} to {}".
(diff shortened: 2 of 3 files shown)


jluebbe (approved):

JoshuaWatt:
I suspect this will cause us problems for our testing. Could you run the command in you shell fixture instead:

```python
def shell_command(target, strategy):
    strategy.transition('shell')
    shell = target.get_active_driver("ShellDriver")
    shell.run('systemctl is-system-running --wait')
    return shell
```

Bastian-Krause:
@JoshuaWatt Which kind of problems do you suspect? Note that `systemctl` is neither required to exist nor to succeed. Running this in the shell fixture seems surprising to me. I would describe the shell state as "having shell access to an operational system", so having this in the strategy makes more sense to me.

Bastian-Krause:
@JoshuaWatt Please tell me how this PR breaks something at your end. I can't see how. This will probably get merged otherwise..

JoshuaWatt:
I don't know for 100% sure it will break something, but I don't like a policy that says that every user of one of these strategies *must* wait for systemd to be in an operational state; it seems like that isn't a policy that should be forced on everyone when there are several alternatives that can be easily implemented.

Emantor:
> I don't know for 100% sure it will break something, but I don't like a policy that says that every user of one of these strategies _must_ wait for systemd to be in an operational state; it seems like that isn't a policy that should be forced on everyone when there are several alternatives that can be easily implemented.

Keep in mind that these strategies are simple examples in most cases and users are still free to implement their own handling in their own strategies. I'd also advise to _NOT_ inherit any of the example strategies. While it may look easy enough at first, we can't guarantee consistent behaviour for these since we do adjust them and complex device strategies often result in situations where you skip between files and the strategy behaviour is hard to figure out.

I'm in favor of adding the wait.

PR #818, opened 2021-08-26
remote: client: bootstrap: pass args to load function

bitshifter1:
**Description**
We would like to be able to pass multiple arguments to our bootstrap driver. This allows us to configure it without having to modify the config, e.g. bootstrapping different kind of images that require different logic.

**Checklist**
- [ ] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/labgrid/driver/flashromdriver.py
+++ b/labgrid/driver/flashromdriver.py
@@ -53,7 +53,7 @@ class FlashromDriver(Driver, BootstrapProtocol):
 
     @Driver.check_active
     @step(args=['filename'])
-    def load(self, filename=None):
+    def load(self, filename=None, **args):
         if filename is None and self.image is not None:
             filename = self.target.env.config.get_image_path(self.image)
         filename = os.path.abspath(filename)
--- a/labgrid/driver/openocddriver.py
+++ b/labgrid/driver/openocddriver.py
@@ -104,7 +104,7 @@ class OpenOCDDriver(Driver, BootstrapProtocol):
 
     @Driver.check_active
     @step(args=['filename'])
-    def load(self, filename=None):
+    def load(self, filename=None, **args):
         if filename is None and self.image is not None:
             filename = self.target.env.config.get_image_path(self.image)
         mf = ManagedFile(filename, self.interface)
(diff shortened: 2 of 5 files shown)


Emantor:
I assume your bootstrap driver is out of tree? Can you find an in-tree usage for this PR?

bitshifter1:
> I assume your bootstrap driver is out of tree? Can you find an in-tree usage for this PR?

Hi @Emantor , sorry for my late responds. Yes we have an off-tree driver. I am not exactly sure what you mean by in-tree usage :sweat_smile: The OpenOCDDriver already makes use of the generic **args arguments, but it does it by passing them through the constructor. It would be nice that also the of-tree bootstrapdrivers can make use of the self.args.bootstrap_args arguments passed to the client. Our client call looks like this.

`labgrid-client bootstrap barebox.img type=barebox device=flash`

bitshifter1:
> I assume your bootstrap driver is out of tree? Can you find an in-tree usage for this PR?

Hi @Emantor , sorry for my late responds. Yes we have an off-tree driver. I am not exactly sure what you mean by in-tree usage :sweat_smile: The OpenOCDDriver already makes use of the generic **args arguments, but it does it by passing them through the constructor. It would be nice that also the of-tree bootstrapdrivers can make use of the self.args.bootstrap_args arguments passed to the client. Our client call looks like this.

`labgrid-client bootstrap barebox.img type=barebox device=flash`

Emantor:
My real intention was to nudge you to submit your own bootstrapdriver to upstream, improving labgrid support and your own driver through the review :grin:

bitshifter1:
> My real intention was to nudge you to submit your own bootstrapdriver to upstream, improving labgrid support and your own driver through the review grin

Our bootstrap driver is very specific and wouldn't serve much purpose for anyone else 😊

jluebbe:
Merging this would mean that we can never add more keyword arguments to the `load()` function without potentially breaking a use-case we don't know about. Also, passing `filename=foo` would overwrite the normal filename argument which is at least confusing.

The benefit of passing them via the constructor (like with OpenOCDDriver) is that you have an interface specific to that driver in the environment yaml. The labgrid-client commands are more generic.

@bitshifter1 Your example looks like you're writing an image to flash? Please note that the Bootstrap protocol is intended for loading and executing a bootloader in RAM, not for writing it to some storage.

bitshifter1:
> The benefit of passing them via the constructor (like with OpenOCDDriver) is that you have an interface specific to that driver in the environment yaml. The labgrid-client commands are more generic.

@jluebbe yes makes sense, no problem. Thanks anyway!!!

PR #820, opened 2021-08-28
remote/client: add write-image-like write-file command

a3f:
**Description**

- what do you use the feature for?

Interactive use with `labgrid-client`. It's useful for bootloader development for processors with bootroms that load code from file systems (e.g. EFI system partition, OMAP, AT91, Zynq, Raspberry ... etc.)

- how does labgrid benefit as a testing library from the feature?

I guess a unit test could place e.g. a broken configuration file and see how it behaves?

Remote usbsdmux that I usually use with `write-image`. Tried writing non-existing files and overwriting existing files with `write-file`. Works well.

- if hardware is needed for the feature, which hardware is supported and which
  hardware did you test with?

Anything exposing a `USBStorageDriver`. I tested with a usbsdmux.

**Checklist**
- [-] Tests for the feature 
- [-] Add a section on how to use the feature to doc/usage.rst
- [-] Add a section on how to use the feature to doc/development.rst
- [x] PR has been tested

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -10,7 +10,7 @@ Package: labgrid
 Architecture: any
 Pre-Depends: dpkg (>= 1.16.1), python3, ${misc:Pre-Depends}
 Depends: ${python3:Depends}, ${misc:Depends}, ${shlibs:Depends}
-Recommends: openssh-client, microcom, socat, sshfs, rsync
+Recommends: openssh-client, microcom, socat, sshfs, rsync, pmount
 Description: embedded board control python library
  Labgrid is a embedded board control python library with a focus on testing,
  development and general automation. It includes a remote control layer to
--- a/labgrid/driver/usbstoragedriver.py
+++ b/labgrid/driver/usbstoragedriver.py
@@ -25,6 +25,12 @@ class Mode(enum.Enum):
         return self.value
 
 
+def escape(path):
+    if path.startswith("/dev/"):
+        path = path[len("/dev/"):]
+    return path.replace('/', '_')
+
+
 @target_factory.reg_driver
 @attr.s(eq=False)
 class USBStorageDriver(Driver):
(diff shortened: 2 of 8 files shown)


Emantor:
#548

debian/control
```diff
@@ -10,7 +10,7 @@ Package: labgrid
 Architecture: any
 Pre-Depends: dpkg (>= 1.16.1), python3, ${misc:Pre-Depends}
 Depends: ${python3:Depends}, ${misc:Depends}, ${shlibs:Depends}
-Recommends: openssh-client, microcom, socat, sshfs, rsync
+Recommends: openssh-client, microcom, socat, sshfs, rsync, pmount
```
jluebbe, line 13: Why do you prefer `pmount` over `udiskctl`?

labgrid/driver/usbstoragedriver.py
```diff
@@ -152,6 +160,67 @@ def get_size(self):
...
+        self.await_medium()
+
+        partition = "" if partition is None else partition
+        remote_path = mf.get_remote_path()
+        target = "{}{}".format(self.storage.path, partition)
+
+        label = escape(target)
+        dstpath = os.path.normpath(dstpath)
+        if not dstpath.startswith("/"):
+            raise ValueError("dstpath must be ab absolute path")
```
jluebbe, line 189:
```suggestion
            raise ValueError("dstpath must be an absolute path")
```

labgrid/driver/usbstoragedriver.py
```diff
@@ -152,6 +160,67 @@ def get_size(self):
...
+
+        dstpath = "/media/{}/{}".format(label, dstpath)
```
jluebbe, line 191: Why can we expect `/media/{label}/` to be writable? Shouldn't the mount helper tell us where the filesystem was mounted?

labgrid/driver/usbstoragedriver.py
```diff
@@ -152,6 +160,67 @@ def get_size(self):
...
+
+        args = ["pumount {}".format(target)]
+
+        processwrapper.check_output(
+            self.storage.command_prefix + args,
+            print_on_silent_log=True
+        )
+
+        if caught:
+            raise caught
```
jluebbe, line 222: Please use a more idiomatic try:/finally: structure.

jluebbe (changes requested):

Emantor:
Closing this, please add udisks support instead of pmount and open a new PR.

labgrid/driver/usbstoragedriver.py
a3f: I might've never used finally before, so I had forgotten about it. Thanks! :)

debian/control
a3f: Will resubmit with udisksctl

labgrid/driver/usbstoragedriver.py
a3f: udisksctl does, albeit a bit more verbose than necessary. Will parse it.

PR #822, opened 2021-08-30
resource/udev: handle multiple resources on single device correctly

Bastian-Krause:
**Description**
Imagine multiple USBResources match the same device, e.g.:

```
{% for idx, sysfs in [
  (1,   'pci-0000:00:14.0-usb-0:2.3'),
  (2,   'pci-0000:00:14.0-usb-0:2.4'),
  (3,   'pci-0000:00:14.0-usb-0:2.1'),
  (4,   'pci-0000:00:14.0-usb-0:2.2')] %}

  usbhub-b-{{idx}}:
    location: 'on my desk'
    USBSerialPort:
      match:
        '@ID_PATH': '{{sysfs}}'

  {% for port in range(1, 5) %}
  usbhub-b-{{idx}}-outlet{{port}}:
    SiSPMPowerPort:
      match:
        'ID_PATH': '{{sysfs}}'
      index: {{port}}
  {% endfor %}
{% endfor %}
```

This describes a 4 port USB hub. On each port USBSerialPort as well as SiSPMPowerPort resources should be exported. Note that the 4 power sockets of the _SiS-PM_ power outlet are exported separately.

This works if the SiS-PM is plugged before the exporter is started. In case it is then unplugged, not all resources are marked as unavailable. If it is replugged again, not all resources are marked as available.

The reason for this behavior is the premature break in `UdevManager.on_resource_added()` and `UdevManager.poll()` once the first matching device is found. All remaining resources are
ignored. This breaks the concept of allowing multiple resources on a single device.

Fix that by dropping the premature break, so all resources are tried to match against the device.

There is no need to log each and every resource while iterating over them. The `try_match()` method logs the matched USBResource already, so log only whether matching was successful.
Also log the matched device in `on_resource_added()` since we do not log that here before.

**Note:** I am unsure if this breaks assumptions made elsewhere or if the matching all resources impacts the udev processing in negative ways.

**Checklist**
- [x] PR has been tested

Fixes fa5ed01 ("resource: implement ManagedResources and a USBSerialPort")

### DIFF
--- a/labgrid/resource/udev.py
+++ b/labgrid/resource/udev.py
@@ -32,7 +32,7 @@ class UdevManager(ResourceManager):
         devices.match_subsystem(resource.match['SUBSYSTEM'])
         for device in devices:
             if resource.try_match(device):
-                break
+                self.log.debug(" matched successfully against %s", resource.device)
 
     def _insert_into_queue(self, device):
         self.queue.put(device)
@@ -46,9 +46,8 @@ class UdevManager(ResourceManager):
                 break
             self.log.debug("%s: %s", device.action, device)
             for resource in self.resources:
-                self.log.debug(" %s", resource)
                 if resource.try_match(device):
-                    break
+                    self.log.debug(" matched successfully")
 
 @attr.s(eq=False)
 class USBResource(ManagedResource):


labgrid/resource/udev.py
```diff
@@ -47,8 +46,7 @@ def poll(self):
             self.log.debug("%s: %s", device.action, device)
             for resource in self.resources:
                 self.log.debug(" %s", resource)
-                if resource.try_match(device):
-                    break
+                resource.try_match(device)
```
Emantor, line 49: `try_match()` can be simplified by turning `return False` into a `return`.

Emantor (changes requested):

labgrid/resource/udev.py
Bastian-Krause: I guess we can also replace the whole content of the`if self.device is None:` code block with a `return`, right?

Bastian-Krause: On second thought: do we really want to change `try_match()` at all? Maybe we want replace the break by a log message. If a match was found could still be interesting there.

Emantor: Yes, sounds good to me, please do so.

Bastian-Krause: Done, please take another look.

labgrid/resource/udev.py
```diff
@@ -32,7 +32,7 @@ def on_resource_added(self, resource):
         devices.match_subsystem(resource.match['SUBSYSTEM'])
         for device in devices:
             if resource.try_match(device):
-                break
+                self.log.debug(" matched successfully against %s", resource.device)
```
Emantor, line 35: Leading space?

labgrid/resource/udev.py
```diff
@@ -46,9 +46,8 @@ def poll(self):
                 break
             self.log.debug("%s: %s", device.action, device)
             for resource in self.resources:
-                self.log.debug(" %s", resource)
                 if resource.try_match(device):
-                    break
+                    self.log.debug(" matched successfully")
```
Emantor, line 50: same as above.

Emantor (changes requested):

labgrid/resource/udev.py
Bastian-Krause: This is an indentation. The intention here is that the log message "belongs" to the log message in `try_match()`which is indented, too.

labgrid/resource/udev.py
Bastian-Krause: Same as above, the log message extends the device log message a few lines above, on the same level as the message in `try_match()`.

Bastian-Krause:
Force-pushed:
- fixed yaml formatting in the commit message

Emantor (approved):

Issue #748, opened 2021-03-10
Place cannot be acquired because one of its resources is acquired by itself


egrigore:
Hi,

From time to time I meet this issue: that I can't acquire a place because one of its resources is already acquired by the place itself. I don't know in what situation it happens. Has anyone met the same issue? 

Below is the error message I get and how the place resources look like (Matching resource '2'  is acquired by the place):
```
(venv) testlabgrid@client-hostname:~/labgrid-tests/labgrid-config$ labgrid-client -p RO--MX8_evk--1 acquire
Matching resource '2' (exporter-hostname/RO--MX8_evk--1-console-2/NetworkSerialPort/USBSerialPort) already acquired by place 'RO--MX8_evk--1'
labgrid-client: error: failed to acquire place RO--MX8_evk--1
(venv) testlabgrid@client-hostname:~/labgrid-tests/labgrid-config$ labgrid-client -p RO--MX8_evk--1 show
Place 'RO--MX8_evk--1':
  matches:
    exporter-hostname/RO--MX8_evk--1/*
    exporter-hostname/RO--MX8_evk--1-console-0/NetworkSerialPort -> 0
    exporter-hostname/RO--MX8_evk--1-console-1/NetworkSerialPort -> 1
    exporter-hostname/RO--MX8_evk--1-console-2/NetworkSerialPort -> 2
    exporter-hostname/RO--MX8_evk--1-console-3/NetworkSerialPort -> 3
  acquired: None
  acquired resources:
  created: 2021-02-03 14:50:36.272348
  changed: 2021-03-10 00:18:44.230550
Matching resource 'NetworkPowerPort' (exporter-hostname/RO--MX8_evk--1/NetworkPowerPort/NetworkPowerPort):
  {'acquired': None,
   'avail': True,
   'cls': 'NetworkPowerPort',
   'params': {'host': '<pdu-service-IP>:8085/192.168.99.104',
              'index': 7,
              'model': 'simplehttp'}}
Matching resource 'NetworkService' (exporter-hostname/RO--MX8_evk--1/NetworkService/NetworkService):
  {'acquired': None,
   'avail': True,
   'cls': 'NetworkService',
   'params': {'address': '<pdu-service-IP>',
              'extra': {'proxy': 'exporter-hostname', 'proxy_required': False},
              'username': 'bamboo'}}
Matching resource 'USBResetPort' (exporter-hostname/RO--MX8_evk--1/NetworkUSBResetPort/USBResetPort):
[...]

PR #832, opened 2021-09-15
remote/client: allow already locked resources

Emantor:
**Description**

Allow a place to still be locked if one of the resources is acquired but
was previously acquired by the same place.

Not really sure how to test this.

**Checklist**
- [ ] PR has been tested

Fixes https://github.com/labgrid-project/labgrid/issues/748

### DIFF
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -615,6 +615,8 @@ class ClientSession(ApplicationSession):
                     name = resource_name
                     if match.rename:
                         name = match.rename
+                    if resource.acquired == place.name:
+                        continue
                     print("Matching resource '{}' ({}/{}/{}/{}) already acquired by place '{}'"
                           .format(name, exporter, group_name, resource.cls, resource_name,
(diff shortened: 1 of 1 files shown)


jluebbe:
We can't just ignore the inconsistent state here. In the situation reported in #748, this would only allow using this resource by the same place, but if the resource is now configured to match for a different place, it cannot be acquired.
I suspect that the issue already exists before the failing acquire. @egrigore: Do you see any other tracebacks on the exporter or coordinator correlated to the issue?

jluebbe (changes requested):
This should not be merged as is, see my comment above.

Emantor:
> This should not be merged as is, see my comment above.

I agree, closing.

egrigore:
@jluebbe  sorry for the late response . We have seen this problem again today but there is nothing special related to it on exporter or coordinator journalctl . There aren't clear steps to isolate the problem and also it doesn't happen only on a specific board to relate it to that.

PR #835, opened 2021-09-16
Add support for running process asynchronously

JoshuaWatt:
**Description**
Adds support for running processes asynchronously with the main test code. Attempts to incorporate the feed back from #552 

**Checklist**
- [x] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -592,6 +592,8 @@ New Features in 0.4.0
   a jumper.
 - AndroidFastbootDriver now supports booting/flashing images preconfigured in
   the environment configuration.
+- CommandProtocol now supports a start_process() method to run a process on the
+  target asynchronously.
 
 Bug fixes in 0.4.0
 ~~~~~~~~~~~~~~~~~~
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,24 +1,100 @@
-Release 25.1 (Unreleased)
--------------------------
+Release 26.0 (Released Jun 06, 2026)
+------------------------------------
+Sponsored by: Analog Devices GmbH
 
-New Features in 25.1
+New Features in 26.0
 ~~~~~~~~~~~~~~~~~~~~
 
 - The default interrupt key for the `BareboxDriver` is now CTRL+D (``\x04``)
   instead of a line feed (``\n``). barebox v2025.03.0 onwards handles
   CTRL+D specially to halt autoboot countdown without running interactive
   hooks like bringing up network interfaces automatically.
+- Support for the tinycontrol tcPDU has been added as a power backend.
+- Support for shelly gen2+ devices has been added as a power backend.
+- Support for Waveshare Modbus Relays was added.
+- The upload directory for labgrid agent python files can be set via
+  the ``LG_AGENT_PREFIX`` environment variable.
+- The ManagedFile now returns an error when trying to access the underlying
+  file path before the file was synced.
+- The RKUSBLoader match was extended with all currently known USB IDs.
+- The error message for trying to acquire a place that is locked by another
+  user was improved.
+- The ``Strategy`` class has gained a new ``never_retry`` decorator that
+  stores the first exception and raises it for every subsequent call.
+  The pytestplugin also correctly respects that StrategyError.
+- IMXUSBDriver has gained a ``verify`` property that is enabled by default, but
+  can disable the verification step for SoCs that don't support it.
+- The codebase gained a lot of fixes for python 3.14 support.
+- ``labgrid-client`` with a multi-target environment configuration now 
+  correctly tries to acquire/release all targets in the environment
+  configuration.
+- IMXUSBLoader now correctly recognizes i.MX95 & i.MX91.
+- Exporter configuration files can now include drop-in files from .d
+  directories by using jinja templates.
+- Labgrid has gained support for new USB SDWire 3 devices.
+- The BareboxDriver has gained support for the ``boot_command``, similar to
+  the UBootDriver.
+- Labgrid has gained support for Aten PE6216 PDUs.
 - The `QEMUDriver` now supports a ``netdev`` argument which can be added to
   ``add_port_forward()`` and ``remove_port_forward()``  in case there is more
   than one network interface defined.
 - Guermok HDMI to USB 3.0 capture dongle supported
+- The UBootDriver can now optionally strip timestamps.
+- Labgrid has partially switched to enforce ruff formatting for some files.
+- Labgrid has gained support for the Linaro Automation Appliance (LAA).
+- The MQTTManager now supports authentication.
+- The `consoleexpectmixin` now supports chunking of multiple bytes
+  with the ``txchunk`` attribute.
+- Labgrid has gained support for network namespace forwarding from exporters
+  to clients. This allows accessing a network as if it is locally connected.
+- HTTPVideoDrivers now optionally allow the specification of a port.
 
-Breaking changes in 25.1
+Breaking changes in 26.0
 ~~~~~~~~~~~~~~~~~~~~~~~~
 - The deprecated pytest plugin option ``--env-config`` has been removed. Use
   ``--lg-env`` instead.
 - The deprecated ``NetworkUSBStorageDriver`` has been removed. Use the
   `USBStorageDriver` instead.
+
+Bug fixes in 26.0
+~~~~~~~~~~~~~~~~~
+- Console activation in the ShellDriver now correctly disables kernel messages
+  on that console.
+- The ShellDriver now correctly handles an existing /tmp/labgrid-ssh directory.
+- An error in the mfi_mpower driver was fixed to correctly use f-strings.
+- An error message in the coordinator was fixed to correctly emit resource and
+  reason.
+- The ShellDriver public key writing has been fixed to write the keys in chunks.
+  This fixes an issue where the number of allowed characters on the command line
+  is too small to contain the full public key.
+- The sync-places contrib script has been fixed to filter tags on creation.
+- Several labgrid client subcommands have been fixed to match the resource name
+  as well.
+- Ordering of ResourceMatches has been fixed.
+- Locking problems in the coordinator with regards to reservations have been
+  fixed.
+
+Release 25.0.1 (Released Jul 07, 2025)
+--------------------------------------
+
+Bug fixes in 25.0.1
+~~~~~~~~~~~~~~~~~~~
+- Fix losing exceptions in `SSHDriver` control master start while handling
+  timeouts in some cases.
+- Make pytestplugin only set a color for the custom console log level if
+  formatter class supports it.
+- Add missing `MatchedSysfsGPIO` binding to `GpioDigitalOutputDriver`.
+- Fix a couple of unevaluated variables in exception and log messages.
+- Use only non-deprecated directives in Dockerfile.
+- Always use latest stable docker compose release in CI.
+- Create symlinks in `ManagedFile` even when skipping sync for local providers.
+- Use ``importlib`` instead of ``pkg_resources`` to retrieve version, because
+  setuptools (providing ``pkg_resources``) is no longer pre-installed in
+  virtualenvs created with the ``venv`` module.
+- Allow passing non-strings to `StepFormatter`.
+- Properly escape file and directory names in the `USBStorageDriver`.
+- Fix ``labgrid-client`` named resource handling for the ``io``, ``bootstrap``,
+  ``sd-mux`` and ``video`` sub commands
 
 Release 25.0 (Released May 7, 2025)
 -----------------------------------
--- a/labgrid/driver/bareboxdriver.py
+++ b/labgrid/driver/bareboxdriver.py
@@ -1,12 +1,14 @@
 import shlex
+from contextlib import contextmanager
 
 import attr
 from pexpect import TIMEOUT
 
+from ..exceptions import CommandProcessBusy
 from ..factory import target_factory
 from ..protocol import CommandProtocol, ConsoleProtocol, LinuxBootProtocol
 from ..step import step
-from ..util import gen_marker, Timeout, re_vt100
+from ..util import gen_marker, Timeout, re_vt100, ConsoleMarkerProcess
 from .common import Driver
 from .commandmixin import CommandMixin
 
@@ -44,6 +46,7 @@ class BareboxDriver(CommandMixin, Driver, CommandProtocol, LinuxBootProtocol):
         self._status = 0
         # barebox' default log level, used as fallback if no log level can be saved
         self.saved_log_level = 7
+        self._process = None
 
     def on_activate(self):
         """Activate the BareboxDriver
@@ -59,6 +62,34 @@ class BareboxDriver(CommandMixin, Driver, CommandProtocol, LinuxBootProtocol):
         Simply sets the internal status to 0
         """
         self._status = 0
+        assert not self._process, "Deactivating while a command process is running is not allowed"
+
+    @contextmanager
+    def _start_process(self, cmd: str, *, adjust_log_level: bool = True):
+        if self._process is not None:
+            raise CommandProcessBusy()
+
+        # FIXME: use codec, decodeerrors
+        marker = gen_marker()
+        # hide marker from expect
+        hidden_marker = f'"{marker[:4]}""{marker[4:]}"'
+        # generate command with marker and log level adjustment
+        cmp_command = f'echo -o /cmd {shlex.quote(cmd)}; echo {hidden_marker};'
+        if self.saved_log_level and adjust_log_level:
+            cmp_command += f' global.loglevel={self.saved_log_level};'
+        cmp_command += f' sh /cmd; echo {hidden_marker} $?;'
+        if self.saved_log_level and adjust_log_level:
+            cmp_command += ' global.loglevel=0;'
+
+        self.console.sendline(cmp_command)
+        self.console.expect(marker)
+
+        with ConsoleMarkerProcess(self.console, marker, self.prompt) as p:
+            self._process = p
+            try:
+                yield p
+            finally:
+                self._process = None
 
     @Driver.check_active
     @step(args=['cmd'])
@@ -76,31 +107,22 @@ class BareboxDriver(CommandMixin, Driver, CommandProtocol, LinuxBootProtocol):
         Returns:
             Tuple[List[str],List[str], int]: if successful, None otherwise
         """
-        # FIXME: use codec, decodeerrors
-        marker = gen_marker()
-        # hide marker from expect
-        hidden_marker = f'"{marker[:4]}""{marker[4:]}"'
-        # generate command with marker and log level adjustment
-        cmp_command = f'echo -o /cmd {shlex.quote(cmd)}; echo {hidden_marker};'
-        if self.saved_log_level and adjust_log_level:
-            cmp_command += f' global.loglevel={self.saved_log_level};'
-        cmp_command += f' sh /cmd; echo {hidden_marker} $?;'
-        if self.saved_log_level and adjust_log_level:
-            cmp_command += ' global.loglevel=0;'
-
         if self._status == 1:
-            self.console.sendline(cmp_command)
-            _, _, match, _ = self.console.expect(
-                rf'{marker}(.*){marker}\s+(\d+)\s+.*{self.prompt}',
-                timeout=timeout)
-            # Remove VT100 Codes and split by newline
-            data = re_vt100.sub('', match.group(1).decode('utf-8')).split('\r\n')[1:-1]
-            self.logger.debug("Received Data: %s", data)
-            # Get exit code
-            exitcode = int(match.group(2))
-            return (data, [], exitcode)
+            with self._start_process(cmd, adjust_log_level=adjust_log_level) as p:
+                output = p.read_to_end(timeout=timeout)
+                # Remove VT100 Codes and split by newline
+                data = re_vt100.sub('', output.decode('utf-8')).split('\r\n')[1:-1]
+                self.logger.debug("Received Data: %s", data)
+                return (data, [], p.exitcode)
 
         return None
+
+    @Driver.check_active
+    @step(args=['cmd'])
+    @contextmanager
+    def start_process(self, cmd: str):
+        with self._start_process(cmd) as p:
+            yield p
 
     @Driver.check_active
     @step()
--- a/labgrid/driver/shelldriver.py
+++ b/labgrid/driver/shelldriver.py
@@ -5,15 +5,17 @@ import io
 import re
 import shlex
 import ipaddress
+from contextlib import contextmanager
 
 import attr
 from pexpect import TIMEOUT
 import xmodem
 
+from ..exceptions import CommandProcessBusy
 from ..factory import target_factory
 from ..protocol import CommandProtocol, ConsoleProtocol, FileTransferProtocol
 from ..step import step
-from ..util import gen_marker, Timeout, re_vt100
+from ..util import gen_marker, Timeout, re_vt100, ConsoleMarkerProcess
 from .commandmixin import CommandMixin
 from .common import Driver
 from .exception import ExecutionError
@@ -61,6 +63,7 @@ class ShellDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
 
         self._xmodem_cached_rx_cmd = ""
         self._xmodem_cached_sx_cmd = ""
+        self._process = None
 
     def on_activate(self):
         if self._status == 0:
@@ -76,6 +79,27 @@ class ShellDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
 
     def on_deactivate(self):
         self._status = 0
+        assert not self._process, "Deactivating while a command process is running is not allowed"
+
+    @contextmanager
+    def _start_process(self, cmd: str):
+        if self._process is not None:
+            raise CommandProcessBusy()
+
+        # FIXME: Handle pexpect Timeout
+        self._check_prompt()
+        marker = gen_marker()
+
+        # hide marker from expect
+        cmp_command = f'''MARKER='{marker[:4]}''{marker[4:]}' run {shlex.quote(cmd)}'''
+        self.console.sendline(cmp_command)
+        self.console.expect(marker)
+        with ConsoleMarkerProcess(self.console, marker, self.prompt) as p:
+            self._process = p
+            try:
+                yield p
+            finally:
+                self._process = None
 
     def _run(self, cmd, *, timeout=30.0, codec="utf-8", decodeerrors="strict"):
         """
@@ -84,29 +108,28 @@ class ShellDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
         Arguments:
         cmd - cmd to run on the shell
         """
-        # FIXME: Handle pexpect Timeout
-        self._check_prompt()
-        marker = gen_marker()
-        # hide marker from expect
-        cmp_command = f'''MARKER='{marker[:4]}''{marker[4:]}' run {shlex.quote(cmd)}'''
-        self.console.sendline(cmp_command)
-        _, _, match, _ = self.console.expect(
-            rf'{marker}(.*){marker}\s+(\d+)\s+{self.prompt}',
-            timeout=timeout
-        )
-        # Remove VT100 Codes, split by newline and remove surrounding newline
-        data = re_vt100.sub('', match.group(1).decode(codec, decodeerrors)).split('\r\n')
-        if data and not data[-1]:
-            del data[-1]
-        self.logger.debug("Received Data: %s", data)
-        # Get exit code
-        exitcode = int(match.group(2))
-        return (data, [], exitcode)
+        with self._start_process(cmd) as p:
+            output = p.read_to_end(timeout=timeout)
+
+            # Remove VT100 Codes, split by newline and remove surrounding newline
+            data = re_vt100.sub('', output.decode(codec, decodeerrors)).split('\r\n')
+            if data and not data[-1]:
+                del data[-1]
+
+            self.logger.debug("Received Data: %s", data)
+            return (data, [], p.exitcode)
 
     @Driver.check_active
     @step(args=['cmd'], result=True)
     def run(self, cmd, timeout=30.0, codec="utf-8", decodeerrors="strict"):
         return self._run(cmd, timeout=timeout, codec=codec, decodeerrors=decodeerrors)
+
+    @Driver.check_active
+    @step(args=['cmd'])
+    @contextmanager
+    def start_process(self, cmd: str):
+        with self._start_process(cmd) as p:
+            yield p
 
     @step()
     def _await_login(self):
--- a/labgrid/driver/sshdriver.py
+++ b/labgrid/driver/sshdriver.py
@@ -9,11 +9,13 @@ import subprocess
 import tempfile
 import time
 from functools import cached_property
+from contextlib import contextmanager
 
 import attr
+import pexpect
 
 from ..factory import target_factory
-from ..protocol import CommandProtocol, FileTransferProtocol
+from ..protocol import CommandProtocol, CommandProcessProtocol, FileTransferProtocol
 from .commandmixin import CommandMixin
 from .common import Driver
 from ..step import step
@@ -22,6 +24,56 @@ from ..util.helper import get_free_port
 from ..util.proxy import proxymanager
 from ..util.timeout import Timeout
 from ..util.ssh import get_ssh_connect_timeout
+from ..util.readmixin import ReadMixIn
+
+
+class SSHDriverProcess(CommandProcessProtocol, ReadMixIn):
+    def __init__(self, sub):
+        self._sub = sub
+
+    @property
+    def exitcode(self):
+        if self._sub.isalive():
+            return None
+
+        if self._sub.exitstatus is None:
+            return -self._sub.signalstatus
+        return self._sub.exitstatus
+
+    def read(self, size=1, timeout=-1):
+        return self._sub.read_nonblocking(size, timeout)
+
+    @step(args=['data'])
+    def write(self, data):
+        self._sub.write(data)
+
+    @step(result=True)
+    def poll(self):
+        return self.exitcode
+
+    @step(result=True)
+    def stop(self):
+        self._sub.close(True)
+
+    @step(args=['pattern', 'timeout'], result=True)
+    def expect(self, pattern, *, timeout=-1):
+        index = self._sub.expect(pattern, timeout=timeout)
+        return index, self._sub.before, self._sub.match, self._sub.after
+
+    @step(result=True)
+    def wait(self):
+        return self._sub.wait()
+
+    @step(args=['char'])
+    def sendcontrol(self, char):
+        self._sub.sendcontrol(char)
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, typ, value, traceback):
+        self.stop()
+        return False
 
 
 @target_factory.reg_driver
@@ -45,6 +97,7 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
         self._scp = self._get_tool("scp")
         self._sshfs = self._get_tool("sshfs")
         self._rsync = self._get_tool("rsync")
+        self._processes = []
 
     def _get_tool(self, name):
         if self.target.env:
@@ -78,6 +131,7 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
         self._start_keepalive()
 
     def on_deactivate(self):
+        assert not self._processes, "Deactivating while a command process is running is not allowed"
         try:
             self._stop_keepalive()
         finally:
@@ -239,6 +293,33 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
             stderr.pop()
         return (stdout, stderr, sub.returncode)
 
+    @Driver.check_active
+    @step(args=['cmd'])
+    @contextmanager
+    def start_process(self, cmd: str):
+        if not self._check_keepalive():
+            raise ExecutionError("Keepalive no longer running")
+
+        complete_cmd = ["ssh", "-o", "LogLevel=QUIET", "-x", *self.ssh_prefix,
+                        "-p", str(self.networkservice.port), "-l", self.networkservice.username,
+                        self.networkservice.address, "-T", "--", '/bin/sh -c {}'.format(shlex.quote(cmd)),
+                        ]
+        self.logger.debug("Sending command: %s", complete_cmd)
+
+        try:
+            sub = pexpect.spawn(complete_cmd[0], complete_cmd[1:])
+        except:
+            raise ExecutionError(
+                "error executing command: {}".format(complete_cmd)
+            )
+
+        with SSHDriverProcess(sub) as p:
+            self._processes.append(p)
+            try:
+                yield p
+            finally:
+                self._processes.remove(p)
+
     def interact(self, cmd=None):
         assert cmd is None or isinstance(cmd, list)
 
@@ -374,7 +455,7 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
                 "-o", f"ControlPath={self.control.replace('%', '%%')}",
                 src, dst,
         ]
-        
+
         if self.explicit_sftp_mode and self._scp_supports_explicit_sftp_mode():
             complete_cmd.insert(1, "-s")
         if self.explicit_scp_mode and self._scp_supports_explicit_scp_mode():
--- a/labgrid/driver/sshdriver.py
+++ b/labgrid/driver/sshdriver.py
@@ -300,6 +300,7 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
         if not self._check_keepalive():
             raise ExecutionError("Keepalive no longer running")
 
+        cmd = f"stty -echo; {cmd}" # Disable input echo from ssh
         complete_cmd = ["ssh", "-o", "LogLevel=QUIET", "-x", *self.ssh_prefix,
                         "-p", str(self.networkservice.port), "-l", self.networkservice.username,
                         self.networkservice.address, "-tt", "--", '/bin/sh -c {}'.format(shlex.quote(cmd)),
@@ -308,6 +309,7 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
 
         try:
             sub = pexpect.spawn(complete_cmd[0], complete_cmd[1:])
+            sub.setecho(False)  # Disable input echo from pexpect
         except:
             raise ExecutionError(
                 "error executing command: {}".format(complete_cmd)
--- a/labgrid/driver/sshdriver.py
+++ b/labgrid/driver/sshdriver.py
@@ -302,7 +302,7 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
 
         complete_cmd = ["ssh", "-o", "LogLevel=QUIET", "-x", *self.ssh_prefix,
                         "-p", str(self.networkservice.port), "-l", self.networkservice.username,
-                        self.networkservice.address, "-T", "--", '/bin/sh -c {}'.format(shlex.quote(cmd)),
+                        self.networkservice.address, "-tt", "--", '/bin/sh -c {}'.format(shlex.quote(cmd)),
                         ]
         self.logger.debug("Sending command: %s", complete_cmd)
 
--- a/labgrid/driver/sshdriver.py
+++ b/labgrid/driver/sshdriver.py
@@ -88,7 +88,7 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
     explicit_sftp_mode = attr.ib(default=False, validator=attr.validators.instance_of(bool))
     explicit_scp_mode = attr.ib(default=False, validator=attr.validators.instance_of(bool))
     username = attr.ib(default="", validator=attr.validators.instance_of(str))
-    password = attr.ib(default="", validator=attr.validators.instance_of(str))
+    password = attr.ib(default=None, validator=attr.validators.optional(attr.validators.instance_of(str)))
 
     def __attrs_post_init__(self):
         super().__attrs_post_init__()
@@ -110,7 +110,9 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
 
     def _get_password(self):
         """Get the password from this class or from NetworkService"""
-        return self.password or self.networkservice.password
+        if self.password is not None:
+            return self.password
+        return self.networkservice.password
 
     def on_activate(self):
         self.ssh_prefix = ["-o", "LogLevel=ERROR"]
@@ -119,7 +121,7 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
             if self.target.env:
                 keyfile_path = self.target.env.config.resolve_path(self.keyfile)
             self.ssh_prefix += ["-i", keyfile_path ]
-        if not self._get_password():
+        if self._get_password() is None:
             self.ssh_prefix += ["-o", "PasswordAuthentication=no"]
 
         self.control = self._start_own_master()
@@ -190,14 +192,15 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
 
         env = os.environ.copy()
         pass_file = ''
-        if self._get_password():
+        password = self._get_password()
+        if password is not None:
             fd, pass_file = tempfile.mkstemp()
             os.fchmod(fd, stat.S_IRWXU)
             #with openssh>=8.4 SSH_ASKPASS_REQUIRE can be used to force SSH_ASK_PASS
             #openssh<8.4 requires the DISPLAY var and a detached process with start_new_session=True
             env = {'SSH_ASKPASS': pass_file, 'DISPLAY':'', 'SSH_ASKPASS_REQUIRE':'force'}
             with open(fd, 'w') as f:
-                f.write("#!/bin/sh\necho " + shlex.quote(self._get_password()))
+                f.write("#!/bin/sh\necho " + shlex.quote(password))
 
         self.process = subprocess.Popen(args, env=env,
                                         stdout=subprocess.PIPE,
@@ -234,7 +237,7 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
                 f"Subprocess timed out [{subprocess_timeout}s] while executing {args}",
             )
         finally:
-            if self._get_password() and os.path.exists(pass_file):
+            if self._get_password() is not None and os.path.exists(pass_file):
                 os.remove(pass_file)
 
         if not os.path.exists(control):
--- a/labgrid/driver/sshdriver.py
+++ b/labgrid/driver/sshdriver.py
@@ -9,8 +9,10 @@ import subprocess
 import tempfile
 import time
 from functools import cached_property
+from contextlib import contextmanager
 
 import attr
+import pexpect
 
 from ..factory import target_factory
 from ..protocol import CommandProtocol, FileTransferProtocol
@@ -22,6 +24,56 @@ from ..util.helper import get_free_port
 from ..util.proxy import proxymanager
 from ..util.timeout import Timeout
 from ..util.ssh import get_ssh_connect_timeout
+from ..util.readmixin import ReadMixIn
+
+
+class SSHDriverProcess(ReadMixIn):
+    def __init__(self, sub):
+        self._sub = sub
+
+    @property
+    def exitcode(self):
+        if self._sub.isalive():
+            return None
+
+        if self._sub.exitstatus is None:
+            return -self._sub.signalstatus
+        return self._sub.exitstatus
+
+    def read(self, size=1, timeout=-1):
+        return self._sub.read_nonblocking(size, timeout)
+
+    @step(args=['data'])
+    def write(self, data):
+        self._sub.write(data)
+
+    @step(result=True)
+    def poll(self):
+        return self.exitcode
+
+    @step(result=True)
+    def stop(self):
+        self._sub.close(True)
+
+    @step(args=['pattern', 'timeout'], result=True)
+    def expect(self, pattern, *, timeout=-1):
+        index = self._sub.expect(pattern, timeout=timeout)
+        return index, self._sub.before, self._sub.match, self._sub.after
+
+    @step(result=True)
+    def wait(self):
+        return self._sub.wait()
+
+    @step(args=['char'])
+    def sendcontrol(self, char):
+        self._sub.sendcontrol(char)
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, typ, value, traceback):
+        self.stop()
+        return False
 
 
 @target_factory.reg_driver
@@ -45,6 +97,7 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
         self._scp = self._get_tool("scp")
         self._sshfs = self._get_tool("sshfs")
         self._rsync = self._get_tool("rsync")
+        self._processes = []
 
     def _get_tool(self, name):
         if self.target.env:
@@ -80,6 +133,7 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
         self._start_keepalive()
 
     def on_deactivate(self):
+        assert not self._processes, "Deactivating while a command process is running is not allowed"
         try:
             self._stop_keepalive()
         finally:
@@ -242,6 +296,40 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
             stderr.pop()
         return (stdout, stderr, sub.returncode)
 
+    @Driver.check_active
+    @step(args=['cmd'])
+    @contextmanager
+    def start_process(self, cmd: str):
+        """Execute ``cmd`` on the target as a process.
+
+        The pseudo-terminal used by SSH merges stdout and stderr into a
+        single output stream.
+        """
+        if not self._check_keepalive():
+            raise ExecutionError("Keepalive no longer running")
+
+        cmd = f"stty -echo; {cmd}"
+        complete_cmd = [self._ssh, "-o", "LogLevel=QUIET", "-x", *self.ssh_prefix,
+                        "-p", str(self.networkservice.port), "-l", self._get_username(),
+                        self.networkservice.address, "-tt", "--", '/bin/sh -c {}'.format(shlex.quote(cmd)),
+                        ]
+        self.logger.debug("Sending command: %s", complete_cmd)
+
+        try:
+            sub = pexpect.spawn(complete_cmd[0], complete_cmd[1:])
+            sub.setecho(False)
+        except:
+            raise ExecutionError(
+                "error executing command: {}".format(complete_cmd)
+            )
+
+        with SSHDriverProcess(sub) as p:
+            self._processes.append(p)
+            try:
+                yield p
+            finally:
+                self._processes.remove(p)
+
     def interact(self, cmd=None):
         assert cmd is None or isinstance(cmd, list)
 
@@ -377,7 +465,7 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
                 "-o", f"ControlPath={self.control.replace('%', '%%')}",
                 src, dst,
         ]
-        
+
         if self.explicit_sftp_mode and self._scp_supports_explicit_sftp_mode():
             complete_cmd.insert(1, "-s")
         if self.explicit_scp_mode and self._scp_supports_explicit_scp_mode():
--- a/labgrid/driver/ubootdriver.py
+++ b/labgrid/driver/ubootdriver.py
@@ -1,11 +1,13 @@
 """The U-Boot Module contains the UBootDriver"""
 import re
+from contextlib import contextmanager
 import attr
 from pexpect import TIMEOUT
 
+from ..exceptions import CommandProcessBusy
 from ..factory import target_factory
 from ..protocol import CommandProtocol, ConsoleProtocol, LinuxBootProtocol
-from ..util import gen_marker, re_vt100, Timeout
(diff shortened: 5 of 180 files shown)


Emantor:
Overall this design looks like what we had in mind in #552. Thank you very much for the implementation, looking forward to testing this on my lab at home.

JoshuaWatt:
It is unfortunately a little hard to write automated tests for most of this since it requires actual hardware interaction, but I'll take a stab at it. I was able to test on actual hardware (Raspberry Pi 4 w/ u-boot) with the following test suite:
```python
import socket
import os
import time

def test_uboot_async(uboot_command):
    with uboot_command.popen("echo hello") as p:
        assert p.read(4).decode("utf-8").splitlines() == ["hell"]
        assert p.read().decode("utf-8").splitlines() == ["o"]

    with uboot_command.popen("echo bzap hello bzap") as p:
        index, before, match, after = p.expect("hello")
        assert index == 0
        assert match.group(0).decode("utf-8") == "hello"
        assert before.decode("utf-8") == "bzap "
        assert after.decode("utf-8") == "hello"
        assert p.read().decode("utf-8").splitlines() == [" bzap"]

def test_ssh_async(ssh_command):
    with ssh_command.popen("echo hello") as p:
        assert p.read(4).decode("utf-8").splitlines() == ["hell"]
        assert p.read().decode("utf-8").splitlines() == ["o"]

    with ssh_command.popen("echo bzap hello bzap") as p:
        index, before, match, after = p.expect("hello")
        assert index == 0
        assert match.group(0).decode("utf-8") == "hello"
        assert before.decode("utf-8") == "bzap "
        assert after.decode("utf-8") == "hello"
        assert p.read().decode("utf-8").splitlines() == [" bzap"]

def test_shell_nc(shell_command, ssh_command):
    with shell_command.popen("nc -l -p 5555") as p:
        # Wait for netcat to start listening
        time.sleep(2)
        with ssh_command.forward_local_port(5555) as local_port:
            with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
                s.connect(("localhost", local_port))
                s.sendall(b"Hello World")
            p.expect(b"Hello World")
            time.sleep(2)

def test_ssh_nc(ssh_command):
    with ssh_command.popen("nc -l -p 5555 127.0.0.1") as p:
        # Wait for netcat to start listening
        time.sleep(2)
        with ssh_command.forward_local_port(5555) as local_port:
            with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
                s.connect(("localhost", local_port))
                s.sendall(b"Hello World")
            p.expect(b"Hello World")
            time.sleep(2)


def test_shell_async(shell_command):
    with shell_command.popen("echo hello") as p:
        assert p.read(4).decode("utf-8").splitlines() == ["hell"]
        assert p.read().decode("utf-8").splitlines() == ["o"]

    with shell_command.popen("echo bzap hello bzap") as p:
        index, before, match, after = p.expect("hello")
        assert index == 0
        assert match.group(0).decode("utf-8") == "hello"
        assert before.decode("utf-8") == "bzap "
        assert after.decode("utf-8") == "hello"
        assert p.read().decode("utf-8").splitlines() == [" bzap"]

def test_shell_interrupt(shell_command):
    with shell_command.popen("yes") as p:
        l = p.read(1024).decode("utf-8").splitlines()
        assert l == ["y"] * len(l)

    shell_command.run_check("true")
```

Note that I wasn't able to test the barebox driver, so manual testing there would be appreciated

JoshuaWatt:
I added some tests, which revealed a few bugs that needed fixed. I think it's all good to go now though (although I still have no way to test barebox)

labgrid/driver/shelldriver.py
```diff
@@ -9,13 +9,13 @@
 import ipaddress
 
 import attr
-from pexpect import TIMEOUT
+from pexpect import TIMEOUT, EOF
```
Bastian-Krause, line 12: `EOF` is unused.

labgrid/driver/sshdriver.py
```diff
@@ -8,18 +8,88 @@
 import subprocess
 import tempfile
 import time
+import pexpect
+import shlex
```
Bastian-Krause, line 12: `shlex` is an import from the Python Standard Library, so move it up (standard imports first, then third party, then local package)

labgrid/protocol/commandprotocol.py
```diff
@@ -1,6 +1,85 @@
...
+        or EOF is encountered.
+
+        If timeout is None, there is no timeout and reading will continue until
+        either size bytes have been read, or EOF occurs. A negative timeout
+        means to use the default timeout for this process
+        """
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def read_nonblocking(size=1, timeout=-1):
```
Bastian-Krause, line 24: If this is a method, it misses the `self` argument.

labgrid/driver/ubootdriver.py
```diff
@@ -4,11 +4,11 @@
 import re
 
 import attr
-from pexpect import TIMEOUT
+from pexpect import TIMEOUT, EOF
```
Bastian-Krause, line 7: `EOF` is unused.

labgrid/util/expect.py
```diff
@@ -33,3 +35,12 @@ def read_nonblocking(self, size=1, timeout=-1):
...
         return self.driver.read(size=size, timeout=timeout)
+
+    def sendcontrol(self, char):
+        char = char.lower()
+        try:
+            ord_ = string.ascii_lowercase.index(char) + 1
+            self.send(bytes([ord_]))
+        except ValueError:
+            raise NotImplementedError("Sending control character {} is not supported yet".format(char))
+
```
Bastian-Krause, line 46: Unnecessary trailing newline.

labgrid/util/marker.py
```diff
@@ -1,6 +1,178 @@
 import random
 import string
+import re
+import time
```
Bastian-Krause, line 4: Unused import.

labgrid/util/marker.py
```diff
@@ -1,6 +1,178 @@
...
+
+from ..protocol import CommandProcessProtocol
+from ..step import step
+from ..util import Timeout, OptionalTimeout
```
Bastian-Krause, line 8: `Timeout` is unused.

labgrid/util/marker.py
```diff
@@ -1,6 +1,178 @@
...
+from pexpect import EOF, TIMEOUT, spawn
```
Bastian-Krause, line 9: This is a third party import. Move it up, so it is above the local package imports (standard imports first, then third party, then local package).

Bastian-Krause (dismissed):
Great work! I found some trivial things. I'll have to dive deeper into this later.

This should be mentioned in `CHANGES.rst` and probably needs some documentation elsewhere?

labgrid/util/marker.py
JoshuaWatt: Done

labgrid/util/marker.py
JoshuaWatt: Done

JoshuaWatt:
> Great work! I found some trivial things. I'll have to dive deeper into this later.
> 
> This should be mentioned in `CHANGES.rst` and probably needs some documentation elsewhere?

I added a little to `CHANGES.rst`; I can add more documentation but I'm not really sure where it should go.

labgrid/protocol/commandprotocol.py
```diff
@@ -1,6 +1,85 @@
...
+    def read_nonblocking(self, size=1, timeout=-1):
+        """
+        This reads at most size characters from the process, or until EOF or a
+        timeout occurs. If no data is read before EOF or a timeout occurs, then
+        pexpect.EOF or pexpect.TIMEOUT are raised respectively.
+
+        A timeout of None means to wait forever for either data or EOF
+
+        Note that the timer is per-character, not for the entire buffer
+        """
```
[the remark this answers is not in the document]
JoshuaWatt: Because I forgot to add it :)

labgrid/util/marker.py
```diff
@@ -1,6 +1,178 @@
...
+
+        self._expect = ConsoleProcessExpect(self, timeout=timeout)
+
+    def _handle_eof(self, match):
+        self.exitcode = int(match.group(1))
+        self._alive = False
+
+    @step(args=['size', 'timeout'], result=True)
+    def read(self, size=-1, *, timeout=None):
+        t = OptionalTimeout(timeout, self._console.timeout)
```
Bastian-Krause, line 83: `QEMUDriver` and `ExternalConsoleDriver` do not implement a timeout attribute.

labgrid/driver/bareboxdriver.py
```diff
@@ -79,25 +91,21 @@ def _run(self, cmd: str, *, timeout: int = 30, codec: str = "utf-8", decodeerror
...
-            data = self.re_vt100.sub('', match.group(1).decode('utf-8')).split('\r\n')[1:-1]
-            self.logger.debug("Received Data: %s", data)
-            # Get exit code
-            exitcode = int(match.group(2))
-            return (data, [], exitcode)
+            with self._popen(cmd) as p:
+                output = p.read(timeout=timeout)
+                # Remove VT100 Codes and split by newline
+                data = self.re_vt100.sub('', output.decode('utf-8')).split('\r\n')[1:-1]
+                self.logger.debug(f"Received Data: {data}")
```
Bastian-Krause, line 99: log functions should not use f-strings, but rather lazy % formatting. There are multiple occurrences of this spread in this PR.

CHANGES.rst
```diff
@@ -20,6 +20,8 @@ New Features in 0.4.0
 - Exporter now checks /usr/sbin/ser2net for SerialPortExport
 - AndroidFastbootDriver now supports booting/flashing images preconfigured in
   the environment configuration.
+- CommandProtocol now supports a popen() method to run a process on the target
+  asynchronously with the pytest code.
```
Bastian-Krause, line 24: Why mention pytest here? The PR does not touch anything pytest-specific.

labgrid/driver/ubootdriver.py
```diff
@@ -69,26 +69,25 @@ def on_deactivate(self):
...
 
+    def _run(self, cmd: str, *, timeout: int = 30, codec: str = "utf-8", decodeerrors: str = "strict"):  # pylint: disable=unused-argument,line-too-long
+        if self._status == 1:
+            with self._popen(cmd) as p:
+                output = p.read(timeout=timeout)
+                # Remove VT100 Codes and split by newline
+                data = self.re_vt100.sub(
+                    '', output.decode('utf-8'), count=1000000
+                ).replace("\r", "").split("\n")
+                self.logger.debug("Received Data: {data}")
```
Bastian-Krause, line 89: I guess this is meant as an f-string. Please convert to lazy % formatting.

labgrid/util/marker.py
```diff
@@ -1,6 +1,178 @@
...
+from pexpect import EOF, TIMEOUT, spawn
+
+from ..protocol import CommandProcessProtocol
+from ..step import step
+from ..util import OptionalTimeout
 
 
 def gen_marker():
-    return ''.join(random.choice(string.ascii_uppercase) for i in range(10))
+    return "".join(random.choice(string.ascii_uppercase) for i in range(10))
```
Bastian-Krause, line 13: No need to change this line.

labgrid/util/timeout.py
```diff
@@ -23,3 +24,43 @@ def remaining(self):
...
     def expired(self):
         return self._deadline <= time.monotonic()
+
+
+@attr.s(eq=False)
+class OptionalTimeout:
+    """
+    Represents an optional timeout. If the timeout value is None, the timer
+    will never expire. If timeout negative and default is default is not None,
+    the default timer value will be used
```
Bastian-Krause, line 34: This sentence needs revisiting.

Bastian-Krause (commented):
I've tested this briefly now.

Issues I stumbled upon:
- `BareboxDriver` now times out on `run()`
- running multiple processes on `BareboxDriver`/`ShellDriver` is not forbidden, but runs in timeouts
- calling `stop()` on `ConsoleMarkerProcess` returned by `ShellDriver.popen()` leads to a timeout (e.g. stopping a "sleep 100") - the implementation looks incomplete, I don't see the prompt being detected on `stop()`

Open questions:
- why is `SSHDriver.run()` not using `popen()`?
- I do not really understand the difference between `read_nonblocking()`/`read()`

Styling:
- we use f-strings where appropriate since #838, please do the same

Note: this is no comprehensive review yet, just a collection of things I saw while testing with a QEMU and a physical target.

JoshuaWatt:
> I've tested this briefly now.
> 
> Issues I stumbled upon:
> 
>     * `BareboxDriver` now times out on `run()`

I don't have any way of testing barebox, can you provide more information so I can fix it?

> 
>     * running multiple processes on `BareboxDriver`/`ShellDriver` is not forbidden, but runs in timeouts

Do we want to explicitly disallow that? I'm fine with that if that's what everyone wants to do.

> 
>     * calling `stop()` on `ConsoleMarkerProcess` returned by `ShellDriver.popen()` leads to a timeout (e.g. stopping a "sleep 100") - the implementation looks incomplete, I don't see the prompt being detected on `stop()`
> 
> 
> Open questions:
> 
>     * why is `SSHDriver.run()` not using `popen()`?

It would be nice, but I can't do that without breaking the SSH driver because of the stderr_merge option, which is not possible to implement with popen()

> 
>     * I do not really understand the difference between `read_nonblocking()`/`read()`
> 

`read_nonblocking` allows a `CommandProcessProcotol` to be directly used as the backend for a `popen.spawn` object, since the latter requires the the `read_nonblocking` function to be implemented. `read` is the API for normal folks :)

> 
> Styling:
> 
>     * we use f-strings where appropriate since #838, please do the same

I tried to use f-strings everywhere.... IHMO it's confusing to developers to require different string formatting for different uses (unless there is some really compelling performance reason to do otherwise).

> 
> 
> Note: this is no comprehensive review yet, just a collection of things I saw while testing with a QEMU and a physical target.

labgrid/driver/bareboxdriver.py
Bastian-Krause: Yes, there is a reason. Here is a short wrap-up: https://medium.com/swlh/why-it-matters-how-you-log-in-python-1a1085851205

TL;DR: formatting is only done if required (i.e. the log level is reached) and errors are handled gracefully.

Bastian-Krause:
> > * `BareboxDriver` now times out on `run()`
> > ```
> 
> I don't have any way of testing barebox, can you provide more information so I can fix it?

Here is my demo script:
```python
#!/usr/bin/env python3

import labgrid

labgrid.StepReporter()

env = labgrid.Environment("beaglebone.yaml")
target = env.get_target()

barebox = target.get_driver("BareboxDriver")
target.activate(barebox)

barebox.run("echo hi")
```

This results in:
[labgrid-barebox-run.log](https://github.com/labgrid-project/labgrid/files/7247463/labgrid-barebox-run.log)

I have not looked into it yet.

labgrid/driver/bareboxdriver.py
JoshuaWatt: Reading the article has confused me more; IIUC, the article is suggesting that logging should use lazy evaluation by passing format arguments as parameters (_not_ using `%`), in which case the log message would be something like `self.logger.debug("Received data: %s", data)`. However, it seems you are suggesting I use `%` instead, which is _not_ a lazy evalutation and therefore not superior (at least in terms of what the article is discussing) from f-strings.

I _think_ what you want to do is either use actual lazy evaluation for log messages, or non-lazy f-strings for consistency with everywhere else f-strings are used

labgrid/driver/ubootdriver.py
JoshuaWatt: Per previous discussion, `%` is not lazt

labgrid/driver/bareboxdriver.py
Bastian-Krause: With "% formatting" I mean `logger.debug("%s", data)`.

JoshuaWatt: OK, I will do that

JoshuaWatt:
> How would you run a command that takes longer than 30 seconds?
> 
> ```
> def test_async_output(strategy):
>     strategy.transition("shell")
>     with strategy.shell.popen("sleep 40") as p:
>         pass
> ```
> 
> This ends up with a `TIMEOUT` exception.
> 
> Another thing: how would you read the output of a long-running command (e.g. `find /`) asynchronously? Could you provide an example? I tried something with `while p.poll() is None`, but the `poll()` method blocks for a second if the process is still alive, so that feels wrong.

`poll` is for checking if the process is still alive; to read output use `p.read()` or `p.expect()`

JoshuaWatt:
> > How would you run a command that takes longer than 30 seconds?
> > ```
> > def test_async_output(strategy):
> >     strategy.transition("shell")
> >     with strategy.shell.popen("sleep 40") as p:
> >         pass
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This ends up with a `TIMEOUT` exception.
> 
> Ah, this has nothing to do with the long-running command. The context manager hits CTRL-c on close, which prevents the second marker from being echoed (on the shell, not in barebox though). That's why we run into the timeout there.

Does something need to change? It seems like it's doing what you are asking for (launching the process, then immediately terminating it). If you want to wait for it to exit normally, use `p.poll()` in a loop

Bastian-Krause:
> > How would you run a command that takes longer than 30 seconds?
> > ```
> > def test_async_output(strategy):
> >     strategy.transition("shell")
> >     with strategy.shell.popen("sleep 40") as p:
> >         pass
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This ends up with a `TIMEOUT` exception.
> > Another thing: how would you read the output of a long-running command (e.g. `find /`) asynchronously? Could you provide an example? I tried something with `while p.poll() is None`, but the `poll()` method blocks for a second if the process is still alive, so that feels wrong.
> 
> `poll` is for checking if the process is still alive; to read output use `p.read()` or `p.expect()`

Okay, will try that.

> > > How would you run a command that takes longer than 30 seconds?
> > > ```
> > > def test_async_output(strategy):
> > >     strategy.transition("shell")
> > >     with strategy.shell.popen("sleep 40") as p:
> > >         pass
> > >   
> > > This ends up with a `TIMEOUT` exception.
> > 
> > 
> > Ah, this has nothing to do with the long-running command. The context manager hits CTRL-c on close, which prevents the second marker from being echoed (on the shell, not in barebox though). That's why we run into the timeout there.
> 
> Does something need to change? It seems like it's doing what you are asking for (launching the process, then immediately terminating it). If you want to wait for it to exit normally, use `p.poll()` in a loop, or `p.wait()`

I think the behavior between barebox and shell should be consistent. I would not expect to get a `TIMEOUT` exception if I do not wait for the process to finish. Do you agree?

JoshuaWatt:
> > > How would you run a command that takes longer than 30 seconds?
> > > ```
> > > def test_async_output(strategy):
> > >     strategy.transition("shell")
> > >     with strategy.shell.popen("sleep 40") as p:
> > >         pass
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > This ends up with a `TIMEOUT` exception.
> > > Another thing: how would you read the output of a long-running command (e.g. `find /`) asynchronously? Could you provide an example? I tried something with `while p.poll() is None`, but the `poll()` method blocks for a second if the process is still alive, so that feels wrong.
> > 
> > 
> > `poll` is for checking if the process is still alive; to read output use `p.read()` or `p.expect()`
> 
> Okay, will try that.
> 
> > > > How would you run a command that takes longer than 30 seconds?
> > > > ```
> > > > def test_async_output(strategy):
> > > >     strategy.transition("shell")
> > > >     with strategy.shell.popen("sleep 40") as p:
> > > >         pass
> > > >   
> > > > This ends up with a `TIMEOUT` exception.
> > > > ```
> > > 
> > > 
> > > Ah, this has nothing to do with the long-running command. The context manager hits CTRL-c on close, which prevents the second marker from being echoed (on the shell, not in barebox though). That's why we run into the timeout there.
> > 
> > 
> > Does something need to change? It seems like it's doing what you are asking for (launching the process, then immediately terminating it). If you want to wait for it to exit normally, use `p.poll()` in a loop, or `p.wait()`
> 
> I think the behavior between barebox and shell should be consistent. I would not expect to get a `TIMEOUT` exception if I do not wait for the process to finish. Do you agree?

Ya, that makes sense

JoshuaWatt:
@Bastian-Krause Ok, I tracked this down to a behavior difference is bash (which I was using) vs. busybox. I've corrected it and now it should work the same for both shells

JoshuaWatt:
For those interested, the difference is:

bash:
```
raspberrypi4-64:~# echo -n 'ABC'; sleep 10; echo 'ABC' $?; 
ABC^C
ABC 130
raspberrypi4-64:~#
```

busybox:
```
raspberrypi4-64:~# echo -n 'ABC'; sleep 10; echo 'ABC' $?;
ABC^C
raspberrypi4-64:~#
```

Note that interrupting the command on busybox prevents all subsequent commands (the `echo` with the exit code) from running, where as they still run under bash.

labgrid/protocol/commandprotocol.py
```diff
@@ -1,6 +1,87 @@
...
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def read_nonblocking(self, size=1, timeout=-1):
+        """
+        This reads at most size characters from the process, or until EOF or a
+        timeout occurs. If no data is read before EOF or a timeout occurs, then
+        pexpect.EOF or pexpect.TIMEOUT are raised respectively.
+
+        A timeout of None means to wait forever for either data or EOF
```
jluebbe, line 30: Why does a 'nonblocking' read support waiting for a timeout?

labgrid/driver/ubootdriver.py
```diff
@@ -69,26 +69,25 @@ def on_deactivate(self):
...
-                '', before.decode('utf-8'), count=1000000
-            ).replace("\r", "").split("\n")
-            self.logger.debug("Received Data: %s", data)
-            # Remove first element, the invoked cmd
-            data = data[data.index(marker) + 1:]
-            data = data[:data.index(marker)]
-            exitcode = int(data[-1])
-            del data[-1]
-            return (data, [], exitcode)
+        cmp_command = f"""echo '{marker[:4]}''{marker[4:]}'\\\\c; {cmd}; echo '{marker[:4]}''{marker[4:]}' $?;"""  # pylint: disable=line-too-long
```
jluebbe, line 76: What does the `\\\\c` do?

labgrid/driver/sshdriver.py
```diff
@@ -211,6 +281,27 @@ def _run(self, cmd, codec="utf-8", decodeerrors="strict", timeout=None): # pylin
...
+    @step(args=['cmd'], result=True)
+    def popen(self, cmd: str):
+        if not self._check_keepalive():
+            raise ExecutionError("Keepalive no longer running")
+
+        complete_cmd = ["ssh", "-o", "LogLevel=QUIET", "-x", *self.ssh_prefix,
+                        "-p", str(self.networkservice.port), "-l", self.networkservice.username,
+                        self.networkservice.address, "-T", "--", '/bin/sh -c {}'.format(shlex.quote(cmd)),
+                        ]
+        self.logger.debug("Sending command: %s", complete_cmd)
```
jluebbe, line 294: In the python standard library, popen doesn't use a shell by default. What is the motivation to do this here? Having the same interface as the `ShellDriver`?

labgrid/protocol/commandprotocol.py
jluebbe: Ah, this matches https://pexpect.readthedocs.io/en/stable/api/pexpect.html#pexpect.spawn.read_nonblocking

labgrid/util/marker.py
```diff
@@ -1,6 +1,187 @@
...
+        if self._alive:
+            return self._console.write(data)
+        return 0
+
+    @step(result=True)
+    def poll(self):
+        if not self._alive:
+            return self.exitcode
+
+        index, _, match, _ = self._console.expect([self._eof_re, TIMEOUT], timeout=1)
```
jluebbe, line 141: This looks like it could drop output from the process. Is that intentional?

labgrid/driver/ubootdriver.py
JoshuaWatt: `\c` is the u-boot method for suppressing new-lines in an echo

labgrid/driver/sshdriver.py
JoshuaWatt: Yes, it should behave the same as ShellDriver; I think specifically you need to use a shell for redirection to work

labgrid/util/marker.py
JoshuaWatt: I don't think it can because the EOF is anchored to the beginning of the string, so unless the output has all been consumed, the process will still be "alive" (see the warning in `wait`)

labgrid/protocol/commandprotocol.py
```diff
@@ -1,6 +1,87 @@
...
+        """
+        Check if the process is alive. If the process is alive, None will be
+        returned. If the process exited normally, the return code will be
+        returned. If the process died from a signal, the negative value of the
+        signal number will be returned.
+        """
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def stop(self):
```
jremmet, line 54: nice work :)
I tested it with my setup and run into a issue while running:
```
def test_shell_interrupt(shell_command):
    with shell_command.popen("yes") as p:
        l = p.read(1024).decode("utf-8").splitlines()
        assert l == ["y"] * len(l)

    shell_command.run_check("true")
```

on the console log I see:
```
root@ksp0627-imx6-1:~# MARKER='XVUS''LJFJHA' run yes
XVUSLJFJHAy
y
y


y
y
^C
root@ksp0627-imx6-1:~#
```

put the test runs into a 60sec  timout
```

        SerialDriver.expect state='start' args={'pattern': [re.compile(b'^(?!YXLELHEROR)(?!YXLELHERO$)(?!YXLELHER$)(?!YXLELHE$)(?!YXLELH$)(?!YXLEL$)(?!YXLE$)(?!YXL$)(?!YX$)(?!Y$).', re.DOTALL), re.compile(b'^YXLELHEROR\\s+(\\d+)\\s+.*root@[\\w-]+:[^ ]+'), <class 'pexpect.exceptions.TIMEOUT'>]}




       SerialDriver.expect state='stop' result=(2, b'', <class 'pexpect.exceptions.TIMEOUT'>, <class 'pexpect.exceptions.TIMEOUT'>) duration='60.007'
     ConsoleMarkerProcess.expect state='stop' result=(1, b'\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\ny\r\n^C\r\n', <re.Match object; span=(711, 733), match=b'root@ksp0627-imx6-1:~#'>, b'root@ksp0627-imx6-1:~#') duration='60.087'
   ConsoleMarkerProcess.stop state='stop' duration='60.088'
```

Seams `read_nonblocking` not give the "promt" only case without a timout for busybox shells

jremmet (commented):
Nice work, found a issue on a busybox system

labgrid/driver/ubootdriver.py
JoshuaWatt: Yes

jluebbe: Perhaps a comment regarding the `\\\\c` would be useful, though. :)

labgrid/protocol/commandprotocol.py
JoshuaWatt: @jremmet Hmm, that suspiciously looks like the problem I fixed yesterday with busybox.... can you please make sure verify you have the latest commits I pushed? If you do please let me know and I'll look more; specifically, there should a additional regex that matches the prompt in the `SerialDriver.expect` (for a total of 3 regex it's looking for)

jremmet: Thanks for the fast response. I'm on 83c64e9 JoshuaWatt/labgrid.git@command-background. The timed out expext seams to come from `ConsoleMarkerProcess` `read_nonblocking`. There I found no `_prompt_re`.

JoshuaWatt: Huh, I'm not sure what is going on here; I've been testing with busybox and can't seem to reproduce what's going on here. The weird part is that `stop` is waiting 60 seconds but then returning that if found the prompt anyway

Can you do a little more digging and provides some more logs?

jremmet: After updating to 58d847d to timeout is gone :+1:

JoshuaWatt:
Ping

JoshuaWatt:
Ping

JoshuaWatt:
Ping

JoshuaWatt:
Ping

Emantor:
Hi,

I deployed this internally and let it run over our internal test suites and didn't find an error related to this PR.
Thats great, thanks a lot for the work!
The issue I see at the moment is that the verbose pytest output becomes unreadable:
```
10:13:34      SerialDriver.expect state='start' args={'pattern': [re.compile(b'^(?!WASPAEMYUO)(?!WASPAEMYU$)(?!WASPAEMY$)(?!WASPAEM$)(?!WASPAE$)(?!WASPA$)(?!WASP$)(?!WAS$)(?!WA$)(?!W$).', re.DOTALL), re.compile(b'^WASPAEMYUO\\s+(\\d+)\\s+.*root@[^:]+:[^ ]+'), <class 'pexpect.exceptions.TIMEOUT'>]}
10:13:34      SerialDriver.expect state='stop' result=(0, b'', <re.Match object; span=(0, 1), match=b'y'>, b'y')
10:13:34      SerialDriver.expect state='start' args={'pattern': [re.compile(b'^(?!WASPAEMYUO)(?!WASPAEMYU$)(?!WASPAEMY$)(?!WASPAEM$)(?!WASPAE$)(?!WASPA$)(?!WASP$)(?!WAS$)(?!WA$)(?!W$).', re.DOTALL), re.compile(b'^WASPAEMYUO\\s+(\\d+)\\s+.*root@[^:]+:[^ ]+'), <class 'pexpect.exceptions.TIMEOUT'>]}
10:13:34      SerialDriver.expect state='stop' result=(0, b'', <re.Match object; span=(0, 1), match=b's'>, b's')
10:13:34      SerialDriver.expect state='start' args={'pattern': [re.compile(b'^(?!WASPAEMYUO)(?!WASPAEMYU$)(?!WASPAEMY$)(?!WASPAEM$)(?!WASPAE$)(?!WASPA$)(?!WASP$)(?!WAS$)(?!WA$)(?!W$).', re.DOTALL), re.compile(b'^WASPAEMYUO\\s+(\\d+)\\s+.*root@[^:]+:[^ ]+'), <class 'pexpect.exceptions.TIMEOUT'>]}
10:13:34      SerialDriver.expect state='stop' result=(0, b'', <re.Match object; span=(0, 1), match=b't'>, b't')
10:13:34      SerialDriver.expect state='start' args={'pattern': [re.compile(b'^(?!WASPAEMYUO)(?!WASPAEMYU$)(?!WASPAEMY$)(?!WASPAEM$)(?!WASPAE$)(?!WASPA$)(?!WASP$)(?!WAS$)(?!WA$)(?!W$).', re.DOTALL), re.compile(b'^WASPAEMYUO\\s+(\\d+)\\s+.*root@[^:]+:[^ ]+'), <class 'pexpect.exceptions.TIMEOUT'>]}
10:13:34      SerialDriver.expect state='stop' result=(0, b'', <re.Match object; span=(0, 1), match=b'e'>, b'e')
10:13:34      SerialDriver.expect state='start' args={'pattern': [re.compile(b'^(?!WASPAEMYUO)(?!WASPAEMYU$)(?!WASPAEMY$)(?!WASPAEM$)(?!WASPAE$)(?!WASPA$)(?!WASP$)(?!WAS$)(?!WA$)(?!W$).', re.DOTALL), re.compile(b'^WASPAEMYUO\\s+(\\d+)\\s+.*root@[^:]+:[^ ]+'), <class 'pexpect.exceptions.TIMEOUT'>]}
10:13:34      SerialDriver.expect state='stop' result=(0, b'', <re.Match object; span=(0, 1), match=b'm'>, b'm')
10:13:34      SerialDriver.expect state='start' args={'pattern': [re.compile(b'^(?!WASPAEMYUO)(?!WASPAEMYU$)(?!WASPAEMY$)(?!WASPAEM$)(?!WASPAE$)(?!WASPA$)(?!WASP$)(?!WAS$)(?!WA$)(?!W$).', re.DOTALL), re.compile(b'^WASPAEMYUO\\s+(\\d+)\\s+.*root@[^:]+:[^ ]+'), <class 'pexpect.exceptions.TIMEOUT'>]}
10:13:34      SerialDriver.expect state='stop' result=(0, b'', <re.Match object; span=(0, 1), match=b'/'>, b'/')
10:13:34      SerialDriver.expect state='start' args={'pattern': [re.compile(b'^(?!WASPAEMYUO)(?!WASPAEMYU$)(?!WASPAEMY$)(?!WASPAEM$)(?!WASPAE$)(?!WASPA$)(?!WASP$)(?!WAS$)(?!WA$)(?!W$).', re.DOTALL), re.compile(b'^WASPAEMYUO\\s+(\\d+)\\s+.*root@[^:]+:[^ ]+'), <class 'pexpect.exceptions.TIMEOUT'>]}
10:13:34      SerialDriver.expect state='stop' result=(0, b'', <re.Match object; span=(0, 1), match=b's'>, b's')
10:13:34      SerialDriver.expect state='start' args={'pattern': [re.compile(b'^(?!WASPAEMYUO)(?!WASPAEMYU$)(?!WASPAEMY$)(?!WASPAEM$)(?!WASPAE$)(?!WASPA$)(?!WASP$)(?!WAS$)(?!WA$)(?!W$).', re.DOTALL), re.compile(b'^WASPAEMYUO\\s+(\\d+)\\s+.*root@[^:]+:[^ ]+'), <class 'pexpect.exceptions.TIMEOUT'>]}
10:13:34      SerialDriver.expect state='stop' result=(0, b'', <re.Match object; span=(0, 1), match=b'y'>, b'y')
10:13:34      SerialDriver.expect state='start' args={'pattern': [re.compile(b'^(?!WASPAEMYUO)(?!WASPAEMYU$)(?!WASPAEMY$)(?!WASPAEM$)(?!WASPAE$)(?!WASPA$)(?!WASP$)(?!WAS$)(?!WA$)(?!W$).', re.DOTALL), re.compile(b'^WASPAEMYUO\\s+(\\d+)\\s+.*root@[^:]+:[^ ]+'), <class 'pexpect.exceptions.TIMEOUT'>]}
10:13:34      SerialDriver.expect state='stop' result=(0, b'', <re.Match object; span=(0, 1), match=b's'>, b's')
10:13:34      SerialDriver.expect state='start' args={'pattern': [re.compile(b'^(?!WASPAEMYUO)(?!WASPAEMYU$)(?!WASPAEMY$)(?!WASPAEM$)(?!WASPAE$)(?!WASPA$)(?!WASP$)(?!WAS$)(?!WA$)(?!W$).', re.DOTALL), re.compile(b'^WASPAEMYUO\\s+(\\d+)\\s+.*root@[^:]+:[^ ]+'), <class 'pexpect.exceptions.TIMEOUT'>]}
10:13:34      SerialDriver.expect state='stop' result=(0, b'', <re.Match object; span=(0, 1), match=b't'>, b't')
10:13:34      SerialDriver.expect state='start' args={'pattern': [re.compile(b'^(?!WASPAEMYUO)(?!WASPAEMYU$)(?!WASPAEMY$)(?!WASPAEM$)(?!WASPAE$)(?!WASPA$)(?!WASP$)(?!WAS$)(?!WA$)(?!W$).', re.DOTALL), re.compile(b'^WASPAEMYUO\\s+(\\d+)\\s+.*root@[^:]+:[^ ]+'), <class 'pexpect.exceptions.TIMEOUT'>]}
10:13:34      SerialDriver.expect state='stop' result=(0, b'', <re.Match object; span=(0, 1), match=b'e'>, b'e')
10:13:34      SerialDriver.expect state='start' args={'pattern': [re.compile(b'^(?!WASPAEMYUO)(?!WASPAEMYU$)(?!WASPAEMY$)(?!WASPAEM$)(?!WASPAE$)(?!WASPA$)(?!WASP$)(?!WAS$)(?!WA$)(?!W$).', re.DOTALL), re.compile(b'^WASPAEMYUO\\s+(\\d+)\\s+.*root@[^:]+:[^ ]+'), <class 'pexpect.exceptions.TIMEOUT'>]}
10:13:34      SerialDriver.expect state='stop' result=(0, b'', <re.Match object; span=(0, 1), match=b'm'>, b'm')
10:13:34      SerialDriver.expect state='start' args={'pattern': [re.compile(b'^(?!WASPAEMYUO)(?!WASPAEMYU$)(?!WASPAEMY$)(?!WASPAEM$)(?!WASPAE$)(?!WASPA$)(?!WASP$)(?!WAS$)(?!WA$)(?!W$).', re.DOTALL), re.compile(b'^WASPAEMYUO\\s+(\\d+)\\s+.*root@[^:]+:[^ ]+'), <class 'pexpect.exceptions.TIMEOUT'>]}
10:13:34      SerialDriver.expect state='stop' result=(0, b'', <re.Match object; span=(0, 1), match=b'd'>, b'd')
10:13:34      SerialDriver.expect state='start' args={'pattern': [re.compile(b'^(?!WASPAEMYUO)(?!WASPAEMYU$)(?!WASPAEMY$)(?!WASPAEM$)(?!WASPAE$)(?!WASPA$)(?!WASP$)(?!WAS$)(?!WA$)(?!W$).', re.DOTALL), re.compile(b'^WASPAEMYUO\\s+(\\d+)\\s+.*root@[^:]+:[^ ]+'), <class 'pexpect.exceptions.TIMEOUT'>]}
10:13:34      SerialDriver.expect state='stop' result=(0, b'', <re.Match object; span=(0, 1), match=b'-'>, b'-')
10:13:34      SerialDriver.expect state='start' args={'pattern': [re.compile(b'^(?!WASPAEMYUO)(?!WASPAEMYU$)(?!WASPAEMY$)(?!WASPAEM$)(?!WASPAE$)(?!WASPA$)(?!WASP$)(?!WAS$)(?!WA$)(?!W$).', re.DOTALL), re.compile(b'^WASPAEMYUO\\s+(\\d+)\\s+.*root@[^:]+:[^ ]+'), <class 'pexpect.exceptions.TIMEOUT'>]}
10:13:34      SerialDriver.expect state='stop' result=(0, b'', <re.Match object; span=(0, 1), match=b'a'>, b'a')
10:13:34      SerialDriver.expect state='start' args={'pattern': [re.compile(b'^(?!WASPAEMYUO)(?!WASPAEMYU$)(?!WASPAEMY$)(?!WASPAEM$)(?!WASPAE$)(?!WASPA$)(?!WASP$)(?!WAS$)(?!WA$)(?!W$).', re.DOTALL), re.compile(b'^WASPAEMYUO\\s+(\\d+)\\s+.*root@[^:]+:[^ ]+'), <class 'pexpect.exceptions.TIMEOUT'>]}
10:13:34      SerialDriver.expect state='stop' result=(0, b'', <re.Match object; span=(0, 1), match=b'u'>, b'u')
10:13:34      SerialDriver.expect state='start' args={'pattern': [re.compile(b'^(?!WASPAEMYUO)(?!WASPAEMYU$)(?!WASPAEMY$)(?!WASPAEM$)(?!WASPAE$)(?!WASPA$)(?!WASP$)(?!WAS$)(?!WA$)(?!W$).', re.DOTALL), re.compile(b'^WASPAEMYUO\\s+(\\d+)\\s+.*root@[^:]+:[^ ]+'), <class 'pexpect.exceptions.TIMEOUT'>]}
```
Since we match on a single byte a lot of the steps don't contain any meaningful information. I am not sure how to solve this, one way would be to create a helper which does not contain the step decorator to supress this output. Note that `expect()` is still called by drivers and I wouldn't want to get rid of this as is.

JoshuaWatt:
> Hi,
> 
> I deployed this internally and let it run over our internal test suites and didn't find an error related to this PR. Thats great, thanks a lot for the work! The issue I see at the moment is that the verbose pytest output becomes unreadable:
> 
> ```
> 10:13:34      SerialDriver.expect state='start' args={'pattern': [re.compile(b'^(?!WASPAEMYUO)(?!WASPAEMYU$)(?!WASPAEMY$)(?!WASPAEM$)(?!WASPAE$)(?!WASPA$)(?!WASP$)(?!WAS$)(?!WA$)(?!W$).', re.DOTALL), re.compile(b'^WASPAEMYUO\\s+(\\d+)\\s+.*root@[^:]+:[^ ]+'), <class 'pexpect.exceptions.TIMEOUT'>]}
> 10:13:34      SerialDriver.expect state='stop' result=(0, b'', <re.Match object; span=(0, 1), match=b'y'>, b'y')
> 10:13:34      SerialDriver.expect state='start' args={'pattern': [re.compile(b'^(?!WASPAEMYUO)(?!WASPAEMYU$)(?!WASPAEMY$)(?!WASPAEM$)(?!WASPAE$)(?!WASPA$)(?!WASP$)(?!WAS$)(?!WA$)(?!W$).', re.DOTALL), re.compile(b'^WASPAEMYUO\\s+(\\d+)\\s+.*root@[^:]+:[^ ]+'), <class 'pexpect.exceptions.TIMEOUT'>]}
> 10:13:34      SerialDriver.expect state='stop' result=(0, b'', <re.Match object; span=(0, 1), match=b's'>, b's')
> 10:13:34      SerialDriver.expect state='start' args={'pattern': [re.compile(b'^(?!WASPAEMYUO)(?!WASPAEMYU$)(?!WASPAEMY$)(?!WASPAEM$)(?!WASPAE$)(?!WASPA$)(?!WASP$)(?!WAS$)(?!WA$)(?!W$).', re.DOTALL), re.compile(b'^WASPAEMYUO\\s+(\\d+)\\s+.*root@[^:]+:[^ ]+'), <class 'pexpect.exceptions.TIMEOUT'>]}
> 10:13:34      SerialDriver.expect state='stop' result=(0, b'', <re.Match object; span=(0, 1), match=b't'>, b't')
> 10:13:34      SerialDriver.expect state='start' args={'pattern': [re.compile(b'^(?!WASPAEMYUO)(?!WASPAEMYU$)(?!WASPAEMY$)(?!WASPAEM$)(?!WASPAE$)(?!WASPA$)(?!WASP$)(?!WAS$)(?!WA$)(?!W$).', re.DOTALL), re.compile(b'^WASPAEMYUO\\s+(\\d+)\\s+.*root@[^:]+:[^ ]+'), <class 'pexpect.exceptions.TIMEOUT'>]}
> 10:13:34      SerialDriver.expect state='stop' result=(0, b'', <re.Match object; span=(0, 1), match=b'e'>, b'e')
> 10:13:34      SerialDriver.expect state='start' args={'pattern': [re.compile(b'^(?!WASPAEMYUO)(?!WASPAEMYU$)(?!WASPAEMY$)(?!WASPAEM$)(?!WASPAE$)(?!WASPA$)(?!WASP$)(?!WAS$)(?!WA$)(?!W$).', re.DOTALL), re.compile(b'^WASPAEMYUO\\s+(\\d+)\\s+.*root@[^:]+:[^ ]+'), <class 'pexpect.exceptions.TIMEOUT'>]}
> 10:13:34      SerialDriver.expect state='stop' result=(0, b'', <re.Match object; span=(0, 1), match=b'm'>, b'm')
> 10:13:34      SerialDriver.expect state='start' args={'pattern': [re.compile(b'^(?!WASPAEMYUO)(?!WASPAEMYU$)(?!WASPAEMY$)(?!WASPAEM$)(?!WASPAE$)(?!WASPA$)(?!WASP$)(?!WAS$)(?!WA$)(?!W$).', re.DOTALL), re.compile(b'^WASPAEMYUO\\s+(\\d+)\\s+.*root@[^:]+:[^ ]+'), <class 'pexpect.exceptions.TIMEOUT'>]}
> 10:13:34      SerialDriver.expect state='stop' result=(0, b'', <re.Match object; span=(0, 1), match=b'/'>, b'/')
> 10:13:34      SerialDriver.expect state='start' args={'pattern': [re.compile(b'^(?!WASPAEMYUO)(?!WASPAEMYU$)(?!WASPAEMY$)(?!WASPAEM$)(?!WASPAE$)(?!WASPA$)(?!WASP$)(?!WAS$)(?!WA$)(?!W$).', re.DOTALL), re.compile(b'^WASPAEMYUO\\s+(\\d+)\\s+.*root@[^:]+:[^ ]+'), <class 'pexpect.exceptions.TIMEOUT'>]}
> 10:13:34      SerialDriver.expect state='stop' result=(0, b'', <re.Match object; span=(0, 1), match=b's'>, b's')
> 10:13:34      SerialDriver.expect state='start' args={'pattern': [re.compile(b'^(?!WASPAEMYUO)(?!WASPAEMYU$)(?!WASPAEMY$)(?!WASPAEM$)(?!WASPAE$)(?!WASPA$)(?!WASP$)(?!WAS$)(?!WA$)(?!W$).', re.DOTALL), re.compile(b'^WASPAEMYUO\\s+(\\d+)\\s+.*root@[^:]+:[^ ]+'), <class 'pexpect.exceptions.TIMEOUT'>]}
> 10:13:34      SerialDriver.expect state='stop' result=(0, b'', <re.Match object; span=(0, 1), match=b'y'>, b'y')
> 10:13:34      SerialDriver.expect state='start' args={'pattern': [re.compile(b'^(?!WASPAEMYUO)(?!WASPAEMYU$)(?!WASPAEMY$)(?!WASPAEM$)(?!WASPAE$)(?!WASPA$)(?!WASP$)(?!WAS$)(?!WA$)(?!W$).', re.DOTALL), re.compile(b'^WASPAEMYUO\\s+(\\d+)\\s+.*root@[^:]+:[^ ]+'), <class 'pexpect.exceptions.TIMEOUT'>]}
> 10:13:34      SerialDriver.expect state='stop' result=(0, b'', <re.Match object; span=(0, 1), match=b's'>, b's')
> 10:13:34      SerialDriver.expect state='start' args={'pattern': [re.compile(b'^(?!WASPAEMYUO)(?!WASPAEMYU$)(?!WASPAEMY$)(?!WASPAEM$)(?!WASPAE$)(?!WASPA$)(?!WASP$)(?!WAS$)(?!WA$)(?!W$).', re.DOTALL), re.compile(b'^WASPAEMYUO\\s+(\\d+)\\s+.*root@[^:]+:[^ ]+'), <class 'pexpect.exceptions.TIMEOUT'>]}
> 10:13:34      SerialDriver.expect state='stop' result=(0, b'', <re.Match object; span=(0, 1), match=b't'>, b't')
> 10:13:34      SerialDriver.expect state='start' args={'pattern': [re.compile(b'^(?!WASPAEMYUO)(?!WASPAEMYU$)(?!WASPAEMY$)(?!WASPAEM$)(?!WASPAE$)(?!WASPA$)(?!WASP$)(?!WAS$)(?!WA$)(?!W$).', re.DOTALL), re.compile(b'^WASPAEMYUO\\s+(\\d+)\\s+.*root@[^:]+:[^ ]+'), <class 'pexpect.exceptions.TIMEOUT'>]}
> 10:13:34      SerialDriver.expect state='stop' result=(0, b'', <re.Match object; span=(0, 1), match=b'e'>, b'e')
> 10:13:34      SerialDriver.expect state='start' args={'pattern': [re.compile(b'^(?!WASPAEMYUO)(?!WASPAEMYU$)(?!WASPAEMY$)(?!WASPAEM$)(?!WASPAE$)(?!WASPA$)(?!WASP$)(?!WAS$)(?!WA$)(?!W$).', re.DOTALL), re.compile(b'^WASPAEMYUO\\s+(\\d+)\\s+.*root@[^:]+:[^ ]+'), <class 'pexpect.exceptions.TIMEOUT'>]}
> 10:13:34      SerialDriver.expect state='stop' result=(0, b'', <re.Match object; span=(0, 1), match=b'm'>, b'm')
> 10:13:34      SerialDriver.expect state='start' args={'pattern': [re.compile(b'^(?!WASPAEMYUO)(?!WASPAEMYU$)(?!WASPAEMY$)(?!WASPAEM$)(?!WASPAE$)(?!WASPA$)(?!WASP$)(?!WAS$)(?!WA$)(?!W$).', re.DOTALL), re.compile(b'^WASPAEMYUO\\s+(\\d+)\\s+.*root@[^:]+:[^ ]+'), <class 'pexpect.exceptions.TIMEOUT'>]}
> 10:13:34      SerialDriver.expect state='stop' result=(0, b'', <re.Match object; span=(0, 1), match=b'd'>, b'd')
> 10:13:34      SerialDriver.expect state='start' args={'pattern': [re.compile(b'^(?!WASPAEMYUO)(?!WASPAEMYU$)(?!WASPAEMY$)(?!WASPAEM$)(?!WASPAE$)(?!WASPA$)(?!WASP$)(?!WAS$)(?!WA$)(?!W$).', re.DOTALL), re.compile(b'^WASPAEMYUO\\s+(\\d+)\\s+.*root@[^:]+:[^ ]+'), <class 'pexpect.exceptions.TIMEOUT'>]}
> 10:13:34      SerialDriver.expect state='stop' result=(0, b'', <re.Match object; span=(0, 1), match=b'-'>, b'-')
> 10:13:34      SerialDriver.expect state='start' args={'pattern': [re.compile(b'^(?!WASPAEMYUO)(?!WASPAEMYU$)(?!WASPAEMY$)(?!WASPAEM$)(?!WASPAE$)(?!WASPA$)(?!WASP$)(?!WAS$)(?!WA$)(?!W$).', re.DOTALL), re.compile(b'^WASPAEMYUO\\s+(\\d+)\\s+.*root@[^:]+:[^ ]+'), <class 'pexpect.exceptions.TIMEOUT'>]}
> 10:13:34      SerialDriver.expect state='stop' result=(0, b'', <re.Match object; span=(0, 1), match=b'a'>, b'a')
> 10:13:34      SerialDriver.expect state='start' args={'pattern': [re.compile(b'^(?!WASPAEMYUO)(?!WASPAEMYU$)(?!WASPAEMY$)(?!WASPAEM$)(?!WASPAE$)(?!WASPA$)(?!WASP$)(?!WAS$)(?!WA$)(?!W$).', re.DOTALL), re.compile(b'^WASPAEMYUO\\s+(\\d+)\\s+.*root@[^:]+:[^ ]+'), <class 'pexpect.exceptions.TIMEOUT'>]}
> 10:13:34      SerialDriver.expect state='stop' result=(0, b'', <re.Match object; span=(0, 1), match=b'u'>, b'u')
> 10:13:34      SerialDriver.expect state='start' args={'pattern': [re.compile(b'^(?!WASPAEMYUO)(?!WASPAEMYU$)(?!WASPAEMY$)(?!WASPAEM$)(?!WASPAE$)(?!WASPA$)(?!WASP$)(?!WAS$)(?!WA$)(?!W$).', re.DOTALL), re.compile(b'^WASPAEMYUO\\s+(\\d+)\\s+.*root@[^:]+:[^ ]+'), <class 'pexpect.exceptions.TIMEOUT'>]}
> ```
> 
> Since we match on a single byte a lot of the steps don't contain any meaningful information. I am not sure how to solve this, one way would be to create a helper which does not contain the step decorator to supress this output. Note that `expect()` is still called by drivers and I wouldn't want to get rid of this as is.

Yes, I noticed that; do you have any suggestions for how to make it more reasonable?

labgrid/util/readmixin.py
```diff
@@ -0,0 +1,77 @@
+from .timeout import Timeout
+from pexpect import EOF, TIMEOUT
```
Bastian-Krause, line 2:
```suggestion
from pexpect import EOF, TIMEOUT

from .timeout import Timeout
```
nitpick: wrong import order

JoshuaWatt:
> Yes, I noticed that; do you have any suggestions for how to make it more reasonable?

Hah, you did give a suggestion! My fault for not reading thoroughly. I'll look into doing that.

JPEWdev:
@Bastian-Krause I think this is ready, would you mind looking again?

Bastian-Krause:
Thanks for keeping at it! I'll try to have a look at this and run our network test suite against this.

labgrid/util/marker.py
```diff
@@ -1,9 +1,174 @@
...
 
 def gen_marker():
     return ''.join(random.choice(MARKER_POOL) for i in range(10))
+
+
+class ConsoleProcessExpect(spawn):
+    """labgrid Wrapper of the pexpect module.
+
+    This class provides pexpect functionality for the ConsoleProtocol classes.
+    driver: ConsoleProtocol object to be passed in
```
jluebbe: This should explain why we need this in addition to the PtxExpect wrapper in `labgrid/util/expect.py`. It seems this allows the user to use expect *only on the process output*, without having to care about the markers.

labgrid/util/marker.py
```diff
@@ -1,9 +1,174 @@
...
+    """
+
+    def __init__(self, process, timeout):
+        "Initializes a pexpect spawn instanse with required configuration"
+        super().__init__(None, timeout=timeout)
+        self.process = process
+
+    def send(self, s):
+        raise NotImplementedError("Sending not implemented")
```
jluebbe: Is there a reason why we don't want to implement send?

JoshuaWatt: No

labgrid/util/marker.py
JoshuaWatt: Done

JoshuaWatt:
Would someone mind looking at this change again? Thanks

Emantor (dismissed):
I gave this a once over and everything seems to be addressed. We just had a release, so its a good time to get this in for additional testing.

Emantor:
We took a deeper look with @jluebbe, @Bastian-Krause and me and taking the state of our serial stack into account, we are not really comfortable to add so much additional complexity on top of the rather fragile expect matching. Nevertheless, the changes for the SSHDriver are much less invasive and more straightforward. What do you think about splitting these out and getting them merged?

While reviewing we also wondered whether the background functionality for the bootloader drivers is really necessary, do you have examples for usage or were those implemented for CommandProtocol completeness?

JoshuaWatt:
> We took a deeper look with @jluebbe, @Bastian-Krause and me and taking the state of our serial stack into account, we are not really comfortable to add so much additional complexity on top of the rather fragile expect matching. Nevertheless, the changes for the SSHDriver are much less invasive and more straightforward. What do you think about splitting these out and getting them merged?
> 
> While reviewing we also wondered whether the background functionality for the bootloader drivers is really necessary, do you have examples for usage or were those implemented for CommandProtocol completeness?

Yes, it was mostly done for completeness. I don't beleive we would need serial async execution. Do you want the serial protocols to raise a `NotImplementedError` for `CommandProtocol.start_process()` or would you prefer `start_process()` be part of a new protocol?

Emantor:
We should just implement it on top of the SSHDriver and not use a protocol at all. If we ever find a second user we can still extend/create a protocol.

Bastian-Krause (commented):
Since `SSHDriver.start_process()` merges stdout and stderr, I'm wondering if it should raise `NotImplementedError` if its `stderr_merge` attribute is False?

JoshHollis251:
> Since `SSHDriver.start_process()` merges stdout and stderr, I'm wondering if it should raise `NotImplementedError` if its `stderr_merge` attribute is False?

If we require stderr_merge=false and the ssh_driver instance were to perform a .run() following a start_process() operation it would have the side effect of maybe unintentionally merging those fields unless it is set back to the default. Is that acceptable behavior, and should we include a unit test for the not_implemented check?

JoshuaWatt:
FYI: @JoshHollis251 is helping me finish this PR :)

Bastian-Krause:
> > Since `SSHDriver.start_process()` merges stdout and stderr, I'm wondering if it should raise `NotImplementedError` if its `stderr_merge` attribute is False?
> 
> If we require stderr_merge=false and the ssh_driver instance were to perform a .run() following a start_process() operation it would have the side effect of maybe unintentionally merging those fields unless it is set back to the default. Is that acceptable behavior, and should we include a unit test for the not_implemented check?

After discussing this with @Emantor: we're not happy with the `stderr_merge` attribute in general, it should have been a kwarg in `run()`/`run_check()`/`wait_for()` from the start. So if we ever get rid of it in the long run, we should not rely on it here.
Also: Since separating stdout from stderr is worth keeping and the default, we should not artificially enforce `stderr_merge=False` just so `start_process()` can be used. So: let's not check that attribute and let's not raise `NotImplementedError`. Let us instead add a docstring documenting that stdout and stderr are merged in `start_process()`.

tests/test_sshdriver.py
```diff
@@ -218,3 +219,92 @@ def test_unix_socket_forward(ssh_localhost, tmpdir):
...
+        code = p.poll()
+        assert code is not None
+        assert code != 0
+
+
+@pytest.mark.sshusername
+def test_start_process_read_to_end(ssh_localhost):
+    with ssh_localhost.start_process("echo Hello World") as p:
+        p.read_to_end() == b"Hello World"
+        p.read_to_end() == b""
```
nick-potenski: Missing assert on both of these lines?

labgrid/util/readmixin.py
```diff
@@ -0,0 +1,81 @@
...
+from .timeout import Timeout
+
+class ReadMixIn:
+    """
+    This class make it more convenient to deal with reading from devices. A
```
nick-potenski:
```suggestion
    This class makes it more convenient to deal with reading from devices. A
```

nick-potenski (changes requested):

nick-potenski (approved):

labgrid/driver/sshdriver.py
```diff
@@ -22,6 +24,56 @@
...
+
+    @step(args=['pattern', 'timeout'], result=True)
+    def expect(self, pattern, *, timeout=-1):
+        index = self._sub.expect(pattern, timeout=timeout)
+        return index, self._sub.before, self._sub.match, self._sub.after
+
+    @step(result=True)
+    def wait(self):
+        return self._sub.wait()
+
```
ozan956: This can lead to a deadlock:

https://pexpect.readthedocs.io/en/stable/api/pexpect.html?utm_source=chatgpt.com#pexpect.spawn.wait

labgrid/driver/sshdriver.py
```diff
@@ -242,6 +296,40 @@ def _run(self, cmd, codec="utf-8", decodeerrors="strict", timeout=None):
...
+        cmd = f"stty -echo; {cmd}"
+        complete_cmd = [self._ssh, "-o", "LogLevel=QUIET", "-x", *self.ssh_prefix,
+                        "-p", str(self.networkservice.port), "-l", self._get_username(),
+                        self.networkservice.address, "-tt", "--", '/bin/sh -c {}'.format(shlex.quote(cmd)),
+                        ]
+        self.logger.debug("Sending command: %s", complete_cmd)
+
+        try:
+            sub = pexpect.spawn(complete_cmd[0], complete_cmd[1:])
+            sub.setecho(False)
```
ozan956:
```suggestion
            sub = pexpect.spawn(complete_cmd[0], complete_cmd[1:], echo=False)
```

If there is no specific reason to do it seperately

labgrid/driver/sshdriver.py
```diff
@@ -242,6 +296,40 @@ def _run(self, cmd, codec="utf-8", decodeerrors="strict", timeout=None):
...
 
+    @Driver.check_active
+    @step(args=['cmd'])
+    @contextmanager
+    def start_process(self, cmd: str):
+        """Execute ``cmd`` on the target as a process.
+
+        The pseudo-terminal used by SSH merges stdout and stderr into a
+        single output stream.
+        """
```
ozan956: Could we expand this to make it clear that start_process() is PTY-backed, not pipe-backed?

Besides merging stdout/stderr, terminal processing can affect the byte stream (as I saw in the tests, the b"Hello World\r\n"). I think this is worth making explicit to callers.

labgrid/driver/sshdriver.py
```diff
@@ -22,6 +24,56 @@
...
+    def write(self, data):
+        self._sub.write(data)
+
+    @step(result=True)
+    def poll(self):
+        return self.exitcode
+
+    @step(result=True)
+    def stop(self):
+        self._sub.close(True)
```
ozan956: One thing I'm unsure about is this

this guarantees termination of the local ssh client, but do we want to guarantee that the remote command itself has terminated as well?

The PTY hangup should handle normal foreground commands, but commands which daemonize/ignore SIGHUP may survive. If remote termination isn't guaranteed, I think that should be documented.

PR #841, opened 2021-09-23
ID_SERIAL_SHORT documentation fix.

calucita:
Small fix for the ID_SERIAL_SHORT documentation.

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -74,7 +74,7 @@ This allows identification through hot-plugging or rebooting.
 
    USBSerialPort:
      match:
-       'ID_SERIAL_SHORT': 'P-00-00682'
+       ID_SERIAL_SHORT: 'P-00-00682'
      speed: 115200
 
 The example would search for a USB serial converter with the key
--- a/man/labgrid-suggest.rst
+++ b/man/labgrid-suggest.rst
@@ -51,7 +51,7 @@ in the USB-serial converter, the resource match configuration is:
 
    USBSerialPort:
      match:
-       '@ID_SERIAL_SHORT': 'P-00-00682'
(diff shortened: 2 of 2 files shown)


man/labgrid-suggest.rst
```diff
@@ -51,7 +51,7 @@ in the USB-serial converter, the resource match configuration is:
 
    USBSerialPort:
      match:
-       '@ID_SERIAL_SHORT': 'P-00-00682'
+       ID_SERIAL_SHORT: 'P-00-00682'
```
jluebbe, line 54: This was actually correct. `'@ID_SERIAL_SHORT'` means that his match must apply to a parent device.

doc/configuration.rst
```diff
@@ -74,7 +74,7 @@ This allows identification through hot-plugging or rebooting.
 
    USBSerialPort:
      match:
-       'ID_SERIAL_SHORT': 'P-00-00682'
+       ID_SERIAL_SHORT: 'P-00-00682'
```
jluebbe, line 77: This should be quoted and needs a `@` as well (see below).

jluebbe (changes requested):

doc/configuration.rst
calucita: On the examples I was working on the code would crash if it had the ''
I didn't try with the @ tho. 
However, there are code examples (commited by you :P) that do not use the quotes... 
https://github.com/labgrid-project/labgrid/blob/a6571a44501df93c649fafe315d75097d95d4941/examples/barebox/local-usb.yaml

My wider issue is that the pattern seems to be different on every file I checked, which made it hard for me to tell what the expected code was.

jluebbe: The output from the labgrid-suggest tool should be correct for your local system.

PR #843, opened 2021-09-23
config: allow types other than string in target/global options

Bastian-Krause:
**Description**
There is no reason to limit options to strings. YAML supports scalars, lists and dictionaries. As there is no way to write the options set during runtime back to YAML, we should not limit the types for the setter either.

Lift this limitation, so YAML-supported data types can be retrieved and all data types can be set.

Note that this changes previous behavior, because now, options are not converted to strings anymore.

Previous discussion about this topic happened in #490.

**Checklist**
- [x] PR has been tested


Bastian-Krause:
Maybe @tobsan has an opinion on this one?

tobsan:
> Maybe @tobsan (author of #490) has an opinion on this one?

While I'm not actively using labgrid anymore, I do remember this being mildly annoying, having to cast the value back to an int again (which I think was my use-case then). So I'm all for this!

Emantor (approved):

PR #849, opened 2021-10-06
doc/development: suggest tail -F rather than -f

Bastian-Krause:
**Description**
The console log file is created once the first line is received. Suggest tail's `-F` (meaning `-f --retry`) to keep trying to open the console log file if it is not yet accessible.


ejoerns (approved):
Nice! Did not know that `-F` exist but ran into the problem some times already ;) 
Tested and works like a charm :+1:

PR #870, opened 2021-11-21
ubootdriver: add configurable sleep before sending interrupt chars

rohieb:
Some U-Boots are configured not to print the "press any key to stop autoboot" message, but still allow interrupting the boot flow with key press. In those cases, another string must be matched (maybe even the U-Boot version string usually intended to be matched by `boot_expression`), which might be printed by U-Boot at a time when the interrupt prompt is not active yet. This PR adds a configurable sleep interval after matching the pattern before sending the interrupt to catch those use cases.

**Checklist**
- [x] PR has been tested

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,9 @@ New Features in 0.5.0
   considering only the closest marker.
 - The labgrid client SSH command is now able to instantiate the SSHDriver when
   there are multiple NetworkService resources available.
+- The UBootDriver acquired the new parameter ``preinterruptsleep`` which
+  configures the number of seconds to sleep after matching the ``autoboot``
+  prompt but before sending the ``interrupt`` characters.
 
 Bug fixes in 0.5.0
 ~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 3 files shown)


Emantor:
Why not modify the `autoboot` parameter of the UBootDriver instead? AFAICS settings this to the last string matched before interruption can be done should be enough.

rohieb:
As I wrote, the last string printed is matched too quickly, and the interrupt char is sent when the prompt is not active yet, so U-Boot happily goes into autoboot.

Emantor (approved):
Bah, this is terrible. But I guess we don't get around it.

jluebbe:
Basing this on timing and delays seems very fragile and hard to debug. I'd suggest fixing U-Boot to print the string a the correct place.

rohieb:
OK. Can you still cherry-pick the first commit? I don't want to force-push this branch as I've already referenced it elsewhere.

Emantor:
> OK. Can you still cherry-pick the first commit? I don't want to force-push this branch as I've already referenced it elsewhere.

Done, I pushed it to master.

PR #872, opened 2021-12-08
pyproject.toml: specify build-system requirements

S4mw1s3:
Using pyproject.toml to specify build system requirements is now the
preferred way. From the setuptools documentation[1]:

"This used to be accomplished with the setup_requires keyword but is now
considered deprecated in favor of the PEP 517 style described above."

I didn't remove the 'setup_requires' from setup.py because some users
who are using older versions of setuptools might still rely on it if
their version of setuptools has no support for pyproject.toml.

Without this pyproject.toml, the following command fails on a local
network without internet access:

[edit: correct command as explained below in the comments]
  `pip install --index-url <local index> -e .`

because setuptools does not take the index into account when fetching
the 'setup_requires' dependencies defined in setup.py.

However, with the correct pyproject.toml available, the build system
requirements are fetched first, using the correct index, and then later
at the point where setuptools tries to install the 'setup_requires'
dependencies, they are already installed so no need to fetch anything
anymore.

[1]: https://setuptools.pypa.io/en/latest/userguide/dependency_management.html

**Description**

**Checklist**
- [ ] Tests for the feature 

- [ ] PR has been tested

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,3 @@
+[build-system]
+# Minimum requirements for the build system to execute.
+requires = ["setuptools", "wheel", "setuptools_scm"]


Emantor (approved):

Bastian-Krause:
> pip install --index -e .

Where is the `--index` option documented?

S4mw1s3:
> Where is the --index option documented?

Eh well... I always thought it was a valid argument to pip install but I just learned that python by default allows argument abbreviations if the abbreviation is unambiguous: https://docs.python.org/3.10/library/argparse.html#argument-abbreviations-prefix-matching so in fact what I meant here is the '--index-url' option.

Also, bad copy/paste from my commit message because in there in fact I wrote:

`pip install --index <local index> -e .` but looks like github removed the `<local index>` part.

If you want I can update the commit message so it states --index-url.

Bastian-Krause:
@S4mw1s3 Okay, yes, please update the commit message accordingly. Thanks!

S4mw1s3:
> @S4mw1s3 Okay, yes, please update the commit message accordingly. Thanks!

Done. The commit message now clearly states
`pip install --index-url <local index> -e .`

Bastian-Krause:
Thanks!

Emantor (approved):

jluebbe (approved):

PR #875, opened 2021-12-14
driver/usbstoragedriver: Accept `dd` options as `write_image()` arguments

MaroMetelski:
Allow to set currently used arguments (conv, status, oflag, bs, skip,
seek) through kwargs instead of hardcoding the defaults. This allows
to set them to custom values while allowing to keep the defaults
for consistent and backwards-compatible behavior.
This also makes place for adding more arguments if needed.

Describe what your pull request does:
Adds a possiblity to set currently hardcoded dd arguments to USBMassStorage.write_image() function

Reasons behind the change:
When using other dd implementation (BusyBox) currently hardcoded arguments were not supported.
This change will allow to avoid such cases and also set custom arguments if needed.

**Checklist**
 Adjust function docstring
- [ ] Tests for the feature 

No need for update here

- [x] PR has been tested
Verify whether dd arguments are properly set by printing the list with different input arguments. Use feature to write_image to sd card with different arguments.

### DIFF
--- a/labgrid/driver/usbstoragedriver.py
+++ b/labgrid/driver/usbstoragedriver.py
@@ -55,7 +55,7 @@ class USBStorageDriver(Driver):
 
     @Driver.check_active
     @step(args=['filename'])
-    def write_image(self, filename=None, mode=Mode.DD, partition=None, skip=0, seek=0):
+    def write_image(self, filename=None, mode=Mode.DD, partition=None, **kwargs):
         """
         Writes the file specified by filename or if not specified by config image subkey to the
         bound USB storage root device or partition.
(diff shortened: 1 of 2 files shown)


MaroMetelski:
Added sign-off and changes.rst line

jluebbe:
Regarding your use-case (busybox's `dd`): this should not need to be handled at the user/test-case level. How would the user know which `dd` version is installed, especially when you have a lab with many exporters?

I'd argue that you'd probably want to use the full `dd`, as these options are there to ensure the data is full written and the page cache is not influenced by image writing.

MaroMetelski:
Hey, 
sorry for the the late response but here it is:

> Regarding your use-case (busybox's dd): this should not need to be handled at the user/test-case level. How would the user know which dd version is installed, especially when you have a lab with many exporters?

User still could want to set preferred `bs` or `status` for example, since he know he uses `Mode.DD`. I used it as a 'hack' to overcome busybox limitations, but thinking more about that, someone could want to remove progress output, as example.

> I'd argue that you'd probably want to use the full dd, as these options are there to ensure the data is full written and the page cache is not influenced by image writing.

Still busybox dd works just fine by removing `status=progress` and changing `conv=fdatasync` to `conv=fsync`. That case could be easily supported. This is the first and only time that we found some limitation while running labgrid plugin on such systems.

Concluding - I'd suggest to still make the arguments more flexible while running from plugin. I could work on that. That's not that necessary for me though, so can be as well closed for now too.

labgrid/driver/usbstoragedriver.py
```diff
@@ -93,21 +106,33 @@ def write_image(self, filename=None, mode=Mode.DD, partition=None, skip=0, seek=
...
                 f"if={remote_path}",
                 f"of={target}",
-                "oflag=direct",
-                "status=progress",
-                f"bs={block_size}",
-                f"skip={skip}",
-                f"seek={seek}",
-                "conv=fdatasync"
             ]
+            if skip is not None:
```
JoshuaWatt, line 122: These will never be `None`, why are you checking for it?

PR #890, opened 2022-02-11
add initial GitLab CI instructions

leiflm:
**Description**

The proposed changes allow users of GitLab CI infrastructure to mirror this repository and have labgrid images built and pushed to their local registry if they want to do that.

**Checklist**
- [ ] PR has been tested


leiflm:
The implementation follows the way (by using Kaniko) to create docker images that is [recommended by GitLab](https://docs.gitlab.com/ee/ci/docker/using_kaniko.html).

jluebbe:
It looks to me like this creates three independent jobs, which would cause the common stage (labgrid-base) to be built multiple times. The existing build.sh runs the stages sequentially in the same environment, so it's able to share the labgrid-base layer.

leiflm:
> It looks to me like this creates three independent jobs, which would cause the common stage (labgrid-base) to be built multiple times.

That is correct.

> The existing `build.sh` runs the stages sequentially in the same environment, so it's able to share the `labgrid-base` layer.

https://github.com/labgrid-project/labgrid/pull/890/commits/b8592a0ad7995b275b8abb28dd12b65c08d3764b should do the same.

PR #894, opened 2022-02-25
Add release-from

JoshuaWatt:
Adds coordinator and client API to release a place, but *only* if it is
acquired by a specific user in an atomic manner.

The intention is that this can be used by cleanup procedures to ensure
that there is not a race condition between figuring out which user has a
resource locked, and actually kicking that user. For example, some CI
systems may leave a place locked if they don't get a chance to cleanup
properly. A cleanup procedure might cross reference the user that has a
place locked with a list of active CI jobs, and if a user is found
that doesn't match any running CI job, the place it has acquired is
released. This presents a race however if this procedure is occuring in
multiple places simultaneously, as it's possible one of them releases
the place, then it gets acquired by an active CI job, then another
cleanup procedure releases it again.

This API prevents the above case because the cleanup procedure can
ensure that the place will *only* be released if it was acquired by the
specific user that it thinks should have it.

**Description**

**Checklist**
- [x] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -625,6 +625,17 @@ class ClientSession(ApplicationSession):
 
         print(f"released place {place.name}")
 
+    async def release_from(self):
+        """Release a place, but only if acquired by a specific user"""
+        place = self.get_place()
+        res = await self.call(
+            'org.labgrid.coordinator.release_place_from', place.name, self.args.acquired,
+        )
+        if not res:
+            raise ServerError(f"failed to release place {place.name}")
+
+        print(f"{self.args.acquired} has released place {place.name}")
+
     async def allow(self):
         """Allow another use access to a previously acquired place"""
         place = self.get_place()
@@ -1624,6 +1635,13 @@ def main():
     subparser.add_argument('-k', '--kick', action='store_true',
                            help="release a place even if it is acquired by a different user")
     subparser.set_defaults(func=ClientSession.release)
+
+    subparser = subparsers.add_parser('release-from',
+                                     help="atomically release a place, but only if locked by a specific user")
+    subparser.add_argument("acquired",
+                           metavar="HOST/USER",
+                           help="User and host to match against when releasing")
+    subparser.set_defaults(func=ClientSession.release_from)
 
     subparser = subparsers.add_parser('allow', help="allow another user to access a place")
     subparser.add_argument('user', help="<host>/<username>")
--- a/labgrid/remote/coordinator.py
+++ b/labgrid/remote/coordinator.py
@@ -195,6 +195,11 @@ class CoordinatorComponent(ApplicationSession):
             options=RegisterOptions(details_arg='details')
         )
         await self.register(
+            self.release_place_from,
+            'org.labgrid.coordinator.release_place_from',
+            options=RegisterOptions(details_arg='details')
+        )
+        await self.register(
             self.allow_place,
             'org.labgrid.coordinator.allow_place',
             options=RegisterOptions(details_arg='details')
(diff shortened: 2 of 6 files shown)


jluebbe (dismissed):
In the longer term, an integration with reservations might be better, but this a reasonable solution for now. @Emantor, what do you think?

labgrid/remote/coordinator.py
```diff
@@ -672,6 +677,27 @@ async def release_place(self, name, details=None):
...
+    @locked
+    async def release_place_from(self, name, acquired, details=None):
+        try:
+            place = self.places[name]
+        except KeyError:
+            return False
+        if not place.acquired:
+            return True
+        if place.acquired != acquired:
+            return True
```
Bastian-Krause, line 689: Could you explain why you return `True` in these cases?

JoshuaWatt: The logic is that the operation returns `True` if the supplied user no longer owns the place, and `False` if there was an error. The user could no longer own the place either because it's not acquired, another user has it acquired, or it was released for the requested user. Either way, the end result is the same, so `True` is returned. Put another way, returning `False` in these cases isn't helpful because we don't care if the place is acquired, or acquired by another user; as long as it's not acquired by the provided user when this function returns, we consider that success.

Bastian-Krause: Hm, that differs from the "normal" release logic, right? Releasing an already released place is considered an error.

Bastian-Krause: I don't have a strong opinion regarding this, I just want to make sure these things follow a consistent logic.

JoshuaWatt: That is correct.... however, based on the intention of this API, returning False in that case makes the return value useless for scripting purposes; you would basically have to ignore the return because it's impossible to predict if it will return True or False, and impossible to know if False is an error, or just a race releasing the place.

Bastian-Krause: Yes, agreed. I'm okay with this then.

jluebbe: Perhaps a comment here describing the reasoning would be useful to avoid confusion in the future. Or even better a short section in the docs?

Emantor:
> In the longer term, an integration with reservations might be better, but this a reasonable solution for now. @Emantor, what do you think?

Even if we integrate with reservations, how do you want to do that? Do we want to allow the release-from mechanic only for places which were acquired using a reservation? Or do your want to integrate the release-from mechanic into the reservation system, i.e. you have to request a reservation to release a place acquired by a specific user?

jluebbe (approved):

jluebbe:
> > In the longer term, an integration with reservations might be better, but this a reasonable solution for now. @Emantor, what do you think?
> 
> Even if we integrate with reservations, how do you want to do that? Do we want to allow the release-from mechanic only for places which were acquired using a reservation? Or do your want to integrate the release-from mechanic into the reservation system, i.e. you have to request a reservation to release a place acquired by a specific user?

We @JoshuaWatt and myself talked about that in the chat (https://matrix.to/#/!NqKpzIiEICmrNQppEy:matrix.org/$164495425918645Rupme:matrix.org). We could either use the LG_TOKEN to find the reservation with the crashed CI job or add a way to add some additional K/V-Pairs to a reservation. Perhaps add a simple (HTTP-JSON?) API to get/modifiy the current reservations from the coordinator. That should be possible with crossbar, if i remember correctly.

labgrid/remote/client.py
```diff
@@ -1625,6 +1636,13 @@ def main():
                            help="release a place even if it is acquired by a different user")
     subparser.set_defaults(func=ClientSession.release)
 
+    subparser = subparsers.add_parser('release-from',
```
Emantor, line 1639: I am not in favor of adding another top level command, I think adding a `--from` parameter to the release command is clearer and does not clutter the interface as much.

jluebbe: As above, the use case and behavior (especially error reporting) is different, so it's find it confusing to have it as a flag to relase.

Emantor (approved):

PR #901, opened 2022-03-11
remote: add -f,--from for release command

Emantor:
Adds coordinator and client API to release a place, but *only* if it is
acquired by a specific user in an atomic manner.

The intention is that this can be used by cleanup procedures to ensure
that there is not a race condition between figuring out which user has a
resource locked, and actually kicking that user. For example, some CI
systems may leave a place locked if they don't get a chance to cleanup
properly. A cleanup procedure might cross reference the user that has a
place locked with a list of active CI jobs, and if a user is found
that doesn't match any running CI job, the place it has acquired is
released. This presents a race however if this procedure is occuring in
multiple places simultaneously, as it's possible one of them releases
the place, then it gets acquired by an active CI job, then another
cleanup procedure releases it again.

This API prevents the above case because the cleanup procedure can
ensure that the place will *only* be released if it was acquired by the
specific user that it thinks should have it.

Alternative to #894


labgrid/remote/client.py
```diff
@@ -613,18 +613,29 @@ async def release(self):
...
+        """Release a place, but only if acquired by a specific user"""
+        place = self.get_place()
+        res = await self.call(
+            'org.labgrid.coordinator.release_place_from', place.name, self.args.acquired,
+        )
+        if not res:
+            raise ServerError(f"failed to release place {place.name}")
+
+        print(f"{self.args.acquired} has released place {place.name}")
+
```
jluebbe, line 638: Why is this still needed?

labgrid/remote/coordinator.py
```diff
@@ -653,12 +653,19 @@ async def acquire_place(self, name, details=None):
         return True
 
     @locked
-    async def release_place(self, name, details=None):
+    async def release_place(self, name, acquired, details=None):
```
jluebbe, line 656: Does this break compatibility with older clients?

labgrid/remote/client.py
Emantor: Its not, left over and forgot to remove.

labgrid/remote/coordinator.py
Emantor: Yes, but we can catch this by assigning `acquired=None`.

jluebbe: Does that fix the other combination as well?

jluebbe:
I'm skeptical if reusing the existing API is useful, as the the use-cases differ significantly and especially the error handling is completely different. A separate call would make that explicit and avoid any possible compatibility issues.

labgrid/remote/coordinator.py
Emantor: which other combination?

Emantor:
Superseded by https://github.com/labgrid-project/labgrid/pull/894.

PR #902, opened 2022-03-11
modbus: add support for Papouch Quido ETH I/O modules

ynezz:
**Description**

[Quido ETH](https://en.papouch.com/io-modules/quido/ethernet/) is a family of I/O modules with up to 32 digital outputs (relays) and ethernet interface which makes them quite nice devices for DUT power control.

image

In order to control the state of the relay outputs, firmware needs to be instructed using Modbus function code 15, multiple coils write method as single coil write method is not supported, only single coil state reading is. Write using single method ends with `Could not write coil (code=7/exception=1)` error.

So in order to support those modules, new `write_multiple` boolean config option has been added to `ModbusTCPCoil` resource. When enabled, the Modbus driver is going to use `write_multiple_coils()` method instead of defaul `write_single_coil()`.

Example usage tested with Quido ETH 8/8 (coils 0-7) module:

```yaml
    resources:
      ModbusTCPCoil:
        host: 192.168.1.254
        coil: 7
        write_multiple: True
    drivers:
      ModbusCoilDriver: {}
      DigitalOutputPowerDriver: {}
```

**Checklist**
- [ ] Tests for the feature 
- [x] PR has been tested

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -362,6 +362,8 @@ Arguments:
   - coil (int): index of the coil e.g. 3
   - invert (bool, default=False): whether the logic level is inverted
     (active-low)
+  - write_multiple_coils (bool, default=False): whether to perform write
+    using "write multiple coils" method instead of "write single coil"
 
 Used by:
   - `ModbusCoilDriver`_
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,6 +14,8 @@ New Features in 0.5.0
 - The client and coordinator learned of a new "release-from" operation that only releases a place
   if it acquired by a specific user. This can be used to prevent race conditions when attempting to
   automate the cleanup of unused places (e.g. in CI jobs)
+- ModbusTCPCoil driver supports writing using multiple coils write method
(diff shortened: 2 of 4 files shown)


doc/configuration.rst
```diff
@@ -363,6 +363,8 @@ Arguments:
   - coil (int): index of the coil e.g. 3
   - invert (bool, default=False): whether the logic level is inverted
     (active-low)
+  - write_multiple (bool, default=False): optional, whether perform write
+    using multiple coils method
```
Bastian-Krause, line 367:
```suggestion
  - write_multiple (bool, default=False): whether to perform write
    using "write multiple coils" method instead of "write single coil"
```

Bastian-Krause: Maybe rename the argument `write_multiple` -> `write_multiple_coils`? I think that makes it even more clear.

Bastian-Krause (approved):
Thanks! Looks good to me. The docker issue is unrelated (see #904).

Bastian-Krause:
Rebased on current master.

Bastian-Krause (approved):

PR #903, opened 2022-03-16
Minor Configuration Documentation Fixes/Improvements

Bastian-Krause:
**Description**
Collection of fixes for the configuration section of the documentation fixing/improving the resource/driver arguments (drop inexistent argument, fix wrong types, use generic default notation). Stumbled upon those while looking up arguments for a resource.


Bastian-Krause:
Docker issue should be fixed with #904.

jluebbe (approved):

PR #905, opened 2022-03-27
Update XMODEM command handling in ShellDriver

### LINKED ISSUE #405: Xmodem executables autodetection fails
commit 7521f159250 added XMODEM support for ShellDriver.
Some systems (checked for Debian, Arch Linux, FreeBSD) rename the original lrzsz executable file names in their packages (removing the leading l) so that lrz and lsz are not available. rx is then wrongly detected as Busybox's rx and called with wrong arguments.

Checking for [l]sz respectively [l]rz and then for rx as Busybox version should do the trick for most systems.

cyanidium:
**Description**
There is a bug in the way ShellDriver detects XMODEM for get_bytes and put_bytes that leads to strange errors if the required shell commands are missing. This is because the tests run the 'which' command and test the stdout, but stdout may include error messages such as "<command> not found" which allow the tests to pass. The tests have been changed to rely on the return code instead which is far more consistent.

This PR started as a fix for just that issue, but then I saw #405 that overlaps and added the changes. Given the overlap they can't be split into different commits.

Tested on my local environment and detecting both presence and absence of all four non-BusyBox shell commands. It detects rx as well, but it isn't the BusyBox version, so I'm just assuming that it was working and left it the same way it was before.

**Checklist**
- [ ] Tests for the feature 
- [X] PR has been tested

Fixes #405


labgrid/driver/shelldriver.py
```diff
@@ -318,29 +318,35 @@ def _start_xmodem_transfer(self, cmd):
...
     def _get_xmodem_sx_cmd(self, filename):
         """ Detect which XMODEM send command can be used on the target, and cache the result. """
         if not self._xmodem_cached_sx_cmd:
-            if self._run('which lsz')[0]:
+            if self._run('which lsz')[2] == 0:
                 # redirect stderr to prevent lsz from printing "Give XMODEM receive
                 # cmd now", which will confuse the XMODEM instance
                 self._xmodem_cached_sx_cmd = "lsz -b -X -m 1200 -M 10 '{filename}' 2>/dev/null"
+            elif self._run('which sz')[2] == 0:
+                # ranamed binaries packaged by some distros
```
Bastian-Krause, line 346:
```suggestion
                # renamed binaries packaged by some distros
```

Bastian-Krause (commented):
Please rebase on current master to fix the docker test.

cyanidium:
Rebased and typo suggestions merged, ready to test again

Bastian-Krause:
See #906 regarding the CI failure.

Emantor (approved):

Emantor (approved):
Rebased on master, lgtm.

PR #909, opened 2022-04-15
Add support for more webcams and a generic fallback pipeline

dwagenk:
This
- adds support for two additional webcam models
- fixes up the USBVideoDevice resource to export the correct video device for webcams that have multiple device nodes
- adds a generic fallback pipeline so many webcams are usable without having to adapt labgrid. The parameters are what I found in common in the `gst-device-monitor-1.0` output for both added webcams and my laptops internal webcam.
- a commit that i deem optional. It produces a verbose warning when using the fallback pipeline. I dislike the string indentation mess it leaves in the source code, but the error message looks as intended:
    ```
  WARNING: Unkown USB video device 1224:2825, using fallback pipeline. This might not
             work as expected.
             Consider adding support for your camera model to labgrid as described here:
             https://labgrid.readthedocs.io/en/latest/configuration.html#usbvideodriver
    ```
    This commit should either be dropped or squashed into the preciding one 

All changes have been tested locally.
- Before the [`resource/udev: filter USBVideoDevice to skip metadata only devices`](https://github.com/labgrid-project/labgrid/commit/b2a5436637079c2bf9a3686aaca1fe08512054bb) change the pipeline occasionally failed to start due to the wrong `/dev/videoX` being used. This has not happend since.
- The fallback pipeline works fine with the webcams I have available. Tested by removing the support for those models again. 

**Checklist**
- [ ] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/labgrid/driver/usbvideodriver.py
+++ b/labgrid/driver/usbvideodriver.py
@@ -1,4 +1,5 @@
 # pylint: disable=no-member
+import logging
 import subprocess
 
 import attr
@@ -15,6 +16,11 @@ class USBVideoDriver(Driver, VideoProtocol):
     bindings = {
         "video": {"USBVideo", "NetworkUSBVideo"},
     }
+
+    def __attrs_post_init__(self):
+        super().__attrs_post_init__()
+        self.logger = logging.getLogger(f"{self}")
+        self._prepared = False
 
     def get_qualities(self):
         match = (self.video.vendor_id, self.video.model_id)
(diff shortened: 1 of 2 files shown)


Emantor:
IMO we can drop the optional commit for the fallback pipeline, the rest looks good to me. Especially the fix to always select the correct `/dev/videoX` device for capturing.

labgrid/driver/usbvideodriver.py
```diff
@@ -46,7 +67,18 @@ def get_qualities(self):
                 ("mid", "video/x-h264,width=1280,height=720,framerate=25/1"),
                 ("high", "video/x-h264,width=1920,height=1080,framerate=25/1"),
                 ])
-        raise InvalidConfigError("Unknown USB video device {:04x}:{:04x}".format(*match))
+        else:
```
Bastian-Krause, line 70: My linter says: Unnecessary "else" after "return", remove the "else" and de-indent the code inside it.

Emantor:
I force pushed a version which removes the OPTIONAL warning and the `else`.

Emantor (approved):

Bastian-Krause (approved):

PR #912, opened 2022-05-04
Add max_size option to ConsoleExpectMixin read

marc-roeoesli:
Extend read function of serialdriver and externalconsoledriver with 'max_size' argument that limits the number of read bytes to 'max_size'. Set 'size' = 'max_size' to read exactly 'size' bytes. The values 'None' and '0' result in an unlimited read.
Extend 'test_read' in test_serialdriver.py and test_externalconsoledriver.py  to check 'max_size' argument.
Implement 'max_size' explicitely in consoleexpectmixin.

- We use the feature to read a fixed number of bytes on the serial port during testing the serial communication. The current read function returns the full buffer of the serial port (maximum of size and in_waiting). We use labgrid to test embedded software on hardware devices.
- We think that for certain tests it can be useful to read a well defined and fixed size from the serial port. We hope that this could also be useful for other developers. We are open to discuss other possibilities/suggestions.
- We tested the feature by extending the existing test for the serial driver and external console driver 'read'-function.

**Checklist**
- [x] Tests for the feature
- [x] PR has been tested

Signed-off-by: Marc Roeoesli <marc.roeoesli@mt.com>

### DIFF
--- a/labgrid/driver/serialdriver.py
+++ b/labgrid/driver/serialdriver.py
@@ -83,14 +83,19 @@ class SerialDriver(ConsoleExpectMixin, Driver, ConsoleProtocol):
             vars["protocol"] = self.port.protocol
         return vars
 
-    def _read(self, size: int = 1, timeout: float = 0.0):
+    def _read(self, size: int = 1, timeout: float = 0.0, max_size: int = None):
         """
         Reads 'size' or more bytes from the serialport
 
         Keyword Arguments:
         size -- amount of bytes to read, defaults to 1
+        max_size -- maximal amount of bytes to read, values 'None' or '0' do not restrict the read
+                    length, defaults to None
+        if size == max_size: read and return exactly size = max_size bytes
         """
         reading = max(size, self.serial.in_waiting)
+        if max_size:  # limit reading to max_size if provided
+            reading = min(reading, max_size)
         self.serial.timeout = timeout
         res = self.serial.read(reading)
         if not res:
--- a/tests/test_serialdriver.py
+++ b/tests/test_serialdriver.py
@@ -23,14 +23,18 @@ class TestSerialDriver:
         serial_mock.return_value.open.assert_called_once_with()
         serial_mock.return_value.write.assert_called_once_with(b"testdata")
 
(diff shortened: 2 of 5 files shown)


tests/test_serialdriver.py
```diff
@@ -23,14 +23,20 @@ def test_write(self, target, serial_port, mocker):
...
         serial_mock = mocker.patch('serial.Serial')
-        serial_mock.return_value.in_waiting = 0
+        serial_mock.return_value.in_waiting = param[1]
         s = SerialDriver(target, "serial")
         target.activate(s)
-        s.read()
+        s.read(size=param[0], read_block=param[2])
         serial_mock.return_value.open.assert_called_once_with()
-        serial_mock.return_value.read.assert_called_once_with(1)
+        if not param[2]:
```
AndreHochuliMT, line 36: I would split this test into two tests, so this if/else is not needed.

marc-roeoesli: I refined the testing such that no if/else condition is needed.
The test now asserts the number of read function calls and the returned values.

Emantor:
IMO the RFC from the commit message can be removed.

labgrid/driver/serialdriver.py
```diff
@@ -83,14 +83,19 @@ def get_export_vars(self):
...
         return vars
 
-    def _read(self, size: int = 1, timeout: float = 0.0):
+    def _read(self, size: int = 1, timeout: float = 0.0, max_size: int = None):
         """
         Reads 'size' or more bytes from the serialport
 
         Keyword Arguments:
         size -- amount of bytes to read, defaults to 1
+        max_size -- maximal amount of bytes to read, values 'None' or '0' do not restrict the read length,
```
Bastian-Krause, line 92: nitpick: this line shouldn't be longer than 100 characters.

marc-roeoesli: fixed this line with a line break, and another line in test_serialdriver.py that was too long

jluebbe:
I've squashed the commits together and made minor changes to the message. I've also removed the RFC tag.

jluebbe (approved):

PR #913, opened 2022-05-04
remote/exporter: enable multiple ser2net connections

khilman:
ser2net allows multiple, multiplexed connections.  Enable by default
with a max of 10.

RFC: this should probably be an optional feature, with the number a parameter, but this is just to show how it can be done.

Signed-off-by: Kevin Hilman <khilman@baylibre.com>


Emantor:
I don't think this needs to be optional. Since we use a random port and limit interactions to the acquired user or other users which are explicitly allowed, I don't think there should be too many problems.

khilman:
> I don't think this needs to be optional. Since we use a random port and limit interactions to the acquired user or other users which are explicitly allowed, I don't think there should be too many problems.

@Emantor so do you want me to re-submit without the RFC comments?

Emantor:
Yeah, pretty much. We leave the old ser2net versions behind this way, but we'll remove that when most of the distributions have switched over.

khilman:
OK, rebased onto current master, and removed the RFC.

Emantor (approved):

PR #914, opened 2022-05-13
remote/client: configure txaio earlier

jluebbe:
If twisted is installed, txaio must be configured before importing autobahn.
Otherwise we get an error:

  RuntimeError: To use txaio, you must first select a framework with
  .use_twisted() or .use_asyncio()


khilman (approved):
Tested-by: Kevin Hilman <khilman@baylibre.com>

jluebbe:
Added @khilman's Tested-by.

Emantor (approved):

Emantor:
Thanks!

PR #915, opened 2022-05-13
remote/exporter: support /usr/local/sbin/ser2net

khilman:
Locally modified/compiled installs of ser2net could be in
/usr/local, allow that as a fallback when not found in
the other locations.

Signed-off-by: Kevin Hilman <khilman@baylibre.com>

### DIFF
--- a/labgrid/remote/exporter.py
+++ b/labgrid/remote/exporter.py
@@ -196,6 +196,8 @@ class SerialPortExport(ResourceExport):
         if self.ser2net_bin is None:
             if os.path.isfile("/usr/sbin/ser2net"):
                 self.ser2net_bin = "/usr/sbin/ser2net"
+            elif os.path.isfile("/usr/local/sbin/ser2net"):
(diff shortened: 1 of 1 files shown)


Emantor:
Since this is just a fallback (which is already somewhat distribution specific), I think the correct way to solve this is to setup the correct `PATH` for the user/service calling the exporter and including `/usr/local/bin/` there, since we try to resolve the `PATH` via `which` first.

jluebbe:
@khilman would using $PATH work for you?

khilman:
Using `$PATH` does work, but it's not great IMO.  The problem with `$PATH` is that ser2net installs to `/usr/local/sbin` which is not typically in the path of a normal user.   But I understand why you want to rely on `$PATH` so it's OK for me.  It really only affects me until the ser2net features I'm using get pulled into distro versions of ser2net.

Bastian-Krause:
> It really only affects me until the ser2net features I'm using get pulled into distro versions of ser2net.

I'm assuming this happened in the meantime. So let's close this.

PR #918, opened 2022-05-19
mqtt: Make availability payload more generic

obbardc:
The mqtt availability payload is really device-specific; Tasmota uses `Online` and `Offline`, whereas ESPHome uses `online` and `offline`. Let's make MQTTManager a little bit more device-agnostic and allow both cases to satisfy the criteria.

Signed-off-by: Christopher Obbard <chris.obbard@collabora.com>

---

This is the minimum change required to support ESPhome firmware as a `TasmotaPowerPort`. I think it makes sense to change the availability payload detection where I have done it.

PS: Let me know if also like to have a separate driver called `ESPHomePowerPort` or similar, I can do a separate commit on top

### DIFF
--- a/labgrid/resource/mqtt.py
+++ b/labgrid/resource/mqtt.py
@@ -37,10 +37,10 @@ class MQTTManager(ResourceManager):
     def _on_message(self, client, userdata, msg):
         payload = msg.payload.decode('utf-8')
         topic = msg.topic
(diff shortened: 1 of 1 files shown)


Emantor (approved):

Emantor:
Thanks!

PR #920, opened 2022-05-19
Fix Udev Attribute matching

NiklasReisser:
**Description**

Udev matching by ATTR(S){} does currently not work on my machine.

I found the issue because I have a 4-port usb-to-serial device where I can only determine the ports by looking at the attributes. 
This is also the device I used for testing locally.

After some search I found that pyudev internally stores attributes as bytes, not strings.
To compare them with the expectation when doing udev matching, they need to be encoded.
Luckily, pyudev already has a function for that, just without a default value. 

**Checklist**
- [ ] Tests for the feature 

- [x] PR has been tested


Emantor (approved):

Emantor:
Thanks!

PR #921, opened 2022-05-23
smallubootdriver: allow sending of boot_secret without new line

ynezz:
**Description**

By default SmallUBootDriver sends `boot_secret` terminated with new
line, which works fine on multitude of devices, but fails for example on
quite popular ralink/mt7621 based Ubiquiti ER-X devices, where
bootloader offers following boot menu:

 Please choose the operation:
  1: Load system code to SDRAM via TFTP.
  2: Load system code then write to Flash via TFTP.
  3: Boot system code via Flash (default).
  4: Entr boot command line interface.
  7: Load Boot Loader code then write to Flash via Serial.
  9: Load Boot Loader code then write to Flash via TFTP.
  r: Start TFTP recovery.

and expects only single character choice, otherwise uses default choice
3, but in order to control such DUTs we need to enter boot prompt by
selecting choice 4. and thus sending `4` verbatim, without new line.

So allow usage of SmallUBootDriver on such devices by introducing
`boot_secret_nolf` boolean config option, which when set to `True` sends
`boot_secret` without new line character.

**Checklist**
- [ ] Tests for the feature 
- [x] PR has been tested

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,9 @@ New Features in 0.5.0
   automate the cleanup of unused places (e.g. in CI jobs)
 - ModbusTCPCoil driver supports writing using multiple coils write method
   in order to make driver usable with Papouch Quido I/O modules
+- SmallUBootDriver driver now supports wide range of Ralink/mt7621 devices
+  which expects ``boot_secret`` without new line with new ``boot_secret_nolf``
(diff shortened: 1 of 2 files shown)


Emantor (approved):

Emantor:
Thanks!

PR #925, opened 2022-05-27
sshdriver: Add input argument to run

ColinMcAllister:
**Description**
Added ability to optionally pipe a string or bytes object to stdin for the command ran by labgrid.driver.sshdriver.run.

**Checklist**
- [ ] Tests for the feature 

- [ ] PR has been tested


ColinMcAllister:
As currently implemented, input only works for run and not run_check.  I see two solutions and am not sure how to proceed.  Should cmd_input be supported by all classes that inherit CommandProtocol or should run_check just be overridden for the ssh_driver?

Emantor:
Isn't this also solved by #835 where a handle is given back for the process and you can call `read/write` on the handle?

ColinMcAllister:
#835 does seem like it should work, but based on a test I ran with the changes from #835, the test hangs when I try to write to the process handle.
I did find that changing pexpect.spawn to pexpect.popen_spawn.PopenSpawn() solved the issue for me, where write would not block and the test behaved like I expected.

Bastian-Krause:
> #835 does seem like it should work, but based on a test I ran with the changes from #835, the test hangs when I try to write to the process handle. I did find that changing pexpect.spawn to pexpect.popen_spawn.PopenSpawn() solved the issue for me, where write would not block and the test behaved like I expected.

Could you document what you are doing exactly in #835, so we can improve that?

PR #891, opened 2022-02-15
Bump Requirements to Latest Releases

Bastian-Krause:
**Description**
- bump requirements to recent versions (vmprof tip of master)
- update system dependencies
- fix pytest deprecation warnings caused by the new pytest version
- set crossbar's "user_timeout" in default config

**Known Issues**
- vmprof not pinned to a release, see: https://github.com/vmprof/vmprof-python/issues/240

**Checklist**
- [ ] CHANGES.rst has been updated
- [x] PR has been tested

PR #931, opened 2022-06-14
remote/coordinator: switch to using the custom yaml loader/dumper

benjamb:
**Description**

Using `yaml.load()` without specifying a loader has been deprecated since 5.1. As such this series changes `remote/coordinator` to make use of the custom yaml loader/dumper within `util/yaml`, which ultimately results in use of the SafeLoader and the SafeDumper. Previously however the coordinator will have `yaml.dump`ed tuples with the `python/tuple` tag, which the SafeLoader will not load, in order to not break any existing deployments on upgrade this series also adds the tuple constructor to the custom loader.

Additionally does a little miscellaneous tidying within `util/yaml`, including changing `dump()` to take kwargs.

I've tested this series against a local deployment of a labgrid coordinator and exporter.

**Checklist**

- [x] PR has been tested

### DIFF
--- a/labgrid/remote/coordinator.py
+++ b/labgrid/remote/coordinator.py
@@ -9,14 +9,13 @@ from enum import Enum
 from functools import wraps
 
 import attr
-import yaml
 from autobahn import wamp
 from autobahn.asyncio.wamp import ApplicationRunner, ApplicationSession
 from autobahn.wamp.types import RegisterOptions
 
 from .common import *
 from .scheduler import TagSet, schedule
-from ..util import atomic_replace
+from ..util import atomic_replace, yaml
 
 
 class Action(Enum):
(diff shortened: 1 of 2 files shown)


Bastian-Krause (approved):
Thanks for your PR! I suggested a similar patch in #891, not considering change of the format. So this one is definitely better. Haven't tested this yet.

benjamb:
@Bastian-Krause Ah, apologies for the duplication.

benjamb:
Anything I need to do further to get this merged? I wasn't sure whether adding a section to `CHANGES.rst` made much sense?

Emantor:
> Anything I need to do further to get this merged? I wasn't sure whether adding a section to `CHANGES.rst` made much sense?

No, just me returning from vacation :grin: Thanks!

PR #935, opened 2022-06-15
remote/client: no block in async console, loop even on unavailable resource

### LINKED ISSUE #926: labgrid-client console --loop does not loop
If the underlying `USBSerialPort` is not availble, the `--loop` option doesn't wait or loop it fails, for example:
```
labgrid-client -p am335x-icev2 console --loop
labgrid-client: error: Not all resources are available: {NetworkSerialPort(target=Target(name='am335x-icev2', env=None), name='USBSerialPort', state=<BindingState.bound: 1>, avail=False, host='nuc-1', port=45243, speed=115200, protocol='rfc2217')}
```

Bastian-Krause:
**Description**
`labgrid-client console --loop` should keep trying to connect to the console even if the serial port is currently unvailable.
The console method is declared async because it needs to handle crossbar communication with the coordinator concurrently, see [1] (e.g. place kicks, resouces becoming available).

Until now it only tries again if microcom comes back with a return code != 0. But in case the resource is unavailable, labgrid bails out early because `target.get_resource()` raises a `NoResourceFoundError`.

Calling `target.get_resource()` waits until the resources are available by default. This is done in a blocking `poll()`/`time.sleep()` loop until a timeout is reached allowing no crossbar communication in the meantime. This again means changes in resource availability we wait for cannot arrive. So there is no point in waiting for them at all.

Fix that by awaiting the `NetworkSerialPort` resource while allowing other tasks to run when `--loop` is given. Afterwards "wait" for the `NetworkSerialPort` for 0 seconds. This either succeeds or a `NoResourceFoundError` is raised, both happens instantly. To make this work, instant `Timeout(0.0)` is now allowed. 

[1] 19757a18 ("remote/client: close console connection on place release")

**Checklist**
- [ ] Tests for the feature 
- [x] PR has been tested

Fixes #926

### DIFF
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -31,7 +31,7 @@ from ..util.yaml import dump
 from .. import Target, target_factory
 from ..util.proxy import proxymanager
 from ..util.helper import processwrapper
-from ..util import atomic_replace
+from ..util import atomic_replace, Timeout
 from ..driver import Mode
 
 txaio.config.loop = asyncio.get_event_loop()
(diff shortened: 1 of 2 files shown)


khilman:
Tested-by: Kevin Hilman <khilman@baylibre.com>

Bastian-Krause:
We've stumbled upon blocking code called from async code. So let's not merge this yet, I'll come back with another version later.

Bastian-Krause:
Fixed `s/labgrid/remote/` in the commit's subject line.

Emantor (approved):

khilman:
I tested this, and it works OK the first time (e.g. if board is powered of when `console --loop` is run, and then powered on after `console --loop` has been started.)

However, if the board is then powered off, then `console --loop` exits with
```
Got EOF from port
connection lost
labgrid-client: error: Not all resources are available: {NetworkSerialPort(target=Target(name='am335x-icev2', env=None), name='USBSerialPort', state=<BindingState.bound: 1>, avail=False, host='nuc-1', port=None, speed=115200, protocol='rfc2217')}
```
Ideally, `console --loop` should continue to loop waiting for the serial port to appear again instead of exit.

PR #944, opened 2022-06-25
ubootdriver: allow handling of console output decoding errors 

ynezz:
**Description**

ubootdriver: allow handling of console output decoding errors

Some buggy vendor bootloaders can output non ASCII characters to the
serial console:

```shell
 00001e50  74 68 20 6e 61 6d 65 20  5b 58 38 36 5f 36 34 20  |th name [X86_64 |
 00001e60  4c 45 44 45 20 4c 69 6e  75 78 2d 73 73 68 5f 5f  |LEDE Linux-ssh__|
 00001e70  5f 67 69 74 5f 64 65 73  69 1f 8b 08 5d 0d 0d 0a  |_git_desi...]...|
                                       |
                                       `-- printf buffer overflow starts here
```

Which leads to the following decoding issue as the default error handler
is set to `strict`:

```python
    '', before.decode('utf-8'), count=1000000
 UnicodeDecodeError: 'utf-8' codec can't decode byte 0x8b in position 6453: invalid start byte
```

So fix it by allowing overriding of console output decoding errors via
new `console_decodeerrors` config option, so one could use for example
`backslashreplace` decoding error handler.

**Checklist**

- [x] PR has been tested

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -26,6 +26,8 @@ New Features in 0.5.0
 - `DFUDriver` has been added to communicate with a `DFUDevice`, a device in DFU
   (Device Firmware Upgrade) mode.
 - ``labgrid-client dfu`` added to allow communcation with devices in DFU mode.
+- UBootDriver and SmallUBootDriver now allows overriding of console output
+  decoding errors via new ``console_decodeerrors`` config option.
 
 Bug fixes in 0.5.0
 ~~~~~~~~~~~~~~~~~~
--- a/labgrid/driver/smallubootdriver.py
+++ b/labgrid/driver/smallubootdriver.py
@@ -68,7 +68,7 @@ class SmallUBootDriver(UBootDriver):
         # wait until UBoot has reached it's prompt
         self.console.expect(self.prompt)
         for command in self.init_commands:  #pylint: disable=not-an-iterable
-            self._run(command)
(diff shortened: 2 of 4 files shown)


Emantor (approved):

Bastian-Krause (changes requested):
In #574 I tried something similar (among other things), but I lost the context there.

I think it is surprising that you can now set `console_decodeerrors=backslashreplace` on the (Small)UBootDriver and then call `run_check("foo")` which still uses `strict` instead of `backslashreplace` because of the default kwarg in `CommandMixin.run_check()`.

Why not call `console_decodeerrors` simply `decodeerrors` and really use that where `decodeerrors` was not set explicitly? This would mean we need to change the default of the kwarg in different locations to `None` and use `strict` later if `None` is given.

We should also think about the `codec`/`decodeerrors` behavior globally:
- Should this be configurable on each driver implementing the `CommandProtocol`?
- What about the drivers that take `codec`/`decodeerrors` in their `run()` method? (essentially breaking the `CommandProtocol`, but maybe someone is using that already?)

I never understood why some of the public `run()`/`run_check()` methods take these kwargs and others do not.

Bastian-Krause:
@ynezz Is this something you're still interested in?

ynezz:
Yes, I'm facing those issues from time to time, hopefully soon (tm), sorry.

PR #945, opened 2022-06-25
shelldriver: fixes and improvements

ynezz:
**Description**

 * shelldriver: make run wrapper injecting more reliable
 * shelldriver: make kernel silencing more reliable
 * shelldriver: fix missing kernel silencing on password less consoles
 * shelldriver: make post_login_settle_time generally available

**Checklist**
- [x] PR has been tested


labgrid/driver/shelldriver.py
```diff
@@ -180,9 +180,10 @@ def _await_login(self):
             if timeout.expired:
                 raise TIMEOUT(f"Timeout of {self.login_timeout} seconds exceeded during waiting for login")  # pylint: disable=line-too-long
 
+        if self.post_login_settle_time > 0:
+            self.console.settle(self.post_login_settle_time, timeout=timeout.remaining)
```
JoshuaWatt, line 213: Similar thing here: I think we really want the delay once per activation, but the `self.status` foils that.

labgrid/driver/shelldriver.py
```diff
@@ -139,9 +168,9 @@ def _await_login(self):
             )
 
             if index == 0:
-                if did_login and not did_silence_kernel:
+                if not did_silence_kernel:
```
JoshuaWatt, line 171: The `did_login` was an attempt to ensure that this only ran once per activation, otherwise it runs every time, which isn't desirable. I _think_ this might actually be a bug in that `self.status` on line 178 below should be `self._status` so that the code in `on_activate` will only run once per activation.

Perhaps one of the Labgrid devs can clarify

JoshuaWatt (changes requested):

labgrid/driver/shelldriver.py
ynezz: > The `did_login` was an attempt to ensure that this only ran once per activation

Shouldn't that `did_silence_kernel` be enough for that purpose? I never hit that `did_login=True` code path as there is no `login_prompt` on my system, so this is actually a regression for me, introduced in 40154266ba3.

See it yourself, with the following changes applied:

```diff
diff --git a/labgrid/driver/shelldriver.py b/labgrid/driver/shelldriver.py
index aeb3e54..7b54190 100644
--- a/labgrid/driver/shelldriver.py
+++ b/labgrid/driver/shelldriver.py
@@ -166,21 +166,25 @@ class ShellDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
                 expectations,
                 timeout=self.await_login_timeout
             )
+            self.logger.info(f"shelldriver: while loop, index={index}")
 
             if index == 0:
                 if not did_silence_kernel:
                     # Silence the kernel and wait for another prompt
+                    self.logger.info("shelldriver: silencing kernel")
                     self._silence_kernel(timeout)
                     did_silence_kernel = True
                 else:
                     # we got a prompt. no need for any further action to
                     # activate this driver.
+                    self.logger.info("shelldriver: got a prompt, self.status=1")
                     self.status = 1
                     break
 
             elif index == 1:
                 # we need to login
                 self.console.sendline(self.username)
+                self.logger.info("shelldriver: did login, did_login=True")
                 did_login = True
 
             elif index == 2:
@@ -190,8 +194,10 @@ class ShellDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
                     raise Exception("Password entry needed but no password set")
 
             elif index == 3:
+                self.logger.info("shelldriver: expect hit a timeout")
                 # expect hit a timeout while waiting for a match
                 if before == last_before:
+                    self.logger.info("shelldriver: expect did not received anything, sending empty line")
                     # we did not receive anything during
                     # self.await_login_timeout.
                     # let's assume the target is idle and we can safely issue a
@@ -199,6 +205,7 @@ class ShellDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
                     self.console.sendline("")
 
             elif index == 4:
+                self.logger.info("shelldriver: console activated")
                 # we have just activated a console here
                 # lets start over again and see if login or prompt will appear
                 # now.
@@ -213,6 +220,7 @@ class ShellDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
             self.console.settle(self.post_login_settle_time, timeout=timeout.remaining)
 
         if did_login:
+            self.logger.info("shelldriver: did_login=True, executing _check_prompt()")
             self._check_prompt()
 
     @step()
```

I'm seeing following:

```shell
   INFO: shelldriver: while loop, index=4
   INFO: shelldriver: console activated
   INFO: shelldriver: while loop, index=0
   INFO: shelldriver: silencing kernel
   INFO: shelldriver: while loop, index=3
   INFO: shelldriver: expect hit a timeout
   INFO: shelldriver: while loop, index=3
   INFO: shelldriver: expect hit a timeout
   INFO: shelldriver: expect did not received anything, sending empty line
   INFO: shelldriver: while loop, index=0
   INFO: shelldriver: got a prompt, self.status=1
```

ynezz: > I think this might actually be a bug in that self.status on line 178 below should be self._status so that the code in on_activate will only run once per activation.

Indeed, this looks like a bug to me as well. Anyway, in your use case, when you've console with login, you should get [`self._status = 1`](https://github.com/labgrid-project/labgrid/blob/5d78a63b27b281b99f23971e597966d0c29a1751/labgrid/driver/shelldriver.py#L207) set via `_check_prompt()`:

https://github.com/labgrid-project/labgrid/blob/5d78a63b27b281b99f23971e597966d0c29a1751/labgrid/driver/shelldriver.py#L183-L186

JoshuaWatt: Ah, I missed that `_check_prompt` did that also. Maybe the line in `self.status` line in `_await_login` should just be removed?

Emantor: That `self.status` line doesn't do anything since the variable that is used by the driver is `self._status`.

ynezz: So I've fixed it by renaming it.

labgrid/driver/shelldriver.py
```diff
@@ -110,6 +110,35 @@ def _run(self, cmd, *, timeout=30.0, codec="utf-8", decodeerrors="strict"):
...
 
+    def _raw_run(self, cmd, timeout=2, codec="utf-8", decodeerrors="strict"):
+        """Runs command without run() wrapper as it's not yet available"""
+        marker = gen_marker()
+        self.console.sendline(f"echo {marker[:4]}; {cmd} ; echo {marker[4:]}")
+        _, _, match, _ = self.console.expect(
+            rf"{marker[:4]}(.*){marker[4:]}\s+{self.prompt}",
+            timeout=timeout
+        )
+        data = self.re_vt100.sub('', match.group(1).decode(codec, decodeerrors)).split('\r\n')
```
aparcar, line 121:
```suggestion
        data = re_vt100.sub('', match.group(1).decode(codec, decodeerrors)).split('\r\n')
```

This seems to have changed in 6743c69

aparcar:
Tested this with a Turris Omnia and it fixes the ShellDriver issues as described.

aparcar (changes requested):

aparcar:
While this fixes "real world" devices, it break the QEMU driver for me, muting all shell output. I'll investigate further and give more details once I got them.

labgrid/driver/shelldriver.py
```diff
@@ -172,7 +172,7 @@ def _await_login(self):
                 else:
                     # we got a prompt. no need for any further action to
                     # activate this driver.
-                    self.status = 1
+                    self._status = 1
```
Bastian-Krause, line 175: That's already fixed with 8bc78588.

labgrid/driver/shelldriver.py
```diff
@@ -236,11 +236,24 @@ def _check_prompt(self):
...
+            self.console.expect(self.prompt)
+            try:
+                data = self._raw_run("type run")
+                if data[2] and "function" in data[2]:
+                    return
+            except TIMEOUT:
+                pass
+
+            if timeout.expired:
+                raise TIMEOUT(f"Timeout of {inject_timeout} seconds exceeded during waiting for run injection")  # pylint: disable=line-too-long
```
Bastian-Krause, line 256: If your serial console does not work reliably (even after specifying a certain settle time), I don't think retrying is the correct approach. Even if this works after the nth try, too many other things will fail in unexpected ways later. Better make your serial console work reliably or prefer the SSHDriver in these situations.

labgrid/driver/shelldriver.py
```diff
@@ -110,6 +110,35 @@ def _run(self, cmd, *, timeout=30.0, codec="utf-8", decodeerrors="strict"):
...
+                data = self._raw_run("cat /proc/sys/kernel/printk")
+                if data[2] and data[2].startswith(f'{kernel_log_level}\t'):
+                    return
+
+            except TIMEOUT:
+                pass
+
+            if timeout.expired:
+                raise TIMEOUT(f"Timeout of {self.login_timeout} seconds exceeded during waiting for kernel silencer")  # pylint: disable=line-too-long
+
```
Bastian-Krause, line 141: I'm not sure we want more retry code for certain devices behaving strangely. Shouldn't these devices wait until input can be handled reliably before starting login/shell?

From the commit message:
> Those issues mostly happen frequently during early boot process, where there is a lot of output
going on.

Could `await_login_timeout` help here instead?

labgrid/driver/shelldriver.py
ynezz: > If your serial console does not work reliably (even after specifying a certain settle time)

Ok, fair enough, but my goal is to make it generic and reusable across different use cases, instead of fiddling with variable settle time, fighting crazy applications logging into `/dev/console` directly etc.

> If your serial console does not work reliably (even after specifying a certain settle time), I don't think retrying is the correct approach. 

Well, this approach works reliably for over than 2 years, with a mix of dozen different devices, bootloaders, kernels/drivers and operating systems. So the idea here is to get as fast (and reliably) as possible into the SSH and profit.   

> SSHDriver in these situations.

Indeed, its the only reliable solution if there is no dedicated serial console available.

ynezz:
@Bastian-Krause thanks a lot for a review, really appreciate it! I'll rebase and retest soon.

aparcar:
Hi @ynezz 

Since I faced the same issue and gave this another try. I found that the simplest fix was to drop the `console_ready` parameter and instead use `logIn_prompt`. It's a little bit mystical to me but it turned out to work fine. I assume Labgrids internal state machine for silencing the Kernel got confused by OpenWrt not requiring a proper login.

Below is my working example which did not require any code modifications:

```yaml
targets:
  main:
    features:
      - wifi
      - online
    resources:
      RemotePlace:
        name: !template '$LG_PLACE'
    drivers:
      NetworkPowerDriver: {}
      SerialDriver:
        txdelay: 0.05
      ShellDriver:
        prompt: 'root@[\w()]+:[^ ]+ '
        login_prompt: Please press Enter to activate this console.
        await_login_timeout: 15
        login_timeout: 120
        post_login_settle_time: 5
        username: root
      UBootDriver:
        prompt: '=> '
        init_commands:
          - setenv kernel_addr_r 0x2000000
          - setenv bootargs earlyprintk console=ttyS0,115200
        boot_command: >
          dhcp $kernel_addr_r turris-omnia/openwrt-mvebu-cortexa9-cznic_turris-omnia-initramfs-kernel.bin;
          bootm $kernel_addr_r
      TFTPProviderDriver: {}
      UBootStrategy: {}
```

SmithChart:
I am also testing OpenWRT (specifically Gluon) devices with labgrid.
And I can report, that for me silencing of the log-messages with `dmesg -n 1` is currently broken.

In my use case using `console_ready` means that i am directly logged in afterwards. So I would either assume that we should set `did_login = True` after https://github.com/labgrid-project/labgrid/blob/1ab0abda791c00e448cbbb8c44a61564efbd16e9/labgrid/driver/shelldriver.py#L171 or use https://github.com/labgrid-project/labgrid/pull/945/commits/1209a65e18402386d5e7054a5bad25c918bf8b91 .

Cherrypicking 1209a65e18402386d5e7054a5bad25c918bf8b91 onto current master at least fixes my issue.

But looking at the code, as it is now, I think I would prefer to set `did_login = True` after logging in with `index == 4`. This way we make sure, that we only issue the `dmesg -n 1` a single time - even if we maybe extend the state machine later on.

PR #949, opened 2022-07-04
path too long error fix

### LINKED ISSUE #948: too long for Unix domain socket when running labgrid-client natively on MacOS
if running `labgrid-client` nativly on MacOS the ssh-connection to the exporter will fail due to an error called `too long for Unix domain socket`

I'll link a pull-request

danielschickmunit:
Fixes #948

changes the names for the temporary file_paths from 
`labgrid-ssh-tmp-` to `lg-ssh-` 
and
 `labgrid-connection-` to `lg-con-`

in order to be compatible with MacOS and its longer tempdir-path

### DIFF
--- a/labgrid/driver/sshdriver.py
+++ b/labgrid/driver/sshdriver.py
@@ -81,7 +81,7 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
 
     def _start_own_master_once(self, timeout):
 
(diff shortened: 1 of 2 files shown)


Emantor:
FWIW this also happens on Linux, so its not macOS exclusive. Can you fix the DCO by adding a signed off by?

danielschickmunit:
@Emantor added a signed off by ✅

Emantor (approved):

PR #954, opened 2022-07-29
agentwrapper: Bump SSH timeout

JoshuaWatt:
Bumps the SSH connection timeout to 30 seconds for the agent wrapper to
match the other SSH timeouts

**Checklist**
- [ ] Tests for the feature 

- [x] PR has been tested


Emantor (approved):

jluebbe:
The idea behind the lower timeout for the agent was that the exporter must already be running at this point and that the network connection to an exporter should be reliable.

Did you run into the 5 second timeout?

JoshuaWatt:
Yes. Due to many factors, we need a longer delay when connecting to the exporters (internal networking, multiple places per exporter, etc.). The 30 seconds on the normal connection timeout appears to be sufficient, although ideally it would be globally adjustable which I'm working on proposal to do that which I'll submit as a PR soon.

JoshuaWatt:
#958 effectively does the same thing, but is my attempt to make it globally controlled. If you would prefer submitting that instead, that is fine with me

JoshuaWatt:
Superseded by #958

Issue #489, opened 2019-08-30
labgrid-client commands fail but coordinator running


spark5:
When running labgrid-client commands, the coordinator is up and running, exporter the same, but this is thrown:

```
 ERROR: Task exception was never retrieved
future: <Task finished coro=<onJoin() done, defined at labgrid-tests/venv/lib/python3.5/site-packages/labgrid-0.1.dev1453+g1b56ade.d20190830-py3.5.egg/labgrid/remote/client.py:69> exception=ApplicationError('no callee registered for procedure <org.labgrid.coordinator.get_resources>',)>
Traceback (most recent call last):
  File "/usr/lib/python3.5/asyncio/tasks.py", line 241, in _step
    result = coro.throw(exc)
  File "labgrid-tests/venv/lib/python3.5/site-packages/labgrid-0.1.dev1453+g1b56ade.d20190830-py3.5.egg/labgrid/remote/client.py", line 72, in onJoin
    'org.labgrid.coordinator.get_resources'
  File "/usr/lib/python3.5/asyncio/futures.py", line 361, in __iter__
    yield self  # This tells Task to wait for completion.
  File "/usr/lib/python3.5/asyncio/tasks.py", line 296, in _wakeup
    future.result()
  File "/usr/lib/python3.5/asyncio/futures.py", line 274, in result
    raise self._exception
autobahn.wamp.exception.ApplicationError: ApplicationError(error=<wamp.error.no_such_procedure>, args=['no callee registered for procedure <org.labgrid.coordinator.get_resources>'], kwargs={}, enc_algo=None, callee=None, callee_authid=None, callee_authrole=None, forward_for=None)
```

The partial output from coordinator:

```
aug 30 14:01:03 : 2019-08-30T14:00:57+0300 [Router      19901 crossbar.router.protocol.WampWebSocketServerProtocol] Connection to/from tcp4:IP:45296 was closed cleanly
aug 30 14:01:03 : 2019-08-30T14:00:57+0300 [Router      19901 crossbar.router.protocol.WampWebSocketServerProtocol] _connectionLost: [Failure instance: Traceback (failure with no frames): <class 'twisted.internet.error.ConnectionDone'>: Connection was closed cleanly.
aug 30 14:01:03 : ]
[...]

PR #955, opened 2022-08-01
Integrate nxp uuu into labgrid.

atline:
This PR to add nxp uuu support to labgrid, we use uuu to flash images to i.mx boards.

**Description**

Some considerations here:

* I notice there is realization for uuu driver [here](https://github.com/labgrid-project/labgrid/blob/master/labgrid/driver/usbloader.py#L151), but it's really not how we use it in NXP internal, see [official example](https://github.com/NXPmicro/mfgtools#examples), also how we use it in [lava](https://validation.linaro.org/static/docs/v2/actions-boot.html#contents)
* Usually user need to use uuu to remotely flash a bootloader into broken board, so `labgrid-client uuu` added.
* We don't design `uuu resource` based on `udev`, this is because `uuu` will be separated into several stages: SDP, FB, etc. For `SDP`, uuu will use usb2.0 protocol, while for `FB`, if the board connected to a `usb3.0`, uuu will use usb3.0 protocol. That means the otg path may jump from `3:8` to `2:8` during uuu run. But if we use udev method we are not able to get some like `uuu -m 3:8 -m 2:8` before run uuu, so we choose to statically specify `["3:8", "2:8"]`.
* In our scenario, some boards not directly linked to exporter, so a `NetworkUUU` could be specified for such case.
* When use ssh to call uuu, we will have to allocate a pseudo terminal there as when uuu flash, it will has a progress bar. Without `-t`, the screen will be chaos. 

Local test here:

image

image

**Checklist**
- [√] Documentation for the feature
- [√] Tests for the feature
- [√] The arguments and description in doc/configuration.rst have been updated
- [√ ] CHANGES.rst has been updated
- [√ ] PR has been tested

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -26,6 +26,7 @@ New Features in 0.5.0
 - `DFUDriver` has been added to communicate with a `DFUDevice`, a device in DFU
   (Device Firmware Upgrade) mode.
 - ``labgrid-client dfu`` added to allow communcation with devices in DFU mode.
+- Add NXP uuu resource/driver, add labgrid-client uuu command.
 
 Bug fixes in 0.5.0
 ~~~~~~~~~~~~~~~~~~
@@ -299,7 +300,7 @@ New and Updated Drivers
 - The `SerialDriver` now supports using plain TCP instead of RFC 2217, which is
   needed from some console servers.
 - The `ShellDriver` has been improved:
-  
+
   - It supports configuring the various timeouts used during the login process.
   - It can use xmodem to transfer file from and to the target.
 
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -625,6 +625,45 @@ Arguments:
 Used by:
   - `AndroidFastbootDriver`_
 
+UUU
+~~~
+An UUU resource describes a i.mx uuu device in the serialdownload mode which attached to exporter.
+
+.. code-block:: yaml
+
+   UUU:
+     usb_otg_path: "3:8"
+
+   UUU:
+     usb_otg_path: ["3:8", "2:8"]
+
+Arguments:
+  - usb_otg_path (str or list): UUU otg paths, could obtain it with ``uuu -lsusb``
+
+Used by:
+  - `UniversalUpdateUtilityDriver`_
+
+NetworkUUU
+~~~~~~~~~~
+An UUU resource describes a i.mx uuu device in the serialdownload mode which not directly attached to exporter.
+
+.. code-block:: yaml
+
+   UUU:
+     host: "10.192.244.109"
+     usb_otg_path: "3:8"
+
+   UUU:
+     host: "10.192.244.109"
+     usb_otg_path: ["3:8", "2:8"]
+
+Arguments:
+  - host (str): the host which with uuu device attached.
+  - usb_otg_path (str or list): UUU otg paths, could obtain it with ``uuu -lsusb``
+
+Used by:
+  - `UniversalUpdateUtilityDriver`_
+
 DFUDevice
 ~~~~~~~~~
 A DFUDevice resource describes a USB device in DFU (Device Firmware Upgrade)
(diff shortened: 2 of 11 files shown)


jluebbe:
I see some issues with this PR, which we'll need to discuss.

The UUU resource doesn't support the normal udev matching, which means that the client can't know whether the device is actually available or not. So, `await_resource` cannot be used with it, which makes some more complex test cases unreliable, as they would need to rely only on timing.

The UUU resource mixes multiple protocols, which already have existing resources in labgrid (IMXUSBLoader, AndroidFastboot).

What's the benefit of using UUU to handle the Fastboot phase (instead of the standard fastboot tool)?

I'd prefer to keep the stages (bootloader upload & flashing) separate, as otherwise you don't have a chance to capture and interact with the bootloader. What is the benefit of mixing that in UUU?

The UUUDriver tries to push any argument that is a file to the exporter. Isn't there a better way to give the arguments more structure, so the driver knows what is happening? 

Did you try to use the existing IMXUSBLoader & AndroidFastboot resources with a new UUUDriver?

atline:
> I see some issues with this PR, which we'll need to discuss.

[A] Yes, thanks for reply,  I'm new to this project, so please guide me :)

> 
> The UUU resource doesn't support the normal udev matching, which means that the client can't know whether the device is actually available or not. So, `await_resource` cannot be used with it, which makes some more complex test cases unreliable, as they would need to rely only on timing.

[A]
* I never touch the concept `await_resource`, pretty new to learn labgrid detail :) . But if we use `NetworkSerialPort` or `PDUDaemon`, could labgrid assure the resource actually available or not? So, looks `UUU` without udev is not the only exception, you have more exceptions already?
* Why we not based on `udev`?
  * We have a hybrid farm in NXP, lava for CI test(support release, nightly and kernelCI), while labgrid for engineer debug(support power cycle, console, and maybe UUU later), and they share 1 board.
  * But when we deploy exporter on the same board server which lava-worker also on it, it frequently meet [this problem](https://github.com/labgrid-project/labgrid/issues/489#issuecomment-1021819767), looks lava's pexpect on serial port results in a heavy IO. It's not just happens on one machine, so now we move the labgrid exporter to a separate machine, pretty stable now. Then, we need client not ssh to exporter, but ssh to a specified board server to operate UUU. If that is the case, could we benifit from udev? Exporter even don't know the board...
  * We can't get next `-m 3:8 -m 2:8` with udev, of course you suggest not to flash with UUU, use fastboot.

> 
> The UUU resource mixes multiple protocols, which already have existing resources in labgrid (IMXUSBLoader, AndroidFastboot).

[A] Depends on if we use udev.

> 
> What's the benefit of using UUU to handle the Fastboot phase (instead of the standard fastboot tool)?

[A] 
* Much simpler. When board broken, the i.mx board will automatically switch to serial download mode, Then with just one command, `uuu` will interact with ROM to download a valid bootloader and flash it. Meanwhile, for most of i.mx8 and later i.mx9 chip, we will have BCU support, which could switch the jumper automatically with software, that will be our main solutions later to automate i.mx test. 
* And, uuu is also a test target here, guys may need to use labgrid to directly operate uuu to verify flash function ok or not.
* The embeded uuu script will default combine all modes, users will have to separate them with customized uuu script. As a tool in farm,  I hope user's life much simple when use the tool.
* In our local release, some legacy's board even don't have fastboot command in uboot. Meanwhile, internal, most of guys use this tool, seldom to use fastboot as UUU is much easier.

> 
> I'd prefer to keep the stages (bootloader upload & flashing) separate, as otherwise you don't have a chance to capture and interact with the bootloader. What is the benefit of mixing that in UUU?

[A] 
* What we need to interact with the bootloader? For our cases, we use it to recover boards, you know in a boardfarm, guys may corrupt the boards, we just need to use it recover that. And for detail interact with bootloader, after flash it with UUU, we could open the serial to interact with uboot. For more detail check, we even could write our own `UUUScript`, not use the `Embedded -b emmc` uuu script. In customized uuu script, you still could control not to use FB.
* And I'm not sure, does you suggestion works on a pretty new board without bootloader?

> 
> The UUUDriver tries to push any argument that is a file to the exporter. Isn't there a better way to give the arguments more structure, so the driver knows what is happening?

[A] What we usually used is next:

image

If we have an accurate control, that means we will need to add special arguments for bootloader & rootfs? Is my understanding correct? My initial thought is: the command could have the same usage as the one if we not use labgrid, then everyone could quickly run any uuu command just as no labgrid there. You know when use `fastboot`, it could just use `labgrid -p place fastboot boot boot.bin` without change the habit of non-labgrid user. But for `uuu`, the file numbers are not fixed, so I don't know how to handle that better.

> 
> Did you try to use the existing IMXUSBLoader & AndroidFastboot resources with a new UUUDriver?

[A] No, same as previous, I guess it depends on if we based on udev...... Or, if it's acceptable to afford 2 options: one based on udev to get the exact "usb path" if we link the board directly to exporter, for what you mentioned "await resource"; another one not based on udev to fix our current scenario? It that, I will have a try with `IMXUSBLoader` although I'm not quite sure what it is... Sorry, currently I just use `power` and `console` with labgrid...

atline:
There is another team in NXP which use labgrid, we're contacting now to possible align something internal, other guys may back with other solution.

PR #959, opened 2022-08-01
driver: fastbootdriver: support fastboot over network

alegsan:
Signed-off-by: Oleg Karfich <oleg.karfich@wago.com>

Extend the currently fastboot implementation in labgrid by supporting fastboot over ethernet.

The extension is tested with an WAGO PFC200 (750-8212) with running Barebox v2021.10.0. We also tested the currently implemented USB functionality in case of some breaking changes. 

**Checklist**
- [ ] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -605,13 +605,13 @@ NetworkRKUSBLoader
 ~~~~~~~~~~~~~~~~~~~
 A NetworkRKUSBLoader describes an `RKUSBLoader`_ available on a remote computer.
 
-AndroidFastboot
-~~~~~~~~~~~~~~~
-An AndroidFastboot resource describes a USB device in the fastboot state.
+AndroidUSBFastboot
+~~~~~~~~~~~~~~~~~~
+An AndroidUSBFastboot resource describes a USB device in the fastboot state.
 
 .. code-block:: yaml
 
-   AndroidFastboot:
+   AndroidUSBFastboot:
      match:
        ID_PATH: pci-0000:06:00.0-usb-0:1.3.2:1.0
 
@@ -621,6 +621,25 @@ Arguments:
   - usb_product_id (str, default="0104"): USB product ID, to be compared with
     the ``ID_MODEL_ID`` udev property
   - match (dict): key and value pairs for a udev match, see `udev Matching`_
+
+Used by:
+  - `AndroidFastbootDriver`_
+
+AndroidNetFastboot
+~~~~~~~~~~~~~~~~~~
+An AndroidNetFastboot resource describes a network device in fastboot state.
+
+.. code-block:: yaml
+
+   AndroidNetFastboot:
+     address: "192.168.23.42"
+
+Arguments:
+  - address (str): ip address of the fastboot device
+  - port (int, default=5554): udp/tcp fastboot port that is used in the
+    device. (e.g. Barebox uses port 5554)
+  - protocol (str, default="udp"): which protocol should be used when issuing
+    fastboot commands. (Barebox supports currently only the udp protocol)
 
 Used by:
   - `AndroidFastbootDriver`_
@@ -1159,7 +1178,7 @@ The initial matching and monitoring for udev events is handled by the
 :any:`UdevManager` class.
 This manager is automatically created when a resource derived from
 :any:`USBResource` (such as :any:`USBSerialPort`, :any:`IMXUSBLoader` or
-:any:`AndroidFastboot`) is instantiated.
+:any:`AndroidUSBFastboot`) is instantiated.
 
 To identify the kernel device which corresponds to a configured `USBResource`,
 each existing (and subsequently added) kernel device is matched against the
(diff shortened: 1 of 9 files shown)


doc/configuration.rst
```diff
@@ -605,13 +605,13 @@ NetworkRKUSBLoader
 ~~~~~~~~~~~~~~~~~~~
 A NetworkRKUSBLoader describes an `RKUSBLoader`_ available on a remote computer.
 
-AndroidFastboot
-~~~~~~~~~~~~~~~
-An AndroidFastboot resource describes a USB device in the fastboot state.
+AndroidUsbFastboot
```
Emantor, line 608: This is a breaking change, we want to retain backwards compatibility with a DeprecationWarning, than we can remove `AndroidFastboot` in a later release.

labgrid/resource/remote.py
```diff
@@ -113,12 +113,20 @@ class RemoteUSBResource(NetworkResource, ManagedResource):
...
 
 
+@target_factory.reg_resource
+@attr.s(eq=False)
+class NetworkAndroidNetFastboot(NetworkResource):
+    address = attr.ib(validator=attr.validators.instance_of(str))
+    port = attr.ib(default=5554, validator=attr.validators.instance_of(int))
+    protocol = attr.ib(default="udp", validator=attr.validators.instance_of(str))
+
+
```
jluebbe, line 129: This seems identical to the AndroidNetFastboot resource, so it shouldn't be needed?

doc/configuration.rst
alegsan: Ok. I changed the PR and add a `DeprecationWarning` like in [`usbstoragedriver.py`](https://github.com/labgrid-project/labgrid/blob/dffc61d31db002d674ce2a31cff9795760d65682/labgrid/driver/usbstoragedriver.py#L161)

labgrid/resource/remote.py
alegsan: Yes it seems, because the attributes are identical. But the inherited type is different. Our understanding is that we need two separate ressources. One local and one remote resource. In case of the remote resource we need  to inherit from `NetworkResource` to get the ssh command prefix. Is this correct?

doc/configuration.rst
Bastian-Krause, line 608: I am not really happy with the resource name. It should be "USB" (to keep things consistent with the other resource names). Is there a reason you added "Usb" in the middle?

For the time being we're stuck with:
- AndroidFastboot
- NetworkAndroidFastboot
- AndroidNetFastboot
- AndroidUsbFastboot

Pretty confusing..

alegsan: Hi Bastian,

thx for the review.

> It should be "USB" (to keep things consistent with the other resource names)

Fine for me. We wanted to keep it consistent with the new resource `AndroidNetFastboot`. So we should change this also to `AndroidNETFastboot`?


 
> For the time being we're stuck with:
> 
> * AndroidFastboot
> * NetworkAndroidFastboot
> * AndroidNetFastboot
> * AndroidUsbFastboot
> 
> Pretty confusing..

Our intention was to make it clear that the user knows wich resource to use by only reading the name of the resource in the documentation. But if the maintainers are fine to keep the current resource name `AndroidFastboot` for the USB backend resource, its fine for me too. The benifit is, that no deprecation warning is needed anymore and also no breaking changes for users in the future when die resource `AndroidFastboot` would be removed.

labgrid/resource/remote.py
jluebbe: Ah, I had expected that with network fastboot, you'd just use SSH port forwarding. But now I understand that for UDP, that can't work and that's why you still execute fastboot on the exporter.

To clear up the naming confusion, we've started to name new exported resources Remote* (see RemoteNetworkInterface). So for this case, I'd use RemoteAndroidNetFastboot.

In the medium term, if we get SSH SOCKS5 tunnels, we could run network fastboot over that tunnel, but that can be added later.

doc/configuration.rst
jluebbe: I'd like to take this chance to clean up a bit. We have and keep for compatibility:
- AndroidFastboot
- NetworkAndroidFastboot

New names for the above:
- AndroidUSBFastboot
- RemoteAndroidUSBFastboot

And for the new network functionality:
- AndroidNetFastboot
- RemoteAndroidNetFastboot

alegsan: > New names for the above:
> 
> * AndroidUSBFastboot
> * RemoteAndroidUSBFastboot

By testing the changed PR with the above renaming we get another question we need to discuss here. `AndroidFastboot` and `AndroidUSBFastboot` are exported by the same class `USBGenericExport` in [exporter.py](https://github.com/labgrid-project/labgrid/blob/dffc61d31db002d674ce2a31cff9795760d65682/labgrid/remote/exporter.py#L494). But this class sets the `Network` prefix. So the new resource name `RemoteAndroidUSBFastboot` is not handled here. Changing the class `USBGenericExport` would result in changing all `Network*` resources handled by this class. Creating a new class that handles the `Remote` prefix is not the  most elegant way but would allow us the renaming? (see my updated changes)? What do you think?

jluebbe: Yes, it's not elegant, but at least it's explicit. :) I think we'll have to live with this.

jluebbe (approved):
I've added a small comment that the AndroidFastboot resource still exists for backwards compatibility.

alegsan:
bump :-)

Emantor (approved):

PR #961, opened 2022-08-04
driver/qemudriver: add support for Q35 machine

EmilKronborg:

**Description**
Allow usage of Q35-based machines in QEMU.

**Checklist**
- [ ] Tests for the feature 

- [x] PR has been tested


labgrid/driver/qemudriver.py
```diff
@@ -121,11 +121,15 @@ def on_activate(self):
                 self._cmd.append(
                     f"if=sd,format={disk_format},file={disk_path},id=mmc0")
                 boot_args.append("root=/dev/mmcblk0p1 rootfstype=ext4 rootwait")
-            elif self.machine == "pc":
+            if self.machine == "pc":
```
jluebbe, line 124: Shouldn't this stay as `elif`?

labgrid/driver/qemudriver.py
```diff
@@ -121,11 +121,15 @@ def on_activate(self):
...
                 self._cmd.append("-drive")
                 self._cmd.append(
                     f"if=virtio,format={disk_format},file={disk_path}")
                 boot_args.append("root=/dev/vda rootwait")
+            elif self.machine == "q35":
+                self._cmd.append("-drive")
+                self._cmd.append(
+                    f"format={disk_format},file={disk_path}")
```
jluebbe, line 132: Is `if=virtio` the default now?

We should probably keep the `root=... rootwait` or is there a downside?

jluebbe (changes requested):

labgrid/driver/qemudriver.py
EmilKronborg: No, `if=ide` is default for `q35`. Should I include that?

I think it is fine to include `rootwait`, but I would expect people to define `root` and `rootfstype` in `local.yaml` if it is not handled by their bootloader. By appending them to `boot_args`, you effectively overwrite any `boot` or `rootfstype` defined in `local.yaml` and thus hardcode them or did I miss anything? For example, if i spin up a `pc` machine, I will always have `root=/dev/vda`, no?

labgrid/driver/qemudriver.py
EmilKronborg: That's fair.

labgrid/driver/qemudriver.py
jluebbe: > No, `if=ide` is default for `q35`. Should I include that?

I'd argue that we `q35` should behave mostly similar to the existing `pc` config.

> I think it is fine to include `rootwait`, but I would expect people to define `root` and `rootfstype` in `local.yaml` if it is not handled by their bootloader. By appending them to `boot_args`, you effectively overwrite any `boot` or `rootfstype` defined in `local.yaml` and thus hardcode them or did I miss anything? For example, if i spin up a `pc` machine, I will always have `root=/dev/vda`, no?

The boot_args are only passed to qemu if a kernel is started directly:
https://github.com/labgrid-project/labgrid/blob/master/labgrid/driver/qemudriver.py#L174-L176

Also, custom boot_args are appended after the autogenerated ones
(https://github.com/labgrid-project/labgrid/blob/master/labgrid/driver/qemudriver.py#L172-L173), which allows the user to override them in the yaml.

EmilKronborg: > I'd argue that we q35 should behave mostly similar to the existing pc config.

If we keep the interface default for q35, i.e., `if=ide`, I agree.

> The boot_args are only passed to qemu if a kernel is started directly:
https://github.com/labgrid-project/labgrid/blob/master/labgrid/driver/qemudriver.py#L174-L176

> Also, custom boot_args are appended after the autogenerated ones
(https://github.com/labgrid-project/labgrid/blob/master/labgrid/driver/qemudriver.py#L172-L173), which allows the user to override them in the yaml.

Now it makes sense to me. Thanks for clarifying. I will include `root=... rootwait` to q35 also.

jluebbe (approved):

labgrid/driver/qemudriver.py
jluebbe: Oh, I misread the diff and went ahead with the merge. Could you add the `root=` in a new PR? :)

EmilKronborg: I was confused for a second. Yes, I will make a new PR.

PR #964, opened 2022-08-08
driver/qemudriver: add drive interface and kernel boot parameters

EmilKronborg:
**Description**
Make the `q35` machine behave similar to `pc` if a kernel is started directly. Furthermore, the default drive interface has been added. This commit is an addition to `086ba3b3fe1e04bcc8cba0e6dfe558b7747d372d`.

**Checklist**
- [ ] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/labgrid/driver/qemudriver.py
+++ b/labgrid/driver/qemudriver.py
@@ -124,7 +124,8 @@ class QEMUDriver(ConsoleExpectMixin, Driver, PowerProtocol, ConsoleProtocol):
             if self.machine == "q35":
                 self._cmd.append("-drive")
                 self._cmd.append(
-                    f"format={disk_format},file={disk_path}")
+                    f"if=virtio,format={disk_format},file={disk_path}")
+                boot_args.append("root=/dev/vda rootwait")
             elif self.machine == "pc":
                 self._cmd.append("-drive")
                 self._cmd.append(


jluebbe (changes requested):
I would expect that you'd need `if=virtio` to get `root=/dev/vda` to work.

EmilKronborg:
> I would expect that you'd need `if=virtio` to get `root=/dev/vda` to work.

You're right. Let's keep the interface out for now. I rewrote the commit message to reflect this change.

jluebbe:
Does it work this way? You said that q35 defaults to IDE in https://github.com/labgrid-project/labgrid/pull/961#discussion_r938193620.

My preferred approach would be `if=virtio`, as that avoids the need for legacy IDE drivers.

EmilKronborg:
> 

In my case, I am using a bios image (`.qcow2`) and a raw image (`.wic`) to mount my filesystem. This setup fails with anything but the standard interface, i.e., `if=ide`. If there was a way to override the interface, I could add `virtio` to make q35 and pc more similar, but as far as I can tell you're not able to override it?

jluebbe:
Ah. You could not set the `disk` property, as then no `-drive` property would be generated. Then you could pass that via the `extra_args` property.

EmilKronborg:
OK. That solution is not as elegant as using the `disk` property, but I can live with that. I will rather get going with tests for our project. I added `if=virtio` and rewrote the commit message.

jluebbe (approved):

jluebbe:
Thanks a lot for your contribution!

PR #973, opened 2022-08-18
util/qmp: log representation of line instead of line

Bastian-Krause:
**Description**
Lines end with '\r\n', so an extra empty line is logged after each QMP line received. Log the representation of the line to get rid of extra lines.

**Checklist**
- [x] PR has been tested

### DIFF
--- a/labgrid/util/qmp.py
+++ b/labgrid/util/qmp.py
@@ -27,7 +27,7 @@ class QMPMonitor:
 
     def _read_parse_json(self):
         line = self.monitor_out.readline().decode('utf-8')
-        self.logger.debug("Received line: %s", line)
+        self.logger.debug("Received line: %s", line.rstrip("\r\n"))
         if not line:
             raise QMPError("Received empty response")
         return json.loads(line)


Emantor (dismissed):

labgrid/util/qmp.py
```diff
@@ -27,7 +27,7 @@ def _negotiate_capabilities(self):
 
     def _read_parse_json(self):
         line = self.monitor_out.readline().decode('utf-8')
-        self.logger.debug("Received line: %s", line)
+        self.logger.debug("Received line: %r", line)
```
jluebbe, line 30:
```suggestion
        self.logger.debug("Received line: %s", line.rstrip("\r\n"))
```

jluebbe (changes requested):
As this usually contains JSON and was just produced by readline, we should just strip the line ending and not quote it again. JSON becomes hard to read after repr().

labgrid/util/qmp.py
Bastian-Krause: Good point, I've changed it according to your suggestion.

jluebbe (approved):

Emantor (approved):

PR #975, opened 2022-08-23
driver/shelldriver: handle bracketed-paste mode

### LINKED ISSUE #969: Bracketed-paste mode makes ShellDriver prompt detection fail
I am following the getting starting steps and running the first test

I have local config file setup for my pi zerow:
```
targets:
  main:
    resources:
      RawSerialPort:
        port: "/dev/ttyUSB0"
    drivers:
      ManualPowerDriver:
        name: "example"
      SerialDriver: {}
      ShellDriver:
        prompt: 'pi@\w+:[^ ]+ '
        login_prompt: 'raspberrypi login: '
        username: 'pi'
        password: 'password'

```

I have verified the port `/dev/ttyUSB0` is correct using dmesg

When I run the test using: 
```
pytest --lg-env local.yaml test_shell.py
```
I consistently get this timeout  error: 
```
**ERROR test_shell.py::test_shell - pexpect.exceptions.TIMEOUT: Timeout of 4.93 seconds exceeded or connection closed by peer**
```
I also see this error in dmesg after test is run (bold):

[670927.706302] usb 1-7: new full-speed USB device number 72 using xhci_hcd
[670927.866892] usb 1-7: New USB device found, idVendor=067b, idProduct=2303, bcdDevice= 4.00
[670927.866900] usb 1-7: New USB device strings: Mfr=1, Product=2, SerialNumber=0
[670927.866903] usb 1-7: Product: USB-Serial Controller
[670927.866906] usb 1-7: Manufacturer: Prolific Technology Inc.
[670927.869636] pl2303 1-7:1.0: pl2303 converter detected
[670927.871602] usb 1-7: pl2303 converter now attached to ttyUSB0
**[670972.423388] pl2303 ttyUSB0: pl2303_get_line_request - failed: -32
[670972.423787] pl2303 ttyUSB0: pl2303_get_line_request - failed: -32
[670980.558787] pl2303 ttyUSB0: error sending break = -32**

I am using the pl2303 to convert serial/usb

One last thing to note is that when the test is over and I minicom to the pi I can see that it did indeed login successfully:
```
pi@raspberrypi:~$ whoami
pi
```

Bastian-Krause:
**Description**
bash >= 5.1 and readline >= 8.1 enable bracketed-paste mode by default, allowing the terminal emulator to tell a program whether input was typed or pasted. To achieve this, `\e[?2004h` is inserted when user input is expected enabling paste detection. `\e[?2004l` is inserted to disable it.

To handle the escape sequences, match the enable sequence between marker and prompt. The disable sequence is already stripped by the regex removing all VT100 codes.

See https://github.com/pexpect/pexpect/issues/669

Tested with readline 8.1 and bash 5.1.16

**Checklist**
- [x] PR has been tested

Fixes #969

### DIFF
--- a/labgrid/driver/shelldriver.py
+++ b/labgrid/driver/shelldriver.py
@@ -95,8 +95,9 @@ class ShellDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
         # hide marker from expect
         cmp_command = f'''MARKER='{marker[:4]}''{marker[4:]}' run {shlex.quote(cmd)}'''
         self.console.sendline(cmp_command)
+        # match output by run function, bracketed-paste escape sequence (optionally), prompt
         _, _, match, _ = self.console.expect(
-            rf'{marker}(.*){marker}\s+(\d+)\s+{self.prompt}',
+            rf'{marker}(.*){marker}\s+(\d+)\s+(\x1b\[\?2004h)?{self.prompt}',
             timeout=timeout
         )
         # Remove VT100 Codes, split by newline and remove surrounding newline
@@ -200,8 +201,9 @@ class ShellDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
         # hide marker from expect
         self.console.sendline(f"echo '{marker[:4]}''{marker[4:]}'")
         try:
+            # match marker, bracketed-paste escape sequence (optionally), prompt
             self.console.expect(
-                rf"{marker}\s+{self.prompt}",
+                rf"{marker}\s+(\x1b\[\?2004h)?{self.prompt}",
                 timeout=5
             )
             self._status = 1
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -29,6 +29,8 @@ New Features in 0.5.0
 - Support for QEMU Q35 machine added.
 - `UBootDriver` now handles idle console, allowing driver activation on
   an interupted U-Boot.
+- `ShellDriver` now handles bracketed-paste mode (default in bash >= 5.1,
+  readline >= 8.1)
 
 Bug fixes in 0.5.0
 ~~~~~~~~~~~~~~~~~~


labgrid/driver/shelldriver.py
```diff
@@ -95,8 +95,9 @@ def _run(self, cmd, *, timeout=30.0, codec="utf-8", decodeerrors="strict"):
         # hide marker from expect
         cmp_command = f'''MARKER='{marker[:4]}''{marker[4:]}' run {shlex.quote(cmd)}'''
         self.console.sendline(cmp_command)
+        # match output by run function, bracketed-paste escape sequence (optionally), prompt
         _, _, match, _ = self.console.expect(
-            rf'{marker}(.*){marker}\s+(\d+)\s+{self.prompt}',
+            rf'{marker}(.*){marker}\s+(\d+)\s+(\x1b\[\?2004h)?{self.prompt}',
```
jluebbe, line 100: This only matches the escape sequence at a specific location, which seems fragile.

jluebbe (commented):
Isn't it possible to disable this feature?

Bastian-Krause:
> Isn't it possible to disable this feature?

We could run `bind 'set enable-bracketed-paste off'` (right after `dmesg -n 1`) to disable it. But we would still need to relax the pattern in `_check_prompt()`, because that's expected before the commands can be issued.

labgrid/driver/shelldriver.py
Bastian-Krause: Agreed.

jluebbe: If (on a given system) this always appears at the start of the prompt, the user could add this to their prompt regex instead.

Bastian-Krause: Correct. On the one hand it feels a bit like a work-around, but on the other you could probably argue that this belongs to the prompt somewhat. So I agree that we consider this a valid solution.

ynezz:
(Was directed here from the discussion in #1397, so just my 2 cents)

> On the one hand it feels a bit like a work-around

+2 (myself and @khilman https://github.com/labgrid-project/labgrid/issues/1397#issuecomment-2101570850)

> We could run bind 'set enable-bracketed-paste off' (right after dmesg -n 1) to disable it. 

This feature is supposed to be functional on the DUT; therefore, disabling the feature and conducting the test without it contradicts the purpose of the testing, doesn't it?

> but on the other you could probably argue that this belongs to the prompt somewhat

IMO it belongs to the underlying shell/terminal, not the prompt itself.

PR #976, opened 2022-08-24
Qemu Improvements

JoshuaWatt:
**Description**
Adds display support and networking support to QEMU

**Checklist**
- [x] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/labgrid/driver/sshdriver.py
+++ b/labgrid/driver/sshdriver.py
@@ -12,7 +12,7 @@ import time
 import attr
 
 from ..factory import target_factory
-from ..protocol import CommandProtocol, FileTransferProtocol
+from ..protocol import CommandProtocol, FileTransferProtocol, DynamicNetworkServiceProtocol
 from .commandmixin import CommandMixin
 from .common import Driver
 from ..step import step
@@ -27,7 +27,7 @@ from ..util.ssh import get_ssh_connect_timeout
 @attr.s(eq=False)
 class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
     """SSHDriver - Driver to execute commands via SSH"""
-    bindings = {"networkservice": "NetworkService", }
+    bindings = {"networkservice": {"NetworkService", "DynamicNetworkServiceProtocol",},}
     priorities = {CommandProtocol: 10, FileTransferProtocol: 10}
     keyfile = attr.ib(default="", validator=attr.validators.instance_of(str))
     stderr_merge = attr.ib(default=False, validator=attr.validators.instance_of(bool))
@@ -37,6 +37,7 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
         super().__attrs_post_init__()
         self.logger = logging.getLogger(f"{self}({self.target})")
         self._keepalive = None
+        self._networkservice = None
 
     def on_activate(self):
         self.ssh_prefix = ["-o", "LogLevel=ERROR"]
@@ -45,7 +46,13 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
             if self.target.env:
                 keyfile_path = self.target.env.config.resolve_path(self.keyfile)
             self.ssh_prefix += ["-i", keyfile_path ]
-        if not self.networkservice.password:
+
+        if isinstance(self.networkservice, DynamicNetworkServiceProtocol):
+            self._networkservice = self.networkservice.get_network_service()
+        else:
+            self._networkservice = self.networkservice
+
+        if not self._networkservice.password:
             self.ssh_prefix += ["-o", "PasswordAuthentication=no"]
 
         self.control = self._start_own_master()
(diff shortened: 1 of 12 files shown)


JoshuaWatt:
Ping?

Emantor:
Sorry, I was on vacation, had our yearly tech week and we are now at ELCE, earliest I can look at this is next week.

JoshuaWatt:
> Sorry, I was on vacation, had our yearly tech week and we are now at ELCE, earliest I can look at this is next week.

Ok. Thanks!

labgrid/driver/sshdriver.py
```diff
@@ -44,7 +45,13 @@ def on_activate(self):
             if self.target.env:
                 keyfile_path = self.target.env.config.resolve_path(self.keyfile)
             self.ssh_prefix += ["-i", keyfile_path ]
-        if not self.networkservice.password:
+
+        if isinstance(self.networkservice, DynamicNetworkServiceProtocol):
+            self._networkservice = self.networkservice.get_network_service()
```
Emantor, line 50: Hm, I don't like this indirection. Do we really need to make this dynamic? Why don't we register a NetworkService in the QemuDriver and only fill in the details later? That would be easier than indirecting inside the driver for the function call. As long as the QemuDriver fills in the details before the SSHDriver is activated this should be fine.

JoshuaWatt: Would that mean that the QemuDriver binds to the NetworkService (it can't(?) be the other way around because NetworkService is a Resource).

I think the reason I avoided that was because it seemed weird to have a NetworkService that was a complete lie in the YAML file, then expect QEmuDriver to "fix" it; this is a little more indirect but at least there is no misleading the user about what's going on in the YAML file. The SSHDriver binds to the QEmuDriver which is responsible for telling it the correct details about how to connect.

If we wanted to go that route, maybe it would make more sense to have some.... IndirectNetworkService that doesn't require lies in the YAML file and QEmuDriver and SSDriver can bind to that instead?

Emantor: Programatically it's totally possible to create another resource from a driver by simply binding it with
```
NetworkService(target,…)
```
But I agree that it is very non-obvious that there is an available NetworkService for SSH to connect to in this case.
My ideal vision would be that the QemuDriver exposes multiple properties, because there can be multiple serial ports/network ports/other emulated devices. However with the current architecture we don't support binding a driver multiple times (to expose the property directly via QemuDriver) and the `DynamicResource` concept is new and departs from the strict declarative style of the current resource/driver handling. Maybe the `IndirectNetworkService` is a solution for that. On the other hand we need to parse the qemu commandline either way and might as well create the resources directly. I have no clear idea how to solve this, I'll keep it in mind and think on it for a bit.

JoshuaWatt: Ok. I thought I would give a little more background on why I went this route as some food for thought. With this patch, it's now possible to define a SSH connection to QEMU instance using user mode networking with something like:
```yaml
- QEMUDriver:
    ...
    nic: user,model=virtio-net-pci,net=192.168.0.0/24,dhcpstart=192.168.0.1

# Binds to QEMUDriver
- QEMUNetworkService:
    address: 192.168.0.1
    username: root

# Binds to QEMUNetworkService
- SSHDriver: {}
```

I don't feel like it's particularly unclear what's going on here: the fact that SSHDriver is use a "hidden" NetworkService is more or less an implementation detail (which can probably be eliminated if you think it makes it more clear what is going on); the SSHDriver is still binding to QEMUNetworkService just like it would NetworkService. From the user's perspective, I don't really thing this is unexpected.

My other motive for doing things is that we have devices where it's simply not possible to know the IP address of the device statically in the configuration. Because of this, we've had to make a custom network service that knows how to get the IP address from an interface, and then a custom SSH driver that binds to this, so we end up with:
```yaml
- CustomNetworkInterfaceDriver:
    iface: eth0

# Binds to CustomNetworkInterfaceDriver
- CustomSSHServiceDriver: {}
```
Our `CustomSSHServiceDriver` fabricates can fabricate a correct `SSHDriver` on demand, but this has a few problems:
1. The SSHDriver can't live in the env.yaml file, so _other_ drivers can't statically bind to it
2. Our fixture code has to deal with it specially, like:
```python
@pytest.fixture(scope="function")
def ssh_command(target, strategy, capsys):
    with capsys.disabled():
        strategy.transition("shell")

    try:
        ssh = target.get_driver("CustomSSHServiceDriver").get_ssh()
        target.activate(ssh)
        return ssh
    except NoDriverFoundError:
        pass

    try:
        return target.get_driver("SSHDriver")
    except NoDriverFoundError:
        pytest.skip("No SSHDriver found")
```

Basically, this ends up snowballing where all downstream drivers now have to know about `CustomSSHServiceDriver` which is really messy. With this change, this all goes away though because we can make `CustomNetworkInterfaceDriver` inherit from `DynamicNetworkServiceProtocol` and suddenly a normal `SSHDriver` can bind to it and nothing downstream really needs to care about it.

PR #981, opened 2022-09-02
util: expect: implement sendcontrol()

Bastian-Krause:
**Description**
Sometimes it's useful to send Ctrl-C to interrupt a process or Ctrl-Z to send a hung process to the background and kill it then (e.g. gst-validate-1.0). This is often faster than transitioning to the off/root state and back to the shell.

Although pexpect.spawn has a sendcontrol() method, it relies on ptyprocess which is not useful for labgrid's usecase. So implement it here.

Note: was originally part of #552.

**Checklist**
- [ ] Tests for the feature 
- [x] PR has been tested

### DIFF
--- a/labgrid/util/expect.py
+++ b/labgrid/util/expect.py
@@ -1,3 +1,5 @@
+import string
+
 import pexpect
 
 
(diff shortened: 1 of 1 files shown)


Bastian-Krause:
Force-pushed:
- rebased on master
- dropped CHANGES.rst entry (see #990)

labgrid/util/expect.py
```diff
@@ -22,6 +24,15 @@ def send(self, s):
         b = s
         return self.driver.write(b)
 
+    def sendcontrol(self, char):
+        """Send control character to underlying transport, e.g. Ctrl-C or Ctrl-Z."""
+        char = char.lower()
+        try:
+            ord_ = string.ascii_lowercase.index(char.lower()) + 1
```
jluebbe, line 31: The second call to .lower() is redundant.

jluebbe (changes requested):
Looks fine otherwise.

labgrid/util/expect.py
Bastian-Krause: Fixed.

PR #984, opened 2022-09-05
sshdriver: escape password when writing to pass_file

S4mw1s3:
This avoids shell interpretation in case the password contains special
characters.


Bastian-Krause (approved):
Thanks for your contribution!

Emantor (approved):

Emantor:
Thanks!

PR #987, opened 2022-09-12
Fix simplerest

cbs101:
Hi

We want the python3-powerrelay from meta-labgrid to work. python3-powerrelay has always worked using PUT, and not get, for setting power. Hence patching is needing either in python3-powerrelay or labgrid. Proposing to do the change in labgrid.

If someone are using an alternative to python3-powerrelay that uses GET, I would like to know so we can switch to using that instead.

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -198,7 +198,7 @@ Currently available are:
 
 ``simplerest``
   This is a generic backend for PDU implementations which can be controlled via
-  HTTP GET requests (both set and get).
+  HTTP GET and PUT requests.
   See the `docstring in the module
   <https://github.com/labgrid-project/labgrid/blob/master/labgrid/driver/power/simplerest.py>`__
   for details.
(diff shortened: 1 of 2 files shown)


Bastian-Krause:
Hi, did you try the `rest` power backend?

> **rest**
> This is a generic backend for PDU implementations which can be controlled via HTTP PUT and GET requests. See the [docstring in the module](https://github.com/labgrid-project/labgrid/blob/master/labgrid/driver/power/rest.py) for details.

Source: https://labgrid.readthedocs.io/en/latest/configuration.html#networkpowerport

Bastian-Krause:
@cbs101 Did the `rest` power backend work for you? If so, I'd like to close this.

cbs101:
Hi @Bastian-Krause 

Yes we got it working.
I'll submit a patch to meta-labgrid to get python3-powerrelay with the rest interface as soon as I get the time to do it :)

Bastian-Krause:
Thanks.

PR #994, opened 2022-09-14
Fix relevant pylint warnings, run basic pylint checks in CI

Bastian-Krause:
**Description**
The pylint results were not considered consequently because the amount of messages is pretty high and we did not follow the rules set in our pylintrc. This is why the CI jobs were never configured to run pylint (actually breaking a PR build job on issues).

On the other hand, pylint can reveal important issues. These issues should be addressed before a PR is merged. In the past, this was done on some PRs manually.

Let's change that: Fix relevant pylint warnings. Then configure a basic set of checks in pyproject.toml (no styling checks though, since we want to enforce [black](https://black.readthedocs.io/en/stable/) code style in a future PR). Finally let pylint run these checks in CI.

**Checklist**
- [x] PR has been tested


Bastian-Krause:
Force-pushed:
- rebased on current master
- drop `--rcfile=.pylintrc` from tox.ini

jluebbe (approved):

Emantor:
Maybe we should move the pylint step out of the unittests and into its own CI run. This way there is no confusion between the tests failing and the code having linting errors. AFAICS all the installation steps are not required for pylint anyway.

Bastian-Krause:
> Maybe we should move the pylint step out of the unittests and into its own CI run. This way there is no confusion between the tests failing and the code having linting errors.

I can think of cases where code uses a Python feature that is not available in all supported Python versions. pylint won't find this if it is not run against all supported Python versions.

> AFAICS all the installation steps are not required for pylint anyway.

We still need to install the dependencies. With #1001 merged there is no separation between installing the dependencies and installing labgrid.

While I see your point, considering the arguments above, I don't think it's a good idea to have a separate CI workflow for pylint.

Emantor (approved):

Emantor (approved):
Makes sense, thanks for the explanation.

PR #999, opened 2022-09-19
requirements/setup.cfg: drop yapf

Bastian-Krause:
**Description**
We do not use yapf in our current workflow. To improve code style consistency and reduce discussions about style, we want to switch to [black](https://github.com/psf/black) in a future PR and enforce it via CI step by step.

For now, only drop yapf.


Emantor (approved):

Emantor:
Thanks!

PR #1001, opened 2022-09-20
Switch to pyproject.toml

Bastian-Krause:
**Description**
We decided to move from setup.{py,cfg}/requirements to pyproject.toml.

**Checklist**
- [x] PR has been tested


Bastian-Krause:
- [ ] coverage decrease is unexpected, needs investigation

Bastian-Krause:
> * [x]  coverage decrease is unexpected, needs investigation

pytest-cov only works in editable mode: https://github.com/pytest-dev/pytest-cov/issues/388

So force-pushed:
- re-added the `-e` in `pip install -e .[dev]` in GitHub actions workflow to collect coverage correctly

labgrid/driver/serialdriver.py
```diff
@@ -27,12 +26,6 @@ class SerialDriver(ConsoleExpectMixin, Driver, ConsoleProtocol):
         bindings = {"port": "SerialPort", }
     else:
         bindings = {"port": {"SerialPort", "NetworkSerialPort"}, }
-    if version.parse(serial.__version__) != version.Version('3.4.0.1'):
-        message = ("The installed pyserial version does not contain important RFC2217 fixes.\n"
-                   "You can install the labgrid fork via:\n"
-                   "pip uninstall pyserial\n"
-                   "pip install https://github.com/labgrid-project/pyserial/archive/v3.4.0.1.zip#egg=pyserial\n")  # pylint: disable=line-too-long
-        warnings.warn(message)
```
jluebbe, line 35: > labgrid's pyserial fork is now referenced in the dependencies, so there
> should be no scenario left where the user needs to intervene manually.

This is only correct when using pip to install the package. If a user relies on a distribution (perhaps Debian/Ubuntu or OE), this can still fail.

pyproject.toml
```diff
@@ -1,3 +1,138 @@
 [build-system]
 # Minimum requirements for the build system to execute.
-requires = ["setuptools", "wheel", "setuptools_scm"]
+requires = [
+    "setuptools>=47.2.0",
+    "wheel",
+    "setuptools_scm[toml]",
```
jluebbe, line 6: Why don't we need to add 'pip>=21.2.0' here as described in the commit message?

dockerfiles/Dockerfile
```diff
@@ -23,8 +23,7 @@ ARG VERSION
 RUN set -e ;\
     cd /opt/labgrid ;\
     pip3 install yq ;\
-    pip3 install --no-cache-dir -r requirements.txt ;\
-    SETUPTOOLS_SCM_PRETEND_VERSION="$VERSION" python3 setup.py install ;\
+    SETUPTOOLS_SCM_PRETEND_VERSION="$VERSION" pip3 install --no-cache-dir . ;\
```
jluebbe, line 26: Why is `--no-cache-dir .` needed?

dockerfiles/README.rst
```diff
@@ -33,6 +33,7 @@ with the included script
 
 .. code-block:: bash
 
+   $ pip install --upgrade pip setuptools setuptools_scm wheel
```
jluebbe, line 36: Shouldn't this be covered by the build-system.requires property in pyproject.toml?

.crossbar/config.yaml
```diff
@@ -35,6 +35,7 @@ workers:
     endpoint:
       type: tcp
       port: 20408
+      user_timeout: 5000
```
jluebbe, line 38: Is this a behavior change triggered by the new crossbar?
Note that TCP_USER_TIMEOUT also applies during the normal connection lifetime, not just shutdown.

We already ping the exporters regularly and kick them from the crossbar on timeout. Is this broken with the new crossbar release?

jluebbe (changes requested):
Thanks for the work! I have a few comments, though.

dockerfiles/Dockerfile
Bastian-Krause: No idea, it was there before, so I left it there.

labgrid/driver/serialdriver.py
Bastian-Krause: True, I didn't think of this case. I'll leave it there and update the latest version.

pyproject.toml
Bastian-Krause: I tried adding that and built on a Debian system shipping pip 20.3.4, but it has no effect. The old pip is used for installation.

See also https://github.com/pypa/pip/issues/4145

dockerfiles/README.rst
Bastian-Krause: For `pip`, see above. `setuptools_scm` is only installed as a build requirement, which won't cover `VERSION="$(python -m setuptools_scm)"` in `dockerfiles/build.sh` - so this is should probably be added to the `docker` extra, right? `setuptools` and `wheel` can probably be dropped here.

.crossbar/config.yaml
Bastian-Krause: Context is https://github.com/crossbario/crossbar/pull/1930.

> Is this a behavior change triggered by the new crossbar?

I don't think so.

> Note that TCP_USER_TIMEOUT also applies during the normal connection lifetime, not just shutdown.
> 
> We already ping the exporters regularly and kick them from the crossbar on timeout. Is this broken with the new crossbar release?

This is especially useful for [LXA Test Automation Controller](https://www.linux-automation.com/en/products/lxa-tac.html), which might not get shut down cleanly. The TCP connection is considered alive for quite a long time.

I thought that this might be a sane default. Do you have concerns?

labgrid/driver/serialdriver.py
Bastian-Krause: Done.

pyproject.toml
Bastian-Krause: I've added this information to the commit message.

dockerfiles/README.rst
Bastian-Krause: Dropped `setuptools` and `wheel` in the latest PR revision.

Bastian-Krause:
Force-pushed:
- rebased on master
- fixed typos in commit messages
- dropped unnecessary `pip install` statements from the CI workflow and documentation
- extended commit message with https://github.com/pypa/pip/issues/4145
- re-add/update documentation/warning in case other pyserial version than https://github.com/labgrid-project/pyserial/releases/tag/v3.5.0.1 are used (maybe we should allow the previous v3.4.0.1 version for non-pip installs as well?)

dockerfiles/Dockerfile
benjamb: Presumably because it avoids taking up unnecessary space, thus keeping the docker image that little bit smaller.

.crossbar/config.yaml
jluebbe: This seems unrelated to the pyproject.toml migration.

I'm still not clear on why we'd need both the TCP timeout and the ping/exporter-kick at the same time.

jluebbe:
Discussed this @Bastian-Krause: As the installation time of 1-2 hours in CI is not acceptable, we should split the move to pyproject.toml from the dependency version changes. Also, we should avoid self-dependencies in the extras, as the failure mode of older pip versions (such as the one in debian stable) in this case is very confusing.

Bastian-Krause:
Force-pushed:
- rebased
- dropped dependency version bumps
- dropped dependency version relaxations, use pinned versions instead (like we did before)
- dropped tox workarounds to force new pip
- dropped commit: crossbar: config: set user_timeout in default config (since we don't bump crossbar's version)
- dropped release doc adjustment regarding dependency version relaxations
- added commit: tox: pass TERM environment variable to test env
- added commit: debian: use recent pip, drop obsolete setuptools requirement
- quoted pip installs with extras in the docs (due to square brackets)

.crossbar/config.yaml
Bastian-Krause: I dropped this because we decided to drop the version bumps and dependency relaxations in this particular PR altogether.

Bastian-Krause:
Hm, this seems to break coverage. No idea why, yet.

jluebbe:
> Hm, this seems to break coverage. No idea why, yet.
> 
> Edit: coverage is also broken on master, so that's unrelated.

I worked in https://github.com/labgrid-project/labgrid/actions/runs/3345404124/jobs/5540884396, so perhaps the update to pytest-cov-4.0.0 fixed it.

jluebbe (approved):

Bastian-Krause:
> > Hm, this seems to break coverage. No idea why, yet.
> > Edit: coverage is also broken on master, so that's unrelated.
> 
> It worked in https://github.com/labgrid-project/labgrid/actions/runs/3345404124/jobs/5540884396, so perhaps the update to pytest-cov-4.0.0 fixed it.

See https://github.com/labgrid-project/labgrid/pull/1029#issuecomment-1300692866.

Bastian-Krause:
Rebased (due to previous conflicts caused by #1032).

jluebbe (approved):

Emantor (approved):

Emantor:
Thanks a lot!

PR #1003, opened 2022-09-21
drivers/sshdriver: add support for the SFTP protocol for file transfers

ekronborg:
Recent versions of OpenSSH allow using the SFTP protocol in favor of the deprecated SCP protocol, and from OpenSSH 9, SCP will use the SFTP protocol by default. Therefore, `put()` and `get()` can now be configured to use SFTP by setting `sftp` to true for `SSHDriver`.

**Description**
I tested the change locally using an image that only supports SFTP. Without this change, I am unable to transfer files using `put()` and `get()`. However, applying this change, the methods work as expect.

**Checklist**
- [ ] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -1491,6 +1491,8 @@ Arguments:
     stdout, and an empty list as second element.
   - connection_timeout (float, default=30.0): timeout when trying to establish connection to
     target.
+  - explicit_sftp_mode (bool, default=False): if set to True, `put()` and `get()` will
+    explicitly use the SFTP protocol for file transfers instead of scp's default protocol
 
 UBootDriver
 ~~~~~~~~~~~
--- a/labgrid/driver/sshdriver.py
+++ b/labgrid/driver/sshdriver.py
@@ -2,6 +2,7 @@
 import contextlib
 import logging
 import os
+import re
 import stat
 import shlex
 import shutil
@@ -32,6 +33,7 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
     keyfile = attr.ib(default="", validator=attr.validators.instance_of(str))
     stderr_merge = attr.ib(default=False, validator=attr.validators.instance_of(bool))
     connection_timeout = attr.ib(default=float(get_ssh_connect_timeout()), validator=attr.validators.instance_of(float))
+    explicit_sftp_mode = attr.ib(default=False, validator=attr.validators.instance_of(bool))
 
     def __attrs_post_init__(self):
         super().__attrs_post_init__()
@@ -397,12 +399,28 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
         """The SSHDriver is always connected, return 1"""
         return 1
 
+    def _scp_supports_explicit_sftp_mode(self):
+        version = subprocess.run(["ssh", "-V"], capture_output=True, text=True)
+        version = re.match(r"^OpenSSH_(\d+)\.(\d+)", version.stderr)
+        major, minor = map(int, version.groups())
+
+        # OpenSSH >= 8.6 supports explicitly using the SFTP protocol via -s
+        if major == 8 and minor >= 6:
+            return True
+        # OpenSSH >= 9.0 default to the SFTP protocol
+        if major >= 9:
+            return False
+        raise Exception(f"OpenSSH version {major}.{minor} does not support explicit SFTP mode")
+
     @Driver.check_active
     @step(args=['filename', 'remotepath'])
     def put(self, filename, remotepath=''):
+        ssh_prefix = self.ssh_prefix
+        if self.explicit_sftp_mode and self._scp_supports_explicit_sftp_mode():
+            ssh_prefix.append("-s")
         transfer_cmd = [
             "scp",
-            *self.ssh_prefix,
+            *ssh_prefix,
             "-P", str(self.networkservice.port),
             "-r",
             filename,
@@ -425,9 +443,12 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
     @Driver.check_active
     @step(args=['filename', 'destination'])
     def get(self, filename, destination="."):
+        ssh_prefix = self.ssh_prefix
+        if self.explicit_sftp_mode and self._scp_supports_explicit_sftp_mode():
+            ssh_prefix.append("-s")
         transfer_cmd = [
             "scp",
-            *self.ssh_prefix,
+            *ssh_prefix,
             "-P", str(self.networkservice.port),
             "-r",
             f"{self.networkservice.username}@{self.networkservice.address}:{filename}",


Emantor (dismissed):

labgrid/driver/sshdriver.py
```diff
@@ -392,9 +393,12 @@ def get_status(self):
     @Driver.check_active
     @step(args=['filename', 'remotepath'])
     def put(self, filename, remotepath=''):
+        ssh_prefix = self.ssh_prefix
+        if self.sftp:
+            ssh_prefix.append("-s")
         transfer_cmd = [
             "scp",
-            *self.ssh_prefix,
+            *ssh_prefix,
```
Bastian-Krause, line 401: If I read this correctly, you are passing `-s` to scp. Looking at `man scp`, I don't see such an option. I guess you're referring to ssh's `-s` here? No idea how this should work, though. How exactly did you test this?

labgrid/driver/sshdriver.py
```diff
@@ -417,9 +421,12 @@ def put(self, filename, remotepath=''):
     @Driver.check_active
     @step(args=['filename', 'destination'])
     def get(self, filename, destination="."):
+        ssh_prefix = self.ssh_prefix
+        if self.sftp:
+            ssh_prefix.append("-s")
         transfer_cmd = [
             "scp",
-            *self.ssh_prefix,
+            *ssh_prefix,
```
Bastian-Krause, line 429: Same as above.

labgrid/driver/sshdriver.py
ekronborg: I just checked on my up-to-date Arch machine and `-s` indeed does not exist for `scp` here. However, on my other machine `-s` does exist. According the the release notes from OpenSSH, `-s` is removed in version 9 (https://www.openssh.com/releasenotes.html#9.0). So this will be fine on targets with OpenSSH < 9. However, it will probably not work on targets with OpenSSH > 9, since `scp` uses the `sftp` protocol by default here.

Sorry I missed this detail. What do you think the best solution is to handle both protocols?

Bastian-Krause: > I just checked on my up-to-date Arch machine and `-s` indeed does not exist for `scp` here. However, on my other machine `-s` does exist. According the the release notes from OpenSSH, `-s` is removed in version 9 (https://www.openssh.com/releasenotes.html#9.0). So this will be fine on targets with OpenSSH < 9. However, it will probably not work on targets with OpenSSH > 9, since `scp` uses the `sftp` protocol by default here.

Are you sure about this? I tested `scp` from Debian's `openssh-client` package in versions 8.4 and 9.0. Neither have the `-s` option.

> Sorry I missed this detail. What do you think the best solution is to handle both protocols?

`-o SessionType=subsystem` should work, right?

Bastian-Krause: > > I just checked on my up-to-date Arch machine and `-s` indeed does not exist for `scp` here. However, on my other machine `-s` does exist. According the the release notes from OpenSSH, `-s` is removed in version 9 (https://www.openssh.com/releasenotes.html#9.0). So this will be fine on targets with OpenSSH < 9. However, it will probably not work on targets with OpenSSH > 9, since `scp` uses the `sftp` protocol by default here.
> 
> Are you sure about this? I tested `scp` from Debian's `openssh-client` package in versions 8.4 and 9.0. Neither have the `-s` option.

Oh, that's not true. scp from `openssh-client` 9.0 has it, but the man page seems outdated. We should still stick to:

> > Sorry I missed this detail. What do you think the best solution is to handle both protocols?
> 
> `-o SessionType=subsystem` should work, right?

ekronborg: > I guess this will still be around after 2022..

Fingers crossed!

I tried substituting `-s` for `-o SessionType=subsystem`,  but it doesn't work even though the documentation states that it should be the same. I will troubleshoot more tomorrow.

Bastian-Krause: Could you push your WIP state so I can have a look?

ekronborg: Done

Bastian-Krause: No luck here either:
```
$ scp -v -F /dev/null -o SessionType=subsystem somefile root@target:/tmp/
subsystem request failed on channel 0
lost connection
```
On the target sshd side:
```
subsystem request for scp -v -t /tmp/ by user root failed, subsystem not found
```

So `-s` and `-o SessionType=subsystem` do not seem to be interchangeable.

In this case I'd opt for passing `-s` and rename the SSHDriver's `sftp` attribute to something like `explicit_sftp_mode`. We should document that this option should only by passed for OpenSSH version >=8.6, <9 and that the scp flag will be removed in 2022.

ekronborg: I agree with this approach and I just pushed some new changes.

ekronborg: @Bastian-Krause Did you have a chance to look at these changes and [1009](https://github.com/labgrid-project/labgrid/pull/1009) yet?

labgrid/driver/sshdriver.py
```diff
@@ -389,12 +391,28 @@ def get_status(self):
         """The SSHDriver is always connected, return 1"""
         return 1
 
+    def _ssh_prefix_append_s(self):
```
Bastian-Krause, line 394: The method name is a bit off, it does not append anything, but rather determines whether we can append `-s`. Maybe something like..

```suggestion
    def _scp_supports_explicit_sftp_mode(self):
```

labgrid/driver/sshdriver.py
```diff
@@ -389,12 +391,28 @@ def get_status(self):
...
+        version = re.match(r'^OpenSSH_(\d+)\.(\d+)(?:\.\d+){0,2}\w*', version.stderr)
+        major = int(version[1])
+        minor = int(version[2])
+
+        if major == 8 and minor >= 6:
+            return True
+        elif major >=9:
+            return False
+        else:
+            raise Exception(f'OpenSSH version {major}.{minor} is not supported. It must be at least version 8.6')
```
Bastian-Krause, line 405:
```suggestion
        # OpenSSH >= 8.6 supports explicitly using the SFTP protocol via -s
        if major == 8 and minor >= 6:
            return True
        # OpenSSH >= 9.0 default to the STFTP protocol
        if major >= 9:
            return False
        raise Exception(f"OpenSSH version {major}.{minor} does not support explicit SFTP mode")
```
Replaced an unnecessary `elif` with `if` and dropped an `else`. Added comments to the cases and simplified the exception.

Bastian-Krause (changes requested):
Thanks for working on this! Here's some feedback.

CHANGES.rst
```diff
@@ -29,6 +29,8 @@ New Features in 0.5.0
 - Support for QEMU Q35 machine added.
 - `UBootDriver` now handles idle console, allowing driver activation on
   an interupted U-Boot.
+- The `SSHDriver` now supports using the SFTP protocol for file transfers as an alternative
+  to the deprecated SCP protocol.
```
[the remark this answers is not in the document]
ekronborg: OK

doc/configuration.rst
```diff
@@ -1443,6 +1443,8 @@ Arguments:
     (has precedence over `NetworkService`'s password)
   - stderr_merge (bool, default=False): set to True to make `run()` return stderr merged with
       stdout, and an empty list as second element.
+  - explicit_sftp_mode (bool, default=False): if set to True, `put()` and `get()` will use the
+    SFTP protocol for file transfers instead of the legacy SCP protocol
```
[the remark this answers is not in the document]
ekronborg: OK

labgrid/driver/sshdriver.py
```diff
@@ -389,12 +391,28 @@ def get_status(self):
         """The SSHDriver is always connected, return 1"""
         return 1
 
+    def _ssh_prefix_append_s(self):
+        version = subprocess.run("ssh -V", capture_output=True, shell=True, text=True)
```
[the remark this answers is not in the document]
ekronborg: You are right.

labgrid/driver/sshdriver.py
```diff
@@ -389,12 +391,28 @@ def get_status(self):
...
+        version = re.match(r'^OpenSSH_(\d+)\.(\d+)(?:\.\d+){0,2}\w*', version.stderr)
+        major = int(version[1])
+        minor = int(version[2])
```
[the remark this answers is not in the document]
ekronborg: Yes. I might have went a bit overboard.

labgrid/driver/sshdriver.py
ekronborg: I agree. Let's go with your suggestion.

labgrid/driver/sshdriver.py
ekronborg: OK for me.

labgrid/driver/sshdriver.py
```diff
@@ -389,12 +391,28 @@ def get_status(self):
...
+    def _scp_supports_explicit_sftp_mode(self):
+        version = subprocess.run(["ssh", "-V"], capture_output=True, text=True)
+        version = re.match(r"^OpenSSH_(\d+)\.(\d+)", version.stderr)
+        major, minor = map(int, version.groups())
+
+        # OpenSSH >= 8.6 supports explicitly using the SFTP protocol via -s
+        if major == 8 and minor >= 6:
+            return True
+        # OpenSSH >= 9.0 default to the SFTP protocol
+        if major >=9:
```
Bastian-Krause, line 403:
```suggestion
        if major >= 9:
```
Small nitpick, rest looks good.

ekronborg: I'm annoyed I missed that one. It's fixed now.

doc/configuration.rst
```diff
@@ -1443,6 +1443,8 @@ Arguments:
     (has precedence over `NetworkService`'s password)
   - stderr_merge (bool, default=False): set to True to make `run()` return stderr merged with
       stdout, and an empty list as second element.
+  - explicit_sftp_mode (bool, default=False): if set to True, `put()` and `get()` will
+    explicitly use the SFTP protocol for file transfers instead of the default SCP protocol
```
Bastian-Krause, line 1447:
```suggestion
    explicitly use the SFTP protocol for file transfers instead of the default protocol
```
or:
```suggestion
    explicitly use the SFTP protocol for file transfers instead of scp's default protocol
```

I did not notice this earlier: "default SCP protocol" is still misleading. The default protocol might be SFTP, depending on the version.

ekronborg: I guess we could be even more specific and write something like "... explicitly use the SFTP protocol for file transfers instead of scp's default protocol if 8.6 >= OpenSSH < 9.0". I am not sure if this will cause confusion though. Maybe it is better to keep the description simple like your suggestion. What do you think?

Bastian-Krause: It's okay for me either way.

Bastian-Krause (approved):
Thanks!

Bastian-Krause:
Force-pushed:
- rebased on latest master

Bastian-Krause (dismissed):

Emantor (dismissed):

ekronborg:
@jluebbe Hi Jan. Did you have a chance to look at this PR and [this PR](https://github.com/labgrid-project/labgrid/pull/1009) yet?

Emantor (approved):

jluebbe (approved):
Hmm, it seems this PR has no way to request SCP mode for older targets that do not support SFTP (since 9.0, scp supports a `-O` option to force SCP mode: https://man.openbsd.org/scp.1).

That change wait until that's actually needed, though.

ejoerns:
> Hmm, it seems this PR has no way to request SCP mode for older targets that do not support SFTP (since 9.0, scp supports a `-O` option to force SCP mode: https://man.openbsd.org/scp.1).
> 
> That change wait until that's actually needed, though.

I had this requirement due to an update of my environment to Debian Bookwork combined with an old non-sftp target. I have drafted a solution for this in #1281

PR #1004, opened 2022-09-22
client: show names for multiple matching resources

Emantor:
**Description**

When matching multiple resources but not having a name, we raise an unintuitive error about not finding a resource, but don't indicate the available names. Bubble up the found resources and provide a neat list to the user to select from the available names.

**Checklist**
- [x] PR has been tested

### DIFF
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -1957,7 +1957,12 @@ def main():
                 traceback.print_exc()
             else:
                 print(f"{parser.prog}: error: {e}", file=sys.stderr)
-            print("This may be caused by disconnected exporter or wrong match entries.\nYou can use the 'show' command to review all matching resources.", file=sys.stderr)  # pylint: disable=line-too-long
+            if e.found:
+                print("Found multiple resources but no name was given, available names:")
+                for res in e.found:
+                    print(f"{res.name}")
+            else:
+                print("This may be caused by disconnected exporter or wrong match entries.\nYou can use the 'show' command to review all matching resources.", file=sys.stderr)  # pylint: disable=line-too-long
             exitcode = 1
         except NoDriverFoundError as e:
             if args.debug:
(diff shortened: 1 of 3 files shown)


labgrid/remote/client.py
```diff
@@ -1958,6 +1958,10 @@ def main():
             else:
                 print(f"{parser.prog}: error: {e}", file=sys.stderr)
             print("This may be caused by disconnected exporter or wrong match entries.\nYou can use the 'show' command to review all matching resources.", file=sys.stderr)  # pylint: disable=line-too-long
+            if e.found:
+                print("Found multiple resources but no name was given, available names:")
+                for res in e.found:
+                    print(f"{res.name}")
```
Bastian-Krause, line 1964: I think it would be better to implement this in `__str__()` of `NoResourceFoundError`. Suggestion:
```python
def __str__(self):
    if self.found:
        names = [res.name for res in self.found]
        return f"{', '.join(self.args)}; available names: {', '.join(names)}"

    return self.message
```
Output looks like this:
```
$ labgrid-client --place myplace console
labgrid-client: error: multiple resources matching <class 'labgrid.resource.serialport.NetworkSerialPort'> found in Target(name='myplace', env=None); available names: console, acm
```
Maybe we could show the class name there instead of its whole representation for even better readability?

Bastian-Krause (commented):
I like the general idea. See the suggestion below.

labgrid/remote/client.py
Emantor: I think there is a case to be made for both. As a library user descriptive error cases are helpful and including the names definitely helps there, but for client usage I really want to enumerate the individual names on different lines since it makes it a lot more explicit and user friendly.

Bastian-Krause: Okay, agreed. Is the "This may be caused by disconnected exporter" message needed in that case or should that happen in an `else` path? This kind of error generates seems to generate a whole mixture of helpful and not so helpful messages. Maybe we can drop some for the multiple resources case?

Emantor: Good point, we should do that since it wasn't a connection error.

Bastian-Krause (approved):
Looks good.

PR #1005, opened 2022-09-22
processwrapper: Add input argument

JoshuaWatt:
**Description**

Adds an input argument that allows a string to be passed as input to the process, matching subprocess.run().

**Checklist**
- [x] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/labgrid/util/helper.py
+++ b/labgrid/util/helper.py
@@ -4,6 +4,7 @@ import logging
 import pty
 import select
 import subprocess
+import errno
 from socket import socket, AF_INET, SOCK_STREAM
 from contextlib import closing
 
@@ -26,21 +27,34 @@ def get_user():
     import pwd
     return pwd.getpwuid(os.getuid())[0]
 
+def set_nonblocking(fd):
+    flags = fcntl.fcntl(fd, fcntl.F_GETFL)
+    fcntl.fcntl(fd, fcntl.F_SETFL, flags | os.O_NONBLOCK)
+
 @attr.s
 class ProcessWrapper:
     callbacks = attr.ib(default=attr.Factory(list))
     loglevel = logging.INFO
 
     @step(args=['command'], result=True, tag='process')
-    def check_output(self, command, *, print_on_silent_log=False):
+    def check_output(self, command, *, print_on_silent_log=False, input=None): # pylint: disable=redefined-builtin
         """Run a command and supply the output to callback functions"""
         logger = logging.getLogger("Process")
         res = []
         mfd, sfd = pty.openpty()
-        flags = fcntl.fcntl(mfd, fcntl.F_GETFL)
-        fcntl.fcntl(mfd, fcntl.F_SETFL, flags | os.O_NONBLOCK)
+        set_nonblocking(mfd)
+
+        kwargs = {}
+
+        stdin_r = None
+        stdin_w = None
+        if input is not None:
+            stdin_r, stdin_w = os.pipe()
+            kwargs['stdin'] = stdin_r
+            set_nonblocking(stdin_w)
+
         process = subprocess.Popen(command, stderr=sfd,
-                                   stdout=sfd, bufsize=0)
+                                   stdout=sfd, bufsize=0, **kwargs)
 
         logger.log(ProcessWrapper.loglevel, "[%d] command: %s", process.pid, " ".join(command))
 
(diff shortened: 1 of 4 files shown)


Emantor (dismissed):
Overall design looks good to me.

labgrid/util/helper.py
```diff
@@ -26,21 +27,34 @@ def get_user():
...
+    fcntl.fcntl(fd, fcntl.F_SETFL, flags | os.O_NONBLOCK)
+
 @attr.s
 class ProcessWrapper:
     callbacks = attr.ib(default=attr.Factory(list))
     loglevel = logging.INFO
 
     @step(args=['command'], result=True, tag='process')
-    def check_output(self, command, *, print_on_silent_log=False):
+    def check_output(self, command, *, print_on_silent_log=False, input=None):
```
Bastian-Krause, line 40:
```suggestion
    def check_output(self, command, *, print_on_silent_log=False, input=None):  # pylint: disable=redefined-builtin
```
Or is there variable name you could use instead? If not, just disable the warning.

labgrid/util/helper.py
```diff
@@ -51,33 +65,67 @@ def check_output(self, command, *, print_on_silent_log=False):
...
+        if stdin_r is not None:
+            os.close(stdin_r)
+
         # close sfd so we notice when the child is gone
         os.close(sfd)
         # get a file object from the fd
         buf = b""
+        read_fds = [mfd]
+        write_fds = []
+        except_fds = []
```
Bastian-Krause, line 77:
```suggestion
```
We can drop that, right?

labgrid/util/helper.py
```diff
@@ -51,33 +65,67 @@ def check_output(self, command, *, print_on_silent_log=False):
...
-                raise
-
-            if raw:
-                buf += raw
-                *parts, buf = buf.split(b'\r')
-                res.extend(parts)
-                for part in parts:
-                    for callback in self.callbacks:
-                        callback(part, process)
+            ready_r, ready_w, ready_x = select.select(read_fds, write_fds, except_fds, 0.1)
```
Bastian-Krause, line 82:
```suggestion
            ready_r, ready_w, _ = select.select(read_fds, write_fds, [], 0.1)
```
..and provide an empty list here, just like we did in the `select()` previously. Also drop another unused variable.

labgrid/util/helper.py
```diff
@@ -51,33 +65,67 @@ def check_output(self, command, *, print_on_silent_log=False):
...
+
+            if mfd in ready_r:
+                raw = None
+
+                try:
+                    raw = os.read(mfd, 4096)
+                except BlockingIOError as e:
```
Bastian-Krause, line 89:
```suggestion
                except BlockingIOError:
```
Drop another unused variable.

Bastian-Krause (changes requested):
Thanks! Looks good to me, just some minor suggestions. Please rebase (note that pylint warnings are now fatal in CI).

labgrid/util/helper.py
JoshuaWatt: `input` is the name of the argument in python `subprocess.check_output()` so I think we want to keep it the same?

labgrid/util/helper.py
JoshuaWatt: Done

labgrid/util/helper.py
JoshuaWatt: Done

labgrid/util/helper.py
Emantor: Yeah, disabling the warning is the correct move here, we want to keep it as is.

Bastian-Krause: Agreed.

Bastian-Krause (approved):

labgrid/util/helper.py
```diff
@@ -26,21 +27,34 @@ def get_user():
...
-        fcntl.fcntl(mfd, fcntl.F_SETFL, flags | os.O_NONBLOCK)
+        set_nonblocking(mfd)
+
+        kwargs = {}
+
+        stdin_r = None
+        stdin_w = None
+        if input is not None:
+            stdin_r, stdin_w = os.pipe()
+            kwargs['stdin'] = stdin_r
```
jluebbe, line 53: I'm probably missing something... Why can't we use the pty for stdin as well?

JoshuaWatt: In order to inject a string as stdin (which is what `input` does), you need to be able to independently close stdin after the input is consumed; think of a program like `cat` that terminates on input EOF. If you tied it into the the pty, you would also close stdout when you did that.

JoshuaWatt: Or put another way, `input` should be more like doing `echo <INPUT> | PROGRAM` rather than interactive (where I think you would do it through the pty).

jluebbe (approved):

PR #1009, opened 2022-09-29
dockerfiles: change debian to ubuntu as base image

ekronborg:
* Using `scp -s` with the SSHDriver requires OpenSSH >= 8.6, but Debian Bullseye only provides OpenSSH 8.4, while Ubuntu 22.04 provides OpenSSH 8.9. Therefore, changing the base image from Debian to Ubuntu allows using the SFTP protocol in containers. For more information, see https://github.com/labgrid-project/labgrid/pull/1003.
* This will increase the image sizes slightly. For example, the labgrid client image is 869 MB with Debian as base image, while it is 952 MB with Ubuntu as base image.

**Description**

**Checklist**
- [ ] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/dockerfiles/Dockerfile
+++ b/dockerfiles/Dockerfile
@@ -52,54 +52,6 @@ EXPOSE 20408
 CMD ["crossbar", "start", "--config", "/opt/labgrid/.crossbar/config.yaml"]
 
 #
-# ser2net
-#
-# These have to be built from source to bring in a newer version that has a
-# needed bugfix. This can be removed once the images are switched to
-# debian:bullseye, as it has a new enough version
-#
-FROM debian:buster-slim AS ser2net
-
-RUN apt update && \
-    apt install --yes --no-install-recommends \
-        build-essential \
-        cmake \
-        python3-dev \
-        wget \
-        ca-certificates \
-        libsctp-dev \
-        libssl-dev \
-        pkg-config \
-        file \
-        libyaml-dev \
-    && \
-    apt clean && \
-    rm -rf /var/lib/apt/lists/*
-
-RUN mkdir -p /src
-
-RUN cd /src && \
-    wget https://downloads.sourceforge.net/project/ser2net/ser2net/gensio-2.2.4.tar.gz && \
-    tar -xvzf gensio-2.2.4.tar.gz && \
-    cd gensio-2.2.4 && \
-    mkdir build && \
-    cd build && \
-    ../configure --prefix=/usr && \
-    make && \
-    make install && \
-    make install DESTDIR=/opt/ser2net
-
-RUN cd /src && \
-    wget https://downloads.sourceforge.net/project/ser2net/ser2net/ser2net-4.3.3.tar.gz && \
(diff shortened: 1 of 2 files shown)


krevsbech (dismissed):

Bastian-Krause (changes requested):
I have no opinion on the base image change, but there are a few things to consider:
- https://github.com/labgrid-project/labgrid/commit/30a0e38ca5f38a5e048b47a6d09371e411655112 can probably be dropped now
- what about the base image version in `dockerfiles/staging/dut/Dockerfile`?

ekronborg:
>     * [30a0e38](https://github.com/labgrid-project/labgrid/commit/30a0e38ca5f38a5e048b47a6d09371e411655112) can probably be dropped now
You are right. Joshua made this commit to build version 4.3.3. In Ubuntu 22.04, ser2net is version 4.3.4.

>     * what about the base image version in `dockerfiles/staging/dut/Dockerfile`?
I changed that as well.

Bastian-Krause (commented):
Please add a short explanation why you revert the commit, so the git history contains all necessary information and we do not need to dig into the PR comments later.

Apart from that, this looks okay to me. I'd still like a comment from @Emantor or @jluebbe.

ekronborg:
> Please add a short explanation why you revert the commit, so the git history contains all necessary information and we do not need to dig into the PR comments later.

I changed the commit message to include an explanation of the revert.

Bastian-Krause (approved):
Thanks. Note that I don't have an opinion on the debian:buster-slim -> ubuntu:22.04 move.

obbardc:
Why the move from debian -> ubuntu ?

Emantor:
@obbardc The reasoning concerning SSH support. However I am not sure that we want to switch over the default docker images. IMO the next debian release shouldn't be too far away and it's always possible to build your own containers to get the necessary software.

Bastian-Krause:
As a first step, I've bumped the docker base image to Debian bullseye: #1129. I guess Debian bookworm will become stable around July.

ekronborg:
> As a first step, I've bumped the docker base image to Debian bullseye: #1129. I guess Debian bookworm will become stable around July.

Bumping the image is fine, but please note that Bullseye still uses `OpenSSH 8.4`. Hopefully, it will be bumped to 8.6+ in the Bookworm release.

Bastian-Krause:
> > As a first step, I've bumped the docker base image to Debian bullseye: #1129. I guess Debian bookworm will become stable around July.
> 
> Bumping the image is fine, but please note that Bullseye still uses `OpenSSH 8.4`. Hopefully, it will be bumped to 8.6+ in the Bookworm release.

[Bookworm contains OpenSSH 9.2](https://packages.debian.org/bookworm/openssh-client).

jluebbe:
Bookworm will likely be released in less than a month, so we shouldn't switch all containers to Ubuntu just for that. If you need a newer OpenSSH, you can just build the container locally from bookworm even before the release.

PR #1014, opened 2022-10-01
Add NetworkInterface resource

Bastian-Krause:
**Description**
Until now only USB network interfaces are supported. The NetworkInterface class already exists, but there is no corresponding
resource export.

Change that by adding such a resource export and document the new NetworkInterface resource. Also mention that a RemoteNetworkInterface can either be a remote USBNetworkInterface or a NetworkInterface.

Now allow binding of the NetworkInterfaceDriver against it.

Note: Only tested on REPL by instantiating a NetworkInterface resource and binding a NetworkInterfaceDriver against it

**Checklist**
- [ ] Tests for the feature 
- [x] PR has been tested (by @nick-potenski, see below)

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -663,6 +663,19 @@ Arguments:
 Used by:
   - `DFUDriver`_
 
+NetworkInterface
+~~~~~~~~~~~~~~~~
+A NetworkInterface resource describes a network adapter (such as Ethernet or
+WiFi)
+
+.. code-block:: yaml
+
+   NetworkInterface:
+     ifname: eth0
+
+Arguments:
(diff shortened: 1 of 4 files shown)


nick-potenski:
I think the `NetworkInterface` class needs the `@target_factory.reg_resource` decorator, perhaps?  That seems to get things working on my end.

Bastian-Krause:
@nick-potenski Ah right, thanks for testing this :)

Force-pushed:
- register NetworkInterface with factory

nick-potenski:
@Bastian-Krause  - the update works for all my needs.  Let me know if there is anything else I can do to help here.

jluebbe (approved):

PR #1018, opened 2022-10-20
Fix ResourceWarnings

Bastian-Krause:
**Description**
Collection of fixes for various ResourceWarnings:
- unclosed files
- unclosed sockets
- unclosed pipes

Found by running labgrid-client and labgrid's pytest plugin in [Python Development Mode](https://docs.python.org/3/library/devmode.html) during day-to-day use as well as the test suite:
```
python3 -X dev -X tracemalloc=50 -m pytest
```

Test suite warnings are down from 90 warnings to 15. Only labgrid module code was touched, warnings in test code were ignored.

**Checklist**
- [.] PR has been tested (not all changes were manually tested)

### DIFF
--- a/labgrid/driver/qemudriver.py
+++ b/labgrid/driver/qemudriver.py
@@ -183,6 +183,11 @@ class QEMUDriver(ConsoleExpectMixin, Driver, PowerProtocol, ConsoleProtocol):
     def on_deactivate(self):
         if self.status:
             self.off()
+        if self._clientsocket:
+            self._clientsocket.close()
+            self._clientsocket = None
+        self._socket.close()
+        self._socket = None
         shutil.rmtree(self._tempdir)
 
     @step()
--- a/labgrid/driver/qemudriver.py
+++ b/labgrid/driver/qemudriver.py
@@ -87,10 +87,10 @@ class QEMUDriver(ConsoleExpectMixin, Driver, PowerProtocol, ConsoleProtocol):
             return
         self._child.terminate()
         try:
-            self._child.wait(1.0)
+            self._child.communicate(timeout=1)
         except subprocess.TimeoutExpired:
             self._child.kill()
-            self._child.wait(1.0)
+            self._child.communicate(timeout=1)
 
     def on_activate(self):
         self._tempdir = tempfile.mkdtemp(prefix="labgrid-qemu-tmp-")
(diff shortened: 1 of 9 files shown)


labgrid/util/ssh.py
```diff
@@ -472,9 +476,7 @@ def _stop_keepalive(self):
             self._keepalive.communicate(timeout=60)
         except subprocess.TimeoutExpired:
             self._keepalive.kill()
-
-        try:
-            self._keepalive.wait(timeout=60)
+            self._keepalive.communicate()
```
jluebbe, line 479: Is dropping the timeout here intentional?

labgrid/driver/qemudriver.py
```diff
@@ -183,6 +183,9 @@ def on_activate(self):
     def on_deactivate(self):
         if self.status:
             self.off()
+        if self._clientsocket:
+            self._clientsocket.close()
+        self._socket.close()
```
jluebbe, line 188: We should set them to `None` after closing, to match the initial state.

jluebbe (changes requested):
I find the `communicate()` calls a bit confusing at the points where the process is no longer expected to be running, but it seems simpler than the alternatives.

labgrid/util/ssh.py
Bastian-Krause: Yes, I expect the process that received SIGKILL to terminate soon, so we don't need a timeout.

labgrid/driver/qemudriver.py
Bastian-Krause: Agreed.

Bastian-Krause: Done.

Bastian-Krause:
> I find the `communicate()` calls a bit confusing at the points where the process is no longer expected to be running, but it seems simpler than the alternatives.

I agree.

Rebased, set QEMUDriver's sockets to `None` on deactivate.

labgrid/util/ssh.py
jluebbe: Still, it should not wait forever if something goes wrong.

Bastian-Krause: Okay, re-added `timeout=60`.

jluebbe (approved):

PR #1019, opened 2022-10-20
Make sure PyYAML 6 can be used

bgermann:
PyYAML 6.0's changelog says:
"always require `Loader` arg to `yaml.load()"

Use the safe_load() function instead.

Link: https://github.com/yaml/pyyaml/releases/tag/6.0
Signed-off-by: Bastian Germann <bastiangermann@fishpost.de>

- [ ] PR has been tested


Bastian-Krause:
This does not look right. Have a look at the yaml import:
```python
from ..util import atomic_replace, yaml
```
Our load function already uses the `yaml.SafeLoader`.

bgermann:
Never mind.

PR #1024, opened 2022-10-26
driver/power/netio_kshell: pexpect.spawn() telnet rather than deprecated telnetlib

Bastian-Krause:
**Description**
telnetlib is deprecated since Python 3.11 [1] and will be removed in Python 3.13 [2]. Use pexpect.spawn() with telnet instead. This requires matching more explicit strings and matches here and there.

While at it, send line separators as expected by NetIO devices [3].

Note: I have no NetIO devices available, so I tested the functions with increased timeout against a listening netcat pasting the expected answers [3] manually.

[1] https://docs.python.org/3/library/telnetlib.html
[2] https://peps.python.org/pep-0594/#deprecated-modules
[3] https://www.netio-products.com/files/NETIO-M2M-API-Protocol-Telnet.pdf

**Checklist**
- [x] PR has been tested (see above)

### DIFF
--- a/labgrid/driver/power/netio_kshell.py
+++ b/labgrid/driver/power/netio_kshell.py
@@ -1,7 +1,8 @@
 """ tested with NETIO 4C, should be compatible with all NETIO 4-models """
 
 import re
-import telnetlib
+
+import pexpect
 
 PORT = 1234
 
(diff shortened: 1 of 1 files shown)


labgrid/driver/power/netio_kshell.py
```diff
@@ -1,37 +1,46 @@
...
-    tn = telnetlib.Telnet(host, port, 1)
-    tn.read_until(b"\r\n", 0.5)
-    tn.write(b"login admin admin\n")
-    tn.read_until(b"250 OK\r\n", 0.5)
-    tn.write(f"port {index} {value}\n".encode())
-    tn.read_until(b"250 OK\r\n", 0.5)
-    tn.write(b"quit\n")
-    tn.close()
+
+    with pexpect.spawn(f"telnet {host} {port}", timeout=1) as tn:
```
jluebbe, line 14: This timeout now applies to any `expect()` call, instead of only the connection creation.

labgrid/driver/power/netio_kshell.py
```diff
@@ -1,37 +1,46 @@
...
+        tn.expect(b"100 HELLO .*\r\n")
```
jluebbe, line 15: This wasn't in the old version. Where did you find it?

jluebbe (changes requested):

labgrid/driver/power/netio_kshell.py
Bastian-Krause: Correct, that's intentional. Prior to this, all calls had their own timeout. I think it's more readable this way.

labgrid/driver/power/netio_kshell.py
Bastian-Krause: In the manual linked above. If that does not apply to all versions, we need something else. As I don't have a NetIO device, I have no idea what to use else.

jluebbe (approved):

PR #1025, opened 2022-10-26
Test/Announce Python 3.11 Support

Bastian-Krause:
**Description**
Python 3.11 was released on October 24th [1]. Let's add a test job for it and announce support in the setup.py.

[1] https://blog.python.org/2022/10/python-3110-is-now-available.html

Note: Deprecation warning leading to an error of the Python 3.11 job will be fixed by #1024.


jluebbe (approved):

jluebbe:
I'll merge it, it is only a deprecation *warning*.

PR #1033, opened 2022-11-03
client: fix help parser of labgrid-client tmc

userid0x0:
**Description**
The command
```
  labgrid-client tmc
```
showed the help of last subparser (=export). Now it shows the help of TMC including all subcommands.

Signed-off-by: Alexander Merkle <alexander.merkle@lauterbach.com>

**Checklist**
- [ ] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -1815,9 +1815,9 @@ def main():
     subparser = subparsers.add_parser('audio', help="start a audio stream")
     subparser.set_defaults(func=ClientSession.audio)
 
-    subparser = subparsers.add_parser('tmc', help="control a USB TMC device")
-    subparser.set_defaults(func=lambda _: subparser.print_help())
-    tmc_subparsers = subparser.add_subparsers(
+    tmc_parser = subparsers.add_parser('tmc', help="control a USB TMC device")
+    tmc_parser.set_defaults(func=lambda _: tmc_parser.print_help())
+    tmc_subparsers = tmc_parser.add_subparsers(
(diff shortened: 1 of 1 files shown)


jluebbe:
Do you have an explanation of why using a copy of the subparser fixes this?

userid0x0:
The lambda gets evaluated with the `args.func(session)` call https://github.com/labgrid-project/labgrid/blob/76101cd02d8d4585abb31a28878945b6dd6f361b/labgrid/remote/client.py#L1956 . At this point the `subparser` variable holds a reference to the export subparser created here https://github.com/labgrid-project/labgrid/blob/76101cd02d8d4585abb31a28878945b6dd6f361b/labgrid/remote/client.py#L1885 . So it's a problem when the lamda gets evaluated.

Thus I am taking a deep-copy of the TMC subparser once it's fully created.

userid0x0:
Update: I just see that the deep-copy is a bit of an overkill. Saving a reference of the `subparser` should be enough as well.
```
- tmc_parser = copy.deepcopy(subparser)
+ tmc_parser = subparser
```

jluebbe:
Perhaps we should just use different names for every subparser used in lambdas?

Bastian-Krause:
CI seems broken. Could you wrap your commit message at 74 chars per line and force-push your branch, so CI runs again? Thanks!

Bastian-Krause (approved):
Thanks!

jluebbe (approved):

PR #1035, opened 2022-11-07
shelldriver: increase _check_prompt() timeout

nick-potenski:
As noted by 0993e54c5344a24b92475478e50a6568dde7292e, sometimes heavy load can cause echoing the marker and displaying a prompt to take more time.  Further increased the timeout to 30 seconds to catch more cases.

Signed-off-by: Nick Potenski <nick.potenski@garmin.com>

**Description**

**Checklist**
- [ ] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/labgrid/driver/shelldriver.py
+++ b/labgrid/driver/shelldriver.py
@@ -202,7 +202,7 @@ class ShellDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
         try:
             self.console.expect(
                 rf"{marker}\s+{self.prompt}",
-                timeout=5
+                timeout=30
             )
             self._status = 1
         except TIMEOUT:


Bastian-Krause:
Hm, I'm wondering if this justifies another ShellDriver argument. Maybe we should simply increase this value even further? What value are you passing for your use case?

nick-potenski:
Heh, there are a lot of args to that driver already, aren't there?

I bumped the timeout to 30 seconds and haven't seen any check_prompt timeouts since, but this is a very low reproducible bug for us.  I haven't been able to actually quantify its frequency, so I don't necessarily have confidence in the solution yet.

There is probably very little to no harm in increasing that timeout for everyone, I suppose.  So, if you are cool with it I can update this PR to do that instead.

Emantor:
Increasing the timeout for everyone sounds fine to me, lets not make this configurable for now.

nick-potenski:
This has been reworked to increase the timeout to 30 seconds for everyone.  Tested and verified as working on my local setup.

Emantor (approved):

Bastian-Krause (approved):

PR #1042, opened 2022-11-23
Introduce anonymous static authentication, deprecate dynamic ticket authentication

Bastian-Krause:
**Description**
labgrid performs dynamic clear text challenge authentication with unchecked dummy tickets [1]. Add the anonymous static authentication method [2] and deprecate the previous one.

The only change required apart from the crossbar config is the location of the authid. Previously we could set the authid on `ApplicationSession.join()`, but with anonymous static authentication the authid is generated randomly. Since we still need a way to distinguish coordinator, exporter and client, add labgrid's authid to `authextra` [3]:

> All authenticators can return `authextra` auxiliary information that is returned both to the authenticating client, as well as clients which get disclosed the identity of a calling/publishing peer, as well as in the meta API when querying a session.

Dropping dynamic clear text challenge authentication will allow us to install crossbar in a dedicated virtualenv in the future, since `labgrid.remote.authenticator.AuthenticatorSession` is the only reference into labgrid code except for the coordinator itself, which is called with a dedicated python executable that can live inside a separate virtualenv. This again should reduce dependency conflicts caused by shared dependencies between crossbar and labgrid.

[1] https://crossbar.io/docs/Ticket-Authentication/
[2] https://crossbar.io/docs/Anonymous-Authentication/
[3] https://github.com/crossbario/crossbar/blob/ccbae0a56ef331fe00192455837aa72578eadc54/docs-old/pages/Features.md#authentication

**Checklist**
- [x] Tests for the feature 
- [x] PR has been tested

### DIFF
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -68,11 +68,17 @@ class ClientSession(ApplicationSession):
         self.monitor = self.config.extra.get('monitor', False)
         enable_tcp_nodelay(self)
         self.join(
-            self.config.realm, authmethods=["ticket"],
-            authid=f"client/{self.gethostname()}/{self.getuser()}"
+            self.config.realm,
+            authmethods=["anonymous", "ticket"],
+            authid=f"client/{self.gethostname()}/{self.getuser()}",
+            authextra={"authid": f"client/{self.gethostname()}/{self.getuser()}"},
         )
 
     def onChallenge(self, challenge):
+        import warnings
+        warnings.warn("Ticket authentication is deprecated. Please update your coordinator.",
+                      DeprecationWarning)
+        logging.warning("Ticket authentication is deprecated. Please update your coordinator.")
         return "dummy-ticket"
 
     async def onJoin(self, details):
--- a/.crossbar/config-anonymous.yaml
+++ b/.crossbar/config-anonymous.yaml
@@ -1,24 +1,9 @@
-# Legacy crossbar configuration with dynamic ticket authentication.
-# Please use .crossbar/config-anonymous.yaml instead.
 version: 2
 workers:
 - type: router
   realms:
   - name: realm1
     roles:
-    - name: authenticator
-      permissions:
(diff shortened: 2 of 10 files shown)


jluebbe:
Hmm, can we enable both methods in parallel for some time to simplify migration? Then, we could warn on the coordinator which exporters and clients still need to be updated.

Bastian-Krause:
Force-pushed:
- add new crossbar config with static anonymous authentication separately
- allow both authentication methods (allowing migration)
- do not drop `labgrid.remote.authenticator` for now
- add deprecation notice to the previous config
- add deprecation warnings to coordinator, client and exporter
- add deprecation warning for every client/exporter using ticket authentication
- recommend the new config in docs, use it in test suite and docker

Emantor (approved):

labgrid/remote/client.py
```diff
@@ -68,11 +68,17 @@ def onConnect(self):
...
+            self.config.realm,
+            authmethods=["anonymous", "ticket"],
+            authid=f"client/{self.gethostname()}/{self.getuser()}",
+            authextra={"authid": f"client/{self.gethostname()}/{self.getuser()}"},
         )
 
     def onChallenge(self, challenge):
+        import warnings
+        warnings.warn("Ticket authentication is deprecated. Please update your coordinator.",
+                      DeprecationWarning)
```
Bastian-Krause, line 80: FYI, this is only visible when used as library code.

labgrid/remote/client.py
```diff
@@ -68,11 +68,17 @@ def onConnect(self):
...
+        logging.warning("Ticket authentication is deprecated. Please update your coordinator.")
```
Bastian-Krause, line 81: I am not sure if this is too verbose. On the other hand, users might miss the deprecation if we don't warn loud enough.

Emantor: Yes, Warning load and clear is the best move here.

jluebbe (approved):

PR #1043, opened 2022-11-28
contrib: coordinator-statsd: Usability improvements

JoshuaWatt:
**Description**

Based on feedback from our statsd team after trying out the reporting, it has been re-worked with several major improvements
 1) Sets are now replaced with gauges. While sets are really cool from a
    reporting perspective, they are difficult to deal with.
    Specifically, it's very hard to report and graph "0" with a set
    since that's just the absence of any set items, which could also be
    "no data". Instead, use gauges and persistently record which metrics
    were reported so that they can be reported as 0 next collection
    iteration
 2) Add tags to the place reporting, so it can more easily be correlated
    to the reservation reporting. It's fairly trivial to sum these out
    if desired.
 3) Improve failure correction. If either the connection to the statsd
    server or labgrid coordinator fails, the program will continue on
    and attempt to reconnect instead of exiting.

**Checklist**
- [ ] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/contrib/coordinator-statsd.py
+++ b/contrib/coordinator-statsd.py
@@ -14,27 +14,28 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+# All metrics are optionally prefixed by a set of tag values which can be
+# specified with the --tag command line option.
 #
 # This program will connect to a labgrid coordinator and report the following
 # stats about places to a statsd server on a periodic basis:
 #
-#   labgrid.places_acquired (set): The number of places that are currently
-#                                  acquired
-#   labgrid.places_total (set): The number of places attached to the
-#                               coordinator
+#   labgrid.places.[TAG[.TAG]].acquired (gauge): The number of places that are
+#                                                currently acquired with this
+#                                                set of tags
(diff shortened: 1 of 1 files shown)


contrib/coordinator-statsd.py
```diff
@@ -59,22 +66,25 @@ async def report_reservations(session, statsd_client, tags):
...
+                    ["reservations", group_name]
+                    + [group.get(t, "") for t in tags]
+                    + [state]
+                ),
             )
-            statsd_client.set(".".join(path), token)
 
 
-async def report_places(session, statsd_client):
+async def report_places(session, tags, gauges):
```
Bastian-Krause, line 79: There are still calls to `report_places()` with 2 arguments left in the code.

Bastian-Krause (changes requested):
Your Signed-off-by is missing.

contrib/coordinator-statsd.py
JoshuaWatt: Ah, ya. The `async def report()` isn't used anymore (in favor of `asyncio.gather` and I forgot to remove it :)

PR #1045, opened 2022-12-02
CI: add scheduled tests

Bastian-Krause:
**Description**
Test results may vary due to new dependency versions. Daily CI runs should help us find issues early.

This becomes especially important once all of labgrid's dependencies get unpinned.

**Checklist**
- [x] PR has been tested

### DIFF
--- a/.github/workflows/push-pr-unit-tests.yml
+++ b/.github/workflows/push-pr-unit-tests.yml
@@ -1,0 +1,17 @@
+name: unit tests
+
+on: [push, pull_request]
+
+jobs:
+  push-pr-unit-tests:
+    name: unit tests
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
+    uses: ./.github/workflows/reusable-unit-tests.yml
+    with:
+      python-version: ${{ matrix.python-version }}
+  push-pr-unit-tests-docker:
+    name: Docker Unit Tests
+    uses: ./.github/workflows/reusable-unit-tests-docker.yml
--- a/.github/workflows/push-pr-unit-tests.yml
+++ b/.github/workflows/push-pr-unit-tests.yml
@@ -1,6 +1,6 @@
 name: unit tests
 
-on: [push, pull_request]
+on: [push, pull_request, workflow_dispatch]
 
 jobs:
   push-pr-unit-tests:
--- a/.github/workflows/reusable-unit-tests-docker.yml
+++ b/.github/workflows/reusable-unit-tests-docker.yml
@@ -1,0 +1,28 @@
+name: reusable docker tests
+
+on:
+  workflow_call:
+    inputs:
+      branch:
+        type: string
+        required: false
+
+jobs:
+  docker:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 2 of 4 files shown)


.github/workflows/push-pr-unit-tests.yml
```diff
@@ -0,0 +1,20 @@
...
+        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
+    uses: ./.github/workflows/reusable-unit-tests.yml
+    with:
+      python-version: ${{ matrix.python-version }}
+      branch:
+  push-pr-unit-tests-docker:
+    name: Docker Unit Tests
+    uses: ./.github/workflows/reusable-unit-tests-docker.yml
+    with:
+      branch:
```
Emantor, line 20: This looks weird, is this required?

Bastian-Krause: Passing an empty branch makes `actions/checkout@v3` use the default one. That's what we want here. If we mark the input as `required: false`, we can probably drop this. I'd have to look what is passed in that case.

Bastian-Krause: Fixed.

Bastian-Krause:
Force-pushed:
- made branch input optional, default is an empty string, see [docs](https://docs.github.com/en/enterprise-cloud@latest/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_callinputs)
- dropped empty branch inputs

Emantor (approved):

.github/workflows/push-pr-unit-tests.yml
```diff
@@ -0,0 +1,17 @@
+name: unit tests
+
+on: [push, pull_request, workflow_dispatch]
+
+jobs:
+  push-pr-unit-tests:
```
Emantor, line 6: Isn't the trick here to name this `build` so it still fits the repo settings? This way we shouldn't need to reconfigure the repo and rebase old PRs?

.github/workflows/push-pr-unit-tests.yml
```diff
@@ -0,0 +1,17 @@
...
+    name: unit tests
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
+    uses: ./.github/workflows/reusable-unit-tests.yml
+    with:
+      python-version: ${{ matrix.python-version }}
+  push-pr-unit-tests-docker:
```
Emantor, line 15: Same here, shouldn't this be named `docker`?

Emantor (changes requested):

Emantor:
Ah, I guess this may be more than simple string matching, but I'll test it nevertheless.

Bastian-Krause:
The checks now have a 3 component identifier, previously it consisted of 2 components. I don't know if renaming makes a difference. You can try that, though. I don't have the access rights to see the configuration of the required checks anyway.

Emantor:
It doesn't matter, the extra component removes the match.

jluebbe:
LGTM, but we should check if we can keep the existing check associations.

jluebbe:
Due to the different workflow call chain, it doesn't seem possible to keep the check names the same.

jluebbe (approved):

PR #1046, opened 2022-12-02
Github Release workflow

Emantor:
Add a build and release workflow which runs after successful tests and if a Tag was pushed.

Depends on #1045

### DIFF
--- a/.github/workflows/build-and-release.yml
+++ b/.github/workflows/build-and-release.yml
@@ -1,0 +1,22 @@
+name: build and publish
+
+on: workflow_call
+
+jobs:
+  release:
+    runs-on: ubuntu-22.04
+    continue-on-error: false
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Python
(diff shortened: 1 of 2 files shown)


.github/workflows/build-and-release.yml
```diff
@@ -0,0 +1,28 @@
...
+        type: string
+        required: false
+jobs:
+  release:
+    runs-on: ubuntu-22.04
+    continue-on-error: false
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        ref: ${{ inputs.branch }}
```
jluebbe, line 16: Shouldn't this be the tag ref?

jluebbe (changes requested):
Looks good otherwise.

.github/workflows/build-and-release.yml
jluebbe: We don't need explicit inputs here.

jluebbe (approved):

PR #1050, opened 2022-12-08
sshdriver: Escape control path on cleanup

JoshuaWatt:
**Description**
The control file needs the "%" escaped when cleaning up like all the other places it is passed to SSH

**Checklist**
- [ ] Tests for the feature 

- [x] PR has been tested


Bastian-Krause (approved):
Looks good. Maybe it's even better to replace this once right before it's initially returned in `_start_own_master_once()` instead of replacing this on every use?

JoshuaWatt:
> Looks good. Maybe it's even better to replace this once right before it's initially returned in `_start_own_master_once()` instead of replacing this on every use?

Ya, I thought about that; wasn't sure if anyone outside of this class was looking at the variable

jluebbe (approved):

Emantor (approved):

PR #1059, opened 2023-01-09
exporter/coordinator/client: create/set event loop explicitly

### LINKED ISSUE #1056: DeprecationWarning: There is no current event loop
I get an annoying warning at the end of my pytest session, when running on the latest master

```
test_setup/test_setup.py::test_setup_ip
  /usr/local/lib/python3.10/dist-packages/labgrid/remote/client.py:1238: DeprecationWarning: There is no current event loop
    loop = asyncio.get_event_loop()

test_setup/test_setup.py::test_setup_ip
  /usr/local/lib/python3.10/dist-packages/autobahn/asyncio/websocket.py:300: DeprecationWarning: There is no current event loop
    self.loop = loop or asyncio.get_event_loop()

test_setup/test_setup.py::test_setup_ip
  /usr/local/lib/python3.10/dist-packages/autobahn/asyncio/wamp.py:238: DeprecationWarning: There is no current event loop
    loop = asyncio.get_event_loop()
```
I tried to update autobahn to latest version, but the error stil exists. Anybody knows why this warning comes.

Bastian-Krause:
**Description**
Calling `asyncio.get_event_loop()` with no current event loop is deprecated since Python 3.10 and will be an error in Python >=3.12, see [docs](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_event_loop).

So create a new event loop explicitly and set it as the current event loop for the current OS thread.

**Checklist**
- [x] PR has been tested

Fixes #1056

### DIFF
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -32,8 +32,6 @@ from ..util.proxy import proxymanager
 from ..util.helper import processwrapper
 from ..driver import Mode, ExecutionError
 
-txaio.config.loop = asyncio.get_event_loop()  # pylint: disable=no-member
-
 
 class Error(Exception):
     pass
@@ -1235,7 +1233,10 @@ class ClientSession(ApplicationSession):
 def start_session(url, realm, extra):
     from autobahn.asyncio.wamp import ApplicationRunner
 
-    loop = asyncio.get_event_loop()
+    loop = asyncio.new_event_loop()
+    asyncio.set_event_loop(loop)
+    txaio.config.loop = loop  # pylint: disable=no-member
+
     ready = asyncio.Event()
 
     async def connected(session):  # pylint: disable=unused-argument
--- a/labgrid/remote/coordinator.py
+++ b/labgrid/remote/coordinator.py
@@ -888,6 +888,7 @@ class CoordinatorComponent(ApplicationSession):
         return {k: v.asdict() for k, v in self.reservations.items()}
 
 if __name__ == '__main__':
+    asyncio.set_event_loop(asyncio.new_event_loop())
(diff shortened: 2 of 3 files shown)


labgrid/remote/client.py
```diff
@@ -1235,7 +1233,10 @@ async def export(self, place, target):
 def start_session(url, realm, extra):
     from autobahn.asyncio.wamp import ApplicationRunner
 
-    loop = asyncio.get_event_loop()
+    loop = asyncio.new_event_loop()
```
jluebbe, line 1236: Doesn't this break use-cases where the caller already has a running event loop?

Bastian-Krause: This is not the case in our codebase. But it could be the case in third-party code, though. So maybe add a loop kwarg and use that if it's specified?

jluebbe: Some context for the Python upstream change: https://github.com/python/cpython/issues/83710

kjeldflarup:
I just tried to include this in the build of my test runner container.
```
git clone -b bst/explicit-event-loop https://github.com/Bastian-Krause/labgrid.git /opt/labgrid
```
I however still gets this warning:
```
tests_self/test__ssh.py::test_ssh
  /usr/local/lib/python3.10/dist-packages/autobahn/asyncio/websocket.py:300: DeprecationWarning: There is no current event loop
    self.loop = loop or asyncio.get_event_loop()
```
Do I need to do more than updating my test runner?

Bastian-Krause:
This PR needs more work. We need to understand the reasoning behind this deprecation better and then update this PR with a solution covering all use cases.

PR #1072, opened 2023-01-20
resource/remote: allow instantiation of RemotePlaces without Environment

Bastian-Krause:
**Description**
While it is recommended to use an Environment config, it should still be possible to instantiate a Target and a RemotePlace without it, e.g.:

```python
import labgrid

target = labgrid.Target("main")
remote_place = labgrid.resource.RemotePlace(target, name="my-place-name")
power = labgrid.driver.powerdriver.NetworkPowerDriver(target, name=None)
target.activate(power)
power.cycle()
```

We already have a fallback for crossbar URL and realm if it's not present in the Environment, so stick to that when no Environment is present at all.

Also add a test for this.

**Checklist**
- [x] Tests for the feature 
- [x] PR has been tested

### DIFF
--- a/labgrid/resource/remote.py
+++ b/labgrid/resource/remote.py
@@ -40,13 +40,12 @@ class RemotePlaceManager(ResourceManager):
         # be the same).
         if not self.session:
             self.env = remote_place.target.env
-            config = self.env.config
-            self.url = config.get_option(
-                'crossbar_url',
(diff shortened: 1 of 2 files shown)


Bastian-Krause:
Hm, running the new test exclusively works, but it does not work when run with the other crossbar tests.

Bastian-Krause:
Rebased.

jluebbe (approved):

PR #1073, opened 2023-01-20
add custom options to sshdriver

b2vn:
**Description**

added the possibility to define custom option for the SSHDriver in the environment yaml. The motivation for this change was to be able to overwrite host key algorithms in test context.

**Checklist**

- [x] Tests for the feature 
- [x] PR has been tested


labgrid/driver/sshdriver.py
```diff
@@ -48,6 +49,9 @@ def on_activate(self):
         if not self.networkservice.password:
             self.ssh_prefix += ["-o", "PasswordAuthentication=no"]
 
+        if self.options:
+            self.ssh_prefix += self.options.split(" ")
```
jluebbe, line 53: https://docs.python.org/3/library/shlex.html#shlex.split is probably better in this case.

As some special cases around quoting are not easily handled even with `shlex.split`, we should support optionally having a list instead of a string. In that case, they could simply be passed along.

jluebbe (changes requested):
Your commit is missing the Signed-off-by line. See https://labgrid.readthedocs.io/en/latest/development.html#developer-s-certificate-of-origin

labgrid/driver/sshdriver.py
```diff
@@ -30,6 +30,7 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
     bindings = {"networkservice": "NetworkService", }
     priorities = {CommandProtocol: 10, FileTransferProtocol: 10}
     keyfile = attr.ib(default="", validator=attr.validators.instance_of(str))
+    options = attr.ib(default="", validator=attr.validators.instance_of(str))
```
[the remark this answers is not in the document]
b2vn: I agree, will be done

Emantor (approved):

Emantor (approved):

doc/configuration.rst
```diff
@@ -1483,6 +1483,10 @@ Implements:
 
    SSHDriver:
      keyfile: example.key
+     extra_options:
+       - "-o option1"
+       - "-o options2"
+       - "-4"
```
jluebbe: This seems inconsistent. Why use an explicit list if not to exactly define the argv elements?

jluebbe (changes requested):

labgrid/driver/sshdriver.py
```diff
@@ -48,6 +49,12 @@ def on_activate(self):
         if not self.networkservice.password:
             self.ssh_prefix += ["-o", "PasswordAuthentication=no"]
 
+        if isinstance(self.extra_options, str) and len(self.extra_options) > 0:
+            self.ssh_prefix += shlex.split(self.extra_options)
+        elif isinstance(self.extra_options, list) and len(self.extra_options) > 0:
+            for eo in self.extra_options:
+                self.ssh_prefix += shlex.split(eo)
```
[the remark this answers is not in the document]
Bastian-Krause: Fixed.

doc/configuration.rst
Bastian-Krause: Should have been
```yaml
SSHDriver:
  extra_options: ["-o", "option1", "-o", "option2", "-4"]
```

Since this is a rather special use case, I've dropped it from the docs example altogether.

PR #1078, opened 2023-01-25
Actions: build and release fixes

Emantor:
**Description**

Adjust the build and release workflow to work:
- upload to testpypi if it is not a tag, to ensure the release pipeline works
- set an explicit python version
- inherit secrets from the main run so API tokens can be accessed in the build and release job.
- remove local versioning for `setuptools_scm`

### DIFF
--- a/.github/workflows/build-and-release.yml
+++ b/.github/workflows/build-and-release.yml
@@ -16,7 +16,14 @@ jobs:
     - name: Build sdist
       run: |
         python -m build -s
-    - name: Publish package
+    - name: Publish distribution 📦 to Test PyPI
+      uses: pypa/gh-action-pypi-publish@release/v1
+      if: startsWith(github.ref, 'refs/tags') != true
+      with:
+        password: ${{ secrets.TEST_PYPI_API_TOKEN }}
+        repository_url: https://test.pypi.org/legacy/
+    - name: Publish distribution 📦 to PyPI
+      if: startsWith(github.ref, 'refs/tags')
       uses: pypa/gh-action-pypi-publish@release/v1
       with:
         password: ${{ secrets.PYPI_API_TOKEN }}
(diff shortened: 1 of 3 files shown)


jluebbe (approved):

.github/workflows/build-and-release.yml
```diff
@@ -10,13 +10,22 @@ jobs:
...
     - name: Install python dependencies
       run: |
         python -m pip install --upgrade pip build
     - name: Build sdist
       run: |
         python -m build -s
-    - name: Publish package
+    - name: Publish distribution 📦 to Test PyPI
+      uses: pypa/gh-action-pypi-publish@release/v1
+      if: startsWith(github.ref, 'refs/tags') != true
```
Bastian-Krause, line 23:
```suggestion
      if: "!startsWith(github.ref, 'refs/tags')"
```
[Source](https://github.com/orgs/community/discussions/25017#discussioncomment-3246242) - maybe it's a bit more idiomatic?

.github/workflows/build-and-release.yml
```diff
@@ -16,7 +16,14 @@ jobs:
...
+      with:
+        password: ${{ secrets.TEST_PYPI_API_TOKEN }}
+        repository_url: https://test.pypi.org/legacy/
```
Bastian-Krause, line 26: Is there a non-legacy variant of the pypi test instance?

jluebbe: See https://github.com/pypa/gh-action-pypi-publish, it seems that this API is the one currently supported by the official publish action.

Bastian-Krause: Okay.

.github/workflows/build-and-release.yml
Emantor: I think `!= true` is fine and more verbose since the `!` can be easily missed.

Bastian-Krause: As you like.

Bastian-Krause (approved):

PR #1079, opened 2023-01-25
Install Crossbar Into Separate Virtualenv

Bastian-Krause:
**Description**
After dynamic clear text challenge authentication has been deprecated in #1042, we can now install crossbar in a dedicated virtualenv. This is possible because `labgrid.remote.authenticator.AuthenticatorSession` is the only reference into labgrid code except for the coordinator itself, which is called with a dedicated python executable that can live inside a separate virtualenv. This should reduce dependency conflicts caused by shared dependencies between crossbar and labgrid.

In a future PR we can finally unpin labgrid's dependencies.

**Note: Does not work with crossbar ticket authentication (see #1081)**

**Checklist**
- [x] Documentation has been adjusted
- [x] PR has been tested


tests/conftest.py
```diff
@@ -123,14 +123,16 @@ def crossbar_config(tmpdir, pytestconfig):
...
 
 @pytest.fixture(scope='function')
-def crossbar(tmpdir, crossbar_config):
-    if not find_spec('crossbar'):
-        pytest.skip("crossbar not found")
+def crossbar(tmpdir, pytestconfig, crossbar_config):
+    crossbar_venv = Path(pytestconfig.getoption("--crossbar-venv"))
+    if not crossbar_venv.is_absolute():
+        crossbar_venv = pytestconfig.rootdir / crossbar_venv
+    crossbar_bin = crossbar_venv / "bin/crossbar"
```
jluebbe, line 130: Wouldn't it be simpler to just pass the crossbar_bin path via an option? We're not really interested in the venv otherwise.

doc/RELEASE.rst
```diff
@@ -81,6 +81,10 @@ Test the upload by using pypi dev as a download source
 
 ::
 
+   virtualenv -p python3 labgrid-crossbar-release-<your-version-number>
```
jluebbe, line 84: Why use the version number in the venv path?

examples/usbpower/README.rst
```diff
@@ -116,7 +116,7 @@ Remote Setup
 ------------
 
 To access resources remotely, you first need to start the coordinator::
-  $ crossbar start --logformat none --config config-anonymous.yaml
+  $ crossbar-venv/bin/crossbar start --logformat none --config config-anonymous.yaml
```
jluebbe, line 119: This path doesn't match the one you used above.

jluebbe (changes requested):

jluebbe:
Don't we need to modify the crossbar config to start the coordinator component in the venv containing labgrid?

doc/RELEASE.rst
Bastian-Krause: The other venv a few lines below does the same, that's why I've done the same here.

examples/usbpower/README.rst
Bastian-Krause: Which one? `crossbar-venv/bin/crossbar` is used everywhere except in the release step-by-step guide.

tests/conftest.py
Bastian-Krause: True, I'll change that.

Bastian-Krause:
> Don't we need to modify the crossbar config to start the coordinator component in the venv containing labgrid?

When starting crossbar we are actually in the venv containing labgrid, so..

https://github.com/labgrid-project/labgrid/blob/5ca7395675303b3d369078421bfad543c2e924a7/.crossbar/config-anonymous.yaml#L37
https://github.com/labgrid-project/labgrid/blob/5ca7395675303b3d369078421bfad543c2e924a7/.crossbar/config-anonymous.yaml#L39
https://github.com/labgrid-project/labgrid/blob/5ca7395675303b3d369078421bfad543c2e924a7/.crossbar/config-anonymous.yaml#L40

..works as expected, only crossbar itself is called from within its own venv:

https://github.com/labgrid-project/labgrid/blob/5b43e8ecbe73c7f608b153890088b5dc5803f93f/doc/getting_started.rst#L176

jluebbe:
> > Don't we need to modify the crossbar config to start the coordinator component in the venv containing labgrid?
> 
> When starting crossbar, we are actually in the venv containing labgrid, so..

Hmm, I'm not sure if leaving that implicit is a good idea. It's easy to become confused as to what's happening in which venv.

Bastian-Krause:
> > > Don't we need to modify the crossbar config to start the coordinator component in the venv containing labgrid?
> > 
> > 
> > When starting crossbar, we are actually in the venv containing labgrid, so..
> 
> Hmm, I'm not sure if leaving that implicit is a good idea. It's easy to become confused as to what's happening in which venv.

What would be your preferred way of doing this? Changing the python3 executable in the crossbar config to an absolute path in the labgrid venv via something like [yq](https://github.com/mikefarah/yq)?

jluebbe:
> > > > Don't we need to modify the crossbar config to start the coordinator component in the venv containing labgrid?
> > > 
> > > When starting crossbar, we are actually in the venv containing labgrid, so..
> > 
> > Hmm, I'm not sure if leaving that implicit is a good idea. It's easy to become confused as to what's happening in which venv.
> 
> What would be your preferred way of doing this? Changing the python3 `executable` in the crossbar config to an absolute path in the labgrid venv via something like [yq](https://github.com/mikefarah/yq)?

Hmm. At least we should document that the default config file expects that the crossbar wrapper is stared in the labgrid venv. Then the user could set an explicit path into the other venv manually in case using a "nested" venv is not working. 

On the other hand, I suspect that nested setups are not really supported upstream and are likely to break at some point. So perhaps we should just use sed to replace something like /path/to/labgrid-venv/bin/python?

Bastian-Krause:
> > > > > Don't we need to modify the crossbar config to start the coordinator component in the venv containing labgrid?
> > > > 
> > > > 
> > > > When starting crossbar, we are actually in the venv containing labgrid, so..
> > > 
> > > 
> > > Hmm, I'm not sure if leaving that implicit is a good idea. It's easy to become confused as to what's happening in which venv.
> > 
> > 
> > What would be your preferred way of doing this? Changing the python3 `executable` in the crossbar config to an absolute path in the labgrid venv via something like [yq](https://github.com/mikefarah/yq)?
> 
> Hmm. At least we should document that the default config file expects that the crossbar wrapper is stared in the labgrid venv. Then the user could set an explicit path into the other venv manually in case using a "nested" venv is not working.
> 
> On the other hand, I suspect that nested setups are not really supported upstream and are likely to break at some point. So perhaps we should just use sed to replace something like /path/to/labgrid-venv/bin/python?

Yes, I think setting the absolute path to the python interpreter in the crossbar config is the most stable and least confusing way. I'll update this.

Emantor (approved):

jluebbe (dismissed):

jluebbe:
@Bastian-Krause We need to mention in the release notes that the Debian package no longer contains the crossbar. A local venv or the container could be used instead.

Emantor (approved):

PR #1080, opened 2023-01-26
actions: set fetch-depth for build and release

Emantor:
**Description**
Set the fetch-depth to 0 for the build and release workflow. This way we should have the complete tag history of the branch, which should allow setuptools_scm to calculate the correct wanted release name for the testpypi upload.


jluebbe (approved):
At some point we might want to limit this again, but for now it's the simplest fix.

PR #1081, opened 2023-01-26
dockerfiles: revert to previous authentication

Emantor:
**Description**
Updating the docker containers breaks backwards compatibility with older clients. Revert back to ticket authentication.

### DIFF
--- a/.crossbar/config.yaml
+++ b/.crossbar/config.yaml
@@ -44,6 +44,9 @@ workers:
       ws:
         type: websocket
         auth:
+          anonymous:
+            type: static
+            role: public
           ticket:
             type: dynamic
             authenticator: org.labgrid.authenticate
--- a/dockerfiles/Dockerfile
+++ b/dockerfiles/Dockerfile
@@ -47,7 +47,7 @@ VOLUME /opt/crossbar
 
 EXPOSE 20408
 
-CMD ["crossbar", "start", "--config", "/opt/labgrid/.crossbar/config-anonymous.yaml"]
+CMD ["crossbar", "start", "--config", "/opt/labgrid/.crossbar/config.yaml"]
 
 #
 # ser2net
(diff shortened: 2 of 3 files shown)


jluebbe (approved):
OK, this should fix support for "old" clients using ticket auth for now, while allowing new clients to use anonymous (and avoid the warning).

Bastian-Krause:
Note that we can't have ticket auth enabled for #1079 because `labgrid.remote.authenticator.AuthenticatorSession` is required for it. This prevents us from separating labgrid and crossbar venv.

The whole idea regarding deprecating vs. dropping ticket auth always was that we will *require* anonymous auth at some point when the coordinator (crossbar) is updated. On current master (035e857) it is still possible to use both mechanisms. Once #1079 is merged, everybody needs to have a recent versions of coordinator, crossbar and clients.

So I'd opt for merging this, make a release (or two?), then revert the PR again and merge #1079 on top.
I see no other sane way of ever updating our dependencies otherwise..

Note that the suggestion above will mean that we will face the backward incompatibility issue as soon as this PR is reverted. So no idea if this approach really helps, because we build docker images not only from release tags, but from master. My hope would be that people running the latest coordinator notice the warning..

https://github.com/labgrid-project/labgrid/blob/035e857830f53f7ca5c12cdd7d4e749bf8befef6/labgrid/remote/authenticator.py#L17

..in the mean time and are prepared for the breaking change by updating their exporters, clients and before updating the coordinator to the future backwards incompatible version. We should also announce this plan in the release notes of the [next release](https://github.com/labgrid-project/labgrid/milestone/8).

Bastian-Krause (approved):

jluebbe:
> My hope would be that people running the latest coordinator notice the warning..

The warning via the log for each client is probably more obvious:
https://github.com/labgrid-project/labgrid/blob/035e857830f53f7ca5c12cdd7d4e749bf8befef6/labgrid/remote/authenticator.py#L11

Bastian-Krause:
> > My hope would be that people running the latest coordinator notice the warning..
> 
> The warning via the log for each client is probably more obvious:
> 
> https://github.com/labgrid-project/labgrid/blob/035e857830f53f7ca5c12cdd7d4e749bf8befef6/labgrid/remote/authenticator.py#L11

Yes, in case the client version is new enough.

PR #1082, opened 2023-01-26
U-Boot boot commands Support

jremmet:
uboot doesn't have pre defined boot sources. The "boot" command takes no parameters.
Add self.boot_commands to hold different boot_command by name.
They can be used with LinuxBootProtocol boot(name)

**Checklist**
- [X] Tests for the feature 
- [X] PR has been tested

### DIFF
--- a/labgrid/driver/ubootdriver.py
+++ b/labgrid/driver/ubootdriver.py
@@ -43,6 +43,7 @@ class UBootDriver(CommandMixin, Driver, CommandProtocol, LinuxBootProtocol):
     boot_expression = attr.ib(default="", validator=attr.validators.instance_of(str))
     bootstring = attr.ib(default=r"Linux version \d", validator=attr.validators.instance_of(str))
     boot_command = attr.ib(default="run bootcmd", validator=attr.validators.instance_of(str))
+    boot_commands = attr.ib(default=attr.Factory(dict), validator=attr.validators.instance_of(dict))
     login_timeout = attr.ib(default=30, validator=attr.validators.instance_of(int))
     boot_timeout = attr.ib(default=30, validator=attr.validators.instance_of(int))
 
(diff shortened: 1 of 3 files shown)


labgrid/driver/ubootdriver.py
```diff
@@ -206,13 +206,13 @@ def await_boot(self):
         self.console.expect(self.bootstring, timeout=self.boot_timeout)
 
     @Driver.check_active
-    @step(args=['name'])
-    def boot(self, name):
-        """Boot the default or a specific boot entry
+    @step(args=['cmd'])
+    def boot(self, cmd: str = ""):
+        """Boot the default if cmd is empty or a specific boot cmd otherwise.
```
jluebbe, line 211: The LinuxBootProtocol defines this parameter as `cmd`: https://github.com/labgrid-project/labgrid/blob/master/labgrid/protocol/linuxbootprotocol.py#L6

We could have a dict in the UBootDriver configuration which defines commands for each boot "entry". What do you think? @jremmet @Emantor

jluebbe (changes requested):

labgrid/driver/ubootdriver.py
jremmet: I like the idea. I would be configurable, and we can check if exists. Guess `self.boot_command`should continue to work and the default for `name = ""` ? @jluebbe

jluebbe: Yes. Perhaps something like:
```
boot_commands:
  A: "run boot_slot_a"
  B: "run boot_slot_b"
```

jluebbe: This way, you could also reuse test cases and Strategies easier with different boards.

Emantor (approved):

jluebbe (approved):

PR #1089, opened 2023-02-07
resource/udev: allow fastboot with adb_user=yes

### LINKED ISSUE #1086: labgrid-suggest doesn't notice new fastboot devices
labgrid-suggest is a great tool when adding new devices, and it works great for USB serial ports etc.  But it doesn't detect new USB fastboot devices.  

Please consider it a feature request to add detection/report of fastboot devices to labgrid-suggest.  At least ones that are using the standard USB vendor/device IDs from google for this.

Emantor:
**Description**
Distros which implement the android-udev-rules from github use the adb_user property to indicate that the device is valid for fastboot. Make use of this property for labgrid which removes the need for more hardcoded USB IDs.

**Checklist**
- [x] PR has been tested

Fixes #1086


jluebbe (approved):
We will probably need to do something more involved if/when we add a ADB resource.

PR #1090, opened 2023-02-08
labgrid/driver/usbloader: fix uuu command

jremmet:
The current implementation adds the cmd as last arg. There it is ignored by uuu
tool. It runs in the default mode with just an image.

For CPUs relying on a special build-in run scripts it must be added with
-b before the file argument.
Removed the broken cmd and add script as config option. This way 
currently working setups should work further.
```
uuu help text (libuuu_1.4.43)

uuu [-d -m -v -V] <bootloader|cmdlists|cmd>

uuu [-d -m -v] -b[run] <emmc|emmc_all|fat_write|nand|qspi|sd|sd_all|spl> arg...
	spl	boot spl and uboot
		arg0: _flash.bin
```
**Checklist**
- [x] Tests on i.MX6UL and i.MX8MM

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -2210,7 +2210,7 @@ Implements:
 Arguments:
   - image (str): optional, key in :ref:`images <labgrid-device-config-images>` containing the path
     of an image to bootstrap onto the target
-  - cmd (str, default="spl"): single command used for mfgtool
+  - script (str): run built-in script with "uuu -b", called with image as arg0
 
 USBStorageDriver
 ~~~~~~~~~~~~~~~~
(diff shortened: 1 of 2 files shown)


doc/configuration.rst
```diff
@@ -2210,7 +2210,7 @@ Implements:
 Arguments:
   - image (str): optional, key in :ref:`images <labgrid-device-config-images>` containing the path
     of an image to bootstrap onto the target
-  - cmd (str, default="spl"): single command used for mfgtool
+  - script (str): run built-in script with "uuu -b", called with image as arg0
```
Emantor, line 2213: This is a breaking change for the driver, is this strictly necessary?

Emantor:
FYI: I am not too worried about breaking this driver if it was broken before. Just curious :)

jremmet:
> FYI: I am not too worried about breaking this driver if it was broken before. Just curious :)

The idea was if a config use cfg now, and it works as it is ignored, reusing cmd will break this. By adding script this will work further.

Emantor (approved):

jluebbe (approved):

PR #1098, opened 2023-03-06
Allow default configure when export same type of resources.

atline:
**Description**

If `labgrid-client` find multiple matched resources, currently labgrid will require user to specify the exact resource name as next:

> labgrid-client: error: multiple resources matching <class 'labgrid.resource.serialport.NetworkSerialPort'> found in Target(name='imx8mm-evk-sh99', env=None)

However, in our scenarios: `A core serial port` is the most frequently used resource, while `M core, Sec core & other serial port` seldom be used. So we wish `labgrid-client -p imx8mm-evk-sh99 console` can by default open the `A core serial port`. 

This PR allows admin to configure a default resource with something like next:

```
a:
  cls: NetworkSerialPort
  host: 10.192.244.104
  port: 7006
  default: true
m:
  cls: NetworkSerialPort
  host: 10.192.244.104
  port: 7005
```

Afterwards, we possible to have the `default` parameter when export that:

```
Acquired resource 'a' (shubuntu1/imx8mm-evk-sh99/NetworkSerialPort/a):
  {'acquired': 'imx8mm-evk-sh99',
   'avail': True,
   'cls': 'NetworkSerialPort',
   'default': True,
   'params': {'host': '10.192.244.104', 'port': 7006}}
Acquired resource 'm' (shubuntu1/imx8mm-evk-sh99/NetworkSerialPort/m):
  {'acquired': 'imx8mm-evk-sh99',
   'avail': True,
   'cls': 'NetworkSerialPort',
   'default': False,
   'params': {'host': '10.192.244.104', 'port': 7005}}
```

Then, the `labgrid-client` can choose the `default=True` resource even there are multiple matched resources.

**Checklist**
- [√] Documentation for the feature
- [√] PR has been tested

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -3189,6 +3189,23 @@ and another :any:`USBSerialPort` as
        match:
          '@ID_PATH': pci-0000:05:00.0-usb-3-1.4
 
+Additional, you can set ``default`` parameter to specify the default resource
+which ``labgrid-client`` will use once it matches multiple resources. Otherwise,
+you need to specify resource name explicitly in client command:
+
+.. code-block:: yaml
+
+   usb-hub-in-rack12:
+     console-main:
+       cls: USBSerialPort
+       match:
+         '@ID_PATH': pci-0000:05:00.0-usb-3-1.3
+       default: true
+     console-secondary:
+       cls: USBSerialPort
+       match:
+         '@ID_PATH': pci-0000:05:00.0-usb-3-1.4
+
 Note that you could also split the resources up into distinct groups instead
 to achieve the same effect:
 
--- a/labgrid/remote/common.py
+++ b/labgrid/remote/common.py
@@ -45,6 +45,10 @@ class ResourceEntry:
         return self.data['cls']
 
     @property
+    def default(self):
+        return self.data['default']
+
+    @property
     def params(self):
         return self.data['params']
 
@@ -63,6 +67,7 @@ class ResourceEntry:
     def asdict(self):
         return {
             'cls': self.cls,
+            'default': self.default,
             'params': self.params,
             'acquired': self.acquired,
             'avail': self.avail,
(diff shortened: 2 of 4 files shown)


Emantor:
Unfortunately, the exporter configuration is the wrong place for this. While this makes the implementation easy, exported resources can be assigned to many places over time and connected boards to these resources may not always use that same serial port for default communication. This should rather be attachable to the individual created place instead.

atline:
> Unfortunately, the exporter configuration is the wrong place for this. While this makes the implementation easy, exported resources can be assigned to many places over time and connected boards to these resources may not always use that same serial port for default communication. This should rather be attachable to the individual created place instead.

1. Just a discussion, in my opinion from concept, we could treat `default` as a `suggestion` of of `resource group`.  The `resource group` says: I have 2 `NetworkSerialPort`, so when any place want to use these resources. If the place agree to use that `default` suggestion, just use it. While other places don't agree that, afford the name by yourselves. 

This is how `default` works for 
* E.g. labgrid driver priorities, because in fact not all cases will need the high priority driver, then can select one by name, but by default, it will use high priority.
* E.g. how usually `def fun(arg=1)` works, as not all caller will need the `default arg=1`, then the caller can send its own value, but by default, if you not specify it, it will use `arg=1`.

Anyway, as `multiple places to use the same resource model` not used in my settings, so I definitely have some subjective assumption here. Then if above concept not persuade you, let's move to item2.

2.  Do you have any suggestion about `attach this settings to place`?

* I don't know any explicit configure in labgrid like yaml to configure that (Maybe I miss?). So what I know is something like `labgrid-client --place example-place create` and `labgrid-client -p example-place add-match */example-group/*`.

* So you mean we should somewhat to specify that `default` in above similar command? Then admin possible need to do `labgrid-client -p example-place default {cls} {name}`? Then set command should change the structure of `.crossbar/places.yaml` to add that settings? Finally, the `labgrid-client` need to parse the place object to filter that `default`?
Is above the correct way to move?

atline:
> #1112 has implemented this without an exporter side change.

Thanks @Emantor , YES, I confirm that works for me ~

PR #1108, opened 2023-03-24
Add missing get method to sdwire driver

jpuderer:
**Description**

The labgrid-client tried to call the get method when using the sd-mux command with and sdwire device.  This should work since the labgrid-client sd-mux command has code to explictly support sdwire devices.

```
(labgrid-venv) Rocinante:labgrid-sandbox jpuderer$ labgrid-client -x ws://labgrid-devhost:20408/ws -c test.yaml sd-mux get 
Selected role main and place test from configuration file
  File "/Users/jpuderer/labgrid/labgrid/remote/client.py", line 1787, in main
    args.func(session)
  File "/Users/jpuderer/labgrid/labgrid/remote/client.py", line 950, in sd_mux
    print(drv.get_mode())
          ^^^^^^^^^^^^
AttributeError: 'USBSDWireDriver' object has no attribute 'get_mode'
```

**Checklist**
- [x] PR has been tested

### DIFF
--- a/labgrid/driver/usbsdwiredriver.py
+++ b/labgrid/driver/usbsdwiredriver.py
@@ -37,3 +37,15 @@ class USBSDWireDriver(Driver):
             self.mux.control_serial,
         ]
         processwrapper.check_output(cmd)
+
+    @Driver.check_active
+    @step(title='sdmux_get')
+    def get_mode(self):
+        cmd = self.mux.command_prefix + [
+            self.tool,
+            "-e",
+            self.mux.control_serial,
+            "-u",
+        ]
+        result = processwrapper.check_output(cmd)
+        return result.split(b": ", maxsplit=1)[1].strip().decode()


labgrid/driver/usbsdwiredriver.py
```diff
@@ -37,3 +39,19 @@ def set_mode(self, mode):
...
+    @Driver.check_active
+    @step(title='sdmux_get')
+    def get_mode(self):
+        cmd = self.mux.command_prefix + [
+            self.tool,
+            "-e",
+            self.mux.control_serial,
+            "-u",
+        ]
+        proc = subprocess.run(
```
Emantor, line 52: I think you should use the `processwrapper` here.

jpuderer: Updated.  I was using subprocess here, since the sdmux is using subprocess in the same place.  I can create a new pull requests for sdmux after this one, but I'll need someone else to test, since I don't have an sdmux.

labgrid/driver/usbsdwiredriver.py
```diff
@@ -37,3 +38,15 @@ def set_mode(self, mode):
...
+    @step(title='sdmux_get')
+    def get_mode(self):
+        cmd = self.mux.command_prefix + [
+            self.tool,
+            "-e",
+            self.mux.control_serial,
+            "-u",
+        ]
+        result = processwrapper.check_output(cmd)
+        return re.sub('^.*?:\s*', "", result.strip().decode())
```
Bastian-Krause, line 52: Looking at the [code](https://git.tizen.org/cgit/tools/testlab/sd-mux/tree/src/main.cpp?id=9dd189d973da64e033a0c5c2adb3d94b23153d94#n583), I guess this can be simplified:

```suggestion
        return result.split(": ", maxsplit=1)[1].strip().decode()
```
The `re` import can be dropped.

Bastian-Krause (changes requested):
Thank you for your contribution. Please squash your commits into one.

labgrid/driver/usbsdwiredriver.py
jpuderer: Sure.  Do you not have the option to squash when you merge the pull request?

Bastian-Krause (approved):

labgrid/driver/usbsdwiredriver.py
Bastian-Krause: Thanks. We haven't enabled that option. I don't like dropping logical commit separation accidentally and I don't think it's necessary to fiddle with your commits if I don't have to :)

Emantor (approved):

PR #1111, opened 2023-03-27
remote/client: allow specifying resources by name for various sub commands

### LINKED ISSUE #364: client: subcommands working on resources should accept resource name argument
In order to use resources of which multiple instances exist labgrid-client should accept a name argument just like `labgrid-client console` does:
```
usage: labgrid-client console [-h] [-l] [name]

positional arguments:
  name        optional resource name
[...]
```
(or via a `-n`/`--name` option)

### LINKED ISSUE #1023: Second PowerPort
Hi,
I am currently evaluation Labgrid. First of all - good work!
We are having the requirement that we need to control two power outlets using labgrid. Thats what I want to discuss here.
Thoughts:
* the `power` command should be simple to use, thus a (configurable) default powerport would be nice
* of course the `power` command should be backwards compatible
My ideas:
* Similar to the target `main`  in a configuration file, also named resources with a specific name should get priorized
* In order to handle the priorization and still having the opportunity to give the resource a meaningful name a prefix would be nice
* My idea is to allow a default name `main` as well as prefix `main#<name>`, while the prefix can be also omitted. I am using `NetworkSiSPMPowerPort` just as an example below.
```
add-named-match */NetworkSiSPMPowerPort/Port1 main#dut
add-named-match */NetworkSiSPMPowerPort/Port2 tool
# switches on the default/main port
power on
# it's also possible to directly address the resource with/without prefix
power on --name main#dut
power on --name dut
power on --name tool
```
* The same smart matching should work in a config file
```
    drivers:
      SiSPMPowerDriver: {}
```
Should bind against the single `NetworkSiSPMPowerPort`/`SiSPMPowerPort` but take preference of the resource with name `main` / `main#<name>` when there are multiple resource matches available.
* Multiple main/default ports should get reported if there is no config file (with explicit driver instantiation) used.
```
add-named-match */NetworkSiSPMPowerPort/Port1 main#dut
add-named-match */NetworkSiSPMPowerPort/Port2 tool
add-named-match */NetworkUSBPowerPort/hub main#target
# this should report the conflict
power on
```

I have a first implementation of that in my local git together with some tests scripts (bash based using OSHT) that I used to confirm my patches.

What do you think?

Bastian-Krause:
**Description**
There might be multiple instances of a resource type in a single place (either specified via resources with names in the configuration or via named matches: `labgrid-client add-named-match`).

This allows addressing them via `-n`/`--name`. If a configuration is given, the drivers specified there will be used if they have the same name as the resource. Otherwise new driver instances will be created.

These sub commands gained the new argument:
- `power`
- `dfu`
- `flashscript`
- `bootstrap`
- `sd-mux`
- `usb-mux`
- `forward`
- `tmc`
- `write-image`

**Checklist**
- [ ] Tests for the feature 
- [.] PR has been tested (tested power sub command only)

Closes #1026
Closes #1117
Fixes #1023
Fixes #364

### DIFF
--- a/contrib/completion/labgrid-client.bash
+++ b/contrib/completion/labgrid-client.bash
@@ -291,11 +291,15 @@ _labgrid_client_power()
     -t|--delay)
         return
         ;;
+    -n|--name)
+        _labgrid_complete match-names "$cur"
+        return
+        ;;
     esac
 
     case "$cur" in
     -*)
-        COMPREPLY=( $(compgen -W "--delay $_labgrid_shared_options" -- "$cur") )
+        COMPREPLY=( $(compgen -W "--delay --name $_labgrid_shared_options" -- "$cur") )
         ;;
     *)
         local args
@@ -326,9 +330,14 @@ _labgrid_client_io()
         local args
         _labgrid_count_args || return
         # only complete second argument
-        [ "$args" -ne 2 ] && return
-
-        COMPREPLY=( $(compgen -W "high low get" -- "$cur") )
+        case "$args" in
+        2)
+            COMPREPLY=( $(compgen -W "high low get" -- "$cur") )
+            ;;
+        3)
+            _labgrid_complete match-names "$cur"
+            ;;
+        esac
         ;;
     esac
 }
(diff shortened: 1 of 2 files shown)


jpuderer:
Much better.  Thanks!

jluebbe (approved):
I've not reviewed the bash completion. The rest looks good.

Emantor (approved):

PR #1115, opened 2023-03-28
driver/resource: Support dediprog

edersondisouza:
Akin to 'flashrom', the Dediprog resource expects 'dpcmd' to be available on the host. Currently, only parameter that can be set is VCC.

**Description**
The main advantage to use it instead of `flashrom` is that it allows for using images smaller than the SPI memory, which is then automatically filled with `0xff`. For `flashrom`, one needs the image to have the right size - although one can avoid writing the whole of it by having a layout file - which adds to the inconvenience. Using `dpcmd` is more straightforward and faster.
Tested locally using an SF100.

**Checklist**
- [X] Tests for the feature 

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -1025,6 +1025,35 @@ resource available on a remote computer
 Used by:
   - `FlashScriptDriver`_
 
+DediprogFlasher
+~~~~~~~~~~~~~~~
+A DediprogFlasher resource is used to configure the parameters to a locally installed
+dpmcd instance. It is assumed that dpcmd is installed on the host and the
+executable can be configured via:
+
+.. code-block:: yaml
+
+  tools:
+    dpcmd: '/usr/sbin/dpcmd'
+
+Arguments:
+  - vcc (str): '3.5V', '2.5V' or '1.8V'.
+
+For instance, to flash using 3.5V vcc:
+
+.. code-block:: yaml
+
+  DediprogFlasher:
+    vcc: '3.5V'
+
+
+Used by:
+  - `DediprogFlashDriver`_
+
+NetworkDediprogFlasher
+~~~~~~~~~~~~~~~~~~~~~~
+A NetworkDediprogFlasher describes a `DediprogFlasher`_ available on a remote computer.
+
 XenaManager
 ~~~~~~~~~~~
 A XenaManager resource describes a Xena Manager instance which is the instance the
@@ -2623,6 +2652,30 @@ Key        Description
 Properties of these keys can be selected using the Python format string syntax,
 e.g. ``{device.devnode}`` to select the device node path of
 :any:`USBFlashableDevice`
+
+DediprogFlashDriver
+~~~~~~~~~~~~~~~~~~~
+The :any:`DediprogFlashDriver` is used to flash an SPI device using DediprogFlasher dpcmd.
+
+.. code-block:: yaml
+
+   DediprogFlashDriver:
+     image: 'foo'
+   images:
+     foo: ../images/image_to_load.raw
+
+Binds to:
+  DediprogFlasher_resource:
+    - `DediprogFlasher`_
+    - `NetworkDediprogFlasher`_
+
+Arguments:
+  - image (str): optional, key in :ref:`images <labgrid-device-config-images>` containing the path
+    of an image to flash onto the target
+
+The DediprogFlashDriver allows using DediprogFlasher dpcmd to flash or erase SPI
+devices. It is assumed that the device flashing is an exporter wired, via
+DediprogFlasher SF100 for instance, to the device being flashed.
 
 XenaDriver
 ~~~~~~~~~~
--- a/labgrid/driver/dediprogflashdriver.py
+++ b/labgrid/driver/dediprogflashdriver.py
@@ -1,0 +1,70 @@
+import os.path
+import logging
+import attr
+
+from ..resource import NetworkDediprogFlasher
+from ..factory import target_factory
+from ..step import step
+from .common import Driver, check_file
+from ..util.managedfile import ManagedFile
+from ..util.helper import processwrapper
+
+
+@target_factory.reg_driver
+@attr.s(eq=False)
+class DediprogFlashDriver(Driver):
+    """ The DediprogFlashDriver uses the dediprog utility to write an image
+    to a raw rom. The driver is a pure wrapper of the dpcmd utility"""
+    bindings = {
+        'flasher': {"DediprogFlasher", NetworkDediprogFlasher},
+    }
+
+    image = attr.ib(validator=attr.validators.optional(attr.validators.instance_of(str)),
+                    default=None)
+
+    def __attrs_post_init__(self):
+        super().__attrs_post_init__()
+        self.logger = logging.getLogger(f'{self}')
+        if self.target.env:
+            self.tool = self.target.env.config.get_tool('dpcmd')
+        else:
+            self.tool = 'dpcmd'
+
+    def _get_dediprog_prefix(self):
+        return self.flasher.command_prefix + [self.tool]
+
+    def on_activate(self):
+        pass
+
+    def on_deactivate(self):
+        pass
+
+    def map_vcc(self):
+        vcc_map = {'3.5V': '0', '2.5V': '1', '1.8V': '2'}
+        return vcc_map[self.flasher.vcc]
+
+    @Driver.check_active
+    @step(title='call', args=['args'])
+    def __call__(self, *args):
+        vcc = self.map_vcc()
+        arg_list = list(args)
+        arg_list.append('--vcc')
+        arg_list.append(vcc)
+        arg_list.append('--silent')
+        processwrapper.check_output(self._get_dediprog_prefix() + arg_list)
+
+    @Driver.check_active
+    @step(args=['filename'])
+    def flash(self, filename=None):
+        if filename is None and self.image is not None:
+            filename = self.target.env.config.get_image_path(self.image)
+        filename = os.path.abspath(filename)
+        check_file(filename)
+        mf = ManagedFile(filename, self.flasher)
+        mf.sync_to_resource()
+
+        self('--auto', mf.get_remote_path(), '--verify', '-x', 'ff')
+
+    @Driver.check_active
+    def erase(self):
+        self('--erase')
--- a/labgrid/resource/dediprogflasher.py
+++ b/labgrid/resource/dediprogflasher.py
@@ -1,0 +1,15 @@
+import attr
+
+from ..factory import target_factory
+from .common import NetworkResource, Resource
+
+
+@target_factory.reg_resource
+@attr.s(eq=False)
+class DediprogFlasher(Resource):
+    vcc = attr.ib(validator=attr.validators.in_(('3.5V', '2.5V', '1.8V')))
+
+@target_factory.reg_resource
+@attr.s(eq=False)
+class NetworkDediprogFlasher(NetworkResource):
+    vcc = attr.ib(validator=attr.validators.in_(('3.5V', '2.5V', '1.8V')))
--- a/tests/test_dediprogflasher.py
+++ b/tests/test_dediprogflasher.py
@@ -1,0 +1,31 @@
+import pytest
+
+from labgrid.resource.dediprogflasher import DediprogFlasher, NetworkDediprogFlasher
+from labgrid.driver.dediprogflashdriver import DediprogFlashDriver
+
+
+def test_dediprog_network_create(target):
+    r = NetworkDediprogFlasher(target, name=None, vcc='3.5V', host='localhost')
+    assert isinstance(r, NetworkDediprogFlasher)
+    d = DediprogFlashDriver(target, name=None)
+    assert isinstance(d, DediprogFlashDriver)
+
+
+def test_dediprog_driver_create(target):
+    r = DediprogFlasher(target, name=None, vcc='2.5V')
+    assert isinstance(r, DediprogFlasher)
+
+    d = DediprogFlashDriver(target, name=None)
+    assert isinstance(d, DediprogFlashDriver)
+
+def test_dediprog_driver_map_vcc(target):
+    r = DediprogFlasher(target, name=None, vcc='2.5V')
+    d = DediprogFlashDriver(target, name=None)
+
+    m = d.map_vcc()
+    assert isinstance(m, str)
+    assert m == '1'
+
+def test_dediprog_driver_map_vcc_inv(target):
+    with pytest.raises(ValueError):
+        r = DediprogFlasher(target, name=None, vcc='2.6V')
--- a/labgrid/driver/__init__.py
+++ b/labgrid/driver/__init__.py
@@ -44,3 +44,4 @@ from .mqtt import TasmotaPowerDriver
 from .manualswitchdriver import ManualSwitchDriver
 from .usbtmcdriver import USBTMCDriver
 from .deditecrelaisdriver import DeditecRelaisDriver
+from .dediprogflashdriver import DediprogFlashDriver
--- a/labgrid/resource/__init__.py
+++ b/labgrid/resource/__init__.py
@@ -22,3 +22,4 @@ from .pyvisa import PyVISADevice
 from .provider import TFTPProvider
 from .mqtt import TasmotaPowerPort
 from .httpvideostream import HTTPVideoStream
+from .dediprogflasher import DediprogFlasher, NetworkDediprogFlasher
--- a/man/labgrid-device-config.5
+++ b/man/labgrid-device-config.5
@@ -98,6 +98,10 @@ The \fBtools:\fP top key provides paths to binaries such as fastboot.
 Path to the dfu\-util binary, used by the DFUDriver.
 See: \fI\%https://dfu\-util.sourceforge.net/\fP
 .TP
+.B \fBdpcmd\fP
+Path to the dpcmd binary, used by the DediprogFlashDriver.
+See: \fI\%https://github.com/DediProgSW/SF100Linux\fP
+.TP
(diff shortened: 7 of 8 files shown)


edersondisouza:
v2:
 - Fixed documentation build issues.

doc/configuration.rst
```diff
@@ -1025,6 +1025,27 @@ resource available on a remote computer
...
+Dediprog
+~~~~~~~~
+A Dediprog resource is used to configure the parameters to a locally installed
+dpmcd instance. It is assumed that dpcmd is installed on the host and the
+executable is configured in:
+
+.. code-block:: yaml
+
+  tools:
+    dpcmd: '/usr/sbin/dpcmd'
```
Emantor, line 1037: This tools entry belongs to the `DediprogDriver`, since it uses the tool. The example here should be something like:

```
Dediprog:
  vcc: 0
```

Also maybe we should rename this from `Dediprog` to `DediprogFlasher` and the Driver to `DediprogFlashDriver` since dediprog also builds eMMC programmers.

Emantor (changes requested):

labgrid/driver/dediprogdriver.py
```diff
@@ -0,0 +1,67 @@
...
+    def on_deactivate(self):
+        pass
+
+    @Driver.check_active
+    @step(title='call', args=['args'])
+    def __call__(self, *args):
+        arg_list = list(args)
+        arg_list.append('--vcc')
+        arg_list.append(f'{self.dediprog_resource.vcc}')
+        arg_list.append('--silent')
```
jluebbe, line 51: Is there still enough output to debug problems even with `--silent`?

doc/configuration.rst
```diff
@@ -1025,6 +1025,27 @@ resource available on a remote computer
...
+
+Arguments:
+  - vcc (int): 0 (3.5V), 1 (2.5V), 2 (1.8V)
```
jluebbe, line 1040: Hmm, I'd prefer to have the actual voltage in the configuration file, as this makes it notice a wrong configuration. You can map the voltages to the command line arguments in the driver.

jluebbe (changes requested):

edersondisouza:
v3:
 - Name resources `DediprogFlasher` and `NetworkDediprogFlasher`, and driver `DediprogFlashDriver`;
 - Use 'named' vcc argument: `3.5V` instead of `0`, for instance.

labgrid/driver/dediprogdriver.py
edersondisouza: Yep. The `--silent` just prevents a "fancy" time counter.

doc/configuration.rst
edersondisouza: While I did add a `vcc` example usage, I think that the tool really belongs here. This is similar to `Flashrom`, as where the tool is located can vary among exporters, so it's a resource attribute. Or did I understand `Flashrom` case wrong?

doc/configuration.rst
edersondisouza: Fair.

edersondisouza:
v4:
- Add some more tests that will help increase coverage.

edersondisouza:
Any comments? @Emantor @jluebbe

doc/configuration.rst
Emantor: You're right. I'm now wondering how we handle this on the exporter side, but thats not relevant for this PR.

Emantor (dismissed):

labgrid/driver/dediprogflashdriver.py
```diff
@@ -0,0 +1,77 @@
...
+from ..step import step
+from ..protocol import BootstrapProtocol
+from .common import Driver, check_file
+from ..util.managedfile import ManagedFile
+from ..util.helper import processwrapper
+
+
+@target_factory.reg_driver
+@attr.s(eq=False)
+class DediprogFlashDriver(Driver, BootstrapProtocol):
```
Bastian-Krause, line 16: @jluebbe @Emantor I always thought that the BootstrapProtocol was meant to write to non-persistent memory only (i.e. load an image into RAM). But the FlashromDriver seems to also implement it and now DediprogFlashDriver will, too?

In #206 we discussed that the QuartusHPSDriver explicitly does not implement the BootstrapProtocol. The plan was to introduce some kind of FlashProtocol instead later.

edersondisouza: Ok, will drop `BootstrapProtocol` on my next version. And don't really need `labgrid-client` support, as I'm using python library directly.

edersondisouza:
v5:
 - Dropped support for `BootstrapProtocol`, exposing `flash()` method;
 - Added `erase()` method;
 - Rebased.

labgrid/resource/dediprogflasher.py
```diff
@@ -0,0 +1,16 @@
...
+
+from ..factory import target_factory
+from .common import NetworkResource, Resource
+
+
+@target_factory.reg_resource
+@attr.s(eq=False)
+class DediprogFlasher(Resource):
+    """vcc: '3.5V', '2.5V' or '1.8V'"""
+    vcc = attr.ib(validator=attr.validators.instance_of(str))
```
Bastian-Krause, line 11:
```suggestion
    vcc = attr.ib(validator=attr.validators.in_(('3.5V', '2.5V', '1.8V')))
```
See [attrs.validators.in_()](https://www.attrs.org/en/stable/api.html#attrs.validators.in_).

labgrid/resource/dediprogflasher.py
```diff
@@ -0,0 +1,16 @@
+import attr
+
+from ..factory import target_factory
+from .common import NetworkResource, Resource
+
+
+@target_factory.reg_resource
+@attr.s(eq=False)
+class DediprogFlasher(Resource):
+    """vcc: '3.5V', '2.5V' or '1.8V'"""
```
Bastian-Krause, line 10: We can drop that if we use the suggestion below.

labgrid/driver/dediprogflashdriver.py
```diff
@@ -0,0 +1,78 @@
...
+    def on_deactivate(self):
+        pass
+
+    def map_vcc(self):
+        vcc_map = {'3.5V': '0', '2.5V': '1', '1.8V': '2'}
+        vcc = self.dediprog_resource.vcc
+        if vcc not in vcc_map:
+            raise ValueError(f"VCC {vcc} not '1.8V', '2.5V' or '3.5V'")
+
+        return vcc_map[vcc]
```
Bastian-Krause, line 50: I guess it's better to check that in the resource class, suggestion below.

labgrid/resource/dediprogflasher.py
```diff
@@ -0,0 +1,16 @@
...
+    vcc = attr.ib(validator=attr.validators.instance_of(str))
+
+@target_factory.reg_resource
+@attr.s(eq=False)
+class NetworkDediprogFlasher(NetworkResource):
+    vcc = attr.ib(validator=attr.validators.instance_of(str))
```
Bastian-Krause, line 16: Same here.

tests/test_dediprogflasher.py
```diff
@@ -0,0 +1,43 @@
...
+    assert (m == '1')
+
+def test_dediprog_driver_map_vcc_inv(target):
+    r = DediprogFlasher(target,
+                 name=None,
+                 vcc='2.6V')
+    d = DediprogFlashDriver(target, name=None)
+
+    with pytest.raises(ValueError):
+        m = d.map_vcc()
```
Bastian-Krause, line 43: We can drop this if we use the suggestion above.

Bastian-Krause (changes requested):

edersondisouza:
v6:
 - Using attr `in_` to map validate vcc argument;
 - Added `dpcmd` entry to `labgrid-device-config` doc;
 - Rebased.

labgrid/driver/dediprogflashdriver.py
```diff
@@ -0,0 +1,72 @@
...
+@target_factory.reg_driver
+@attr.s(eq=False)
+class DediprogFlashDriver(Driver):
+    """ The DediprogFlashDriver uses the dediprog utility to write an image
+    to a raw rom. The driver is a pure wrapper of the dpcmd utility"""
+    bindings = {
+        'dediprog_resource': {"DediprogFlasher", NetworkDediprogFlasher},
+    }
+
+    image = attr.ib(default=None)
```
Bastian-Krause, line 22:
```suggestion
    image = attr.ib(validator=attr.validators.instance_of(str), default=None)
```
The docs say that we expect a string here, so let's check for that.

labgrid/driver/dediprogflashdriver.py
```diff
@@ -0,0 +1,72 @@
...
+        if self.target.env:
+            self.tool = self.target.env.config.get_tool('dpcmd')
+        else:
+            self.tool = 'dpcmd'
+        self.logger.debug('Tool %s', self.tool)
+
+    def _get_dediprog_prefix(self):
+        return self.dediprog_resource.command_prefix+[
+            self.tool
+        ]
```
Bastian-Krause, line 36:
```suggestion
        return self.dediprog_resource.command_prefix + [self.tool]
```

tests/test_dediprogflasher.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def test_dediprog_network_create(target):
+    r = NetworkDediprogFlasher(target,
+                       name=None,
+                       vcc='3.5V',
+                       host='localhost')
+    assert (isinstance(r, NetworkDediprogFlasher))
+    d = DediprogFlashDriver(target, name=None)
+    assert (isinstance(d, DediprogFlashDriver))
```
Bastian-Krause, line 14:
```suggestion
    assert isinstance(d, DediprogFlashDriver)
```

tests/test_dediprogflasher.py
```diff
@@ -0,0 +1,40 @@
...
+from labgrid.resource.dediprogflasher import DediprogFlasher, NetworkDediprogFlasher
+from labgrid.driver.dediprogflashdriver import DediprogFlashDriver
+
+
+def test_dediprog_network_create(target):
+    r = NetworkDediprogFlasher(target,
+                       name=None,
+                       vcc='3.5V',
+                       host='localhost')
+    assert (isinstance(r, NetworkDediprogFlasher))
```
Bastian-Krause, line 12:
```suggestion
    assert isinstance(r, NetworkDediprogFlasher)
```

tests/test_dediprogflasher.py
```diff
@@ -0,0 +1,40 @@
...
+    d = DediprogFlashDriver(target, name=None)
+    assert (isinstance(d, DediprogFlashDriver))
+
+
+def test_dediprog_driver_create(target):
+    r = DediprogFlasher(target,
+                 name=None,
+                 vcc='2.5V')
+    assert (isinstance(r, DediprogFlasher))
```
Bastian-Krause, line 21:
```suggestion
    assert isinstance(r, DediprogFlasher)
```

tests/test_dediprogflasher.py
```diff
@@ -0,0 +1,40 @@
...
+
+    d = DediprogFlashDriver(target, name=None)
+    assert (isinstance(d, DediprogFlashDriver))
```
Bastian-Krause, line 24:
```suggestion
    assert isinstance(d, DediprogFlashDriver)
```

tests/test_dediprogflasher.py
```diff
@@ -0,0 +1,40 @@
...
+
+def test_dediprog_driver_map_vcc(target):
+    r = DediprogFlasher(target,
+                 name=None,
+                 vcc='2.5V')
+    d = DediprogFlashDriver(target, name=None)
+
+    m = d.map_vcc()
+    assert (isinstance(m, str))
```
Bastian-Krause, line 33:
```suggestion
    assert isinstance(m, str)
```

tests/test_dediprogflasher.py
```diff
@@ -0,0 +1,40 @@
...
+    assert (m == '1')
```
Bastian-Krause, line 34:
```suggestion
    assert m == '1'
```

tests/test_dediprogflasher.py
```diff
@@ -0,0 +1,40 @@
...
+
+from labgrid.resource.dediprogflasher import DediprogFlasher, NetworkDediprogFlasher
+from labgrid.driver.dediprogflashdriver import DediprogFlashDriver
+
+
+def test_dediprog_network_create(target):
+    r = NetworkDediprogFlasher(target,
+                       name=None,
+                       vcc='3.5V',
+                       host='localhost')
```
Bastian-Krause, line 11:
```suggestion
    r = NetworkDediprogFlasher(target, name=None, vcc='3.5V', host='localhost')
```

tests/test_dediprogflasher.py
```diff
@@ -0,0 +1,40 @@
...
+    assert (isinstance(r, NetworkDediprogFlasher))
+    d = DediprogFlashDriver(target, name=None)
+    assert (isinstance(d, DediprogFlashDriver))
+
+
+def test_dediprog_driver_create(target):
+    r = DediprogFlasher(target,
+                 name=None,
+                 vcc='2.5V')
```
Bastian-Krause, line 20:
```suggestion
    r = DediprogFlasher(target, name=None, vcc='2.5V')
```

tests/test_dediprogflasher.py
```diff
@@ -0,0 +1,40 @@
...
+    assert (isinstance(r, DediprogFlasher))
+
+    d = DediprogFlashDriver(target, name=None)
+    assert (isinstance(d, DediprogFlashDriver))
+
+def test_dediprog_driver_map_vcc(target):
+    r = DediprogFlasher(target,
+                 name=None,
+                 vcc='2.5V')
```
Bastian-Krause, line 29:
```suggestion
    r = DediprogFlasher(target, name=None, vcc='2.5V')
```

tests/test_dediprogflasher.py
```diff
@@ -0,0 +1,40 @@
...
+
+    m = d.map_vcc()
+    assert (isinstance(m, str))
+    assert (m == '1')
+
+def test_dediprog_driver_map_vcc_inv(target):
+    with pytest.raises(ValueError):
+        r = DediprogFlasher(target,
+                     name=None,
+                     vcc='2.6V')
```
Bastian-Krause, line 40:
```suggestion
        r = DediprogFlasher(target, name=None, vcc='2.6V')
```

labgrid/driver/dediprogflashdriver.py
```diff
@@ -0,0 +1,72 @@
...
+    image = attr.ib(default=None)
+
+    def __attrs_post_init__(self):
+        super().__attrs_post_init__()
+        self.logger = logging.getLogger(f'{self}')
+        if self.target.env:
+            self.tool = self.target.env.config.get_tool('dpcmd')
+        else:
+            self.tool = 'dpcmd'
+        self.logger.debug('Tool %s', self.tool)
```
Bastian-Krause, line 31:
```suggestion
```
I don't think this is useful, `processwrapper.check_output()` logs the command anyway and passes it to the StepReporter.

labgrid/driver/dediprogflashdriver.py
```diff
@@ -0,0 +1,72 @@
...
+from ..util.helper import processwrapper
+
+
+@target_factory.reg_driver
+@attr.s(eq=False)
+class DediprogFlashDriver(Driver):
+    """ The DediprogFlashDriver uses the dediprog utility to write an image
+    to a raw rom. The driver is a pure wrapper of the dpcmd utility"""
+    bindings = {
+        'dediprog_resource': {"DediprogFlasher", NetworkDediprogFlasher},
```
Bastian-Krause, line 19: Hm, I'm not too happy about the name of the binding. Maybe `flasher` is better?

Bastian-Krause (changes requested):
Thanks for working on this. I've added some more comments.

tests/test_dediprogflasher.py
Bastian-Krause: I confused the check for valid values and the mapping here.

edersondisouza:
v7:
  - Fixed styling issues;
  - Ensure `image` attr is actually a string;
  - Removed redundant log message;
  - Renamed resource to  `flasher`;
  - Rebased.

jluebbe (approved):

jluebbe:
Note that there is no support for udev-matching and so no support in the exporter. The `NetworkDediprogFlasher` needs to be configured manually in the local environments.

labgrid/driver/dediprogflashdriver.py
```diff
@@ -0,0 +1,69 @@
...
+@target_factory.reg_driver
+@attr.s(eq=False)
+class DediprogFlashDriver(Driver):
+    """ The DediprogFlashDriver uses the dediprog utility to write an image
+    to a raw rom. The driver is a pure wrapper of the dpcmd utility"""
+    bindings = {
+        'flasher': {"DediprogFlasher", NetworkDediprogFlasher},
+    }
+
+    image = attr.ib(validator=attr.validators.instance_of(str), default='')
```
Bastian-Krause, line 22:
```suggestion
    image = attr.ib(validator=attr.validators.optional(attr.validators.instance_of(str)), default=None)
```
Oh, I'm sorry that my suggestion did not work. That's what I meant, actually.

Bastian-Krause (changes requested):
Everything else looks good.

labgrid/driver/dediprogflashdriver.py
Bastian-Krause: Fixed it myself, so we can get this merged.

Bastian-Krause (approved):

jluebbe (approved):

PR #1116, opened 2023-03-28
driver/power: Add support for Digital Loggers PDU with HTTP API

edersondisouza:
Tested with Web Power Switch 7.

**Description**
Some Digital Loggers PDU have a simple HTTP GET API to control and get outlet status. This patch adds support to it, based on `simplerest.py`. Tested locally with Web Power Switch 7.

**Checklist**
- [X] Tests for the feature 


edersondisouza:
I get that dropping coverage isn't desirable, but how to tackle that? There seems to be no mocking mechanism to exercise those lines. (And drivers for other PDUs don't have more specific tests I could find.) Any hints here?

Emantor:
> I get that dropping coverage isn't desirable, but how to tackle that? There seems to be no mocking mechanism to exercise those lines. (And drivers for other PDUs don't have more specific tests I could find.) Any hints here?

There is no mocking yet. This could be added but is out of scope for this PR. We generally write tests for core functionality changes, but as soon as you have to interact with hardware some leniency and trust into the PR submitter that his change was tested is required.

doc/configuration.rst
```diff
@@ -160,6 +160,9 @@ Currently available are:
 ``digipower``
   Controls a DigiPower PDU via a simple HTTP API.
 
+``digitalloggers_http``
+  Control a Digital Loggers PDU that use the legacy HTTP API.
```
jluebbe, line 164: This should mention the expected format of the `host` parameter. It looks like it should be `http://mypdu`?

edersondisouza:
> The backend comment mentions the 'legacy HTTP API'. If there is a newer API, why did you use the legacy API?
> 
> Otherwise, the PR looks good.

I used Digital Loggers terminology. And the new one isn't present on older PDUs - those only have the "legacy" one. And this patch is for the legacy because, well, that's the PDU I'm targetting =D

edersondisouza:
Any comments? @jluebbe

Emantor (approved):

edersondisouza:
Gentle reminder about this PR - any comments, issues, suggestions @jluebbe?

jluebbe (approved):

PR #1117, opened 2023-03-29
Fix client to pass in driver name, when getting driver for 'io' command.

jpuderer:
**Description**

The 'io' command in the labgrid-client accepts 'name' as an optional parameter, but only uses it when it creates a driver for a remote resource.  

When invoking the 'io' command in the labgrid client with an environment that already defines drivers, labgrid-client returns the first matching driver it finds, and not the one matching the 'name' parameter on the command line.

**Checklist**
- [x] PR has been tested


Emantor:
See #1111 for a complete client conversion for named resources that is likely getting merged soon :+1:

jpuderer:
Much better, thanks.  I'll give #1111 a try.

PR #1126, opened 2023-04-05
Add support for removing remote files when no longer needed

JoshuaWatt:
**Description**
Files controlled with `ManagedFile` can sometimes be quite large and consume a large amount of space on the exporter, potentially causing it to run out and fail tests. In these cases, it is preferred to transfer the files every time they are needed instead of keeping them around in the cache.

This adds a new `cleanup` API to `ManagedFile` to allow the remote file to be cleaned up manually when no longer required. The `FlashScriptDriver` has been extended to support this new API as well

**Checklist**
- [x] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,6 +3,8 @@ Release 23.0 (unreleased)
 
 - `ManagedFile` now has the ability to remove the file from the remote exporter
   when longer needed
+- `FlashScriptDriver` now has the ability to remote the flash script from the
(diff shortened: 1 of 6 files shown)


Bastian-Krause:
Could you explain why the tmpfiles.d suggestion in the "attention box" in the [exporter docs](https://labgrid.readthedocs.io/en/latest/getting_started.html#exporter) does not work for you?

PR #1130, opened 2023-04-06
Add sshfs exporter class

JoshuaWatt:
Adds a class that allows tests to export a local directory to an exporter via sshfs.

In order to prevent the exporter from being able to access any file that the user has permission for on the file system (which would allow it to read SSH keys for example), the SFTP server is run in an isolated container by default using podman, although this can be overridden to use docker or directly execute the SFTP server if the exporters are trusted.

**Description**

**Checklist**
- [ ] Tests for the feature 

- [ ] PR has been tested

### DIFF
--- a/dockerfiles/Dockerfile
+++ b/dockerfiles/Dockerfile
@@ -121,3 +121,10 @@ COPY --from=ser2net /opt/ser2net /
 VOLUME /opt/conf
 
 CMD ["/entrypoint.sh"]
+
+#
+# SFTP server
+#
+FROM alpine:3.17 as labgrid-sftp-server
+
+RUN apk add openssh-sftp-server
--- a/labgrid/util/sshfs.py
+++ b/labgrid/util/sshfs.py
@@ -1,0 +1,152 @@
+import logging
+import os
+import subprocess
+import random
+import string
+import time
+import shlex
+from contextlib import contextmanager
+
+import attr
+
+from .ssh import sshmanager
+from .timeout import Timeout
+from ..driver.exception import ExecutionError
+from ..resource.common import Resource, NetworkResource
+
+
+DEFAULT_SFTP_SERVER = "podman run --rm -i --mount type=bind,source={path},target={path} labgrid/sftp-server:latest usr/lib/ssh/sftp-server -e {readonly}"
+DEFAULT_RO_OPT = "-R"
+
+
+@attr.s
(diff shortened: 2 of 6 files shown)


dockerfiles/Dockerfile
```diff
@@ -121,3 +121,10 @@ COPY --from=ser2net /opt/ser2net /
 VOLUME /opt/conf
 
 CMD ["/entrypoint.sh"]
+
+#
+# SFTP server
+#
+FROM alpine:3.17 as labgrid-sftp-server
+
+RUN apk add openssh-sftp-server
```
jluebbe, line 130: It seems that the podman in Ubuntu 22.04 tries to build the other targets as well, so this should be a different Dockerfile.

labgrid/util/sshfs.py
```diff
@@ -0,0 +1,152 @@
...
+
+import attr
+
+from .ssh import sshmanager
+from .timeout import Timeout
+from ..driver.exception import ExecutionError
+from ..resource.common import Resource, NetworkResource
+
+
+DEFAULT_SFTP_SERVER = "podman run --rm -i --mount type=bind,source={path},target={path} labgrid/sftp-server:latest usr/lib/ssh/sftp-server -e {readonly}"
```
jluebbe, line 18: Hmm, how do we want to tell the user that we expect this image to exist locally? Do we want to have a runtime check for the image or is documentation enough?

labgrid/util/sshfs.py
```diff
@@ -0,0 +1,152 @@
...
+    Exports a local directory to a remote device using "reverse" SSH FS
+    """
+
+    local_path = attr.ib(
+        validator=attr.validators.instance_of(str),
+        converter=lambda x: os.path.abspath(str(x)),
+    )
+    resource = attr.ib(
+        validator=attr.validators.instance_of(Resource),
+    )
```
jluebbe, line 34: We might want to connect to an exporter for a specific resource (similar to ManagedFile) or to the target as well. So I'd prefer to pass host/port instead of a resource. That would mean passing a config separately as well.

@JoshuaWatt Do you see a downside?

jluebbe (changes requested):
Looks good otherwise.

jluebbe:
Thanks for the update. I'll close this for now, but we can reopen when it becomes relevant again.

PR #1131, opened 2023-04-07
remote/client: use crossbar_url from configuration file

nlabriet:
**Description**
Add the ability for labgrid-client to use the crossbar URL coming from the configuration file.

The crossbar URL come from (first wins):
 - -x/--crossbar parameter
 - crossbar_url in configuration file
 - LG_CROSSBAR environment variable
 - default: "ws://127.0.0.1:20408/ws"

**Checklist**
  - The documentation is already there (man/labgrid-device-config.rst)
- [X] Tests for the feature 
  - Only manual test, with a configuration file having the crossbar_url option.

- [X] PR has been tested

### DIFF
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -1386,8 +1386,7 @@ def main():
         '--crossbar',
         metavar='URL',
         type=str,
-        default=os.environ.get("LG_CROSSBAR", "ws://127.0.0.1:20408/ws"),
-        help="crossbar websocket URL (default: %(default)s)"
+        help="crossbar websocket URL (default: value from env variable LG_CROSSBAR, otherwise ws://127.0.0.1:20408/ws)"
     )
     parser.add_argument(
         '-c',
@@ -1795,8 +1794,22 @@ def main():
         try:
             signal.signal(signal.SIGTERM, lambda *_: sys.exit(0))
 
-            session = start_session(args.crossbar, os.environ.get("LG_CROSSBAR_REALM", "realm1"),
-                                    extra)
+            try:
+                crossbar_url = args.crossbar or env.config.get_option('crossbar_url')
+            except (AttributeError, KeyError):
+                # in case of no env or not set, use LG_CROSSBAR env variable or default
+                crossbar_url = os.environ.get("LG_CROSSBAR", "ws://127.0.0.1:20408/ws")
+
+            try:
+                crossbar_realm = env.config.get_option('crossbar_realm')
+            except (AttributeError, KeyError):
+                # in case of no env, use LG_CROSSBAR_REALM env variable or default
+                crossbar_realm = os.environ.get("LG_CROSSBAR_REALM", "realm1")
+
+            logging.debug('Starting session with "%s", realm: "%s"', crossbar_url,
+                    crossbar_realm)
+
+            session = start_session(crossbar_url, crossbar_realm, extra)
             try:
                 if asyncio.iscoroutinefunction(args.func):
                     if getattr(args.func, 'needs_target', False):


Emantor (dismissed):

Emantor (dismissed):

labgrid/remote/client.py
```diff
@@ -1795,6 +1794,15 @@ def main():
...
             signal.signal(signal.SIGTERM, lambda *_: sys.exit(0))
 
+            if args.crossbar is None and env is not None:
+                # If no crossbar URL provided, try config file crossbar_url
+                args.crossbar = env.config.get_option('crossbar_url', None)
+
+            if args.crossbar is None:
+                # then LG_CROSSBAR environment variable
+                # fallback to default
+                args.crossbar = os.environ.get("LG_CROSSBAR", "ws://127.0.0.1:20408/ws")
```
Bastian-Krause, line 1804: The `crossbar_realm` option should also be read from the environment configuration.

labgrid/remote/client.py
```diff
@@ -1795,6 +1794,15 @@ def main():
...
+
```
Bastian-Krause, line 1805: Since the crossbar URL and realm can be defined in several places, we should (debug) log the URL and realm that is finally used before starting the session.

Bastian-Krause (changes requested):
Thanks for contributing to labgrid. A few suggestions below..

labgrid/remote/client.py
nlabriet: I've added reading the realm from the configuration file.

labgrid/remote/client.py
```diff
@@ -1795,8 +1794,32 @@ def main():
...
+            if realm is None:
+                # If not set, get crossbar realm from LG_CROSSBAR_REALM environment variable
+                # fallback to default
+                realm = os.environ.get("LG_CROSSBAR_REALM", "realm1")
+
+            if args.verbose:
+                print(f"Starting session to crossbar (url '{args.crossbar}' realm '{realm}')")
+            session = start_session(args.crossbar, realm, extra)
```
Bastian-Krause, line 1822: This can be simplified to:
```suggestion
            try:
                crossbar_url = args.crossbar or env.config.get_option('crossbar_url')
            except (AttributeError, KeyError):
                # in case of no env or not set, use LG_CROSSBAR env variable or default
                crossbar_url = os.environ.get("LG_CROSSBAR", "ws://127.0.0.1:20408/ws")

            try:
                crossbar_realm = env.config.get_option('crossbar_realm')
            except (AttributeError, KeyError):
                # in case of no env or not set, use LG_CROSSBAR_REALM env variable or default
                crossbar_realm = os.environ.get("LG_CROSSBAR_REALM", "realm1")

            logging.debug('Starting crossbar session with "%s", realm: "%s")', crossbar_url,
                          crossbar_realm)
            session = start_session(crossbar_url, crossbar_realm, extra)
```

labgrid/remote/client.py
```diff
@@ -1386,8 +1386,7 @@ def main():
         '--crossbar',
         metavar='URL',
         type=str,
-        default=os.environ.get("LG_CROSSBAR", "ws://127.0.0.1:20408/ws"),
-        help="crossbar websocket URL (default: %(default)s)"
+        help="crossbar websocket URL (default: ws://127.0.0.1:20408/ws)"
```
Bastian-Krause, line 1389: I guess we should mention how the default works here:
```suggestion
        help="crossbar websocket URL (default: value from env variable LG_CROSSBAR_REALM, otherwise ws://127.0.0.1:20408/ws)"
```

labgrid/remote/client.py
Bastian-Krause: I've added a suggestion that actually uses logging instead of printing.

Bastian-Krause (changes requested):
When you're okay with the suggestions, please squash them into a single commit.

labgrid/remote/client.py
nlabriet: Right, it's simpler that way.

labgrid/remote/client.py
nlabriet: Ok, I've updated the comment

PR #1133, opened 2023-04-12
remote/client: fix using token

nlabriet:
**Description**
Make it possible to use tokens.
- When using `labgrid-client reserve --shell`, the output also contained the selected role, preventing eval from assigning LG_TOKEN.
  - The proposed fix is to disable selected role output for the reserve command
- If a configuration file is used, the role was selected before considering the token.
  - The proposed fix is the leave the role to None if the place starts with '+', and determine the role later on

**Checklist**
  - Already exists, now it's usable
- [X] Tests for the feature
  - Tested in my setup with a configuration file and tokens
```
$ eval `labgrid-client -c ~/boards.yaml reserve --shell --wait family=essential`
$ labgrid-client -c ~/boards.yaml -p + lock
acquired place board1
```

- [X] PR has been tested


Emantor (approved):

Emantor:
Thanks!

nlabriet:
Thank you

PR #1135, opened 2023-04-12
Allow setting initial Strategy state in client

Bastian-Krause:
**Description**
This is the equivalent of the pytest plugin's `--lg-initial-state` option. It can be used to specify an initial state the device is known to be in. This is useful during development. The Strategy used must implement the `force()` method.

**Checklist**
- [ ] Tests for the feature 
- [x] PR has been tested

### DIFF
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -664,6 +664,9 @@ class ClientSession(ApplicationSession):
                     from .. import StepReporter
                     StepReporter.start()
                 strategy = target.get_driver("Strategy")
+                if self.args.initial_state:
+                    print(f"Setting initial state to {self.args.initial_state}")
+                    strategy.force(self.args.initial_state)
                 print(f"Transitioning into state {self.args.state}")
                 strategy.transition(self.args.state)
                 # deactivate console drivers so we are able to connect with microcom later
@@ -1378,6 +1381,7 @@ def main():
     place = os.environ.get('LG_PLACE', place)
     state = os.environ.get('STATE', None)
     state = os.environ.get('LG_STATE', state)
+    initial_state = os.environ.get('LG_INITIAL_STATE', None)
     token = os.environ.get('LG_TOKEN', None)
 
     parser = argparse.ArgumentParser()
@@ -1408,6 +1412,13 @@ def main():
         type=str,
         default=state,
         help="strategy state to switch into before command"
+    )
+    parser.add_argument(
+        '-i',
+        '--initial-state',
+        type=str,
+        default=initial_state,
+        help="strategy state to force into before switching to desired state"
     )
     parser.add_argument(
         '-d',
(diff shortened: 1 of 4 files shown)


Bastian-Krause:
CI tests will be fixed by #1134.

labgrid/remote/client.py
```diff
@@ -664,6 +664,9 @@ def _get_target(self, place):
                     from .. import StepReporter
                     StepReporter.start()
                 strategy = target.get_driver("Strategy")
+                if self.args.initial_state:
+                    print(f"Setting initial state to {self.args.initial_state}")
+                    strategy.force(self.args.initial_state)
```
Emantor, line 669: We should check and error out with a warning if the strategy doesn't implement the `force()` method.

Emantor (changes requested):

man/labgrid-client.1
```diff
@@ -88,6 +92,11 @@ for the \fB+\fP place expansion.
 This variable can be used to specify a state which the device transitions into
 before executing a command. Requires a configuration file and a Strategy
 specified for the device.
+.SS LG_INITIAL_STATE
+.sp
+This variable can be used to specify an initial state the device is known to
+be in. This is useful during development. The Strategy used must implement the
+\fBforce()\fP method. Requires a desired state.
```
jluebbe, line 99: What do you mean with 'Requires a desired state'? In addition to the `force()` method, it has the same requirements as LG_STATE (environment and strategy).

man/labgrid-client.rst
```diff
@@ -45,6 +45,9 @@ OPTIONS
 -s STATE, --state STATE
     set an initial state before executing a command, requires a configuration
     file and strategy
+-i INITIAL_STATE, --initial-state INITIAL_STATE
+    strategy state to force into before switching to desired state, requires a
+    desired state
```
jluebbe, line 50: As above.

man/labgrid-client.rst
```diff
@@ -75,6 +78,12 @@ This variable can be used to specify a state which the device transitions into
 before executing a command. Requires a configuration file and a Strategy
 specified for the device.
 
+LG_INITIAL_STATE
+~~~~~~~~~~~~~~~~
+This variable can be used to specify an initial state the device is known to
+be in. This is useful during development. The Strategy used must implement the
+``force()`` method. Requires a desired state.
```
jluebbe, line 85: Here as well. Please also use semantic newlines.

man/labgrid-client.1
Bastian-Krause: "Desired state" means -s/--state. Any suggestions how to improve the wording?

man/labgrid-client.rst
Bastian-Krause: We didn't follow that rule in the man page, but I'll start now.

jluebbe (approved):

Emantor (approved):

PR #1136, opened 2023-04-18
Unpin labgrid's dependencies

Bastian-Krause:
**Description**
With #1079, crossbar is no longer installed in the same virtualenv as labgrid and its dependencies. We can now loosen labgrid's dependencies.

Use the known-good previously pinned versions as minimum versions.

We cannot unpin crossbar's version because of continuing issues: [dependencies being incompatible with Python > 3.11](https://github.com/erikrose/parsimonious/issues/225), [PyPI packages depending on non-PyPI packages](https://github.com/crossbario/crossbar/issues/2069) etc.

Based on #1079.

**Checklist**
- [ ] PR has been tested (CI only)

Closes #1153

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,18 +27,18 @@ classifiers = [
     "Programming Language :: Python :: 3.11",
 ]
 dependencies = [
-    "ansicolors==1.1.8",
-    "attrs==21.4.0",
-    "autobahn==21.3.1",
-    "jinja2==3.0.2",
-    "pexpect==4.8.0",
+    "ansicolors>=1.1.8",
+    "attrs>=21.4.0",
+    "autobahn>=21.3.1",
+    "jinja2>=3.0.2",
+    "pexpect>=4.8.0",
     "pyserial-labgrid>=3.4.0.1",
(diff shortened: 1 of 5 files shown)


jluebbe (approved):

Bastian-Krause:
I'll rebase and undraft this once #1152 is merged.

pyproject.toml
```diff
@@ -27,18 +27,18 @@ classifiers = [
...
+    "jinja2>=3.0.2",
+    "pexpect>=4.8.0",
     "pyserial-labgrid>=3.4.0.1",
-    "pytest==7.2.2",
-    "pyudev==0.22.0",
-    "pyusb==1.2.1",
-    "PyYAML==5.4.1",
-    "requests==2.26.0",
-    "xmodem==0.4.6",
+    "pytest>=7.2.2",
```
Emantor, line 36: 7.2.2 is unnecessary strict, the commit message that introduced this mentions `>=7.0.0`, we should fix that.

Bastian-Krause: Done.

Emantor (approved):

PR #1138, opened 2023-04-19
labgrid: util: marker: use safe marker

jremmet:
The marker can randomly contain some key words. As they show up in the serial logs they may trigger false positive hits while searching.

Remove the character R, I and D to avoid substrings like ERROR, FAIL, WARN, INFO or DEBUG

### DIFF
--- a/labgrid/util/marker.py
+++ b/labgrid/util/marker.py
@@ -2,5 +2,8 @@ import random
 import string
 
 
+# Remove RID to avoid markers containing substrings like ERROR, FAIL, WARN, INFO or DEBUG
+MARKER_POOL = tuple(c for c in string.ascii_uppercase if c not in 'RID')
+
 def gen_marker():
-    return ''.join(random.choice(string.ascii_uppercase) for i in range(10))
+    return ''.join(random.choice(MARKER_POOL) for i in range(10))


labgrid/util/marker.py
```diff
@@ -2,5 +2,14 @@
...
+BLOCKLIST = ("ERR", "FAIL", "WARN", "INFO", "DBG", "DEBUG")
+
+def _gen_marker():
     return ''.join(random.choice(string.ascii_uppercase) for i in range(10))
+
+def gen_marker():
+    """Generates marker not containing BLOCKLIST strings."""
+    while True:
+        marker = _gen_marker()
+        if not any(b in marker for b in BLOCKLIST):
```
jluebbe, line 14: Is is necessary to do substring matching here? Why would a `XYZINFOABC` marker be a problem?

jremmet: Yes if we `grep INFO console_main_uut_console`

jremmet: If we want to avoid the delay we could also  remove "RID" from the random pool

jluebbe: Ah, making a simple grep work sounds reasonable.

I'd prefer removing "RID" from the `choice()` parameter. Just make a local fixed list in the module and add a comment why these characters are missing.

labgrid/util/marker.py
```diff
@@ -2,5 +2,8 @@
 import string
 
 
+# Remove RID to avoid markers containing substrings like ERROR, FAIL, WARN, INFO or DEBUG
+MARKER_POOL = ''.join(c for c in string.ascii_uppercase if c not in 'RID')
```
Bastian-Krause, line 6:
```suggestion
MARKER_POOL = tuple(c for c in string.ascii_uppercase if c not in 'RID')
```

jremmet: Yes it can be any sequence for `choice()`

Bastian-Krause (approved):

Emantor (approved):

PR #953, opened 2022-07-15
driver: ykushpower: support remote control via ssh.

chen-png:
enable driver for resource YKUSHPowerPort,
change to use ykushcmd as control driver,
support remote control via ssh.

Signed-off-by: Chen Peng1 <peng1.chen@intel.com>

**Description**

support to control YKUSHPowerPort device remotely via ssh.

PR #1143, opened 2023-04-21
Support remote controlling Ykush devices

edersondisouza:
Currently, ykush devices can only be controlled 'locally'. This PR adds support to control them on remote devices. It is based on the seemingly abandoned https://github.com/labgrid-project/labgrid/pull/953 - I basically updated it with the comments there plus a minor fix - first patch, that could go on a different PR if deemed best.

Note that I also removed the 'Export' class created because I'm still figuring out its need - and the commentary for it. If it's deemed necessary, I can bring it back.

**Checklist**

- [X] PR has been tested

### DIFF
--- a/labgrid/driver/powerdriver.py
+++ b/labgrid/driver/powerdriver.py
@@ -271,27 +271,37 @@ class DigitalOutputPowerDriver(Driver, PowerResetMixin, PowerProtocol):
 @attr.s(eq=False)
 class YKUSHPowerDriver(Driver, PowerResetMixin, PowerProtocol):
     """YKUSHPowerDriver - Driver using a YEPKIT YKUSH switchable USB hub
-        to control a target's power - https://www.yepkit.com/products/ykush"""
-    bindings = {"port": "YKUSHPowerPort", }
+        to control a target's power - https://www.yepkit.com/products/ykush.
+        Uses ykushcmd tool to control the hub."""
+    bindings = {"port": {"YKUSHPowerPort", "NetworkYKUSHPowerPort"} }
     delay = attr.ib(default=2.0, validator=attr.validators.instance_of(float))
-
 
     def __attrs_post_init__(self):
         super().__attrs_post_init__()
-        # uses the YKUSH pykush interface from here:
-        # https://github.com/Yepkit/pykush
-        self.pykush_mod = import_module('pykush.pykush')
-        self.pykush = self.pykush_mod.YKUSH(serial=self.port.serial)
+        if self.target.env:
+            self.tool = self.target.env.config.get_tool('ykushcmd') or 'ykushcmd'
+        else:
+            self.tool = 'ykushcmd'
 
     @Driver.check_active
     @step()
     def on(self):
-        self.pykush.set_port_state(self.port.index, self.pykush_mod.YKUSH_PORT_STATE_UP)
+        cmd = [
+            self.tool,
+            "-s", f"{self.port.serial}",
+            "-u", f"{self.port.index}"
+        ]
+        processwrapper.check_output(self.port.command_prefix + cmd)
 
     @Driver.check_active
     @step()
     def off(self):
-        self.pykush.set_port_state(self.port.index, self.pykush_mod.YKUSH_PORT_STATE_DOWN)
+        cmd = [
+            self.tool,
+            "-s", f"{self.port.serial}",
+            "-d", f"{self.port.index}"
+        ]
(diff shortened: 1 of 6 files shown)


jluebbe:
> Note that I also removed the 'Export' class created because I'm still figuring out its need - and the commentary for it. If it's deemed necessary, I can bring it back.

The ResourceExport classes are needed to add the host and proxy information which makes local resources accessible via the network. That way, the exporter config can contain a `YKUSHPowerPort` entry, which is then converted to a `NetworkYKUSHPowerPort` by the exporter (which has the correct host and proxy information).

edersondisouza:
v2:
- Re-add ResourceExport class for YKUSHPowerPower.

edersondisouza:
Any comments about v2, @jluebbe?

labgrid/driver/powerdriver.py
```diff
@@ -271,27 +271,37 @@ def get(self):
 @attr.s(eq=False)
 class YKUSHPowerDriver(Driver, PowerResetMixin, PowerProtocol):
     """YKUSHPowerDriver - Driver using a YEPKIT YKUSH switchable USB hub
-        to control a target's power - https://www.yepkit.com/products/ykush"""
+        to control a target's power - https://www.yepkit.com/products/ykush.
+        Uses ykushcmd tool to control the hub."""
     bindings = {"port": "YKUSHPowerPort", }
```
jluebbe, line 276: This needs to explicitly bind to `NetworkYKUSHPowerPort` as well.
```suggestion
    bindings = {"port": {"YKUSHPowerPort", "NetworkYKUSHPowerPort"} }
```
See for example the `FlashromDriver`.

labgrid/resource/ykushpowerport.py
```diff
@@ -15,3 +15,8 @@ class YKUSHPowerPort(Resource):
     serial = attr.ib(validator=attr.validators.instance_of(str))
     index = attr.ib(validator=attr.validators.instance_of(int),
                     converter=int)
+
+@target_factory.reg_resource
+@attr.s(eq=False)
+class NetworkYKUSHPowerPort(NetworkResource, YKUSHPowerPort):
```
jluebbe, line 21: We don't use multiple inheritance for resources, so die users need explicitly choose which to support.

labgrid/remote/client.py
```diff
@@ -744,6 +746,8 @@ def power(self):
                     drv = self._get_driver_or_new(target, "PDUDaemonDriver", name=name)
                 elif isinstance(resource, TasmotaPowerPort):
                     drv = self._get_driver_or_new(target, "TasmotaPowerDriver", name=name)
+                elif isinstance(resource, YKUSHPowerPort):
```
jluebbe, line 749:
```suggestion
                elif isinstance(resource, NetworkYKUSHPowerPort):
```

labgrid/remote/client.py
```diff
@@ -727,13 +727,15 @@ def power(self):
         target = self._get_target(place)
         from ..resource.power import NetworkPowerPort, PDUDaemonPort
         from ..resource.remote import NetworkUSBPowerPort, NetworkSiSPMPowerPort
-        from ..resource import TasmotaPowerPort
+        from ..resource import TasmotaPowerPort, YKUSHPowerPort
```
jluebbe, line 730:
```suggestion
        from ..resource import TasmotaPowerPort, NetworkYKUSHPowerPort
```
Also add the import the the resource module.

jluebbe (changes requested):
Please also add a hint to the changelog that it's now necessary to install the ykushcmd tool.

In `doc/configuration.rst`, there's still a recommendation to use `pykush -l` for the serial. Is there an alternative with ykushcmd?

labgrid/resource/ykushpowerport.py
edersondisouza: Ack.

edersondisouza:
v3:
  - No more inheritance from `YKUSHPowerPort` in `NetworkYKUSHPowerPort`;
  - Using `NetworkYKUSHPowerPort` directly on `client.py`;
  - `YKUSHPowerDriver` explicitly binds to `NetworkYKUSHPowerPort`;
  - Updated `configuration.rst` example to use `ykushcmd`;
  - Updated commit message to inform about installing `ykushcmd`.

edersondisouza:
Any more coments, @jluebbe?

edersondisouza:
Gentle reminder about this PR, @jluebbe =D

PR #1153, opened 2023-04-26
pyproject.toml: Accept newer versions of PyYAML

edersondisouza:
Instead of requiring exactly PyYAML 5.4.1, use that version as the minimum requirement. This will help avoid conflicts with other packages that usually require a newer version of PyYAML.

**Description**

Requiring an specific version of PyYAML is causing conflicts with other project requirements, that are requiring PyYAML 6.0 (released on October 2021) as minimum. Running the tests, I didn't see any issue, and https://github.com/labgrid-project/labgrid/pull/1019 also increases the confidence that there are no issues.

**Checklist**
- [X] Tests for the feature 
- [X] PR has been tested

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,7 @@ dependencies = [
     "pytest==7.2.2",
     "pyudev==0.22.0",
     "pyusb==1.2.1",
-    "PyYAML==5.4.1",
+    "PyYAML>=5.4.1",
     "requests==2.26.0",
     "xmodem==0.4.6",
 ]


Emantor:
See #1136 which unpins all dependencies. This is finally possible since we removed crossbar from the optional installs, which affords us the freedom to have unpinned dependencies again (which is correct for a library)

Emantor:
I just merged unpinned dependencies and tagged a first alpha release which contains the change.
You can test this by using `pip install --upgrade --pre labgrid`, which will install the latest pre-release version.

edersondisouza:
> See #1136 which unpins all dependencies. This is finally possible since we removed crossbar from the optional installs, which affords us the freedom to have unpinned dependencies again (which is correct for a library)

Oh, that's great! Sorry missed that...

Issue #1031, opened 2022-11-03
Deprecate remote power backends supported by pdudaemon


Bastian-Krause:
[pdudaemon](https://github.com/pdudaemon/pdudaemon) already supports a wide range of PDUs. labgrid supports interaction with pduaemon ([PDUDaemonPort](https://labgrid.readthedocs.io/en/latest/configuration.html#pdudaemonport), [PDUDaemonDriver](https://labgrid.readthedocs.io/en/latest/configuration.html#pdudaemondriver)).

We don't want to duplicate functionality. Issues like Gude's IP blocking (#586) will not happen because all requests to a PDU come from the same daemon.

- [ ] Add support for synchronous pdudaemon behavior to PDUDaemonDriver, see https://github.com/pdudaemon/pdudaemon/pull/78
- [ ] Document that we won't merge any new remote power backends
- [ ] Deprecate remote power backends supported by pdudaemon

See also #109, #161, #464, #595, #606, #635, #638

PR #1154, opened 2023-04-27
PDUDaemonDriver improvements

Bastian-Krause:
**Description**
Now that [1] is merged, we can drop the sleeps that waited for pdudaemon to execute the request. Note that [1] is not part of a release, yet.

pdudaemon does not implement port state retrieval. PDUDaemonDriver.get() returns None. This leads to cases where None is interpreted as "off", i.e. with constructs relying on `bool(drv.get())`. `labgrid-client power get` is one of those cases.

[1] https://github.com/pdudaemon/pdudaemon/pull/78

**Checklist**
- [x] PR has been tested

### DIFF
--- a/labgrid/driver/powerdriver.py
+++ b/labgrid/driver/powerdriver.py
@@ -411,4 +411,4 @@ class PDUDaemonDriver(Driver, PowerResetMixin, PowerProtocol):
 
     @Driver.check_active
     def get(self):
-        return None
+        raise NotImplementedError("pdudaemon does not support retrieving the port's state")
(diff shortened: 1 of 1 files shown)


obbardc (changes requested):
It'd be great to link (in code or commi message) to an upstream pdudaemon issue about the `get()` not implemented issue

Bastian-Krause:
> It'd be great to link (in code or commi message) to an upstream pdudaemon issue about the `get()` not implemented issue

I'd like to add a way to pdudaemon to retrieve the port's status. Probably by remembering the latest state set in the common driver, which could be overwritten in each driver individually by an actual status retrieval from the PDU. I'll add a link here when I get to it.

Bastian-Krause:
I've updated #1031 with the latest info and discussion state about network power drivers and pdudaemon.

PR #1156, opened 2023-04-27
driver/power: add raritan pdu network power driver

LukaszMrugala:
Based on the Eaton PDU NetworkPowerDriver implementation, uses PySNMP.

**Description**

This PR adds `raritan` as a possible PDU NetworkPowerPort model by adding an appropriate NetworkPowerDriver using SNMP via PySNMP.

The feature has been tested on our Raritan PDU. Changing the power state for an outlet via labgrid has worked properly with this change.

Documentation sources for Raritan PDU, i.a. its SNMP protocol OIDs:
  * http://support.raritan.com/px3/version-3.0.3/user-guides/PX3-0B-v3.0-E.pdf 
  * https://cdn1.raritan.com/download/pdu-g2/4.0.20/MIB_Usage_4.0.20_49038.pdf
  * https://www.circitor.fr/Mibs/Html/P/PDU2-MIB.php#switchingOperation

**Checklist**
- [x] Documentation for the feature - I feel that the `configuration.rst` change is sufficient

- [ ] Tests for the feature - May benefit from being included in the test_powerdriver.test_import_backend_eaton(), as it uses the same backend (PySNMP).

- [x] PR has been tested

### DIFF
--- a/labgrid/driver/power/raritan.py
+++ b/labgrid/driver/power/raritan.py
@@ -1,0 +1,38 @@
+"""Driver for the Raritan PDUs
+
+Documentation sources:
+* https://www.circitor.fr/Mibs/Html/P/PDU2-MIB.php
+* http://support.raritan.com/px3/version-3.0.3/user-guides/PX3-0B-v3.0-E.pdf
+* https://cdn1.raritan.com/download/pdu-g2/4.0.20/MIB_Usage_4.0.20_49038.pdf
+"""
+from ..exception import ExecutionError
+from ...util.snmp import SimpleSNMP
+
+
+OID = ".1.3.6.1.4.1.13742.6.4.1.2.1"
+NUMBER_OF_OUTLETS = 16
+
+
+def power_set(host, port, index, value):
+    assert 1 <= int(index) <= NUMBER_OF_OUTLETS
+
+    _snmp = SimpleSNMP(host, 'private', port=port)
+    outlet_control_oid = "{}.2.1.{}".format(OID, index)
+
+    _snmp.set(outlet_control_oid, str(int(value)))
+
+
+def power_get(host, port, index):
+    assert 1 <= int(index) <= NUMBER_OF_OUTLETS
+
+    _snmp = SimpleSNMP(host, 'public', port=port)
+    output_status_oid = "{}.3.1.{}".format(OID, index)
+
+    value = _snmp.get(output_status_oid)
+
+    if value == 7:  # On
+        return True
+    if value == 8:  # Off
+        return False
+
+    raise ExecutionError("failed to get SNMP value")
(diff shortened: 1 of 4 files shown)


labgrid/driver/power/raritan.py
```diff
@@ -0,0 +1,60 @@
...
+        if error_indication or error_status:
+            raise ExecutionError("Failed to get SNMP value.")
+        return res[0][1]
+
+    def set(self, oid, value):
+        identify = hlapi.ObjectType(hlapi.ObjectIdentity(oid),
+                   hlapi.Integer(value))
+        g = hlapi.setCmd(self.engine, self.community, self.transport,
+            self.context, identify, lookupMib=False)
+        next(g)
```
jluebbe, line 35: This duplicates the code in `eaton.py`. Please move the class as `SimpleSNMP` to a newly created `labgrid/util/snmp.py` first. Then a second commit would use it from a the new raritan backend.

jluebbe (changes requested):
Please also add a docstring to the backend module with the information where you found the necessary documentation.

Looks good otherwise.

Bastian-Krause:
Maybe this is a good opportunity to move from [pysnmp](https://pypi.org/project/pysnmp/) to [pysnmp-lextudio](https://pypi.org/project/pysnmp-lextudio/). The original author of pysnmp passed away and the lextudio folks took over maintenance: https://github.com/etingof/pysnmp/issues/429.

It should be sufficient to replace `pysnmp` with `pysnmp-lextudio` in our pyproject.toml. While we're at it, we can drop `pysnmp-mibs` [altogether](https://github.com/lextudio/pysnmp-mibs/blob/master/README.md). If you'd be willing to do that, you could test this PR against pysnmp-lextudio right away. That would be great.

LukaszMrugala:
> Maybe this is a good opportunity to move from [pysnmp](https://pypi.org/project/pysnmp/) to [pysnmp-lextudio](https://pypi.org/project/pysnmp-lextudio/). The original author of pysnmp passed away and the lextudio folks took over maintenance: [etingof/pysnmp#429](https://github.com/etingof/pysnmp/issues/429).
> 
> It should be sufficient to replace `pysnmp` with `pysnmp-lextudio` in our pyproject.toml. While we're at it, we can drop `pysnmp-mibs` [altogether](https://github.com/lextudio/pysnmp-mibs/blob/master/README.md). If you'd be willing to do that, you could test this PR against pysnmp-lextudio right away. That would be great.

Sorry for uploading changes without first answering you, I've noticed your comment only after having already committed them. I'll incorporate your propositions right away, although I can't really test the results right now.

jluebbe:
> Sorry for uploading changes without first answering you, I've noticed your comment only after having already committed them. I'll incorporate your propositions right away, although I can't really test the results right now.

If you can't test them, we should make that a separate PR and get this one in as-is.

PR #1157, opened 2023-04-28
Add RawNetworkInterfaceDriver, labgrid-raw-interface helper and example

Bastian-Krause:
**Description**
The RawNetworkInterfaceDriver allows "raw" control of a network interface (such as Ethernet or WiFi).

The newly added labgrid-raw-interface helper (``helpers/labgrid-raw-interface``) needs to be installed in the PATH and usable via sudo without password.

It supports:
- recording traffic
- replaying traffic
- basic statistic collection

**Checklist**
- [ ] Tests for the feature 
- [x] PR has been tested

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -2967,6 +2967,48 @@ Implements:
 Arguments:
   - None
 
+RawNetworkInterfaceDriver
+~~~~~~~~~~~~~~~~~~~~~~~~~
+This driver allows "raw" control of a network interface (such as Ethernet or
+WiFi).
+
+The labgrid-raw-interface helper (``helpers/labgrid-raw-interface``) needs to
+be installed in the PATH and usable via sudo without password.
+A configuration file ``/etc/labgrid/helpers.yaml`` must be installed on hosts
+exporting network interfaces for the RawNetworkInterfaceDriver, e.g.:
+
+.. code-block:: yaml
+
+   raw-interface:
+     denied-interfaces:
+       - eth1
+
+It supports:
+- recording traffic
+- replaying traffic
+- basic statistic collection
+
+For now, the RawNetworkInterfaceDriver leaves pre-configuration of the exported
+network interface to the user, including:
+- disabling DHCP
+- disabling IPv6 Duplicate Address Detection (DAD) by SLAAC (Stateless
+Address Autoconfiguration) and Neighbor Discovery
+- disabling Generic Receive Offload (GRO)
+
+This might change in the future.
+
+Binds to:
+  iface:
+    - `NetworkInterface`_
+    - `USBNetworkInterface`_
+    - `RemoteNetworkInterface`_
+
+Implements:
+  - None yet
+
+Arguments:
+  - None
+
 .. _conf-strategies:
 
 Strategies
--- a/examples/network-test/pkg-replay-record.py
+++ b/examples/network-test/pkg-replay-record.py
@@ -1,0 +1,55 @@
+#!/usr/bin/env python3
+# Generates an Ethernet frame via scapy using pcap, copies pcap to DUT, replays pcap on interface,
+# records frame locally (or on exporter, adjust env.yaml accordingly), and compares both.
+
+import logging
+import os
+from tempfile import NamedTemporaryFile, TemporaryDirectory
+
+from scapy.all import Ether, Raw, rdpcap, wrpcap, conf
+
+from labgrid import Environment
+from labgrid.logging import basicConfig, StepLogger
+
+def generate_frame():
+    frame = [Ether(dst="11:22:33:44:55:66", src="66:55:44:33:22:11", type=0x9000)]
+    padding = "\x00" * (conf.min_pkt_size - len(frame))
+    frame = frame[0] / Raw(load=padding)
+    return frame
+
+
+basicConfig(level=logging.INFO)
+StepLogger.start()
+env = Environment("env.yaml")
+target = env.get_target()
+
+netdrv = target.get_driver("RawNetworkInterfaceDriver")
+ssh = target.get_driver("SSHDriver")
+
+# get DUT interface
+exporter_iface = netdrv.iface.ifname
+dut_iface = env.config.get_target_option(target.name, "local_iface_to_dut_iface")[exporter_iface]
+
+# generate test frame
+generated_frame = generate_frame()
+
+# write pcap, copy to DUT
+remote_pcap = "/tmp/pcap"
+with NamedTemporaryFile() as pcap:
+    wrpcap(pcap.name, generated_frame)
+    ssh.put(pcap.name, remote_pcap)
+
+# copy recorded pcap from DUT, compare with generated frame
+with TemporaryDirectory() as tempdir:
+    # start record on exporter
+    tempf = os.path.join(tempdir, "record.pcap")
+    with netdrv.record(tempf, count=1) as record:
+        # replay pcap on DUT
+        ssh.run_check(f"ip link set {dut_iface} up")
+        ssh.run_check(f"tcpreplay -i {dut_iface} {remote_pcap}")
+
+    remote_frame = rdpcap(tempf)
+    assert remote_frame[0] == generated_frame[0]
+
+print("statistics", netdrv.get_statistics())
+print("address", netdrv.get_address())
--- a/helpers/labgrid-raw-interface
+++ b/helpers/labgrid-raw-interface
@@ -1,0 +1,96 @@
+#!/usr/bin/env python3
+#
+# Wrapper script to be deployed on machines whose network interfaces should be
+# controllable via the RawNetworkInterfaceDriver. A /etc/labgrid/helpers.yaml
+# can deny access to network interfaces. See below.
+#
+# This is intended to be used via sudo. For example, add via visudo:
+# %developers ALL = NOPASSWD: /usr/sbin/labgrid-raw-interface
+
+import argparse
+import os
+import sys
+
+import yaml
+
+
+def get_denylist():
+    denylist_file = "/etc/labgrid/helpers.yaml"
+    try:
+        with open(denylist_file) as stream:
+            data = yaml.load(stream, Loader=yaml.SafeLoader)
+    except (PermissionError, FileNotFoundError, AttributeError) as e:
+        raise Exception(f"No configuration file ({denylist_file}), inaccessable or invalid yaml") from e
+
+    denylist = data.get("raw-interface", {}).get("denied-interfaces", [])
+
+    if not isinstance(denylist, list):
+        raise Exception("No explicit denied-interfaces or not a list, please check your configuration")
+
+    denylist.append("lo")
+
+    return denylist
+
+
+def main(program, ifname, count):
+    if not ifname:
+        raise ValueError("Empty interface name.")
+    if any((c == "/" or c.isspace()) for c in ifname):
+        raise ValueError(f"Interface name '{ifname}' contains invalid characters.")
+    if len(ifname) > 16:
+        raise ValueError(f"Interface name '{ifname}' is too long.")
+
+    denylist = get_denylist()
+
+    if ifname in denylist:
+        raise ValueError(f"Interface name '{ifname}' is denied in denylist.")
+
+    programs = ["tcpreplay", "tcpdump"]
+    if program not in programs:
+        raise ValueError(f"Invalid program {program} called with wrapper, valid programs are: {programs}")
+
+    args = [
+        program,
+    ]
+
+    if program == "tcpreplay":
+        args.append(f"--intf1={ifname}")
+        args.append('-')
+
+    if program == "tcpdump":
+        args.append("-n")
+        args.append(f"--interface={ifname}")
+        args.append("-w")
+        args.append('-')
+
+        if count:
+            args.append("-c")
+            args.append(str(count))
+
+    try:
+        os.execvp(args[0], args)
+    except FileNotFoundError as e:
+        raise RuntimeError(f"Missing {program} binary") from e
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        '-d',
+        '--debug',
+        action='store_true',
+        default=False,
+        help="enable debug mode"
+    )
+    parser.add_argument('program', type=str, help='program to run, either tcpreplay or tcpdump')
+    parser.add_argument('interface', type=str, help='interface name')
+    parser.add_argument('count', nargs="?", type=int, default=None, help='amount of frames to capture while recording')
+    args = parser.parse_args()
+    try:
+        main(args.program, args.interface, args.count)
+    except Exception as e:  # pylint: disable=broad-except
+        if args.debug:
+            import traceback
+            traceback.print_exc(file=sys.stderr)
+        print(f"ERROR: {e}", file=sys.stderr)
+        exit(1)
--- a/labgrid/driver/rawnetworkinterfacedriver.py
+++ b/labgrid/driver/rawnetworkinterfacedriver.py
@@ -1,0 +1,187 @@
+# pylint: disable=no-member
+import contextlib
+import json
+import subprocess
+
+import attr
+
+from .common import Driver
+from ..factory import target_factory
+from ..step import step
+from ..util.helper import processwrapper
+from ..util.managedfile import ManagedFile
+from ..resource.common import NetworkResource
+
+
+@target_factory.reg_driver
+@attr.s(eq=False)
+class RawNetworkInterfaceDriver(Driver):
+    bindings = {
+        "iface": {"NetworkInterface", "RemoteNetworkInterface", "USBNetworkInterface"},
+    }
+
+    def __attrs_post_init__(self):
+        super().__attrs_post_init__()
+        self._record_handle = None
+        self._replay_handle = None
+
+    def _wrap_command(self, args):
+        wrapper = ["sudo", "labgrid-raw-interface"]
+
+        if self.iface.command_prefix:
+            # add ssh prefix, convert command passed via ssh (including wrapper) to single argument
+            return self.iface.command_prefix + [" ".join(wrapper + args)]
+        else:
+            # keep wrapper and args as-is
+            return wrapper + args
+
+    def _stop(self, proc, *, timeout=None):
+        assert proc is not None
+
+        try:
+            _, err = proc.communicate(timeout=timeout)
+        except subprocess.TimeoutExpired:
+            proc.terminate()
+            _, err = proc.communicate()
+            raise
+
+        if proc.returncode:
+            raise subprocess.CalledProcessError(
+                returncode=proc.returncode,
+                cmd=proc.args,
+                stderr=err,
+            )
+
+    @Driver.check_active
+    @step(args=["filename", "count"])
+    def start_record(self, filename, *, count=None):
+        """
+        Starts tcpdump on bound network interface resource.
+
+        Args:
+            filename (str): name of a file to record to
+            count (int): optional, exit after receiving this many number of packets
+        Returns:
+            Popen object of tcpdump process
+        """
+        assert self._record_handle is None
+
+        cmd = ["tcpdump", self.iface.ifname]
+        if count is not None:
+            cmd.append(str(count))
+        cmd = self._wrap_command(cmd)
+        with open(filename, "wb") as outdata:
+            self._record_handle = subprocess.Popen(cmd, stdout=outdata, stderr=subprocess.PIPE)
+        return self._record_handle
+
+    @Driver.check_active
+    @step(args=["timeout"])
+    def stop_record(self, *, timeout=None):
+        """
+        Stops previously started tcpdump on bound network interface resource.
+
+        Args:
+            timeout (int): optional, maximum number of seconds to wait for the tcpdump process to
+                           terminate
+        """
+        try:
+            self._stop(self._record_handle, timeout=timeout)
+        finally:
+            self._record_handle = None
+
+    @contextlib.contextmanager
+    def record(self, filename, *, count=None, timeout=None):
+        """
+        Context manager to start/stop tcpdump on bound network interface resource.
+
+        Either count or timeout must be specified.
+
+        Args:
+            filename (str): name of a file to record to
+            count (int): optional, exit after receiving this many number of packets
+            timeout (int): optional, maximum number of seconds to wait for the tcpdump process to
+                           terminate
+        """
+        assert count or timeout
+
+        try:
+            yield self.start_record(filename, count=count)
+        finally:
+            self.stop_record(timeout=timeout)
+
+    @Driver.check_active
+    @step(args=["filename"])
+    def start_replay(self, filename):
+        """
+        Starts tcpreplay on bound network interface resource.
+
+        Args:
+            filename (str): name of a file to replay from
+        Returns:
+            Popen object of tcpreplay process
+        """
+        assert self._replay_handle is None
+
+        if isinstance(self.iface, NetworkResource):
+            mf = ManagedFile(filename, self.iface)
+            mf.sync_to_resource()
+            cmd = self._wrap_command([f"tcpreplay {self.iface.ifname} < {mf.get_remote_path()}"])
+            self._replay_handle = subprocess.Popen(cmd, stderr=subprocess.PIPE)
+        else:
+            cmd = self._wrap_command(["tcpreplay", self.iface.ifname])
+            with open(filename, "rb") as indata:
+                self._replay_handle = subprocess.Popen(cmd, stdin=indata)
+
+        return self._replay_handle
+
+    @Driver.check_active
+    @step(args=["timeout"])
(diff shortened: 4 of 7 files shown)


Emantor (dismissed):

helpers/labgrid-raw-interface
```diff
@@ -0,0 +1,94 @@
...
+
+def main(program, ifname, capfile, count):
+    if not ifname:
+        raise ValueError("Empty interface name.")
+    if any((c == "/" or c.isspace()) for c in ifname):
+        raise ValueError(f"Interface name '{ifname}' contains invalid characters.")
+    if len(ifname) > 16:
+        raise ValueError(f"Interface name '{ifname}' is too long.")
+
+    config = configparser.ConfigParser()
```
jluebbe, line 25: I'm not convinced using two different config file formats is a good idea. Can't we use YAML here as well?

labgrid/driver/rawnetworkinterfacedriver.py
```diff
@@ -0,0 +1,97 @@
...
+            out, err = proc.communicate()
+
+        assert proc.returncode == 0, f"tcpdump out={out} err={err}"
+
+        if isinstance(self.iface, RemoteNetworkInterface):
+            # TODO: remove remote capture afterwards
+            sshmanager.get_file(self.iface.host, self.proc_pcap_map[proc], filename)
+        else:
+            os.rename(self.proc_pcap_map[proc], filename)
+            os.sync()
```
jluebbe, line 68: We don't need a full system `sync()` here. If the user needs to persist the file, they should do it in their own code.

labgrid/driver/rawnetworkinterfacedriver.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def __attrs_post_init__(self):
+        super().__attrs_post_init__()
+        self.proc_pcap_map = {}
+
+    def _get_wrapper_prefix(self):
+        return self.iface.command_prefix + ["sudo", "labgrid-raw-interface"]
+
+    def _remote_tmp_file_name(self):
+        cmd = self.iface.command_prefix + ["mktemp", "--dry-run"]
```
jluebbe, line 33: We should use `mktemp -d` here to avoid potential races and security issues. In the created directory, only the user has write access.

labgrid/driver/rawnetworkinterfacedriver.py
```diff
@@ -0,0 +1,97 @@
...
+        try:
+            out, err = proc.communicate(timeout=timeout)
+        except subprocess.TimeoutExpired:
+            proc.send_signal(signal.SIGINT)
+            out, err = proc.communicate()
+
+        assert proc.returncode == 0, f"tcpdump out={out} err={err}"
+
+        if isinstance(self.iface, RemoteNetworkInterface):
+            # TODO: remove remote capture afterwards
```
jluebbe, line 64: The TODO is still open.

jluebbe (changes requested):

labgrid/driver/rawnetworkinterfacedriver.py
Bastian-Krause: Ok, agreed.

labgrid/driver/rawnetworkinterfacedriver.py
Bastian-Krause: The SSHManager should track files that are created on the exporter. Once they have been downloaded, they should be removed from the exporter. labgrid's ManagedFile handles files that exist locally and get copied to
the exporter. This is the roughly the reverse use case.
Since recording a pcap on the exporter and transferring it to the local machine meets this use case, this should be implemented as "RemoteFile" in labgrid and used in the RawNetworkInterfaceDriver.

This should be a separate PR, though.

jluebbe: OK.

helpers/labgrid-raw-interface
jluebbe: Discussed this with @Emantor: We can use a /etc/labgrid/helpers.yaml with a structure like this:
``` yaml
raw-interface:
  denied-interfaces:
    - eth0
```

labgrid/driver/rawnetworkinterfacedriver.py
```diff
@@ -0,0 +1,96 @@
...
+    def __attrs_post_init__(self):
+        super().__attrs_post_init__()
+        self.proc_pcap_map = {}
+
+    def _get_wrapper_prefix(self):
+        return self.iface.command_prefix + ["sudo", "labgrid-raw-interface"]
+
+    def _remote_tmp_file_name(self):
+        cmd = self.iface.command_prefix + ["mktemp", "-d"]
+        return processwrapper.check_output(cmd).strip().decode("utf-8") + "remote.pcap"
```
jluebbe, line 34: On my system, `mktemp -d` does not produce a trailing slash, so we need to use `os.path.join()` instead of string concatenation.

helpers/labgrid-raw-interface
```diff
@@ -0,0 +1,103 @@
...
+    return denylist
+
+
+def main(program, ifname, capfile, count):
+    if not ifname:
+        raise ValueError("Empty interface name.")
+    if any((c == "/" or c.isspace()) for c in ifname):
+        raise ValueError(f"Interface name '{ifname}' contains invalid characters.")
+    if len(ifname) > 16:
+        raise ValueError(f"Interface name '{ifname}' is too long.")
```
jluebbe, line 42: Do we need to check if any parameters (ifname, capfile, count) start with `-`? That might confuse tcpdump/tcpreplay...

jluebbe (changes requested):
Looks good otherwise.

jluebbe:
We discussed this internally: The wrapper can be simplified by using stdin/stdout to pass packet data, as we don't need to create/open paths as root.

Bastian-Krause:
Rebased on master.

Bastian-Krause:
- labgrid-raw-interface: use yaml instead of ini configs
- change argument order
- use shell redirection and stdin/stdout instead of using tcpdump's `-Z` and letting tcpump/tcpreplay write to capfile directly

Bastian-Krause:
- labgrid-raw-interface: fix optional count argument
- networkinterfacedriver: simplify capfile copy/close handling in `get_record()`
- networkinterfacedriver: fix `start_replay()` for remote network interfaces

Bastian-Krause:
- added documentation for `/etc/labgrid/helpers.yaml`
- example: use `basicConfig`/`StepLogger` instead of `StepReporter`
- labgrid-raw-interface: fix deny list defaults, always deny loopback interface
- rawnetworkinterfacedriver: fix `get_statistics()`

Bastian-Krause:
- rawnetworinterfacedriver: stdout is now redirected into a file in `get_record()`, so drop the obsolete `out` variable

helpers/labgrid-raw-interface
Bastian-Krause: Done.

labgrid/driver/rawnetworkinterfacedriver.py
Bastian-Krause: This is not needed anymore, we use stdout and capture into a NamedTemporaryFile instead.

labgrid/driver/rawnetworkinterfacedriver.py
Bastian-Krause: Same as above, no files are created on the exporter anymore with the new approach.

labgrid/driver/rawnetworkinterfacedriver.py
Bastian-Krause: Code was dropped with the change to capturing to stdout, see above.

helpers/labgrid-raw-interface
Bastian-Krause: This should not be an issue anymore, since we use longopts now.

Bastian-Krause:
- labgrid-raw-interface: use longopts for tcpreplay

labgrid/driver/rawnetworkinterfacedriver.py
```diff
@@ -0,0 +1,90 @@
...
+        mf.sync_to_resource()
+
+        if isinstance(self.iface, NetworkResource):
+            cmd = self.iface.command_prefix
+            cmd.append(f"sudo labgrid-raw-interface tcpreplay {self.iface.ifname} < {mf.get_remote_path()}")
+            indata = None
+        else:
+            cmd = self._get_wrapper_prefix() + ["tcpreplay", self.iface.ifname]
+            indata = open(mf.get_remote_path(), "rb").read()
+        return subprocess.run(cmd, input=indata)
```
Bastian-Krause, line 74: This waits for the completion of the subprocess. When I see the method name `start_replay()`, I expect receiving a subprocess handle instead of a CompletedProcess object. I think this is wrong.

Emantor: The interesting problem here is that with Popen, we can't simply pass in input data as with subprocess run. We probably need to `Popen` with `stdin=PIPE`, pipe in the data and than close `stdin` on the other side.

Bastian-Krause: I've integrated the suggested change, please have a look.

Bastian-Krause:
- rawnetworkinterfacedriver: make `start_replay()` return process handle instead of waiting for the process to complete and returning a CompletedProcess

Bastian-Krause:
Rebased one more time now that #1249 was merged.

Emantor (dismissed):

labgrid/driver/rawnetworkinterfacedriver.py
jluebbe: Can't we just pass the open FD as stdin to the subprocess and then close it on our side?

helpers/labgrid-raw-interface
```diff
@@ -0,0 +1,97 @@
...
+    except (PermissionError, FileNotFoundError, AttributeError):
+        warnings.warn("No configuration file (/etc/labgrid/helpers.yaml), inaccessable or invalid yaml, continuing without denylist.")
+        return []
+
+    denylist = data.get("raw-interface", {}).get("denied-interfaces", [])
+
+    if not isinstance(denylist, list):
+        warnings.warn("No explicit denied-interfaces or not a list, please check your configuration.")
+
+    denylist.append("lo")
```
jluebbe, line 31: If `denylist` is not a list, `.append("lo")` will fail. We should just abort here.

labgrid/driver/rawnetworkinterfacedriver.py
```diff
@@ -0,0 +1,94 @@
...
+@target_factory.reg_driver
+@attr.s(eq=False)
+class RawNetworkInterfaceDriver(Driver):
+    bindings = {
+        "iface": {"NetworkInterface", "RemoteNetworkInterface", "USBNetworkInterface"},
+    }
+
+    def __attrs_post_init__(self):
+        super().__attrs_post_init__()
+        self.proc_pcap_map = {}
```
jluebbe, line 27: This should not be needed as we have one Driver per Interface.

Instead remember one process per record and capture for consistency checks on start/stop.

labgrid/driver/rawnetworkinterfacedriver.py
```diff
@@ -0,0 +1,94 @@
...
+
+    @Driver.check_active
+    @step(args=["filename"])
+    def start_replay(self, filename):
+        mf = ManagedFile(filename, self.iface)
+        mf.sync_to_resource()
+
+        if isinstance(self.iface, NetworkResource):
+            cmd = self.iface.command_prefix
+            cmd.append(f"sudo labgrid-raw-interface tcpreplay {self.iface.ifname} < {mf.get_remote_path()}")
```
jluebbe, line 69: This should also use `self._get_wrapper_prefix()`.

labgrid/driver/rawnetworkinterfacedriver.py
```diff
@@ -0,0 +1,94 @@
...
+        capfile = self.proc_pcap_map[proc]
+        shutil.copy(capfile.name, filename)
+        capfile.close()
+
+    @Driver.check_active
+    @step(args=["filename"])
+    def start_replay(self, filename):
+        mf = ManagedFile(filename, self.iface)
+        mf.sync_to_resource()
+
```
jluebbe, line 66: Move this into the remove case.

labgrid/driver/rawnetworkinterfacedriver.py
```diff
@@ -0,0 +1,94 @@
...
+        if isinstance(self.iface, NetworkResource):
+            cmd = self.iface.command_prefix
+            cmd.append(f"sudo labgrid-raw-interface tcpreplay {self.iface.ifname} < {mf.get_remote_path()}")
+            indata = None
+        else:
+            cmd = self._get_wrapper_prefix() + ["tcpreplay", self.iface.ifname]
+            indata = open(mf.get_remote_path(), "rb").read()
```
jluebbe, line 73: Just pass an open file handle for `filename` to Popen as stdin.

labgrid/driver/rawnetworkinterfacedriver.py
```diff
@@ -0,0 +1,94 @@
...
+
+        proc = subprocess.Popen(cmd, stdin=subprocess.PIPE)
+        proc.stdin.write(indata)
+        proc.stdin.close()
```
jluebbe, line 77: Move this into both cases, as no `stdin` handling will be needed and the `Popen` args will be different (stdin only in the local case).

labgrid/driver/rawnetworkinterfacedriver.py
```diff
@@ -0,0 +1,94 @@
...
+    def __attrs_post_init__(self):
+        super().__attrs_post_init__()
+        self.proc_pcap_map = {}
+
+    def _get_wrapper_prefix(self):
+        return self.iface.command_prefix + ["sudo", "labgrid-raw-interface"]
+
+    @Driver.check_active
+    @step(args=["count"])
+    def record(self, *, count=None):
```
jluebbe, line 34: Rename to `start_record` for consistency.

labgrid/driver/rawnetworkinterfacedriver.py
```diff
@@ -0,0 +1,94 @@
...
+        capture = tempfile.NamedTemporaryFile(prefix="lg-raw-int-cap-")
+        cmd = self._get_wrapper_prefix() + ["tcpdump", self.iface.ifname]
+        if count is not None:
+            cmd.append(str(count))
+
+        proc = subprocess.Popen(cmd, stdout=capture, stderr=subprocess.PIPE)
```
jluebbe, line 40: Open the target filename (which should be an argument to this function) directly for use as stdout, avoid both the temporary file and copy operation.

labgrid/driver/rawnetworkinterfacedriver.py
```diff
@@ -0,0 +1,94 @@
...
+            _, err = proc.communicate(timeout=timeout)
+        except subprocess.TimeoutExpired:
+            proc.send_signal(signal.SIGINT)
+            _, err = proc.communicate()
+
+        assert proc.returncode == 0, f"tcpdump err={err}"
+
+        capfile = self.proc_pcap_map[proc]
+        shutil.copy(capfile.name, filename)
+        capfile.close()
```
jluebbe, line 59: This will not be needed any more.

labgrid/driver/rawnetworkinterfacedriver.py
```diff
@@ -0,0 +1,94 @@
...
+    def get_record(self, proc, filename, *, timeout=None):
+        assert proc in self.proc_pcap_map
+
+        try:
+            _, err = proc.communicate(timeout=timeout)
+        except subprocess.TimeoutExpired:
+            proc.send_signal(signal.SIGINT)
+            _, err = proc.communicate()
+
+        assert proc.returncode == 0, f"tcpdump err={err}"
```
jluebbe, line 55: Reuse this as an internal helper for `stop_record`/`stop_replay`. Those would be convenience methods for easier use in test cases.

examples/network-test/pkg-replay-record.py
```diff
@@ -0,0 +1,53 @@
...
+
+# write pcap, copy to DUT
+remote_pcap = "/tmp/pcap"
+with NamedTemporaryFile() as pcap:
+    wrpcap(pcap.name, generated_frame)
+    ssh.put(pcap.name, remote_pcap)
+
+# start record on exporter
+ethers = filter(lambda frame: isinstance(frame, Ether), generated_frame)
+cap_proc = netdrv.record(count=len(list(ethers)))
```
jluebbe, line 42: Why iterate over the layers in the `generated` frame to find the capture count. Isn't this fixed as `1`?

examples/network-test/pkg-replay-record.py
```diff
@@ -0,0 +1,53 @@
...
+
+# replay pcap on DUT
+ssh.run_check(f"ip link set {dut_iface} up")
+ssh.run_check(f"tcpreplay -i {dut_iface} {remote_pcap}")
+
+# copy recorded pcap from DUT, compare with generated frame
```
jluebbe, line 48: This won't copy any file. Instead stop the recording and read the locally written pcap.

jluebbe (changes requested):

helpers/labgrid-raw-interface
Bastian-Krause: Done.

labgrid/driver/rawnetworkinterfacedriver.py
Bastian-Krause: Done.

labgrid/driver/rawnetworkinterfacedriver.py
Bastian-Krause: Done.

labgrid/driver/rawnetworkinterfacedriver.py
Bastian-Krause: btw, s/remove/remote/

examples/network-test/pkg-replay-record.py
Bastian-Krause: Right, fixed.

examples/network-test/pkg-replay-record.py
Bastian-Krause: Yep, comment is outdated, fixed.

helpers/labgrid-raw-interface
```diff
@@ -0,0 +1,97 @@
...
+
+import yaml
+
+
+def get_denylist():
+    try:
+        with open("/etc/labgrid/helpers.yaml") as stream:
+            data = yaml.load(stream, Loader=yaml.SafeLoader)
+    except (PermissionError, FileNotFoundError, AttributeError):
+        warnings.warn("No configuration file (/etc/labgrid/helpers.yaml), inaccessable or invalid yaml, continuing without denylist.")
```
jluebbe, line 23: As we are much stricter below, perhaps it's better to fail here as well, similar to line 29.

doc/configuration.rst
```diff
@@ -2948,6 +2948,48 @@ It supports:
...
+  - None
+
+RawNetworkInterfaceDriver
+~~~~~~~~~~~~~~~~~~~~~~~~~
+This driver allows "raw" control of a network interface (such as Ethernet or
+WiFi).
+
+The labgrid-raw-interface helper (``helpers/labgrid-raw-interface``) needs to
+be installed in the PATH and usable via sudo without password.
+A configuration file ``/etc/labgrid/helpers.yaml`` should be installed on hosts
```
jluebbe, line 2970: `s/should/must/` for the stricter check in the helper.

labgrid/driver/rawnetworkinterfacedriver.py
```diff
@@ -0,0 +1,130 @@
...
+            # keep wrapper and args as-is
+            return wrapper + args
+
+    def _stop(self, proc, *, timeout=None):
+        assert proc is not None
+
+        try:
+            _, err = proc.communicate(timeout=timeout)
+        except subprocess.TimeoutExpired:
+            proc.send_signal(signal.SIGINT)
```
jluebbe, line 45: Why do you use `.send_signal(SIGINT)` instead of `.terminate()`? Does tcpdump/tcpreplay have specific behavior for SIGINT?

labgrid/driver/rawnetworkinterfacedriver.py
```diff
@@ -0,0 +1,130 @@
...
+    def start_record(self, filename, *, count=None):
+        assert self._record_handle is None
+
+        cmd = ["tcpdump", self.iface.ifname]
+        if count is not None:
+            cmd.append(str(count))
+        cmd = self._wrap_command(cmd)
+        outdata = open(filename, "wb")
+        self._record_handle = subprocess.Popen(cmd, stdout=outdata, stderr=subprocess.PIPE)
+        return self._record_handle
```
jluebbe, line 67: Perhaps use a contextmanager for `outdata`, so we explicitly close it in the parent process?

labgrid/driver/rawnetworkinterfacedriver.py
```diff
@@ -0,0 +1,130 @@
...
+
+        if isinstance(self.iface, NetworkResource):
+            mf = ManagedFile(filename, self.iface)
+            mf.sync_to_resource()
+            cmd = self._wrap_command([f"tcpreplay {self.iface.ifname} < {mf.get_remote_path()}"])
+            self._replay_handle = subprocess.Popen(cmd, stderr=subprocess.PIPE)
+        else:
+            cmd = self._wrap_command(["tcpreplay", self.iface.ifname])
+            indata = open(filename, "rb")
+            self._replay_handle = subprocess.Popen(cmd, stdin=indata)
```
jluebbe, line 97: Use a contextmanager here as well.

labgrid/driver/rawnetworkinterfacedriver.py
```diff
@@ -0,0 +1,130 @@
...
+    @Driver.check_active
+    @step(args=["timeout"])
+    def stop_record(self, *, timeout=None):
+        try:
+            self._stop(self._record_handle, timeout=timeout)
+        finally:
+            self._record_handle = None
+
+    @contextlib.contextmanager
+    def record(self, filename, *, count=None, timeout=None):
```
jluebbe, line 78: Do we need to ensure that one of `count` or `timeout` is set? Otherwise tcpdump won't ever stop...

Perhaps some docstrings for these functions would be useful. I think, in most cases a timeout should be specified, as it avoids the possibility of just having a stuck test.

jluebbe (changes requested):

Bastian-Krause:
Rebased.

helpers/labgrid-raw-interface
Bastian-Krause: Yes, done.

doc/configuration.rst
Bastian-Krause: Done.

labgrid/driver/rawnetworkinterfacedriver.py
Bastian-Krause: tcpdump does. From tcpdump(1):
> Tcpdump will, if not run with the -c flag, continue capturing packets until it is interrupted by a SIGINT signal (generated, for example, by typing your interrupt character, typically control-C) or a SIGTERM signal (typically generated with the kill(1) command); if run with the -c flag, it will capture packets until it is interrupted by a SIGINT or SIGTERM signal or the specified number of packets have been processed.

labgrid/driver/rawnetworkinterfacedriver.py
Bastian-Krause: Done.

labgrid/driver/rawnetworkinterfacedriver.py
Bastian-Krause: Done.

labgrid/driver/rawnetworkinterfacedriver.py
Bastian-Krause: > Do we need to ensure that one of `count` or `timeout` is set? Otherwise tcpdump won't ever stop...

I've added an assert to check this.

> Perhaps some docstrings for these functions would be useful. I think, in most cases a timeout should be specified, as it avoids the possibility of just having a stuck test.

Docstrings added.

Bastian-Krause:
Addressed the review feedback by @jluebbe. Thanks!

labgrid/driver/rawnetworkinterfacedriver.py
jluebbe: `.terminate()` sends SIGTERM, so should have the same result. I'd prefer using the standard way of `.terminate()`.

jluebbe (approved):
Besides https://github.com/labgrid-project/labgrid/pull/1157#discussion_r1448936147, this looks good. As that doesn't affect behavior, you could merge this as-is and address it in a separate PR.

labgrid/driver/rawnetworkinterfacedriver.py
Bastian-Krause: Done.

jluebbe (approved):

PR #1166, opened 2023-05-04
 config: return passed in tool by default 

Emantor:
**Description**
The tools in the tools key can either be a single string with the binary
name or a complete path to a binary. This works for every tool expect
for qemu since drivers can provide a fallback to the default tool.
However qemu currently requires the user to supply the full binary path,
which is unintuitive.

Instead return the originally requested tool-name instead if no
specialized path is found. And return the simple unexpanded path if a
local tool with the same name can't be found.

**Checklist**
- [x] PR has been tested

### DIFF
--- a/labgrid/config.py
+++ b/labgrid/config.py
@@ -98,9 +98,14 @@ class Config:
         """
         try:
             path = str(self.data['tools'][tool])
-            return self.resolve_path(path)
         except KeyError:
-            return None
+            return tool
+
+        resolved = self.resolve_path(path)
+        if os.path.exists(resolved):
+            return resolved
+
+        return path
 
     def get_image_path(self, kind):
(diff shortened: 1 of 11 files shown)


Emantor:
I'll rework this according to local feedback.

jluebbe:
This would fail when the tool has a different path on the exporter.

jluebbe:
Should we also drop all the instances of `or 'tool'` in the drivers?

Emantor:
> Should we also drop all the instances of `or 'tool'` in the drivers?

Yeah, we should do that. I'll push another commit.

jluebbe (changes requested):
This looks good, we just need a test for the tool handling.

jluebbe (approved):

PR #1176, opened 2023-05-09
Unique resource name per resource class

### LINKED ISSUE #1173: Resource names must be unique per resource class, not globally
This would allow multiple resources named "default" (which was intended by #1112).

Pointed out by @nlabriet in https://github.com/labgrid-project/labgrid/issues/1163#issuecomment-1534341129:
> I changed my configuration to remove names from the exporter and set them in the coordinator with `add-named-match`. It works, except when I need to have 2 (or more) resources with the same name ("default").
> 
> [`get_target_resources()`](https://github.com/labgrid-project/labgrid/blob/e1585887fde9111e6672a4df4c616fa5f4b54963/labgrid/remote/client.py#L628) creates a dictionary with the name as the key. In my case, the name should be "default", but the resource classes are different.
> 
> I tested using the (name, class) tuple as key and it solves my issue. Does this seem reasonable to you?

nlabriet:
**Description**
Allow using the same resource name across classes.
I tested it in my setup and created a test that fails without this fix.

**Checklist**
  - No documentation on this fix
- [X] Tests for the feature 
- [X] PR has been tested

Fixes #1173

### DIFF
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -630,17 +630,17 @@ class ClientSession(ApplicationSession):
         resources = {}
         for resource_path in place.acquired_resources:
             match = place.getmatch(resource_path)
-            (exporter, group_name, _, resource_name) = resource_path
+            (exporter, group_name, cls, resource_name) = resource_path
             name = resource_name
             if match.rename:
                 name = match.rename
-            resources[name] = self.resources[exporter][group_name][resource_name]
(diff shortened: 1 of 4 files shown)


jluebbe (approved):
Discussed this with @Emantor: While `get_target_resources()` returns a dict, we're never using it for lookups. So we should change this to a list or set of `(resource_entry, name)` tuples in the future.

PR #1181, opened 2023-05-16
driver/usbloader: enable print_on_silent_log for subprocess calls

jluebbe:
The diagnostic output of these tools is often useful for development and debugging.


jluebbe:
@rohieb Could you try this?

Emantor (approved):

PR #1182, opened 2023-05-17
drivers, resource: implement get_bound_resources()

Emantor:
**Description**
Implement a new get_bound_resources() function which returns the underlying resources a driver is bound to. This can also be called on Resources, but they will just return themselves in a set.

Add a new keyword argument to get_driver/get_active_driver(), which allows the user to retrieve a driver using a specific resource it's bound to.

**Checklist**
- [x] Tests for the feature 
- [x] PR has been tested

### DIFF
--- a/labgrid/driver/common.py
+++ b/labgrid/driver/common.py
@@ -67,6 +67,16 @@ class Driver(BindingMixin):
         """
         return False
 
+    def get_bound_resources(self):
+        """Return the bound resources for a driver
+
+        This recursively calls all suppliers and combines the sets of returned resources.
+        """
+        res = set()
+        for supplier in self.suppliers:
+            res |= supplier.get_bound_resources()
+        return res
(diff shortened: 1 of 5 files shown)


jluebbe (approved):

labgrid/driver/common.py
```diff
@@ -67,6 +67,16 @@ def skip_deactivate_on_export(self):
         """
         return False
 
+    def get_bound_resources(self):
+        """Return the bound resources for a driver
+
+        This recursively calls all suppliers to and combines the sets of returned resources.
```
Bastian-Krause, line 73: Something's wrong with this sentence.

Bastian-Krause (commented):
Looks good to me, only a minor comment on a docstring.

Bastian-Krause (approved):

PR #1188, opened 2023-05-26
ubootstrategy: add support for initial state

nlabriet:
**Description**

Update the uboot strategy to feature initial state, usable from pytest and labgrid-client.
Unnecessary actions to reach the requested state are not performed.

Tested with labgrid-client in my setup, transitioning:
- from off to uboot with initial state: off
- from uboot to shell with initial state: uboot

**Checklist**
Documentation already present
- [X] PR has been tested
Manual tests in my setup
No man page impacted, this can serve as an example to implement a strategy with initial state support

### DIFF
--- a/labgrid/strategy/ubootstrategy.py
+++ b/labgrid/strategy/ubootstrategy.py
@@ -57,3 +57,16 @@ class UBootStrategy(Strategy):
         else:
             raise StrategyError(f"no transition found from {self.status} to {status}")
         self.status = status
+
+    def force(self, status):
+        if not isinstance(status, Status):
+            status = Status[status]
+        if status == Status.off:
+            self.target.activate(self.power)
+        elif status == Status.uboot:
+            self.target.activate(self.uboot)
+        elif status == Status.shell:
+            self.target.activate(self.shell)
+        else:
+            raise StrategyError("can not force state {}".format(status))
+        self.status = status


Bastian-Krause (changes requested):
The driver activation should happen inside the `force()` method, not in `transition()`. See `labgrid/strategy/bareboxstrategy.py` or `labgrid/strategy/shellstrategy.py`.

nlabriet:
Right, I've changed the force method to include the driver activations.

labgrid/strategy/ubootstrategy.py
```diff
@@ -41,19 +41,35 @@ def transition(self, status):
             self.target.activate(self.power)
             self.power.off()
         elif status == Status.uboot:
-            self.transition(Status.off)
+            if self.status != Status.off:
+                self.transition(Status.off)
```
Bastian-Krause, line 45: This is not needed. The transition method does nothing if the requested status is the same as the saved `self.status`, see above ("nothing to do"). Think of `transition()` as a state machine.

labgrid/strategy/ubootstrategy.py
```diff
@@ -41,19 +41,35 @@ def transition(self, status):
...
             # cycle power
             self.power.cycle()
             # interrupt uboot
             self.target.activate(self.uboot)
         elif status == Status.shell:
-            # transition to uboot
-            self.transition(Status.uboot)
+            if self.status != Status.uboot:
+                # transition to uboot
+                self.transition(Status.uboot)
```
Bastian-Krause, line 54: Same as above.

Bastian-Krause (changes requested):

labgrid/strategy/ubootstrategy.py
nlabriet: right it's called recursively

labgrid/strategy/ubootstrategy.py
nlabriet: ok, reverting to original

labgrid/strategy/ubootstrategy.py
```diff
@@ -57,3 +57,17 @@ def transition(self, status):
...
             raise StrategyError(f"no transition found from {self.status} to {status}")
         self.status = status
+
+    def force(self, status):
+        if not isinstance(status, Status):
+            status = Status[status]
+        if status == Status.off:
+            self.target.activate(self.power)
+        elif status == Status.uboot:
+            self.target.activate(self.console)
```
Bastian-Krause, line 67: You can drop this line. The activation of the UBootDriver (which binds against the ConsoleProtocol) in the next line activates it anyway. Activating the ConsoleProtocol explicitly is only needed if we need to to it early, i.e. before turning power on, so we don't miss early characters on the serial interface.

Bastian-Krause (commented):
Sorry, this one slipped through earlier.

labgrid/strategy/ubootstrategy.py
nlabriet: Line removed. I tested the transition form initial uboot state to shell state, it works!

Bastian-Krause (approved):

PR #1191, opened 2023-05-30
bootloader: detect boot

jremmet:
The boot_expression feature for u-boot was deprecated to sync barebox with features and to allow to connect to a running bootloader.

Readd the feature for both bootloaders and use the check only to set the `boot_detected` flag.
So after getting a prompt we can check if the bootloader runs trough a reboot.
Show this in the 'reset' method.

**Checklist**
- [X] Tests for the feature 

- [X] PR has been tested on uboot
- [ ] PR has been tested on barebox

### DIFF
--- a/labgrid/driver/bareboxdriver.py
+++ b/labgrid/driver/bareboxdriver.py
@@ -9,6 +9,7 @@ from ..step import step
 from ..util import gen_marker, Timeout, re_vt100
 from .common import Driver
 from .commandmixin import CommandMixin
+from .exception import ExecutionError
 
 
 @target_factory.reg_driver
@@ -25,6 +26,7 @@ class BareboxDriver(CommandMixin, Driver, CommandProtocol, LinuxBootProtocol):
         prompt (str): barebox prompt to match
         autoboot (regex): optional, autoboot message to match
         interrupt (str): optional, string to interrupt autoboot (use "\x03" for CTRL-C)
+        boot_expression (regex): optional, string to search for on barebox start
         bootstring (regex): optional, regex indicating that the Linux Kernel is booting
         password (str): optional, password to use for access to the shell
         login_timeout (int): optional, timeout for access to the shell
@@ -33,9 +35,11 @@ class BareboxDriver(CommandMixin, Driver, CommandProtocol, LinuxBootProtocol):
     prompt = attr.ib(default="", validator=attr.validators.instance_of(str))
     autoboot = attr.ib(default="stop autoboot", validator=attr.validators.instance_of(str))
     interrupt = attr.ib(default="\n", validator=attr.validators.instance_of(str))
+    boot_expression = attr.ib(default=r"[\n]barebox 20\d+", validator=attr.validators.instance_of(str))
     bootstring = attr.ib(default=r"Linux version \d", validator=attr.validators.instance_of(str))
     password = attr.ib(default="", validator=attr.validators.instance_of(str))
     login_timeout = attr.ib(default=60, validator=attr.validators.instance_of(int))
+    boot_detected = attr.ib(default=False, validator=attr.validators.instance_of(bool))
 
     def __attrs_post_init__(self):
         super().__attrs_post_init__()
@@ -108,6 +112,8 @@ class BareboxDriver(CommandMixin, Driver, CommandProtocol, LinuxBootProtocol):
         self._status = 0
         self.console.sendline("reset")
         self._await_prompt()
+        if not self.boot_detected:
+            raise ExecutionError(f"No reboot message detected: {self.boot_expression}")
 
     def get_status(self):
         """Retrieve status of the BareboxDriver
@@ -148,8 +154,9 @@ class BareboxDriver(CommandMixin, Driver, CommandProtocol, LinuxBootProtocol):
         # occours, we can't lose any data this way.
         last_before = None
         password_entered = False
+        self.boot_detected = False
 
-        expectations = [self.prompt, self.autoboot, "Password: ", TIMEOUT]
+        expectations = [self.prompt, self.autoboot, "Password: ", self.boot_expression, TIMEOUT]
         while True:
             index, before, _, _ = self.console.expect(
                 expectations,
@@ -177,6 +184,10 @@ class BareboxDriver(CommandMixin, Driver, CommandProtocol, LinuxBootProtocol):
                 password_entered = True
 
             elif index == 3:
+                # we detect a boot
+                self.boot_detected = True
+
+            elif index == 4:
                 # expect hit a timeout while waiting for a match
                 if before == last_before:
                     # we did not receive anything during the previous expect cycle
--- a/labgrid/driver/ubootdriver.py
+++ b/labgrid/driver/ubootdriver.py
@@ -8,6 +8,7 @@ from ..util import gen_marker, Timeout, re_vt100
 from ..step import step
 from .common import Driver
 from .commandmixin import CommandMixin
+from .exception import ExecutionError
 
 
 @target_factory.reg_driver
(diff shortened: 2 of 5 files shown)


jremmet:
@Bastian-Krause We want to do some Watchdog tests and rely on the bootloader driver to detect the reboot. We could do this in the test but we like the idea of having this beside the other bootloader configs like the prompt.
A 2nd use case is that we have switchable powersupply which has a manual overwrite for always on. So it happens that we don't power on the board via labgrid. Before the deprecation this runs into a bootloader timeout. Now it starts the tests.

So short having the feature back and only disable the check if we force a state  would be good for out use cases. Not sure If this acceptable for the project. This PR tries to solve it without changing too much. (Without addressing accidentally running boards)

Bastian-Krause:
I'd probably implement the `self.console.expect()` directly in `reset()` and drop `self.check_boot` altogether.

I'm torn between having `boot_expression` as a driver attribute and passing it as a kwarg to `reset()` (or `await_boot()`?): I think `boot_expression` does not make it clear that it's only used on `reset()`. Then there's the previous functionality (before #259), then the deprecation (#965) and now we reintroduce it with a subtly different meaning, that's not ideal.
If we allow passing it as a kwarg to `reset()` or `await_boot()`, we'd need to change the LinuxBootProtocol.

@Emantor What do you think?

Emantor:
The decision to remove this support was to let drivers be deactivated and activated even if the bootloader is already running, which is necessary to support reusing the current state of the board within a labgrid client state transition.

In my opionion testing a watchdog reset belongs to the testcase and using labgrid functionality for this is not the correct approach. What you want to do for watchdog testing anway is to set a global variable inside the bootloader, deactivate the driver, wait for the watchdog reset and reactivate the driver. Than the test can check whether the variable is not there any longer and this is a new running instance of the bootloader and even check the reset reason if this is possible with the board.

I don't have a solution for the case where you have a manual override on the power switch. Can this be queried by software? I think this is more of an organizational problem than something that can be solved with a technical solution inside of labgrid.

jremmet:
> The decision to remove this support was to let drivers be deactivated and activated even if the bootloader is already running, which is necessary to support reusing the current state of the board within a labgrid client state transition.
> 
> In my opionion testing a watchdog reset belongs to the testcase and using labgrid functionality for this is not the correct approach. What you want to do for watchdog testing anway is to set a global variable inside the bootloader, deactivate the driver, wait for the watchdog reset and reactivate the driver. Than the test can check whether the variable is not there any longer and this is a new running instance of the bootloader and even check the reset reason if this is possible with the board.
> 
> I don't have a solution for the case where you have a manual override on the power switch. Can this be queried by software? I think this is more of an organizational problem than something that can be solved with a technical solution inside of labgrid.

Yes I agree. Would adding "boot_expression" to the driver again and adding a check for it to https://github.com/labgrid-project/labgrid/blob/ed4e12e35064c857bf2b3db79d1d952fa7f56a2a/labgrid/driver/bareboxdriver.py#L148-L187 ok?
We can set the "boot_detected" flag there and use it in strategies and tests.

jremmet:
@Emantor I added the flag to driver as discussed.

labgrid/driver/ubootdriver.py
```diff
@@ -137,6 +134,8 @@ def reset(self):
         self._status = 0
         self.console.sendline("reset")
         self._await_prompt()
+        if not self.boot_detected:
+            raise Exception(f"No reboot message detected: {self.boot_expression}")
```
Emantor, line 138: Please don't raise a plain Exception, this should be something along the lines of either `ExecutionError` or `BareboxStateError`.

labgrid/driver/bareboxdriver.py
```diff
@@ -108,6 +111,8 @@ def reset(self):
         self._status = 0
         self.console.sendline("reset")
         self._await_prompt()
+        if not self.boot_detected:
+            raise Exception(f"No reboot message detected: {self.boot_expression}")
```
Emantor, line 115: Please don't raise a plain Exception, this should be something along the lines of either ExecutionError or UBootStateError.

labgrid/driver/bareboxdriver.py
```diff
@@ -33,13 +34,15 @@ class BareboxDriver(CommandMixin, Driver, CommandProtocol, LinuxBootProtocol):
...
     interrupt = attr.ib(default="\n", validator=attr.validators.instance_of(str))
+    boot_expression = attr.ib(default=r"[\n]barebox 20\d+", validator=attr.validators.instance_of(str))
     bootstring = attr.ib(default=r"Linux version \d", validator=attr.validators.instance_of(str))
     password = attr.ib(default="", validator=attr.validators.instance_of(str))
     login_timeout = attr.ib(default=60, validator=attr.validators.instance_of(int))
 
     def __attrs_post_init__(self):
         super().__attrs_post_init__()
         self._status = 0
+        self.boot_detected = False
```
Emantor, line 45: please use an `attr.ib` for this as well.

labgrid/driver/ubootdriver.py
```diff
@@ -47,10 +47,7 @@ class UBootDriver(CommandMixin, Driver, CommandProtocol, LinuxBootProtocol):
     def __attrs_post_init__(self):
         super().__attrs_post_init__()
         self._status = 0
-
-        if self.boot_expression:
-            import warnings
-            warnings.warn("boot_expression is deprecated and will be ignored", DeprecationWarning)
+        self.boot_detected = False
```
Emantor, line 50: please use an `attr.ib` for this as well.

Emantor (changes requested):

labgrid/driver/bareboxdriver.py
jremmet: Ok, sure here because it makes this also an Argument. Added it and a note to configuration.rst

jremmet:
@Emantor gently ping

labgrid/driver/bareboxdriver.py
```diff
@@ -33,9 +35,11 @@ class BareboxDriver(CommandMixin, Driver, CommandProtocol, LinuxBootProtocol):
     prompt = attr.ib(default="", validator=attr.validators.instance_of(str))
     autoboot = attr.ib(default="stop autoboot", validator=attr.validators.instance_of(str))
     interrupt = attr.ib(default="\n", validator=attr.validators.instance_of(str))
+    boot_expression = attr.ib(default=r"[\n]barebox 20\d+", validator=attr.validators.instance_of(str))
     bootstring = attr.ib(default=r"Linux version \d", validator=attr.validators.instance_of(str))
     password = attr.ib(default="", validator=attr.validators.instance_of(str))
     login_timeout = attr.ib(default=60, validator=attr.validators.instance_of(int))
+    boot_detected = attr.ib(default=False, validator=attr.validators.instance_of(int))
```
Emantor, line 42: This should be `instance_of(bool)`. We also probably want to set this to `setattr=None`, so users can't change this attribute.

Emantor (changes requested):

labgrid/driver/bareboxdriver.py
jremmet: I fixed the type.
With attr I also found no way to protect it. Not sure if it worth `_boot_detected` and `get_boot_detected` like done with `_status`

jremmet:
> I'm still not convinced why we need to add such functionality to the BareboxDriver/UBootDriver when we can achieve the same with:
> 
> @jremmet What do you think?

@Bastian-Krause Sorry, the message got lost.
We currently use similar function added to the boot loader driver, so the boot_message can be defined there along with the other information about the boot loader. (calling _await_prompt instead of activate / deactivate the driver)

Is there a chance to get something like this in or is a boot detection out of scope of the driver?

PR #1196, opened 2023-06-05
sync-places: add named match support

nlabriet:
**Description**

Update sync-places to support named matches.

**Checklist**
  - updated the help text
- [X] Tests for the feature 
  - Tested in my setup using dump and sync
    - dump of initial setup
    - sync of named setup
    - dump of named setup
    - sync of initial setup
    - sync of dumped (named) setup
    - sync of initial setup


contrib/sync-places.py
```diff
@@ -139,6 +158,7 @@ async def do_dump(session, args):
                 my-place1: # Replace with your place
                   matches: # A list of match patterns. Replace with your match patterns
                     - "*/my-place1/*"
+                    - ["exporter/my-place1/resource", "name"] # named matches supported
```
JoshuaWatt, line 161: I think it would be more convenient for users if this was done as a dictionary mapping:

```yaml
matches:
  - "*/my-place1/*"
  - name: "exporter/my-place1/resource"
```

contrib/sync-places.py
```diff
@@ -63,34 +63,50 @@ async def do_sync(session, args):
 
         for name in config["places"]:
             matches = config["places"][name].get("matches", [])
+            matches = [tuple(m) if isinstance(m, list) else m for m in matches]
```
JoshuaWatt, line 66: I don't like the mixed types in the `matches` variable as it makes iterating a lot harder. It would be better if all entries in the `matches` array were a tuple of `(match_expression, name)` and `name` can just be `None` if there is no name for the match. This also works nicely because passing `None` to `add_place_match` means "don't rename"

So, with that in mind and the comment below, perhaps something like:

```python
matches = []
for m in config["places"][name].get("matches", []):
    if isinstance(m, dict):
        name = m.keys()[0]
        matches.append((m[name], name))
    else:
        matches.append((m, None))

```

contrib/sync-places.py
nlabriet: @JoshuaWatt thanks for the review.
I like the idea of the dictionary, I just need to use the resources as keys as names are not unique

contrib/sync-places.py
nlabriet: I've updated the PR accordingly.

contrib/sync-places.py
JoshuaWatt: It technically doesn't matter because each dictionary is a dedicated list element, so you should only have a single dictionary key-value pair in each dictionary anyway.

I don't really care either way; which ever method makes more sense or has precedence is fine with me

nlabriet: @Bastian-Krause what do you think about the dictionary order?
I lean a bit more in favor of {resource: name} as it matches the order of `labgrid-client -v places`

JoshuaWatt: OK

JoshuaWatt (approved):

jluebbe (approved):

PR #1201, opened 2023-06-09
doc: do not open if already closed -> open

vermaete:


Bastian-Krause (approved):
Good catch, thanks for taking the time to create this PR.

PR #1210, opened 2023-06-15
Fix handling disabled pytest logging plugin

### LINKED ISSUE #1209: labgrid plugin breaks pytest if logging plugin is disabled
Why is it that the pytest plugin is registered by default, and not only on the "dev environment"?

This can cause issues if someone is running an unrelated pytest session on a different project, and labgrid plugin is loaded by pytest but fails (due for instance, the other project setting `-p no:logging`).

Is there a good reason for that? It seems to me that running pytest is more important for development anyway.

Emantor:
**Description**

Adjust our pytestplugin to handle the case where the logging plugin is disabled.

**Checklist**
- [x] PR has been tested

Fixes #1209

### DIFF
--- a/labgrid/pytestplugin/hooks.py
+++ b/labgrid/pytestplugin/hooks.py
@@ -36,28 +36,31 @@ def pytest_cmdline_main(config):
         set_cli_log_level(logging.INFO)
 
 
+def configure_pytest_logging(config, plugin):
+    plugin.log_cli_handler.formatter.add_color_level(logging.CONSOLE, "blue")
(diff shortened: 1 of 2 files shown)


jluebbe:
@Emantor CI reports `labgrid/pytestplugin/hooks.py:45:14: E0602: Undefined variable 'config' (undefined-variable)`

Emantor:
> @Emantor CI reports `labgrid/pytestplugin/hooks.py:45:14: E0602: Undefined variable 'config' (undefined-variable)`

Fixed.

jluebbe (approved):

PR #1211, opened 2023-06-16
qemudriver: support looking up qemu executable in PATH

a3f:
We currently mandate that a tools key records the full path to the Qemu binary to use. This is unfortunate as not all distros may install e.g. qemu-system-arm into the same location. Resolve this by looking up the qemu_bin key in $PATH if no tools key was found.

**- what do you use the feature for?** Use same YAML across distros
**- how does labgrid benefit as a testing library from the feature?** YAMLs are directly usable on NixOS (which doesn't place Qemu in `/usr/bin`.
**- how did you verify the feature works?** I tested it with tools key, without tools key and with invalid qemu_bin.

**Description**

**Checklist**
- [ ] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -2378,7 +2378,7 @@ Implements:
   - :any:`PowerProtocol`
 
 Arguments:
-  - qemu_bin (str): reference to the tools key for the QEMU binary
+  - qemu_bin (str): reference to the tools key for the QEMU binary or executable name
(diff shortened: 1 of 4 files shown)


Emantor:
This was already fixed in https://github.com/labgrid-project/labgrid/pull/1166. Closing this PR.

PR #1212, opened 2023-06-16
qemudriver: export get_qemu_base_args method

a3f:
For debugging, it can be useful to start an interactive session for the
user to access the DUT console. This is possible with real targets
through labgrid-client -s STATE console, but no equivalent exists yet
for targets using QEMUDriver. Resolving that may be a bigger
undertaking, so for now, let's provide a class method
get_qemu_base_args, that returns the list of arguments sans QMP parts.

Users can then initialize labgrid as usual and call the function to
get the command line and start Qemu for interactive use without having
to duplicate the labgrid environment parsing as in [1].

[1]: https://github.com/barebox/barebox/blob/v2023.05.0/test/emulate.pl#L226

**- what do you use the feature for?** Start Qemu interactively with same environment yaml used for testing
**- how does labgrid benefit as a testing library from the feature?** makes debugging failing tests easier
**- how did you verify the feature works?** I verified tests work and that calling the function starts a usable qemu

**Checklist**
- [ ] Tests for the feature 
- [x] PR has been tested

---

THis is rebased on #1211 to avoid a conflict.

### DIFF
--- a/labgrid/driver/qemudriver.py
+++ b/labgrid/driver/qemudriver.py
@@ -120,12 +120,12 @@ class QEMUDriver(ConsoleExpectMixin, Driver, PowerProtocol, ConsoleProtocol):
 
         return (int(m.group('major')), int(m.group('minor')), int(m.group('micro')))
 
-    def on_activate(self):
-        self._tempdir = tempfile.mkdtemp(prefix="labgrid-qemu-tmp-")
-        sockpath = f"{self._tempdir}/serialrw"
-        self._socket = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
-        self._socket.bind(sockpath)
-        self._socket.listen(0)
+    def get_qemu_base_args(self):
+        """Returns the base command line used for Qemu without the options
+        related to QMP. These options can be used to start an interactive
+        Qemu manually for debugging tests
+        """
+        cmd = []
 
         qemu_bin = self.target.env.config.get_tool(self.qemu_bin)
         if qemu_bin is None:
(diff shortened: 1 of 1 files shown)


Emantor (commented):
It's not clear to me how this is used in the supplied test script? The only caller introduced in the commits is the old location you extracted the function from. Please explain how this works in conjunction with your perl script.

a3f:
barebox test/emulate.pl duplicates the YAML parsing to be able to run Qemu interactively. This PR is preparation for using existing labgrid infrastructure for that. See here for how it's used: https://lore.barebox.org/barebox/20230619095240.4168216-4-a.fatoum@pengutronix.de/T/#m6e6deb363efc855b0ec77d46d8f2ba6c0f67ff21

My first PR attempt just started QEMU directly. I am open to implement it either way.

a3f:
Is this ok? Should I split this PR up, so the non-functional changes can be merged first and I can submti PRs for other Qemu related things?

jluebbe:
@a3f https://github.com/labgrid-project/labgrid/pull/1212/commits/104223cb8a83fc7918dfc68d6dfa7ef4743c3be8 should not be needed with https://github.com/labgrid-project/labgrid/commit/552a42c9f47e325e57e5abf4d368ca17417612ef, or am I missing something?

jluebbe (dismissed):
https://github.com/labgrid-project/labgrid/pull/1212/commits/104223cb8a83fc7918dfc68d6dfa7ef4743c3be8 should not be needed with https://github.com/labgrid-project/labgrid/commit/552a42c9f47e325e57e5abf4d368ca17417612ef, or am I missing something?

Looks OK otherwise. Later, we might want to expose this string via the export functionality.

jluebbe:
> @jluebbe Yes, I had rebased on #1211 to avoid the conflict, but I see now that it's not required.
> 
> > Looks OK otherwise. Later, we might want to expose this string via the export functionality.
> 
> What's that?

See https://github.com/labgrid-project/labgrid/commit/70c66cc28fe4d82e9514ee39dbc3f079256b6b5d. It's built into `labgrid-client` so it won't work for your use-case yet.

Emantor (approved):

Emantor (approved):

PR #1213, opened 2023-06-21
Pr changes to support acquisition during test execution

MischaZihler:
**Description**
We want to be able to acquire remote resource during test executions. This PR accommodate to that in repolling the state of the exporters, when a session already exists. 
Additionally this PR modifies the SerialLoggingReporter.flush() method, so it won't throw an exception, when later added serial connections are being tried to flush, but not yet added to the dictionary.

**Checklist**

- [x] PR has been tested

### DIFF
--- a/labgrid/logging.py
+++ b/labgrid/logging.py
@@ -120,6 +120,8 @@ class SerialLoggingReporter:
             "step": self.lastevent.step,
         }
         for source, logger in self.loggers.items():
+            if source not in self.bufs.keys():
+                continue
             data = self.vt100_replace_cr_nl(self.bufs[source])
             if data:
                 logger.log(logging.CONSOLE, self._create_message(self.lastevent, data), extra=extra)
--- a/labgrid/resource/remote.py
+++ b/labgrid/resource/remote.py
@@ -48,6 +48,8 @@ class RemotePlaceManager(ResourceManager):
                 self.url = config.get_option('crossbar_url', self.url)
                 self.realm = config.get_option('crossbar_realm', self.realm)
             self._start()
+        else:
+            self.poll()
         place = self.session.get_place(remote_place.name)  # pylint: disable=no-member
         resource_entries = self.session.get_target_resources(place)  # pylint: disable=no-member
         expanded = []
--- a/tests/test_crossbar.py
+++ b/tests/test_crossbar.py
@@ -293,7 +293,7 @@ def test_remoteplace_target_without_env(request, place_acquire):
     from labgrid import Target
     from labgrid.resource import RemotePlace
 
-    t = Target(request.node.name)
+    t = Target("test")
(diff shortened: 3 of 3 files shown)


MischaZihler:
@jluebbe would you be able to help out on this issue? locally with real hardware, I can execute the failing unittest `test_remoteplace_target_without_env`.

jluebbe:
Here was the discussion to unterstand the background behind this PR: https://matrix.to/#/!NqKpzIiEICmrNQppEy:matrix.org/$16872534391424UJhAh:libera.chat?via=matrix.org

As @Emantor said, the current design doesn't support acquiring (and especially releasing) places during a single python/pytest run. This was done to reduce the complexity of resource handling and limit the possible concurrent state changes. For example, the `ResourceManager` singletons don't support removal of resources, which is not a problem as long as all `RemotePlace`s stay acquired until the Python process exists. With flow suggested in chat (calling `labgrid-client acquire`/`... release` via subprocess), this is no longer the case.

On release, there would still be a `Target` object for that `Place` in memory, potentially even with active `Driver`s. So, we won't merge an incomplete implementation which could lead to these inconsistencies. 

To support acquiring/releasing places at runtime, we would need at least:
- cleanup of `ManagedResource`s and `ResourceManager`s
- related to that, robust shutdown of `Drivers`
- an API to reserve, acquire and release places based on an environment file (note that this needs to cross from sync to async APIs)
- comprehensive tests of the new cleanup and state transition code

As the state synchronization in labgrid is already complex, I'm very skeptical if we should try to take that path. In addition to the implementation effort, it would also need significant review time on our side, and I currently can't promise that we'd be able to find enough time without a commercial project.

Our strong suggestion would be to use the supported way of handling place acquire/release *outside* of the pytest run. If you have different test cases that need to run on different boards, use separate pytest runs (perhaps with separate test suite or by selecting test cases using pytest marks). The results could still be aggregated after the run.

Of course, you can continue using the workarounds added in this PR, but we won't promise that it will keep working in the future. You'll likely find other places which need additional workarounds due to the state inconsistencies from Place state changes at runtime.

jluebbe:
> Thank you very much for the comprehensive response. It seems like I wasn't clear in regards of when what is acquired and when what is released. The approach I've taken is to acquire my DUTs during the test execution on demand and keep them acquired until the whole run and therefore the pytest process is finished.

You had example workflows in the chat, where you had '- release board-N' and after that some test execution. That's what I was referring to.

> Only after that, all places are released. With that approach, I shouldn't run into your described inconsistencies regarding multiple `Targets`, `Places `and `Drivers `acquired one after another in a test run, should I?

If you never release places during a single run, that avoids some of the issues above.

Using a subprocess call to labgrid-client still seems like workaround.

> I understand that labgrid doesn't want to merge a PR that supports a not supported/intended usage.

Thanks.

PR #1214, opened 2023-06-22
CHANGES: add hint regarding deprecated ticket authentication

jluebbe:


Bastian-Krause:
`WARNING: Ticket authentication is deprecated. Please update your coordinator.` can't happen on current master (leading to v23.1) with an updated coordinator: The coordinator is expected to be installed into its own virtualenv which won't work with ticket authentication, see #1079 for details. As far as I understand, simply installing crossbar into the same venv as labgrid won't work anyway anymore, now that we have merged #1136, so that can't happen either.

This means the above error can only happen on v23.0, because that release does not contain #1079 and #1136.

Bastian-Krause (approved):

PR #1216, opened 2023-06-22
remote/authenticator: drop obsolete ticket authenticator

Bastian-Krause:
**Description**
Now that #1079 and #1136 are merged, the coordinator no longer supports ticket authentication. Drop the corresponding authenticator class and other ticket auth remains.

Exporters and clients might still talk to older coordinator versions, so leave ticket authentication for these components enabled.

**Checklist**
- [ ] PR has been tested


Bastian-Krause:
Actually, this does not change anything from the user perspective: An up to date coordinator with an outdated config tries to import the now non-existent class `labgrid.remote.authenticator.AuthenticatorSession`, but fails early because the pexpect, pyserial and xmodem modules are not found. Only if these were installed into the crossbar venv, the user would actually see..
```
ModuleNotFoundError: No module named 'labgrid.remote.authenticator'
```

Nonetheless, this is a good opportunity to add a hint in the breaking changes section of v23.1.

Bastian-Krause:
Added a note on breaking changes as explained above.

Emantor (approved):

PR #1217, opened 2023-06-22
pytest.stash instead of protected member

jremmet:
Since pytest 7.0 stash is suggested to be used for storing private information from a plugin.
https://docs.pytest.org/en/7.0.x/how-to/writing_hook_functions.html#storing-data-on-items-across-hook-functions

Tested by running tests using env fixture.

**Checklist**
- [ ] Tests for the feature 

- [X] PR has been tested


Emantor (approved):

Bastian-Krause (approved):
Nice improvement!

PR #1223, opened 2023-06-25
driver/shelldriver: fix status attribute setting

Bastian-Krause:
**Description**
Since `_check_prompt()` sets the correct `_status` attribute anyway when a login prompt is detected, this only is a problem for cases without login (either because no login is required or it happened before). `get_status()` returns 0 instead of 1 then until `run()` is executed the first time.

Fix this by setting the correct attribute.

**Checklist**
- [ ] PR has been tested

Fixes 822abb3

### DIFF
--- a/labgrid/driver/shelldriver.py
+++ b/labgrid/driver/shelldriver.py
@@ -143,7 +143,7 @@ class ShellDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
                 else:
                     # we got a prompt. no need for any further action to
                     # activate this driver.
-                    self.status = 1
+                    self._status = 1
                     break
 
             elif index == 1:


Emantor:
@SmithChart This should be testable with the Freifunk CI system, AFAIR those don't have a login getty?

SmithChart:
> @SmithChart This should be testable with the Freifunk CI system, AFAIR those don't have a login getty?

They have a getty. We still have to press Enter to log in:
https://gitli.stratum0.org/ffbs/ffbs-ci/-/blob/master/parker/r6120.yaml#L23
Thus I am not sure if we hit this case. I can try it anyway if it helps.

Emantor (approved):

PR #1224, opened 2023-06-27
httpdigitaloutput: support generic digital outputs via HTTP

hnez:
**Description**

Add support and documentation for a generic digital output driver that works via HTTP.
The driver allows setting an URL to `PUT`/`POST`/`PATCH` a new output status to and an URL to `GET` the current status from.
The message body to send when asserting/deasserting the output can be specified independently.
The status returned by a `GET` request can either be matched exactly against the same body used to set the state or more fuzzily using regular expressions.

- What do _I_ use the feature for?
   The [tacd](https://github.com/linux-automation/tacd) system management daemon on the [LXA TAC](https://www.linux-automation.com/de/products/lxa-tac.html) exposes digital outputs via a HTTP API.
    The `HttpDigitalOutputDriver` enables controling these outputs.
- How does labgrid benefit as a testing library from the feature?
    The `HttpDigitalOutputDriver` is not limited to the LXA TAC. Instead it can be used with many APIs that do not require authentication. Authentication support may also be added once a usecase arises.
- How did you verify the feature works?
    The `HttpDigitalOutputDriver` was tested on actual LXA TAC hardware. Both locally and by proxying the HTTP connection via SSH.
- If hardware is needed for the feature, which hardware is supported and which hardware did you test with?
    The feature was tested on a LXA TAC using the most recent `meta-lxatac` and `tacd` versions respectively.

**Checklist**
- [x] Tests for the feature 
- [x] PR has been tested

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -483,6 +483,39 @@ Arguments:
 Used by:
   - `HIDRelayDriver`_
 
+HttpDigitalOutput
++++++++++++++++++
+A ``HttpDigitalOutput`` resource describes a generic digital output that can be
+controlled via HTTP.
+
+.. code-block:: yaml
+
+   HttpDigitalOutput:
+     url: http://host.example/some/endpoint
+     body_asserted: "On"
+     body_deasserted: "Off"
+
+The example assumes a simple scenario where the same URL is used for PUT
+requests that set the output state and GET requests to get the current state.
+It also assumes that the returned state matches either "On" or "Off" exactly.
+
+The `HttpDigitalOutputDriver`_ also supports more advanced use cases where the
+current state is fetched from another URL and is interpreted using regular
+expressions.
+
+Arguments:
+  - url (str): URL to use for setting a new state
+  - body_asserted (str): Request body to send to assert the output
+  - body_deasserted (str): Request body to send to de-assert the output
+  - method (str, default="PUT"): HTTP method to set a new state
+
+  - url_get (str): URL to use instead of ``url`` for getting the state
+  - body_get_asserted (str): Regular Expression that matches an asserted response body
+  - body_get_deasserted (str): Regular Expression that matches a de-asserted response body
+
+Used by:
+  - `HttpDigitalOutputDriver`_
+
 NetworkHIDRelay
 +++++++++++++++
 A NetworkHIDRelay describes an `HIDRelay`_ exported over the network.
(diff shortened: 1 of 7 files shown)


tests/test_httpdigitalout.py
```diff
@@ -0,0 +1,103 @@
...
+    mock_request.side_effect = request
+    mock_get = mocker.patch("requests.get")
+    mock_get.side_effect = get
+
+    return (mock_request, mock_get)
+
+
+@pytest.fixture(scope="function")
+def make_http_driver(request, target):
+    def _make_http_driver(with_tls, with_regex, separate_get, match_error):
```
jluebbe, line 32: This doesn't need to be a fixture returning a function. Instead just have `_make_http_driver` as a helper function and pass in the target explicitly.

jluebbe (changes requested):
Looks good otherwise.

doc/configuration.rst
```diff
@@ -483,6 +483,39 @@ Arguments:
...
+   HttpDigitalOutput:
+     url: http://host.example/some/endpoint
+     body_asserted: "On"
+     body_deasserted: "Off"
+
+The example assumes a simple scenario where the same URL is used for PUT
+requests that set the output state and GET requests to get the current state.
+It also assumes that the returned state matches either "On" or "Off" exactly.
+
+The `HttpDigitalOutputDriver`_ also supports more advanced usecases where the
```
Bastian-Krause, line 502:
```suggestion
The `HttpDigitalOutputDriver`_ also supports more advanced use cases where the
```

doc/configuration.rst
```diff
@@ -483,6 +483,39 @@ Arguments:
...
+current state is fetched from another URL and is interpreted using regular
+expressions.
+
+Arguments:
+  - url (str): URL to use for setting a new state
+  - body_asserted (str): Request body to send to assert the output
+  - body_deasserted (str): Request body to send to de-assert the output
+  - method (str): HTTP method to use instead of PUT (the default) to set a new state
```
Bastian-Krause, line 510:
```suggestion
  - method (str, default="PUT"): HTTP method to use instead of PUT to set a new state
```
Keeps things consistent with other resources and drivers.

Bastian-Krause (changes requested):

tests/test_httpdigitalout.py
hnez: Sounds good. I've updated the PR just now to do so.

Bastian-Krause (approved):

Emantor (approved):

PR #1225, opened 2023-06-27
client: Allow NetworkService to be manually specified

JoshuaWatt:
If a place doesn't have a NetworkService (commonly when the device does some sort of dynamic IP addressing and tests get the IP address at runtime), the user needs to manually specify the username, hostname, port, and password required to SSH into the device. To facilitate this, add a `--host` and `--password` to the labgrid-client command.

**Description**

**Checklist**
- [ ] Tests for the feature 

- [ ] PR has been tested

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,8 @@
 Release 23.1 (unreleased)
 ------------------------------------
+- ``labgrid-client (ssh|scp|sshfs|rsync|forward)`` now accept a ``--host`` and
+  ``--password`` argument to allow the user to connect if the place has no
+  NetworkService
 
 New Features in 23.1
 ~~~~~~~~~~~~~~~~~~~~
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -1031,13 +1031,47 @@ class ClientSession(ApplicationSession):
             return drv
         except NoDriverFoundError:
             from ..resource import NetworkService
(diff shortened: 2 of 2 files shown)


Emantor:
I am not fundamentally opposed to this, however I do wonder what the advantage over plain `ssh` is. Could you elaborate how your usage scenario looks like?

JoshuaWatt:
> I am not fundamentally opposed to this, however I do wonder what the advantage over plain `ssh` is. Could you elaborate how your usage scenario looks like?

Our devices are not directly accessible to be ssh'd into; you have to go through the exporter (`labgrid-client ssh` would make this quite easy). However, our devices also do automatic IP negotiation, so they don't have a fixed IP address that can be put in a NetworkService to make `labgrid-client ssh` work. When running pytest this isn't a problem because our strategy knows how to look up the IP address for the device and create a correct NetworkService, but when users are manually interacting with devices, this doesn't work. Instead, (assuming they shell in on the console and look up the device IP address), this patch allows them to specify that IP address so that `labgrid-client ssh` is useful.

Emantor:
But isn't plain SSH than easier instead of using labgrid?

JoshuaWatt:
Ah of course. `ssh USER@IP_ADDRESS -J EXPORTER` works, I just didn't think to try that; perhaps suggesting this in the docs would be helpful

PR #1226, opened 2023-07-03
github/workflows: add scheduled test jobs for stable branch

Bastian-Krause:
**Description**
The stable branch is typically close to the latest release (i.e. what the user will see on `pip install labgrid`). Testing this regularly allows us to notice issues early.


Emantor (approved):

Bastian-Krause:
I did not think this through: https://github.com/labgrid-project/labgrid/actions/runs/5462091372

It only makes sense to add stable branches to the scheduled builds once they include the setup introduced in #1079 (i.e. stable-23.1).

PR #1234, opened 2023-07-11
[feature] driver/netio: use netio via json-api

ActionHOSchT:
**Description**

Following the contribution-guidelines an additional functionality using the netio-json-api was added to simplify netio handling. To get all actions working powerprotocol.py:NetworkPowerDriver probably needs a refactor.

Tested manually with a Netio-4-Device and in productive-env used with netio-8.

``` python
import logging
import time

import pytest

FEATURE = "power"


@pytest.fixture()
def pdu(target):
    return target.get_driver('NetworkPowerDriver')


@pytest.mark.lg_feature(FEATURE)
def test_something(pdu):
    pdu.on()
    # Note that the pdu might not have switched on when requesting it's
    # status immediately
    time.sleep(5)
    assert pdu.get()['State'] == 1
    logging.info("could turn on")

    pdu.cycle()
    time.sleep(5)
    assert pdu.get()['State'] == 1
    logging.info("could cycle")

    pdu.off()
    time.sleep(5)
    assert pdu.get()['State'] == 0
    logging.info("could turn off")
```
grafik 

Signed-off-by: Raffael Krakau <r.krakau90@gmail.com>

**Checklist**

- [x] PR has been tested
<!---

### DIFF
--- a/labgrid/driver/power/netio_json.py
+++ b/labgrid/driver/power/netio_json.py
@@ -1,0 +1,165 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+"""
+Add HTTP-JSON-API functions for labgrid-netio power driver.
+Set model in YML-CFG to 'netio_json'.
+"""
+__author__ = "Eugen Wiens, Christian Happ, Raffael Krakau"
+__copyright__ = "Copyright 2023, JUMO GmbH & Co. KG"
+__email__ = "Eugen.Wiens@jumo.net, Christian.Happ@jumo.net"
+__date__ = "2023-07-05"
+__status__ = "Production"
+
+import json
+
+import requests
+
+
+class __NetioControl:
+    """
+    Netio-Class to switch sockets via HTTP-JSON-API
+    """
+    __host: str
+    __port: int
+    __username: str
+    __password: str
+    __telnetConnection = None
+
+    def __init__(self, host, port=80, username=None, password=None):
+        """
+        Define connection details
+        """
+        self.__host = host
+        self.__port = port
+        self.__username = username
+        self.__password = password
+
+    def set_state(self, socketID, action):
+        """
+        Set power state by socket port number (e.g. 1 - 8) and an action ('1'-'4').
+
+        - actions:
+            - 0: Turn OFF,
+            - 1: Turn ON,
+            - 2: Short OFF delay (restart),
+            - 3: Short ON delay,
+            - 4: Toggle (invert the state)
+        """
+        state = None
+        response = requests.post(self.create_request_url(), json=self.get_request_json_data(socketID, action))
+
+        if response.ok:
+            responseDict = json.loads(response.text)
+            outputStates = responseDict['Outputs']
+
+            for outputState in outputStates:
+                if outputState['ID'] == socketID:
+                    state = outputState
+                    break
+        else:
+            raise Exception(f"Cannot SET the power state for socket number: {socketID} for action: {action}."
+                            f" Error code: {response.text}")
+
+        return state
+
+    def get_state(self, socketID):
+        """
+        Get current state of a given socket number as json.
+        """
+        state = None
+        response = requests.get(self.create_request_url())
+
+        if response.ok:
+            responseDict = json.loads(response.text)
+
+            for outputState in responseDict['Outputs']:
+                if outputState['ID'] == socketID:
+                    state = outputState
+                    break
+        else:
+            raise Exception(f"Cannot get the power state for socket {socketID}. Error code: {response.text}")
+
+        return state
+
+    def convert_socket_id(self, socketID) -> int:
+        """
+        cast socketID to int
+
+        raises ValueError
+        """
+        try:
+            socketID = int(socketID)
+        except ValueError as e:
+            raise Exception(f"socketID \"{socketID}\" could not be converted to an integer: {e}!")
+
+        return socketID
+
+    def create_request_url(self) -> str:
+        """
+        Build URL from params set in __init__()
+        """
+        requestUrl = 'http://'
+
+        if self.__username and self.__password:
+            requestUrl += f'{self.__username}:{self.__password}@'
+
+        requestUrl += self.__host
+
+        if self.__port:
+            requestUrl += f':{self.__port}'
+
+        requestUrl += '/netio.json'
+
+        return requestUrl
+
+    def get_request_json_data(self, socketID, action) -> str:
+        data = f'{{"Outputs":[{{"ID":{socketID},"Action":{self.convert_action(action)}}}]}}'
+
+        return json.loads(data)
+
+    def to_str(self, toConvert) -> str:
+        if toConvert == 1 or toConvert is True:
+            toConvert = "1"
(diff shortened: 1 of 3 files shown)


ActionHOSchT:
> labgrid already supports these netio APIs:
> 
> > **netio**
> > Controls a NETIO 4-Port PDU via a simple HTTP API.
> > **netio_kshell**
> > Controls a NETIO 4C PDU via a Telnet interface.
> 
> See: https://labgrid.readthedocs.io/en/latest/configuration.html#networkpowerport
> 
> Why do you need to implement an additional API?


We had issues using kshell with the 8-port version of netio. (even changed devices)

When we contacted netio they told us, that we must use e.g. the json-api to control the 8-port-netios. Which we did and implemented this netio-json-driver for labgrid.

Considering the manufacturer telling us to use json-api, this is a mandatory feature for people who want to use 8-port netios. Even if the current netio-docs say, that kshell should be possible with the netio *PowerPDU 8QS*, we use the json-api to control the netios.

labgrid/driver/power/netio_json.py
```diff
@@ -0,0 +1,157 @@
...
+__status__ = "Production"
+
+import json
+
+import requests
+
+PORT = 80
+
+
+class __NetioControl:
```
Emantor, line 21: Why the double underscore?

labgrid/driver/power/netio_json.py
```diff
@@ -0,0 +1,157 @@
...
+    """
+    Netio-Class to switch sockets via HTTP-JSON-API
+    """
+    __host: str
```
Emantor, line 25: `__host` and the variables below are defined as class variables, however they are not used anywhere? Every place uses the instance variables instead.

labgrid/driver/power/netio_json.py
```diff
@@ -0,0 +1,157 @@
...
+    - port: standard is 80
+    - index: depends on netio-device 1-n (n is the size of netio)
+    - action:
+        - 0: Turn OFF,
+        - 1: Turn ON,
+        - 2: Short OFF delay (restart),
+        - 3: Short ON delay,
+        - 4: Toggle (invert the state)
+    """
+    netio = __NetioControl(host, port)
```
Emantor, line 144: Is the class indirection necessary? How are `username` and `password` set?

Emantor (changes requested):

Emantor:
Are the netio devices supported by pdudaemon? In that case it may be easier to use PDUDaemon.

labgrid/driver/power/netio_json.py
ActionHOSchT: Initially dunderscore was added to keep NetioControl private and to prevent it from showing up in the doc.(imo a user/dev does only need this info if he already takes a look in the source-code)

With dunderscore the class will be mangled and show up as power__NetioControl giving a better idea what this exactly is.

labgrid/driver/power/netio_json.py
ActionHOSchT: `__host` is a private copy of `host` (added to prevent side-effects)
At line 105 this copy will be used.
`__host` will be mangled and show up as `NetioControl__host` which prevents any other operation from accidentally modifying it.

labgrid/driver/power/netio_json.py
ActionHOSchT: The class indirection isn't strictly necessary, sure it would also be possible to implement this classless. 

The json-api supports/expects username and password, even if they're empty.
afaik NetworkPowerDriver doesn't support to set username and password because NetworkPowerPort doesn't support it.
We/I couldn't find a way to set `username` and `password` at `exporter-config.yaml`.
Because of this another PR is on it's way (hopefully PR-ready within 2 weeks).

ActionHOSchT:
> Are the netio devices supported by pdudaemon? In that case it may be easier to use PDUDaemon.

I don't see the benefits from using `PDUDaemon` since we look forward to provide the NetworkPowerDriverWithAuth to finally(no offense) set credentials at export-config.yaml.

If you or someone else sees the benefit from using `PDUDaemon` I would be really thankful to get to know them.

labgrid/driver/power/netio_json.py
```diff
@@ -0,0 +1,157 @@
...
+        """
+        state = None
+        response = requests.get(self.generate_request_url())
+
+        if response.ok:
+            responseDict = json.loads(response.text)
+
+            for outputState in responseDict['Outputs']:
+                if outputState['ID'] == socketID:
+                    state = outputState
```
ep1cman, line 76: This should be
```state = outputState['State']```
Otherwise it always shows as on

ActionHOSchT: thank you

jluebbe:
> > Are the netio devices supported by pdudaemon? In that case it may be easier to use PDUDaemon.
> 
> I don't see the benefits from using `PDUDaemon` since we look forward to provide the NetworkPowerDriverWithAuth to finally(no offense) set credentials at export-config.yaml.

Why do you want distribute the credentials via the coordinator in the clear to all clients? PDUDaemon current has no authentication either, but it's a HTTP API, so it should be relatively easy to switch to authenticated HTTPS using a reverse proxy. At that point, every user could have personal credentials (would need small changes to the `PDUDaemonDriver`).
 
> If you or someone else sees the benefit from using `PDUDaemon` I would be really thankful to get to know them.

With PDUDaemon, we can:

- reuse the driver implementations across projects (labgrid, lava and others)
- avoid concurrent access from clients to the same PDU by serializing them (some have bugs with parallel access)
- "hide" the actual PDUs behind a web service that can be protected via firewalls and/or authentication

labgrid/driver/power/netio_json.py
ActionHOSchT: Can this be resolved?

labgrid/driver/power/netio_json.py
ActionHOSchT: Can this be resolved?

labgrid/driver/power/netio_json.py
ActionHOSchT: Username and password are prefabricated.
If you want this to be removed I will remove *username* and *password*.
If it's ok for you I would keep it for possible "future-use".

Bastian-Krause:
Reopened as requested by @JUMO-GmbH-Co-KG.

labgrid/driver/power/netio_json.py
ChHapp: No, the class is not necessary, we can adjust that. Just as note, we use the NetioControl class in other projects as well, we just added the Labgrid API functions `power_set` and `power_get` to reuse that code and get it working with Labgrid. 

How should we continue on that ?

labgrid/driver/power/netio_json.py
SeminSey: Would a single underscore still be prefered?

labgrid/driver/power/netio_json.py
SeminSey: Any further suggestion on how to handle this preferably?

PR #1238, opened 2023-07-20
fix scheduled unit tests:  don't use minimalmodbus 2.1.0 or 2.1.1

jremmet:
Currently "scheduled unit tests"  and new PRs crashes tests/test_modbusrtudriver.py
minimalmodbus.MasterReportedException: Failed to initialise serial port

minimalmodbus-2.0.1 was working before.

Disallow this versions till further investigation. I don't have a setup for this

**Description**

**Checklist**
- [ ] Tests for the feature 

- [ ] PR has been tested


Emantor:
We should use `, <2.1` instead of `, !=2.1.0, !=2.1.1` here to ensure a later version does not catch us out.

Bastian-Krause:
We should investigate this before excluding versions.

Bastian-Krause:
> > We should investigate this before excluding versions.
> 
> I disagree. We should restrict the version to fix our CI and unrestrict when we have fixed the issue.

Already done in #1241.

PR #1261, opened 2023-08-28
shelldriver: Fix xmodem

JoshuaWatt:
xmodem was failing because it would consume the trailing newline character between the `echo` for the marker and the command output. Fix this by suppressing new line on the marker

**Description**

**Checklist**
- [ ] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/labgrid/driver/shelldriver.py
+++ b/labgrid/driver/shelldriver.py
@@ -311,7 +311,7 @@ class ShellDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
         """
 
         marker = gen_marker()
-        marked_cmd = f"echo '{marker[:4]}''{marker[4:]}'; {cmd}"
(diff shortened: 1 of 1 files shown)


Emantor (approved):

Emantor:
Thanks for the fix!

Bastian-Krause:
This should be back ported to stable.

Bastian-Krause:
> This should be back ported to stable.

Done.

rohieb:
> xmodem was failing because it would consume the trailing newline character between the `echo` for the marker and the command output. Fix this by suppressing new line on the marker

@JoshuaWatt do you still know what you were using on the target? Was it lrzsz or the busybox rx/sx tools, or something else? (I'm wondering because I didn't get it to work recently…)

JoshuaWatt:
@rohieb I'm pretty sure it was lrzsz, but it was a while ago so I'm not positive

PR #1265, opened 2023-09-07
ssh: Prevent timeout from deadlock

JoshuaWatt:
Using Popen.wait() on a process that has output sent to a pipe can potentially deadlock if the process produces enough output to fill the pipe, since it will stall and never terminate waiting for the pipe to have more space.

Instead, use Popen.communicate() with the timeout parameter. This will consume all output until EOF (preventing the process from stalling due to a full pipe), and then check the return code. In the event of a timeout error, Popen.communicate() doesn't loose any data, so it's safe to call it again after the Popen.kill() in the exception handler.

This likely was done this way because the timeout parameter was new in Python 3.3, but this shouldn't be a concern anymore

**Description**

**Checklist**
- [ ] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/labgrid/driver/sshdriver.py
+++ b/labgrid/driver/sshdriver.py
@@ -136,9 +136,8 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
 
         try:
             subprocess_timeout = timeout + 5
-            return_value = self.process.wait(timeout=subprocess_timeout)
-            if return_value != 0:
-                stdout, _ = self.process.communicate(timeout=subprocess_timeout)
+            stdout, _ = self.process.communicate(timeout=subprocess_timeout)
+            if self.process.returncode != 0:
                 stdout = stdout.split(b"\n")
                 for line in stdout:
                     self.logger.warning("ssh: %s", line.rstrip().decode(encoding="utf-8", errors="replace"))
@@ -155,12 +154,14 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
                     pass
 
                 raise ExecutionError(
-                    f"Failed to connect to {self.networkservice.address} with {' '.join(args)}: return code {return_value}",  # pylint: disable=line-too-long
+                    f"Failed to connect to {self.networkservice.address} with {' '.join(args)}: return code {self.process.returncode}",  # pylint: disable=line-too-long
                     stdout=stdout,
                 )
         except subprocess.TimeoutExpired:
+            self.process.kill()
+            stdout, _ = self.process.communicate()
             raise ExecutionError(
-                f"Subprocess timed out [{subprocess_timeout}s] while executing {args}",
+                f"Subprocess timed out [{subprocess_timeout}s] while executing {args}: {stdout}",
             )
         finally:
             if self.networkservice.password and os.path.exists(pass_file):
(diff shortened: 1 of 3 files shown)


Bastian-Krause (approved):
Looks good, thanks!

Emantor (approved):

jluebbe:
@JoshuaWatt Did you actually experience the deadlock?

JoshuaWatt:
We have seen timeouts and the speculation was that it was the deadlock. I don't think we've seen any since this patch merged..... but ya we also encountered the SSH <8.5 bug unfortunately.

JoshuaWatt:
And unfortunately, the default SSH on Ubuntu 20.04 has the problem where it doesn't close stdout and there isn't an easy work-around, so it's hard to avoid :(

jluebbe:
I've never seen SSH control master startup produce significant output (i.e. more than 8k), so I'm skeptical if a pipe deadlock is the cause.

In https://github.com/labgrid-project/labgrid/pull/1265/commits/b8f059fb4277e3ca01684b7a71ac89974ff1cc36 you speculate:

> This likely was done this way because the timeout parameter was new in Python 3.3, but this shouldn't be a concern anymore

A wait() with timeout was used since the initial implementation: https://github.com/labgrid-project/labgrid/commit/e4862fa4a167a034c06e747df6190020d106605d#diff-9d05d7cbdae8b8ebd1d2d5d387c295c966943c5133258be844cb4d38aefd90e8R41

We wanted to make sure that we *don't* depend on the state of the pipes, so use of wait() was intentional, as it directly covers the interesting cases:

- immediate abort (due to a config error or similar)
- normal startup (parent process exits after fork)
- hang (via wait timeout)

A way to keep this behavior while handling a full pipe would be to use communicate with a very short timeout to start the reader threads and then using wait() to wait for the fork (instead of waiting for pipe closure).

JoshuaWatt:
> I've never seen SSH control master startup produce significant output (i.e. more than 8k), so I'm skeptical if a pipe deadlock is the cause.
> 
> In [b8f059f](https://github.com/labgrid-project/labgrid/commit/b8f059fb4277e3ca01684b7a71ac89974ff1cc36) you speculate:
> 
> > This likely was done this way because the timeout parameter was new in Python 3.3, but this shouldn't be a concern anymore
> 
> A wait() with timeout was used since the initial implementation: [e4862fa#diff-9d05d7cbdae8b8ebd1d2d5d387c295c966943c5133258be844cb4d38aefd90e8R41](https://github.com/labgrid-project/labgrid/commit/e4862fa4a167a034c06e747df6190020d106605d#diff-9d05d7cbdae8b8ebd1d2d5d387c295c966943c5133258be844cb4d38aefd90e8R41)
> 
> We wanted to make sure that we _don't_ depend on the state of the pipes, so use of wait() was intentional, as it directly covers the interesting cases:
> 
>     * immediate abort (due to a config error or similar)
> 
>     * normal startup (parent process exits after fork)
> 
>     * hang (via wait timeout)
> 
> 
> A way to keep this behavior while handling a full pipe would be to use communicate with a very short timeout to start the reader threads and then using wait() to wait for the fork (instead of waiting for pipe closure).

It feels like that's relying on internal implementation details (that the threads stick reading data after communicate() times out) a little too much. Not sure if that worse than the SSH version check though. The version check at least is pretty clear when it can be dropped.

JoshuaWatt:
Ya, that seems reasonable

jluebbe:
Or, alternatively, I think adding a timeout argument to `ProcessWrapper.check_output()` would do what we need here.

PR #1266, opened 2023-09-07
sshdriver: Log keepalive output

JoshuaWatt:
While it is not normally expected for the keepalive process to produce any output, it can be helpful for debugging purposes to log any output it may have produced. As such, try harder to get output when terminating the process, and also report any output after it exits

**Description**

**Checklist**
- [ ] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/labgrid/driver/sshdriver.py
+++ b/labgrid/driver/sshdriver.py
@@ -527,7 +527,8 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
             args,
             stdin=subprocess.PIPE,
             stdout=subprocess.PIPE,
-            stderr=subprocess.PIPE,
+            stderr=subprocess.STDOUT,
+            encoding="utf-8",
         )
 
         self.logger.debug('Started keepalive for %s', self.networkservice.address)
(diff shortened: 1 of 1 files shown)


Emantor (dismissed):

labgrid/driver/sshdriver.py
```diff
@@ -507,16 +507,35 @@ def _check_keepalive(self):
...
-        try:
-            self._keepalive.wait(timeout=60)
+            try:
+                # Try again to get output
+                stdout, stderr = self._keepalive.communicate(timeout=60)
+            except subprocess.TimeoutExpired:
+                try:
+                    self._keepalive.wait(timeout=60)
+                except subprocess.TimeoutExpired:
+                    self.logger.warning("ssh keepalive for %s timed out during termination", self.networkservice.address)
```
Bastian-Krause, line 536: Why is that needed? Using `communicate` instead of `wait` should not change anything, so adding this is surprising to me.

labgrid/driver/sshdriver.py
```diff
@@ -507,16 +507,35 @@ def _check_keepalive(self):
...
         finally:
             self._keepalive = None
+
+        log_output("stdout", stdout)
+        log_output("stderr", stderr)
```
Bastian-Krause, line 541: As you said, normally no output is generated. Is it worth to distinguish stdout/stderr here? Maybe change the `subprocess.Popen` in `_start_keepalive()` (stdin isn't used):

```diff
         self._keepalive = subprocess.Popen(
             args,
-            stdin=subprocess.PIPE,
             stdout=subprocess.PIPE,
-            stderr=subprocess.PIPE,
+            stderr=subprocess.STDOUT,
         )
 ```
If you also specify `encoding="utf-8"` there, you can probably also drop `log_output()` and do the logging here? The logging could even happen in the `try..except`, so you don't have to initialize the variables beforehand. This would slim down the patch even more.

Bastian-Krause (changes requested):
Seems useful, see some suggestions to slim down the patch below.

labgrid/driver/sshdriver.py
JoshuaWatt: stdin has to be `subprocess.PIPE` because this is how the keep alive is terminated (closing stdin), but otherwise I think I've done all of these.

Bastian-Krause (approved):
Thanks!

Emantor (approved):

Emantor:
Thanks!

PR #1270, opened 2023-09-17
target: improve exceptions

Bastian-Krause:
**Description**
Use short-hand class names and add driver causing `NoSupplierFoundErrors` in error messages.

**Checklist**
- [x] PR has been tested (tested, but did not trigger all exception code paths)


Emantor (approved):

cruise-bryan-murdock (approved):
Looks like it will be much more helpful

Bastian-Krause:
Since this changes the exception string only and the PR is open for a while now, I'll merge this without second approval.

PR #1272, opened 2023-09-25
ykushcmd: Add string support to ykushcmd command

redzynix:
**Description**
By default ykushcmd command supports option to [on, off, cycle] whole ykush yepkit.
> ykushcmd -u a  # 'a' stands for all ports
> ykushcmd -d a  # 'a' stands for all ports  

In current build only int is supported for port number, changing it to string allows to control power on all ports with one command instead of 3 if there are 3 ports.  
Tested locally.

**Checklist**
- [ - ] Documentation for the feature
- [ x ] Tests for the feature 

- [ - ] The arguments and description in doc/configuration.rst have been updated

- [ - ] Add a section on how to use the feature to doc/usage.rst

- [ - ] Add a section on how to use the feature to doc/development.rst

- [ x ] PR has been tested

- [ - ] Man pages have been regenerated


Bastian-Krause:
I don't see how this change would be useful in the context of labgrid. As far as I understand, this would allow exporting a YEPKIT Switchable USB Hub as a single resource. I cannot imagine a case where that would be useful. Could you please elaborate why it is?

Bastian-Krause:
Are you still interested in this PR?

Bastian-Krause:
@cgturner1 Can you explain your board setup a bit more? I can't really see how this would be useful.

KamilxPaszkiet:
> @cgturner1 Can you explain your board setup a bit more? I can't really see how this would be useful.

This functionality would be very useful. Right now I have a case that would save my setup. I need to turn off both USB ports simultaneously, because the devices "power" each other. For now, I have to turn them off individually, which takes more time and unnecessary additional commands.

I don't really understand why this change wasn't implemented from the beginning. It's one of the basic functionalities of ykush.

Bastian-Krause:
> > @cgturner1 Can you explain your board setup a bit more? I can't really see how this would be useful.
> 
> @Bastian-Krause @jluebbe @Emantor This functionality would be very useful. Right now I have a case that would save my setup. I need to turn off both USB ports simultaneously, because the devices "power" each other. For now, I have to turn them off individually, which takes more time and unnecessary additional commands.
> 
> I don't really understand why this change wasn't implemented from the beginning. It's one of the basic functionalities of ykush. For me it's a high priority pr

I think it was not implemented because it does not really match the concept of a YKUSHPowerPort when you actually work on multiple ports.

@jluebbe What do you think about this?

KamilxPaszkiet:
> > > @cgturner1 Can you explain your board setup a bit more? I can't really see how this would be useful.
> > 
> > 
> > @Bastian-Krause @jluebbe @Emantor This functionality would be very useful. Right now I have a case that would save my setup. I need to turn off both USB ports simultaneously, because the devices "power" each other. For now, I have to turn them off individually, which takes more time and unnecessary additional commands.
> > I don't really understand why this change wasn't implemented from the beginning. It's one of the basic functionalities of ykush. For me it's a high priority pr
> 
> I think it was not implemented because it does not really match the concept of a YKUSHPowerPort when you actually work on multiple ports.
> 
> What happens when querying the power state for "a"?
> 
> @jluebbe What do you think about this?

error because "a" it's not a int.

this is small change for You, but big change for us.

Bastian-Krause:
> > > > @cgturner1 Can you explain your board setup a bit more? I can't really see how this would be useful.
> > > 
> > > 
> > > @Bastian-Krause @jluebbe @Emantor This functionality would be very useful. Right now I have a case that would save my setup. I need to turn off both USB ports simultaneously, because the devices "power" each other. For now, I have to turn them off individually, which takes more time and unnecessary additional commands.
> > > I don't really understand why this change wasn't implemented from the beginning. It's one of the basic functionalities of ykush. For me it's a high priority pr
> > 
> > 
> > I think it was not implemented because it does not really match the concept of a YKUSHPowerPort when you actually work on multiple ports.
> > What happens when querying the power state for "a"?
> > @jluebbe What do you think about this?
> 
> error, because "a" it's not a int.

No, I meant with this PR applied (e.g. `labgrid-client power get`).

redzynix:
> > > > > @cgturner1 Can you explain your board setup a bit more? I can't really see how this would be useful.
> > > > 
> > > > 
> > > > @Bastian-Krause @jluebbe @Emantor This functionality would be very useful. Right now I have a case that would save my setup. I need to turn off both USB ports simultaneously, because the devices "power" each other. For now, I have to turn them off individually, which takes more time and unnecessary additional commands.
> > > > I don't really understand why this change wasn't implemented from the beginning. It's one of the basic functionalities of ykush. For me it's a high priority pr
> > > 
> > > 
> > > I think it was not implemented because it does not really match the concept of a YKUSHPowerPort when you actually work on multiple ports.
> > > What happens when querying the power state for "a"?
> > > @jluebbe What do you think about this?
> > 
> > 
> > error, because "a" it's not a int.
> 
> No, I meant with this PR applied (e.g. `labgrid-client power get`).

By doing a static analysis of this function it will fail due to different port numbers. 'a' != 15
https://github.com/labgrid-project/labgrid/blob/master/labgrid/driver/powerdriver.py#L314-L332

By running **ykushcmd -g a** ykushcmd returns the port number of the USB that supplies the board with power, at least in my configuration.

>$ ykushcmd -g a
Downstream port 15 is ON

PR #1274, opened 2023-09-28
[feature] spe_switch: Add spe_switch power feature

ActionHOSchT:
This PR adds the feature to set power for a phoenix_fl_switch over telnet.

Devices connected with single-pair-ethernet can now be set to power off/on/cycle.

To verify, this Driver was tested on a *FL SWITCH 2303-8SP1* with FW-version *3.27.01 BETA*.
Even if it's a BETA, we got the information that the power control won't experience breaking changes.

grafik
grafik

grafik

**Checklist**
- [x] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -195,6 +195,10 @@ Currently available are:
 ``netio_kshell``
   Controls a NETIO 4C PDU via a Telnet interface.
 
+``phoenix_fl_switch``
+  Controls a single-pair-ethernet powerswitch via telnet.
+  Tested on a FL SWITCH 2303-8SP1 with FW-version 3.27.01 BETA
+
 ``raritan``
   Controls Raritan PDUs via SNMP.
 
--- a/labgrid/driver/power/phoenix_fl_switch.py
+++ b/labgrid/driver/power/phoenix_fl_switch.py
@@ -6,10 +6,27 @@ date   2023-08-24
 
 Copyright 2023 JUMO GmbH & Co. KG
 '''
+import pexpect
 
-from telnetlib import Telnet
+PORT = 23
 
-PORT=23
+
+def login_telnet(tn):
+    """
+    Login user with set credentials
+
+    @param tn : pyexpect-telnet-object
+    @returns : logged in pyexpect-telnet-object
+    """
+    username = "admin"
+    password = "private"
+
+    # login user with password
+    tn.expect(b'User: ')
+    tn.send(bytes(f'{username}\r\n', "utf-8"))
+    tn.expect(b'Password: ')
+    tn.send(bytes(f'{password}\r\n', "utf-8"))
+
 
 def power_set(host, port, index: int, value: bool):
     """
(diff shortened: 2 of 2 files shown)


doc/configuration.rst
```diff
@@ -195,6 +195,10 @@ Currently available are:
 ``netio_kshell``
   Controls a NETIO 4C PDU via a Telnet interface.
 
+``phoenix_fl_switch``
+  Controls a single-pair-ethernet powerswitch via telnet.
+  Tested on a FL SWITCH 2303-8SP1 with FW-version 3.27.01 BETA
+
```
Bastian-Krause, line 201: I am not sure about the backend name. The complete name of your device is "Phoenix Contact FL SWITCH 2303-8SP1". Could you find out which products use the same "telnet API"? Maybe have look at the firmware? Maybe something like "phoenixcontact_fl_2300"?

Bastian-Krause (commented):
Looks good to me.

CI test errors are not your fault, https://github.com/labgrid-project/labgrid/pull/1277 fixes them.

FYI: there is also https://github.com/pdudaemon/pdudaemon, a project that focuses only on PDUs. labgrid's PDUDaemonDriver interfaces with it. So it might be worth a thought to implement this rather over there. Let us know how you think about this.

Note: we should wait for https://github.com/labgrid-project/labgrid/pull/1277 and rebase this to make CI run this before merging.

labgrid/driver/power/phoenix_fl_switch.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+def __login_telnet(tn):
+    """
+    Login user with set credentials
+
+    @param tn : pyexpect-telnet-object
+    """
+    username = "admin"
+    password = "private"
```
jluebbe, line 21: Instead of hardcoding the credentials, you could use a URL like `telnet://admin:private@1.2.3.4/` as the host. You'd need to add `telnet` as a scheme in https://github.com/labgrid-project/labgrid/blob/master/labgrid/driver/powerdriver.py#L194

jluebbe (changes requested):
Please also squashfs the existing commits into one with a title following our usual conventions, like "driver/power: add backend for Phoenix Contact FL 2303-8SP1 SPE switch".

Adding support for the telnet scheme in labgrid/driver/powerdriver.py should be a separate commit.

labgrid/driver/power/phoenix_fl_switch.py
ActionHOSchT: This would indeed be nice.

I used _netio_kshell.py_ as inspiration for the _phoenix_fl_switch.py_ .
Implementing a new scheme would be some effort.

If you want me to put this suggestion on our todo this will be noted ?

ActionHOSchT:
Since I will be no longer collaborating with JUMO and therefore have no setup to test changes, I close this PR.

labgrid/driver/power/phoenix_fl_switch.py
SeminSey: Commits have been squashed as requested.

Emantor:
According to the [product page](https://www.phoenixcontact.com/de-de/managed-tsn-switches-2300#ex-oi6en), this device also supports snmp, does it support power switching via SNMP as well? This would be less fragile vs implementing this via pexpect over telnet.

ChHapp:
Hi @Emantor, 

We contacted Phoenix Contact round about seven months ago and asked them how we could remotely control the ports. 
They prepared a test firmware (FW version 3.27.01 BETA) for us and we sent them feedback that everything worked just fine with this beta FW. So they told us that this feature would definitely be part of the next release. I understand your concerns about the Telenet API, to be honest I have no idea if it will break soon or not.  
To put it in a nutshell, as far as we know, that's the way to go at the moment.

doc/configuration.rst
SeminSey: Name has been adjusted.

ChHapp:
@Emantor ping

PR #1278, opened 2023-10-06
sshdriver: fix communicate hang if ssh version < 8.5

atline:
**Description**
It looks there is some change to use `subprocess.communicate` to replace `wait` in ssh driver recently, but it indeed makes sshdriver hang if ssh version < 8.5, then break our usage.

You can quickly check it with docker example which use sshdriver, it will finally fails and print traceback:
```
try:
    subprocess_timeout = timeout + 5
> stdout, _ = self.process.communicate(timeout=subprocess_timeout)

../../labgrid/driver/sshdriver.py:139:
```

I use next example to describe the issue:

__test.py__:

```
import subprocess
import os
import sys
import stat

ssh_user = "my_user"       # change to your account
ssh_passwd = "my_password" # change to your password

try:
    os.remove("/tmp/control-channel")
except:
    pass

# prepare passwd file
file = open("./passwd","w")
passwd_content = f"#!/bin/sh\necho {ssh_passwd}"
file.write(passwd_content)
file.close()
os.chmod("./passwd", stat.S_IRWXU)

# setup shared ssh socket, put it in background
env = {'SSH_ASKPASS': "./passwd", 'DISPLAY':'', 'SSH_ASKPASS_REQUIRE':'force'}
args = ['ssh', '-f', '-o', 'LogLevel=ERROR', '-x', '-o', 'ConnectTimeout=30', '-o', 'ControlPersist=300', '-o', 'UserKnownHostsFile=/dev/null', '-o', 'StrictHostKeyChecking=no', '-o', 'ServerAliveInterval=15', '-MN', '-S', '/tmp/control-channel', '-p', '22', '-l', ssh_user, 'localhost']
process = subprocess.Popen(args, env=env,
        #stdout=subprocess.PIPE,
        stderr=subprocess.PIPE,
        #stdin=subprocess.DEVNULL,
        start_new_session=True)
sout, serr = process.communicate()  # hang here!
print(sout)
print(serr)

# use shared socket
args2 = ['ssh', '-o', 'LogLevel=ERROR', '-o', 'ControlPath=/tmp/control-channel', '-p', '22', '-l', ssh_user, 'localhost', 'uname -a']
process2 = subprocess.Popen(args2,
        stdout=subprocess.PIPE,
        stderr=subprocess.STDOUT,
        stdin=subprocess.DEVNULL)
content, _ = process2.communicate()
print(content)
```

Above example will hang at `process.communicate` when setup master ssh connection if ssh < 8.4, while it's ok if ssh is newer.

1. I see it stuck as next:
```
$ strace python3 test.py
lseek(3, 0, SEEK_CUR)                   = -1 ESPIPE (Illegal seek)
fstat(3, {st_mode=S_IFIFO|0600, st_size=0, ...}) = 0
read(3, 0x1744f00, 8192)                = ? ERESTARTSYS (To be restarted if SA_RESTART is set)
--- SIGCHLD {si_signo=SIGCHLD, si_code=CLD_EXITED, si_pid=26779, si_uid=1001, si_status=0, si_utime=1, si_stime=0} ---
read(3,
```

2.
``` 
$ sudo lsof -a -c python -d 3
COMMAND   PID     USER   FD   TYPE DEVICE SIZE/OFF    NODE NAME
python3 26778 nxa13855    3r  FIFO   0,12      0t0 9186593 pipe
```

3.
```
ls -l "/proc/26778/fd/3"
lr-x------ 1 nxa13855 atg 64 Oct  5 20:22 /proc/26778/fd/3 -> 'pipe:[9186593]'
$ lsof | grep 9186593
python3   26778                   nxa13855    3r     FIFO               0,12      0t0    9186593 pipe
ssh       26783                   nxa13855    2w     FIFO               0,12      0t0    9186593 pipe
```

4.
This means the ssh -f did not close stderr when put the ssh to background, so communicate can't read a EOF of stderr then hangs as next:
```
subprocess.py", line 929, in communicate
stderr = self.stderr.read()
```

After check ssh source tree, I found this commit: https://github.com/openssh/openssh-portable/commit/d14fe25e6c3b89f8af17e2894046164ac3b45688 which fixed from V8.4P1. So, for ssh version <8.4, we shouldn't use `communicate` when `ssh -f` put itself to background.

As a result, this PR distinguish ssh version, for >=8.5, we use `communicate`; while for <8.4 we still back to old ways to use `wait` (This may block if ssh produce lots of information, but for a master connection which did nothing just a communication setup, there is low chance to have any issue. And sooner or later ssh>8.4 will be the main stream. Otherwise, currently totally block for <8.5 user.

**Checklist**
- [✓] Tests for the feature 
- [✓] PR has been tested

Fixes #1265 (for OpenSSH <8.5)

### DIFF
--- a/labgrid/driver/sshdriver.py
+++ b/labgrid/driver/sshdriver.py
@@ -137,8 +137,14 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
 
         try:
             subprocess_timeout = timeout + 5
-            stdout, _ = self.process.communicate(timeout=subprocess_timeout)
+            if self._ssh_supports_fork_with_stderr():
+                stdout, _ = self.process.communicate(timeout=subprocess_timeout)
+            else:
+                self.process.wait(timeout=subprocess_timeout)
+
             if self.process.returncode != 0:
+                if not self._ssh_supports_fork_with_stderr():
+                    stdout, _ = self.process.communicate(timeout=subprocess_timeout)
                 stdout = stdout.split(b"\n")
                 for line in stdout:
                     self.logger.warning("ssh: %s", line.rstrip().decode(encoding="utf-8", errors="replace"))
@@ -437,6 +443,12 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
         version = subprocess.run(["ssh", "-V"], capture_output=True, text=True)
         version = re.match(r"^OpenSSH_(\d+)\.(\d+)", version.stderr)
         return map(int, version.groups())
+
+    def _ssh_supports_fork_with_stderr(self):
+        major, minor = self._ssh_version
+
+        # OpenSSH < 8.5 has bug about stderr close via -f
+        return major > 8 or (major == 8 and minor > 4)
 
     def _scp_supports_explicit_sftp_mode(self):
         major, minor = self._ssh_version
--- a/labgrid/driver/sshdriver.py
+++ b/labgrid/driver/sshdriver.py
@@ -39,6 +39,7 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
         super().__attrs_post_init__()
         self.logger = logging.getLogger(f"{self}({self.target})")
         self._keepalive = None
+        self._ssh_version = self._ssh_version_check()
 
     def on_activate(self):
         self.ssh_prefix = ["-o", "LogLevel=ERROR"]
@@ -432,10 +433,13 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
         """The SSHDriver is always connected, return 1"""
         return 1
 
-    def _scp_supports_explicit_sftp_mode(self):
+    def _ssh_version_check(self):
         version = subprocess.run(["ssh", "-V"], capture_output=True, text=True)
         version = re.match(r"^OpenSSH_(\d+)\.(\d+)", version.stderr)
-        major, minor = map(int, version.groups())
+        return map(int, version.groups())
+
+    def _scp_supports_explicit_sftp_mode(self):
+        major, minor = self._ssh_version
 
         # OpenSSH >= 8.6 supports explicitly using the SFTP protocol via -s
         if major == 8 and minor >= 6:
--- a/tests/test_sshdriver.py
+++ b/tests/test_sshdriver.py
@@ -8,20 +8,26 @@ from labgrid.util.helper import get_free_port
 
 @pytest.fixture(scope='function')
 def ssh_driver_mocked_and_activated(target, mocker):
-    NetworkService(target, "service", "1.2.3.4", "root")
-    call = mocker.patch('subprocess.call')
-    call.return_value = 0
-    popen = mocker.patch('subprocess.Popen', autospec=True)
-    path = mocker.patch('os.path.exists')
-    path.return_value = True
-    instance_mock = mocker.MagicMock()
-    popen.return_value = instance_mock
-    instance_mock.wait = mocker.MagicMock(return_value=0)
-    instance_mock.communicate = mocker.MagicMock(return_value=(b"", b""))
-    instance_mock.returncode = 0
-    SSHDriver(target, "ssh")
-    s = target.get_driver("SSHDriver")
-    return s
+    def _ssh_driver_mocked_and_activated(ssh_info="OpenSSH_8.9p1"):
+        NetworkService(target, "service", "1.2.3.4", "root")
+        call = mocker.patch('subprocess.call')
+        call.return_value = 0
+        popen = mocker.patch('subprocess.Popen', autospec=True)
+        path = mocker.patch('os.path.exists')
+        path.return_value = True
+        instance_mock = mocker.MagicMock()
+        popen.return_value = instance_mock
+        instance_mock.wait = mocker.MagicMock(return_value=0)
+        instance_mock.communicate = mocker.MagicMock(return_value=(b"", b""))
+        instance_mock.returncode = 0
+        run = mocker.patch('subprocess.run')
+        run_mock = mocker.MagicMock()
+        run.return_value = run_mock
+        run_mock.stderr = ssh_info
+        SSHDriver(target, "ssh")
+        s = target.get_driver("SSHDriver")
+        return s
+    return _ssh_driver_mocked_and_activated
 
 def test_create_fail_missing_resource(target):
     with pytest.raises(NoResourceFoundError):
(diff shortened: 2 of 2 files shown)


Bastian-Krause:
Thanks for the PR! We've noticed this, too.

While the commit https://github.com/openssh/openssh-portable/commit/396d32f3a1a16e54df2a76b2a9b237868580dcbe sounds a lot like the problem, we're also seeing this on OpenSSH 8.4 (Debian bullseye). I've bisected the problem to https://github.com/openssh/openssh-portable/commit/396d32f3a1a16e54df2a76b2a9b237868580dcbe. Could you verify that we are in fact seeing the same issue? Should be as simple as running `autoreconf -ivf && ./configure && make -j200` in cloned OpenSSH and replacing the "ssh" with the absolute path to the compiled ssh binary.

By the way, the problem is also going away when omitting the `stderr=subprocess.STDOUT`in `Popen()`.

atline:
> Thanks for the PR! We've noticed this, too.
> 
> While the commit [openssh/openssh-portable@396d32f](https://github.com/openssh/openssh-portable/commit/396d32f3a1a16e54df2a76b2a9b237868580dcbe) sounds a lot like the problem, we're also seeing this on OpenSSH 8.4 (Debian bullseye). I've bisected the problem to [openssh/openssh-portable@396d32f](https://github.com/openssh/openssh-portable/commit/396d32f3a1a16e54df2a76b2a9b237868580dcbe). Could you verify that we are in fact seeing the same issue? Should be as simple as running `autoreconf -ivf && ./configure && make -j200` in cloned OpenSSH and replacing the "ssh" with the absolute path to the compiled ssh binary.
> 
> By the way, the problem is also going away when omitting the `stderr=subprocess.STDOUT`in `Popen()`.

You are correct, the 8.4 did not fix that issue, it indeed be fixed in 8.5. And yes, `remove stderr or stderr=subprocess.DEVNULL` could fix it, as subprocess will not read stderr then. But that way, we won't see any error if the ssh command failed before put itself to background, E.g. any configure error in .ssh/config or other error.

labgrid/driver/sshdriver.py
```diff
@@ -135,28 +135,41 @@ def _start_own_master_once(self, timeout):
...
 
         try:
+            version = subprocess.run(["ssh", "-V"], capture_output=True, text=True)
+            version = re.match(r"^OpenSSH_(\d+)\.(\d+)", version.stderr)
+            major, minor = map(int, version.groups())
+
+            if major < 8 or (major == 8 and minor < 5):
+                ssh_good_version = False
+            else:
+                ssh_good_version = True
```
Bastian-Krause, line 145: We already have `_scp_supports_explicit_sftp_mode()`. We should move the version check to a separate method and add a method `_ssh_supports_fork_with_stderr()`. Performing the version check once in `__attrs_post_init__()` would probably be best.

labgrid/driver/sshdriver.py
```diff
@@ -135,28 +135,41 @@ def _start_own_master_once(self, timeout):
...
-                except FileNotFoundError:
-                    pass
-
-                raise ExecutionError(
-                    f"Failed to connect to {self.networkservice.address} with {' '.join(args)}: return code {self.process.returncode}",  # pylint: disable=line-too-long
-                    stdout=stdout,
-                )
+            if not ssh_good_version:
+                return_value = self.process.wait(timeout=subprocess_timeout)
+            elif ssh_good_version or return_value != 0:
```
Bastian-Krause, line 151: I don't see how `return_value` can be anything else than `0` here.

labgrid/driver/sshdriver.py
```diff
@@ -135,28 +135,41 @@ def _start_own_master_once(self, timeout):
...
+                                f"Failed to connect to {self.networkservice.address} with {' '.join(args)}: error from SSH ProxyCommand: {proxy_error}",  # pylint: disable=line-too-long
+                                stdout=stdout,
+                            )
+                    except FileNotFoundError:
+                        pass
+
+                    raise ExecutionError(
+                        f"Failed to connect to {self.networkservice.address} with {' '.join(args)}: return code {self.process.returncode}",  # pylint: disable=line-too-long
+                        stdout=stdout,
+                    )
```
Bastian-Krause, line 172: Wouldn't it be sufficient if we call `communicate()` prior to `self.process.returncode != 0:` for OpenSSH >=8.5 and in the `self.process.returncode != 0:` block for OpenSSH <8.5? That would make the patch way shorter.

labgrid/driver/sshdriver.py
```diff
@@ -357,7 +370,7 @@ def scp(self, *, src, dst):
                 "-o", f"ControlPath={self.control.replace('%', '%%')}",
                 src, dst,
         ]
-        
+
```
Bastian-Krause, line 373: Seems unrelated.

Bastian-Krause (changes requested):

labgrid/driver/sshdriver.py
atline: done

labgrid/driver/sshdriver.py
atline: If ssh report error before ssh put itself into background, that will return non-zero exit code. Anyway, this code is outdated now.

labgrid/driver/sshdriver.py
atline: Reasonable, see updated.

labgrid/driver/sshdriver.py
atline: lg

My vim plugin automatically delete useless blank in a empty line when save. Ok, I disable my vim plugin and leave this line ASIS.

labgrid/driver/sshdriver.py
```diff
@@ -432,10 +439,23 @@ def get_status(self):
...
+    def _ssh_supports_fork_with_stderr(self):
+        major, minor = self._ssh_version
+
+        # OpenSSH < 8.5 has bug about stderr close via -f
+        if major < 8 or (major == 8 and minor < 5):
+            return False
+        else:
+            return True
+        raise Exception(f"OpenSSH version {major}.{minor} does not support ssh mode")
```
Bastian-Krause, line 455:
```suggestion
        return major < 8 or (major == 8 and minor < 5)
```
This should be sufficient.

Bastian-Krause (changes requested):
I suggest one small simplification, rest looks good, thanks!

labgrid/driver/sshdriver.py
Bastian-Krause: You're right, the other way around :)

Bastian-Krause:
Successfully tested with:
- Debian 11.8 (bullseye): OpenSSH_8.4p1 Debian-5+deb11u2, OpenSSL 1.1.1w  11 Sep 2023
- Debian 12.2 (bookworm): OpenSSH_9.2p1 Debian-2+deb12u1, OpenSSL 3.0.11 19 Sep 2023

ejoerns:
FTR: I have also modified the version check with #1281 (using `@cached_property`) since I wasn't aware that this had been done here, too. Maintainers might choose which variant to prefer :smirk: I am fine with rebasing my work on this anyway.

jluebbe:
I think the simpler fix is to revert https://github.com/labgrid-project/labgrid/pull/1265. It avoids the complexity of having to maintain two different ways to start SSH.

If timeouts still occur after the revert, we should implement https://github.com/labgrid-project/labgrid/pull/1265#issuecomment-1761670396.

atline:
> I think the simpler fix is to revert #1265. It avoids the complexity of having to maintain two different ways to start SSH.
> 
> If timeouts still occur after the revert, we should implement [#1265 (comment)](https://github.com/labgrid-project/labgrid/pull/1265#issuecomment-1761670396).

That's ok for me as I never saw issue before that commit, feel free to close this MR after your revert.

PR #1280, opened 2023-10-10
driver/powerdriver: ykushpower: board_name support

PaulVittorino:
**Description**
When 6425df99a2de31563d96fa33f2ba65a2d669c84d switched from `pykush` to `ykushcmd`, support for the [YKUSH 3](https://www.yepkit.com/product/300110/YKUSH3) and [YKUSH XS](https://www.yepkit.com/product/300115/YKUSHXS) was lost. Restore support for those boards by passing the `board_name` argument to `ykushcmd`. The `board_name` is an optional argument for the `YKUSHPowerPort` and `NetworkYKUSHPowerPort` resources.

**Checklist**
- [x] Tests for the feature 
- [x] PR has been tested


jluebbe:
Isn't it possible to detect the model automatically (i.e. using the USB VID/PID) instead of requiring manual configuration by the user? How did this work previously? @edersondisouza?

PaulVittorino:
> Isn't it possible to detect the model automatically (i.e. using the USB VID/PID) instead of requiring manual configuration by the user?

The user manuals for the YKUSH XS, YKUSH, and YKUSH3 state the products have the same VID but different PIDs. Therefore, I believe it is possible to detect the model automatically. Would you advise changing `YKUSHPowerPort` to inherit from `USBResource`?

>How did this work previously?

`pykush` does not have a `board_type` argument.

```shell
$ python3.8 pykush.py -l
listing YKUSH family devices
  found a YKUSH XS release 2 device with serial number YKU2768
    system device path 1-8.2:1.0, vendor id 0x04d8, product id 0xf0cd
    the device has 1 downstream port
    Checking running power state, port 1 : UP
  found a YKUSH3 release 1 device with serial number Y3N10673
    system device path 1-4.4:1.0, vendor id 0x04d8, product id 0xf11b
    the device is running a v1.0 firmware and has 3 downstream ports
    downstream running power states, port 1 to 3: UP, UP, UP
```

`ykushcmd` requires the `board_name` argument for the YKUSH XS and YKUSH3 boards.

```shell
$ ./bin/ykushcmd -l
Attached YKUSH Boards:
No YKUSH boards found.

$ ./bin/ykushcmd ykush3 -l
Attached YKUSH3 Boards:
1. Board found with serial number: Y3N10673

$ ./bin/ykushcmd ykushxs -l
Attached YKUSH XS Boards:
1. Board found with serial number: YKU2768
```

edersondisouza:
Ouch, this is really inconvenient =/

Wonder if trying to get USB PID will work nice with the remote instance. Maybe require pykush and use the command line tool (not the module, so it's easy to support remote instance)?

An ugly hack would be to try to chain three calls to ykushcmd and use it, like `ykushcmd ... | ykushcmd ykush3 ... | ykushcmd ykushxs...` (/me hides in the corner).

Or fix ykushcmd itself... =/

jluebbe:
Fixing `ykushkmd` would be nice and should be easy based on the VID/PID info.

Otherwise, you could make `YKUSHPowerPort` inherit from `USBResource`, which will give it additional USB properties (`busnum`, `devnum`, `path`, `vendor_id` and `model_id`). NetworkYKUSHPowerPort would then inherit from RemoteUSBResource to store these new properties.

YKUSHPowerPort would match USB HID devices by the given serial. In the Driver, you could then use the USB vendor/model IDs to generate the correct board name  for `ykushcmd`.

The additional resource properties would break backwards compatibility, but the current state is broken anyway. :/

PaulVittorino:
> The additional resource properties would break backwards compatibility, but the current state is broken anyway. :/

I found a way to not break backward compatibility. Use `ykushcmd` to list out the attached boards by model. Then determine the model of a given board by looking for the serial number in the output for each model. It will require adding each new board to the `YKUSHPowerDriver` but the PID/VID solution had the same issue.

doc/configuration.rst
```diff
@@ -264,18 +264,19 @@ Used by:
 
 YKUSHPowerPort
 ++++++++++++++
-A :any:`YKUSHPowerPort` describes a *YEPKIT YKUSH* USB (HID) switchable USB
-hub.
+A :any:`YKUSHPowerPort` describes a *YEPKIT YKUSH* family USB (HID)
```
sjg20, line 267: nit: hyphen before family

labgrid/driver/powerdriver.py
```diff
@@ -283,11 +284,28 @@ def __attrs_post_init__(self):
...
+                model,
+                "-l"
+            ]
+            output = processwrapper.check_output(cmd)
+            if self.port.serial in output.decode("utf-8"):
+                self.model = model
+                break
+        else:
+            raise InvalidConfigError(f"Could not find YKUSH device with serial {self.port.serial}")
+
```
sjg20, line 301: drop extra blank line?

sjg20 (approved):

edersondisouza (commented):
Nice workaround! Minor nit: on commit message the phrase "Added tests for YKUSHPowerPort and YKUSHPowerPort.' seems wrong - maybe you meant YKUSHPowerDriver for the second YKUSHPowerPort?

PaulVittorino:
> Nice workaround! Minor nit: on commit message the phrase "Added tests for YKUSHPowerPort and YKUSHPowerPort.' seems wrong - maybe you meant YKUSHPowerDriver for the second YKUSHPowerPort?

Yes, thanks for catching that.

edersondisouza (approved):

edersondisouza:
Any more comments on this, @jluebbe and @Bastian-Krause?

Bastian-Krause:
Force-pushed: dropped unrelated copy-pasted `test_import_backend_poe_mib`.

PR #1285, opened 2023-10-18
Log verbose scp output

nick-potenski:
**Description**
Adds the `-v` option to all scp calls and saves the output to the logger.  The get and put methods log at warning level on failure and the scp method logs at info level always.  The additional output can help debug scp failures.

**Checklist**
- [ ] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/labgrid/driver/sshdriver.py
+++ b/labgrid/driver/sshdriver.py
@@ -357,6 +357,7 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
         complete_cmd = ["scp",
                 "-F", "none",
                 "-o", f"ControlPath={self.control.replace('%', '%%')}",
+                "-v",
                 src, dst,
         ]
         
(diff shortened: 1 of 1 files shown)


Bastian-Krause:
Thanks for thinking about contributing. I can't really see how this "debugging improvement" benefits labgrid. Even without this change, critical output of scp should still be visible on stdout, right? What kind of errors are you debugging?

nick-potenski:
In this case, a call to `SSHDriver.put()` failed.  The only output was `lost connection`.  I thought that adding the `-v` would be beneficial debugging and might give more details on which connection was lost.  I don't necessarily need it to go to the logger, just thought that might be nicer.

Bastian-Krause:
Hm, I see why you've added that for your debugging session, but I don't think that adding `-v` to scp and logging its output belongs into labgrid.

Bastian-Krause:
As explained earlier, I'll close this.

PR #1290, opened 2023-10-24
exporter: Use fqdn instead of hostname only

legraps:
**Description**
In distributed setups, the coordinator may point to places in different subnets. In order for clients to resolve resources behind the exporter, they need to know its FQDN instead of just the hostname.

Thus, set this to be the default when the exporter is invoked without the --hostname parameter.

**Checklist**
- [ ] Tests for the feature 

- [ ] PR has been tested

### DIFF
--- a/labgrid/remote/exporter.py
+++ b/labgrid/remote/exporter.py
@@ -917,6 +917,12 @@ def main():
         help='hostname (or IP) published for accessing resources (defaults to the system hostname)'
     )
     parser.add_argument(
+        '--fqdn',
+        action='store_true',
+        default=False,
+        help='Use fully qualified domain name as default for hostname'
+    )
+    parser.add_argument(
         '-d',
         '--debug',
         action='store_true',
@@ -943,7 +949,7 @@ def main():
 
     extra = {
         'name': args.name or gethostname(),
-        'hostname': args.hostname or gethostname(),
+        'hostname': args.hostname or getfqdn() if args.fqdn else gethostname(),
         'resources': args.resources,
         'isolated': args.isolated
     }
--- a/man/labgrid-exporter.rst
+++ b/man/labgrid-exporter.rst
@@ -46,6 +46,8 @@ OPTIONS
     the public name of the exporter
 --hostname
     hostname (or IP) published for accessing resources
+--fqdn
+    use fully qualified domain name as default for hostname
 -d, --debug
     enable debug mode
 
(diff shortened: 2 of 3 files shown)


jluebbe:
This would break many existing setups where users have a `.ssh/config` which matches their exporters' host names. If you need to access resources which are only accessible via the exporter, use the proxy/isolated mechanism. The proxy property already uses `getfqdn()`.

legraps:
Interesting, thank you for your quick feedback!
The patch was motivated by my conception that the coordinator is designed to be the central reliable point to accumulate connectivity information rather than every developer maintaining their own config. Is this wrong somehow?

As for the proxy, yes this does work and some of my colleagues use it that way. But I also expect drawbacks in forcing all traffic through the exporter and thought it as solution for some use cases only (e.g. remote/VPN, "difficult" networks outside administration control) but not the standard where name resolution into different subnets works.

legraps:
As suggested by @Bastian-Krause, I just pushed a rework introducing a flag --fqdn to change the default when no --hostname is given.

Emantor:
Please also update the man page under `man/` and regenerate the files 👍

man/labgrid-exporter.rst
```diff
@@ -73,6 +75,12 @@ key.
 If the system hostname is not resolvable via DNS, this option can be used to
 override this default with another name (or an IP address).
 
+-f --fqnd
```
Bastian-Krause, line 78:
```suggestion
-f --fqdn
```
Small typo.

labgrid/remote/exporter.py
```diff
@@ -916,6 +916,13 @@ def main():
         default=None,
         help='hostname (or IP) published for accessing resources (defaults to the system hostname)'
     )
+    parser.add_argument(
+        '-f',
```
Bastian-Krause, line 920: After looking at this again, the feature seems to be required in rather special use cases, so we should drop the shortopt, only use `--fqdn` (same as `--hostname`). I would have fixed this myself and would have approved the PR, but you did not allow maintainer pushes to the PR branch.

Bastian-Krause (changes requested):

labgrid/remote/exporter.py
legraps: ok, reworked the branch to drop the -f-

Bastian-Krause: Thanks!

Bastian-Krause (approved):

Emantor (approved):

PR #1291, opened 2023-10-26
Support multiple credentials for shell and ssh

nlabriet:
**Description**

Add support for having a list of usernames and passwords to try to connect with.
This touches `shelldriver`, `networkservice `and `sshdriver`.

Depending on the software provisioned on a place, credentials may be different. It's my case where 3 different credentials are needed (user1/pass1 user2/pass2 user3/pass3) when soft1 or soft2 or soft3 is on the target.

In order not to have to change the LabGrid configuration (and the exporter as `networkservice `holds the IP and credentials for ssh) I can up with this PR.

The credentials are tried in turn and I don't get a timeout with 3 credentials. More credentials will probably need a timeout tweak or a way to configure it.

I have manually tested login in with `console`, `ssh`, and `-s shell ssh`, I have also tested sending a file with `scp`.

This also enable a smoother test experience. My tests can now use the same configuration for all software.

**Checklist**
- [X] Tests for the feature 
- [X] PR has been tested

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -1557,8 +1557,8 @@ Implements:
 Arguments:
   - prompt (regex): shell prompt to match after logging in
   - login_prompt (regex): match for the login prompt
-  - username (str): username to use during login
-  - password (str): optional, password to use during login
+  - username (str or list): username(s) to use during login
+  - password (str or lits): optional, password(s) to use during login
   - keyfile (str): optional, keyfile to upload after login, making the
     `SSHDriver`_ usable
   - login_timeout (int, default=60): timeout for login prompt detection in
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -544,8 +544,8 @@ These and the sudo configuration needs to be prepared by the administrator.
 
 Arguments:
   - address (str): hostname of the remote system
-  - username (str): username used by SSH
-  - password (str, default=""): password used by SSH
+  - username (str or list): username(s) used by SSH
+  - password (str or list, default=""): password(s) used by SSH
   - port (int, default=22): port used by SSH
 
 Used by:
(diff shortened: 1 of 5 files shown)


Bastian-Krause:
Thanks for thinking about contributing.

We had several discussions about "generic environment configurations" recently. My perspective is: The environment configuration is intended to be DUT-specific (including software). If username/password are different, you _need_ a different configuration. The same applies to cases like "different bootloader", "different prompt" etc.

The situation is different when you connect multiple identical DUTs, but your infrastructure is different, e.g. another power switch is used requiring a different driver in the environment config. For these cases, we think that an environment config managed by the coordinator might be a valid approach. Changing this place-specific, central config could be done via labgrid-client. A local config could still be used, both would be merged as long as they do not conflict.

So I don't think that this change is something that we want to merge into labgrid. What do you think, @Emantor, @jluebbe?

nlabriet:
Thank you for your feedback.
I agree with the DUT-specific configuration needing to be in the environment configuration.
This should of course include credentials and prompts.

But I think the `address` given in `NetworkService` should come from the infrastructure and as of now, SSH credentials are mandatory in this class.

By the way, why keep SSH and console credentials separate?

My PR is meant to highlight and easy my use case, but I could probably manage another way (yaml anchors are great in the environment file).

Bastian-Krause:
> Thank you for your feedback. I agree with the DUT-specific configuration needing to be in the environment configuration. This should of course include credentials and prompts.
> 
> But I think the `address` given in `NetworkService` should come from the infrastructure and as of now, SSH credentials are mandatory in this class.

If the address should come from the infrastructure (i.e. via a [SNMPEthernetPort](https://labgrid.readthedocs.io/en/latest/configuration.html#snmpethernetport)), you can use a NetworkService with a placeholder address and update it during runtime with this information.

> By the way, why keep SSH and console credentials separate?

I think @jluebbe or @Emantor have to comment on that.

nlabriet:
> If the address should come from the infrastructure (i.e. via a [SNMPEthernetPort](https://labgrid.readthedocs.io/en/latest/configuration.html#snmpethernetport)), you can use a NetworkService with a placeholder address and update it during runtime with this information.

The address is provided by the exporter and will always be the same for a given place.

What about adding `username` and `password` to `SSHDriver` and using thoses if they are defined? Falling back on the ones provided by `NetworkService` to maintain compatibility.

Since SSHDriver is in the environment file, credentials could then be completely managed by the environment file.

Issue #980, opened 2022-09-01
AssertionError in step.py when running multiple SSHDrivers in different threads


S4mw1s3:
Hi,

I'm using `labgrid==0.4.1` and I want to query multiple devices in parallel by using the SSHDriver. The following script gives me an AssertionError in step.py though:

```
import logging
import threading
import argparse

from labgrid import Target
from labgrid.resource import NetworkService
from labgrid.driver import SSHDriver

parser = argparse.ArgumentParser()
parser.add_argument('user')
parser.add_argument('passwd')
args = parser.parse_args()


def query_params(ip):
    print(f"Checking {ip}")

    t = Target('DUT')
    nw_service = NetworkService(t, "NS", ip, args.user, args.passwd)
    ssh_d = SSHDriver(t, "SSH Driver")
    t.activate(ssh_d)

    try:
        ssh_d.run_check("ls")
    except Exception as e:
        logging.exception(e)

    t.cleanup()


while True:
    t1 = threading.Thread(target=query_params, args=("10.0.0.152", ))
    t1.start()

    t2 = threading.Thread(target=query_params, args=("10.0.0.153", ))
    t2.start()

    t1.join()
    t2.join()
```

The exception I get is:
```
ERROR:root:
Traceback (most recent call last):
  File "/home/testing/utils/lg_test.py", line 24, in query_params
    ssh_d.run_check("ls")
  File "/home/testing/venv/lib/python3.10/site-packages/labgrid/binding.py", line 96, in wrapper
    return func(self, *_args, **_kwargs)
  File "/home/testing/venv/lib/python3.10/site-packages/labgrid/step.py", line 222, in wrapper
    step.stop()
  File "/home/testing/venv/lib/python3.10/site-packages/labgrid/step.py", line 187, in stop
    steps.pop(self)
  File "/home/testing/venv/lib/python3.10/site-packages/labgrid/step.py", line 28, in pop
    assert self._stack[-1] is step
```

Not sure if it is only related to the SSHDriver or not. When only using 1 thread, there are no issues.

Bastian-Krause:
This is untested and not the recommended approach. #835 allows running a process on a target asynchronously, so this is what you want to use.

S4mw1s3:
Ok. Thanks for the suggestion :)

Issue #604, opened 2020-05-12
Can't simply run pytest in thread


MaroMetelski:
Hi,
While trying to invoke pytest in a thread with labgrid plugin in use
RuntimeError is raised.
Maybe the exeption should be catched in the module, or the module optional?

Can be worked around by adding this to the thread target function:
```python
loop = asyncio.new_event_loop()
    asyncio.set_event_loop(loop)
```
Example code:
```python
from threading import Thread
import pytest

def tt():
    pytest.main(['-s', f'--lg-env=../qemu.yaml'])

t = Thread(target=tt, args=())
t.start()
t.join()
```
The line at fault:
https://github.com/labgrid-project/labgrid/blob/d11a5ecc86decd74d3ff86b6fb8b730c9ea4a962/labgrid/remote/client.py#L30

The error log:
```
Exception in thread Thread-1:
Traceback (most recent call last):
  File "/usr/lib/python3.7/threading.py", line 926, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.7/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "hm.py", line 5, in tt
    pytest.main(['-s', f'--lg-env=../qemu.yaml'])
  File "/home/marek/.local/lib/python3.7/site-packages/_pytest/config/__init__.py", line 105, in main
    config = _prepareconfig(args, plugins)
  File "/home/marek/.local/lib/python3.7/site-packages/_pytest/config/__init__.py", line 258, in _prepareconfig
    pluginmanager=pluginmanager, args=args
  File "/home/marek/.local/lib/python3.7/site-packages/pluggy/hooks.py", line 286, in __call__
    return self._hookexec(self, self.get_hookimpls(), kwargs)
  File "/home/marek/.local/lib/python3.7/site-packages/pluggy/manager.py", line 93, in _hookexec
    return self._inner_hookexec(hook, methods, kwargs)
  File "/home/marek/.local/lib/python3.7/site-packages/pluggy/manager.py", line 87, in <lambda>
    firstresult=hook.spec.opts.get("firstresult") if hook.spec else False,
  File "/home/marek/.local/lib/python3.7/site-packages/pluggy/callers.py", line 203, in _multicall
    gen.send(outcome)
  File "/home/marek/.local/lib/python3.7/site-packages/_pytest/helpconfig.py", line 90, in pytest_cmdline_parse
    config = outcome.get_result()
[...]

PR #1294, opened 2023-11-13
step: allow out of order pop

nvincent-vossloh:
If pytest access labgrid resources from various threads, the step event may pop out of push order.
Allow the pop function to extract an element if it is not in the first place.

### DIFF
--- a/labgrid/step.py
+++ b/labgrid/step.py
@@ -26,8 +26,8 @@ class Steps:
         step.level = len(self._stack)
 
     def pop(self, step):  # pylint: disable=redefined-outer-name
-        assert self._stack[-1] is step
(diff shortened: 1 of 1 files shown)


Emantor:
Hi,
threading is currently not supported with labgrid. Aside from steps being popped out of order, there are other places in the code that assume that no threads are used.

nvincent-vossloh:
> Hi, threading is currently not supported with labgrid. Aside from steps being popped out of order, there are other places in the code that assume that no threads are used.

I agree, that PR is a small step towards threading support in labgrid.

Bastian-Krause:
See also #980, #604.

PR #1295, opened 2023-11-13
managed file: adds api to fetch a file

nvincent-vossloh:
**description**
Use `ManagedFile` to be able to pull file from resource, it is useful when a binary on the resource creates a file which needs to be accessible on the client side.
I tested this feature in my environment which uses labgrid.

**Checklist**
- [ ] Tests for the feature 
- [ ] PR has been tested

I could need help regarding adding a test for this feature and which documentations files needs to be updated.

Thanks.

### DIFF
--- a/labgrid/util/managedfile.py
+++ b/labgrid/util/managedfile.py
@@ -39,12 +39,22 @@ class ManagedFile:
     detect_nfs = attr.ib(default=True, validator=attr.validators.instance_of(bool))
 
     def __attrs_post_init__(self):
-        if not os.path.isfile(self.local_path):
-            raise FileNotFoundError(f"Local file {self.local_path} not found")
         self.logger = logging.getLogger(f"{self}")
         self.hash = None
-        self.rpath = None
         self._on_nfs_cached = None
+
+    @property
+    def rpath(self):
+        if isinstance(self.resource, NetworkResource):
+            host = self.resource.host
+            conn = sshmanager.open(host)
+            if self._on_nfs(conn):
+                return os.path.dirname(self.local_path) + "/"
+            else:
+                return f"{self.get_user_cache_path()}/{self.get_hash()}/"
+        else:
+            return None
+
 
     def sync_to_resource(self, symlink=None):
         """sync the file to the host specified in a resource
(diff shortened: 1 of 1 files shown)


Emantor:
Hi,

all drivers which created files on the exporter side so far took care of either opening a channel back to the client, i.e. via SSH and pipes or sync the requested data back automatically. I don't see why we need a generic mechanism which allows syncing files back to the client, the driver should already take care of that. This also extends the `ManagedFile` scope to be bidirectional.

nvincent-vossloh:
> Hi,
> 
> all drivers which created files on the exporter side so far took care of either opening a channel back to the client, i.e. via SSH and pipes or sync the requested data back automatically. I don't see why we need a generic mechanism which allows syncing files back to the client, the driver should already take care of that. This also extends the `ManagedFile` scope to be bidirectional.

Hi,

Thanks for your reply, I may have misunderstood something.
My use case is a [custom driver](https://github.com/SiemaApplications/labgrid/blob/dda69a2515d18396816a97f63ebcc63bcae8afe4/labgrid/driver/stlinkdriver.py#L60) to manage st-link probe, I wish to dump the firmware on the MCU connected to the probe, therefore the file is created on the remote host by the probe and I want to get the dumped firmware to the client, in that sense it is more a fetch rather than a sync since the file was never on the client in the first place.

Bastian-Krause:
You could open the new file locally and make `STM32_Programmer_CLI` write to stdout and pass the file handle as `stdout=` in `subprocess.Popen()`. No need for any additional file transfer, no tmpfiles. See `start_record()` in #1157 as an example.

nvincent-vossloh:
> You could open the new file locally and make `STM32_Programmer_CLI` write to stdout and pass the file handle as `stdout=` in `subprocess.Popen()`. No need for any additional file transfer, no tmpfiles. See `start_record()` in #1157 as an example.

Thank you for the tip, I will look into it and update this issue.

Emantor:
If you have additional questions or require feedback, either open a new PR or [join the chat](https://app.element.io/#/room/#labgrid:matrix.org).

PR #1303, opened 2023-11-29
contrib: add helper to graph current configuration

liambeguin:
## Description

Add a sub-command to sync-places to generate a graphviz graph of the resources and places in a given environment.

A visual representation can help make sense of a large amount of information quickly.
This tried to put all relevant information in one place without being overly complicated.

I also tired to use a color blind friendly palette, not sure how that turned out.. 

I'm very open to modifying this based on feedback and recommendations.

I also tried to put this into a separate tool using fastapi to have the graph rendered in a browser directly, but ran into trouble. If that approach is preferred I can also push that instead, and maybe with a bit of support we can get it to work consistently.

## Testing
```console
$ ./labgrid/contrib/sync-places.py graph | dot -Tsvg -o labgrid-graph.svg
```
By default the command will just output a dot graph to stdout, piping that to `dot` we can generate images.

As noted in the docstring, this will generate a graph with:
- all resources, grouped by groupname and exporter.
- all places, with a list of tags (a darker shade of blue, if the place is acquired)
- solid edges between places and acquired resources
- dotted edges between places and unacquired resources
- edges between resources and places carry the match name if any.

here's a simple example of what the output looks like:
image

## Checklist
- [x] Documentation for the feature (documented in the docstring)
- [x] Tests for the feature 
- [x] PR has been tested

### DIFF
--- a/contrib/README.rst
+++ b/contrib/README.rst
@@ -1,0 +1,41 @@
+labgrid-webapp
+==============
+
+labgrid-webapp implements a browser interface to access some of labgrid's
+information.
+
+Quick Start
+-----------
+
+.. code-block:: bash
+
+   $ cd labgrid/
+   $ source venv/bin/activate
+   venv $ pip install -r contrib/requirements-webapp.txt
+   venv $ ./contrib/labgrid-webapp --help
+   usage: labgrid-webapp [-h] [--crossbar URL] [--port PORT] [--proxy PROXY]
+
+   Labgrid webapp
+
+   options:
+     -h, --help            show this help message and exit
+     --crossbar URL, -x URL
+                           Crossbar websocket URL (default: ws://127.0.0.1:20408/ws)
+     --port PORT           Port to serve on
+     --proxy PROXY, -P PROXY
+
+   venv $ ./contrib/labgrid-webapp --help
+   INFO:     Available routes:
+   INFO:       - /labgrid/graph
+   INFO:     Started server process [2378028]
+   INFO:     Waiting for application startup.
+   INFO:     Application startup complete.
+   INFO:     Uvicorn running on http://0.0.0.0:8800 (Press CTRL+C to quit)
+   ...
+
+Please note that the graph feature relies on a valid `graphviz` system
+installation.
+
+By default the application will start on port 8800.
+
+See http://0.0.0.0:8800/docs for more information on available endpoints.
--- a/contrib/labgrid-webapp
+++ b/contrib/labgrid-webapp
@@ -1,0 +1,160 @@
+#!/usr/bin/env python3
+import argparse
+import logging
+import os
+import sys
+from typing import Dict
+
+import graphviz
+import uvicorn
+from fastapi import FastAPI
+from fastapi.responses import Response
+
+from labgrid.remote.client import ClientSession, start_session
+from labgrid.remote.common import Place
+from labgrid.resource import Resource
+from labgrid.util.proxy import proxymanager
+
+
+async def do_graph(session: ClientSession) -> bytes:
+    '''Generate a graphviz graph of the current configuration.
+
+    Graph displays:
+    - all resources, grouped by groupname and exporter.
+    - all places, with a list of tags
+    - solid edges between places and acquired resources
+    - dotted edges between places and unacquired resources
+    - edges between resources and places carry the match name if any.
+    '''
+    def res_node_attr(name: str, resource: Resource) -> Dict[str, str]:
+        return {
+            'shape': 'plaintext',
+            'label': f'''<
+            <table bgcolor="peru">
+                <tr>
+                    <td border="0" align="left">Resource</td>
+                </tr>
+                <tr>
+                    <td port="cls">{resource.cls}</td>
+                    <td port="name" bgcolor="white">{name}</td>
+                </tr>
+            </table>>''',
+        }
+
+    def place_node_attr(name: str, place: Place) -> Dict[str, str]:
+        acquired = ''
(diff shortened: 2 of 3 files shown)


Bastian-Krause:
Does this actually use anything from the `sync-places.py`? Maybe it would be better to have this as a separate script in `crontrib/` ?

liambeguin:
> Does this actually use anything from `sync-places.py`?

Not much apart from the "setup". I thought it kind of made sense to group, but I'm happy to separate it. I'll update the PR

liambeguin:
I ended up finding a solution to share the event loop between fastapi and labgrid so I also updated the script to serve the graph in a browser.

contrib/sync-places.py
```diff
@@ -109,6 +109,13 @@ async def do_sync(session, args):
                     changed = True
 
             tags = config["places"][name].get("tags", {}).copy()
+            for k, v in tags.items():
+                if not isinstance(k, str):
+                    tags[str(k)] = v
+                    del(tags[k])
+                if not isinstance(v, str):
+                    tags[k] = str(v)
+
```
Bastian-Krause, line 118: Now that the script resides separated in graph-places.py, this should be a separate PR.

contrib/graph-places.py
```diff
@@ -0,0 +1,154 @@
+import argparse
```
Bastian-Krause, line 1: Since this is meant to be run as a script, we should add a shebang and set the executable bit.

pyproject.toml
```diff
@@ -79,6 +79,10 @@ deb = [
     # labgrid[snmp]
     "pysnmp-lextudio>=4.4.12",
 ]
+contrib = [
+    "fastapi>=0.104.1",
+    "uvicorn>=0.24.0",
+]
```
Bastian-Krause, line 85: I don't think we should add dependencies for contrib scripts in our pyproject.toml. Better document this, either in a module doc string in the script itself or maybe add a README in contrib?

liambeguin: Okay, I'll add a requirements file in contrib/ with a readme.

A colleague also mentioned that this could be extended to serve as a web interface for users that aren't too familiar with command line interfaces. I might rename the script so that other routes can be added later.

Bastian-Krause (approved):

contrib/labgrid-webapp
```diff
@@ -0,0 +1,160 @@
...
+    '''Generate a graphviz graph of the current configuration.
+
+    Graph displays:
+    - all resources, grouped by groupname and exporter.
+    - all places, with a list of tags
+    - solid edges between places and acquired resources
+    - dotted edges between places and unacquired resources
+    - edges between resources and places carry the match name if any.
+    '''
+    def res_node_attr(name: str, resource: Resource) -> dict[str, str]:
```
PaulVittorino, line 29: I got the following error when running the code:

>   File "./contrib/labgrid-webapp", line 180, in get_graph
>     svg = await do_graph(session)
>   File "./contrib/labgrid-webapp", line 35, in do_graph
>     def res_node_attr(name: str, resource: Resource) -> dict[str, str]:
> TypeError: 'type' object is not subscriptable


Adding single quotes around the type annotation resolved it.
`def res_node_attr(name: str, resource: Resource) -> 'dict[str, str]':`

contrib/labgrid-webapp
```diff
@@ -0,0 +1,160 @@
...
+                    <td border="0" align="left">Resource</td>
+                </tr>
+                <tr>
+                    <td port="cls">{resource.cls}</td>
+                    <td port="name" bgcolor="white">{name}</td>
+                </tr>
+            </table>>''',
+        }
+
+    def place_node_attr(name: str, place: Place) -> dict[str, str]:
```
PaulVittorino, line 44: Same issue as above

`def place_node_attr(name: str, place: Place) -> 'dict[str, str]':`

contrib/README.rst
```diff
@@ -0,0 +1,38 @@
+labgrid-webapp
+==============
+
+labgrid-webapp implements a browser interface to access some of labgrid's
+information.
+
```
PaulVittorino, line 6: Should the `graphviz` command line tool requirement be mentioned?

contrib/labgrid-webapp
liambeguin: This comes from using different versions of Python. I'll update to use `typing.Dict` instead

contrib/README.rst
liambeguin: the `graphviz` library is listed in the `requirements-webapp.txt` file, with some instructions on how to get things running. would you want to see an explicit mention of the library? I'm not sure it's really useful, but we could add it.

PaulVittorino: Providing the requirements file made running the application so simple I had forgotten about the `graphviz` Python library. My thought was explicitly calling out the system dependency may help future users.

liambeguin: I might be missing something, but I don't think there's a dependency on the `graphviz` system package. Did you have to install it to get things working? I might've had it installed already.

PaulVittorino: https://graphviz.readthedocs.io/en/stable/manual.html does mention needing an installation of Graphviz. I got a very lengthy error from the `graphviz` Python library when I tried viewing the graph web page before I installed Graphviz.

liambeguin: Oh! thanks for pointing that out. I added a note in the README.

liambeguin:
I just rebased this MR on the latest master. If there are no other comments I think this would be ready to merge.

Bastian-Krause (approved):

PR #1304, opened 2023-12-04
driver/qemudriver: add support for virt machine

johannwie:
The virt machine type is commonly used, when a hardware independent environment is required. Adding virt support to the disk image option, makes it easier to handle frequently changing disk images.

### DIFF
--- a/labgrid/driver/qemudriver.py
+++ b/labgrid/driver/qemudriver.py
@@ -168,6 +168,11 @@ class QEMUDriver(ConsoleExpectMixin, Driver, PowerProtocol, ConsoleProtocol):
                 cmd.append(
                     f"if=virtio,format={disk_format},file={disk_path}{disk_opts}")
                 boot_args.append("root=/dev/vda rootwait")
+            elif self.machine == "virt":
+                cmd.append("-drive")
+                cmd.append(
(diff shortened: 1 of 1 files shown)


labgrid/driver/qemudriver.py
```diff
@@ -168,6 +168,11 @@ def get_qemu_base_args(self):
                 cmd.append(
                     f"if=virtio,format={disk_format},file={disk_path}{disk_opts}")
                 boot_args.append("root=/dev/vda rootwait")
+            elif self.machine == "virt":
+                cmd.append("-drive")
+                cmd.append(
+                    f"if=virtio,format={disk_format},file={disk_path}{disk_opts}")
+                boot_args.append("root=/dev/vda rootwait")
```
Bastian-Krause, line 175: As far as I can see, this is exactly what we do for the "pc" case. Do you expect any differences in the future from "pc"? If not, we could handle "pc" and "virt" in the same clause.

Bastian-Krause: Okay, let's keep it that way for now. We can always create a clean up PR on  top.

Bastian-Krause (approved):

jluebbe (approved):

PR #1305, opened 2023-12-12
driver/sshdriver: Add credential overriding NetworkService's

nlabriet:
**Description**

This is another approche than #1291.

Here, username and password properties are added to `SSHDriver` and used instead of username and password coming from `NetworkService`.

This will answer my original issue:
> Depending on the software provisioned on a place, credentials may be different. It's my case where 3 different credentials are needed (user1/pass1 user2/pass2 user3/pass3) when soft1 or soft2 or soft3 is on the target.

`ShellDriver` and `SSHDriver` will both hold credentials in the environment configuration, where the user can adapt then to its use case.

**Checklist**
- [X] PR has been tested locally with `ssh` and `scp`

Closes #1291


Bastian-Krause (approved):
Looks good to me. I made the getter functions a bit more pythonic.

Emantor (approved):

jluebbe (approved):

Issue #917, opened 2022-05-16
arm docker images


obbardc:
hey!
Seems like the docker images posted to https://hub.docker.com/u/labgrid are amd64 only. Is it possible to push (at least) arm64 images too ?

jluebbe:
They are built using https://github.com/labgrid-project/labgrid/blob/master/.github/workflows/docker.yml.

As we deploy labgrid differently (using Debian packages built via CI or via meta-labgrid), we don't really use the Docker images much and would appreciate contributions there. I'm not sure if they should be build via qemu emulation or if there is a better way. @JPEWdev might have some ideas, as he worked on the images some time ago.

jluebbe:
Thanks.

Please note that for the exporter, you need to allow access to `/dev` and udev.

Bastian-Krause:
Resolved by #917.

PR #1307, opened 2023-12-14
Publish arm64 Docker images

hoyes:
Address #917 by adding support for building and publishing arm64 Docker images on CI.

The most direct way to do this seems to be docker buildx, using Github actions provider by Docker, but buildx isn't able to push multi-platform images to a local image store so some supporting changes are required to build.sh:
  1. Allow additional arguments to be passed
  2.  Use the same image names as on Docker Hub

Note that 1. changes usage for build.sh when building for a specific platform. This is addressed in README.md and hopefully addresses more use-cases, but not sure about Labgrid policy about breaking changes?

Building non-x86 platforms with Github Actions uses emulation, so enable layer caching and tweak the Dockerfile to make better use of caches.

There is an additional patch to enable image builds on release events and manual triggers.

I have run a action on top of these changes (using the manual trigger) over at my fork: https://github.com/hoyes/labgrid/actions/runs/7209149336, resulting in https://hub.docker.com/u/hoyes

**Checklist**
- [x] Tests for the feature 
- [x] PR has been tested

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -5,6 +5,9 @@ on:
     branches: [ master ]
     tags:
       - '*'
+  release:
+    types: [ released ]
+  workflow_dispatch:
 
 env:
     QEMU_PLATFORMS: arm64
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -6,6 +6,11 @@ on:
     tags:
       - '*'
 
+env:
+    QEMU_PLATFORMS: arm64
+    IMAGE_PLATFORMS: linux/amd64,linux/arm64
+    IMAGE_PREFIX: ${{ secrets.DOCKERHUB_PREFIX }}
+
 jobs:
   docker:
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


brherger:
Thanks for adding the tags.  I was going to ask for this.

hoyes:
@krevsbech any thoughts on these Docker build changes?

Emantor:
This only seems to work with docker, podman raises an error when trying to use the build script:
```bash
./dockerfiles/build.sh --platform linux/amd64,linux/arm64
Error: unable to parse value provided `[]` to --cache-from: repository must contain neither a tag nor digest: docker.io/labgrid/client:latest
Error: unable to parse value provided `[]` to --cache-from: repository must contain neither a tag nor digest: docker.io/labgrid/exporter:latest
Error: unable to parse value provided `[]` to --cache-from: repository must contain neither a tag nor digest: docker.io/labgrid/coordinator:latest
```

hoyes:
It seems caching doesn't work consistently between Podman and Docker.

If it's OK, I've dropped the commit that adds the caching arguments. I tried to enable caching just in the CI workflow but `--cache-from` requires the specific image name per image (client, coordinator, expertor) which is tricky to pass through. The workflow takes around 15 minutes without it (https://github.com/hoyes/labgrid/actions/runs/8469983487), which isn't _too_ bad.

`./dockerfiles/build.sh --platform linux/amd64,linux/arm64` now works for me with Podman.

jluebbe:
I'm not a container expert, but this looks to be an improvement to me, so let's get it merged. If this causes problems, hopefully one of the container users can help. :)

jluebbe (approved):

PR #1308, opened 2023-12-14
shelldriver: allow login with empty password

### LINKED ISSUE #441: Fix Login situations which require an empty password
Labgrid can currently not deal with empty passwords, due to `""` and `None` being handled the same. `None` should indicate that no password is required and `""` should indicate that an empty password should be used.

rohieb:
**Description**

Fixes #441

Make `password` an optional field, so when is not set, it defaults to `None`. If we then get a password prompt on login, we can still raise the Exception. Otherwise it is now possible to set the password to an empty string explicitly.

**Checklist**
- [ ] Tests for the feature  (not sure how to test this…?)
- [x] PR has been tested

### DIFF
--- a/labgrid/driver/shelldriver.py
+++ b/labgrid/driver/shelldriver.py
@@ -48,7 +48,7 @@ class ShellDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
     prompt = attr.ib(validator=attr.validators.instance_of(str))
     login_prompt = attr.ib(validator=attr.validators.instance_of(str))
     username = attr.ib(validator=attr.validators.instance_of(str))
-    password = attr.ib(default="", validator=attr.validators.instance_of(str))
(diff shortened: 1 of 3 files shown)


labgrid/driver/shelldriver.py
```diff
@@ -152,7 +152,7 @@ def _await_login(self):
                 did_login = True
 
             elif index == 2:
-                if self.password:
+                if type(self.password) == str:
```
Emantor, line 155: `if self.password is not None:`

Bastian-Krause:
> Make `password` an optional field, so when is not set, it defaults to `None`. If we then get a password prompt on login, we can still raise the Exception. Otherwise it is now possible to set the password to an empty string explicitly.

`password` is already an optional attribute. The commit message is a little misleading.

Emantor (approved):

jluebbe (approved):

PR #1312, opened 2023-12-30
contrib: sync-places: convert tags into strings

liambeguin:
**Description**

When settings tags from the command line, key, value pairs are always strings. Make sure this is the same when reading a YAML file. It can easily be done in the YAML file itself, but this makes the syntax a bit lighter.

**Checklist**
- [x] Tests for the feature 
- [x] PR has been tested

This was originally a part of #1303, but was split up now that that PR is editing another file.

### DIFF
--- a/contrib/sync-places.py
+++ b/contrib/sync-places.py
@@ -109,6 +109,11 @@ def main():
                     changed = True
 
             tags = config["places"][name].get("tags", {}).copy()
+            for k, v in tags.items():
+                if not isinstance(k, str) or not isinstance(v, str):
+                    del(tags[k])
+                    tags[str(k)] = str(v)
+
             if place_tags != tags:
                 print(
                     "Setting tags for place %s to %s"


Bastian-Krause:
@JoshuaWatt Could you please review this one?

contrib/sync-places.py
```diff
@@ -109,6 +109,13 @@ async def do_sync(session, args):
                     changed = True
 
             tags = config["places"][name].get("tags", {}).copy()
+            for k, v in tags.items():
+                if not isinstance(k, str):
+                    tags[str(k)] = v
+                    del(tags[k])
```
JoshuaWatt, line 115: There is a hole in this logic if both the key and value are not strings, in which case the `tags[k] = str(v)` will re-add the non-string key.

JoshuaWatt (changes requested):
Otherwise LGTM

contrib/sync-places.py
liambeguin: oh! good catch. Should be fixed now

JoshuaWatt: `del()` needs to be before the assignment, otherwise it will remove the key if it's already a string :)

JoshuaWatt (approved):

Bastian-Krause (approved):

PR #1316, opened 2024-01-10
labgrid/driver/power: add poe-mib backend

Emantor:
**Description**
The PoE MiB backend uses SNMP enable and disable PoE on a switch port. Compatibility was tested on a Cisco CBS350.

**Checklist**
- [x] Tests for the feature 
- [x] PR has been tested


jluebbe (approved):

labgrid/driver/power/poe_mib.py
```diff
@@ -0,0 +1,27 @@
+""" tested with Cisco CBS350, should be compatible with switches implementing the PoE administration MiB"""
+
+from ..exception import ExecutionError
+from ...util.snmp import SimpleSNMP
+
+OID = "1.3.6.1.2.1.105.1.1.1.3.1"
+
+def power_set(host, port, index, value):
+    _snmp = SimpleSNMP(host, 'private')
```
Bastian-Krause, line 9: Shouldn't we pass the port here?

labgrid/driver/power/poe_mib.py
```diff
@@ -0,0 +1,27 @@
...
+    outlet_control_oid = "{}.{}".format(OID, index)
+
+    oid_value = "1" if value else "2"
+
+    _snmp.set(outlet_control_oid, oid_value)
+
+def power_get(host, port, index):
+    _snmp = SimpleSNMP(host, 'private')
```
Bastian-Krause, line 17: Shouldn't we pass the port here?

labgrid/driver/power/poe_mib.py
```diff
@@ -0,0 +1,27 @@
+""" tested with Cisco CBS350, should be compatible with switches implementing the PoE administration MiB"""
+
+from ..exception import ExecutionError
+from ...util.snmp import SimpleSNMP
+
+OID = "1.3.6.1.2.1.105.1.1.1.3.1"
+
+def power_set(host, port, index, value):
+    _snmp = SimpleSNMP(host, 'private')
+    outlet_control_oid = "{}.{}".format(OID, index)
```
Bastian-Krause, line 10: Could be an f-string.

labgrid/driver/power/poe_mib.py
```diff
@@ -0,0 +1,27 @@
...
+
+    oid_value = "1" if value else "2"
+
+    _snmp.set(outlet_control_oid, oid_value)
+
+def power_get(host, port, index):
+    _snmp = SimpleSNMP(host, 'private')
+    output_status_oid = "{}.{}".format(OID, index)
```
Bastian-Krause, line 18: Could be an f-string.

labgrid/driver/power/poe_mib.py
Emantor: Since the devices are network switches where SNMP may already be enabled, I don't see the value in a configurable port.

labgrid/driver/power/poe_mib.py
Emantor: Yeah, do you want it to be one? I don't care either way tbh.

labgrid/driver/power/poe_mib.py
Bastian-Krause: You can configure the "snmp-agent-port" on the Cisco switches, so it should also be possible to configure it on labgrid's side. The other power backends using SNMP allow the same.

labgrid/driver/power/poe_mib.py
Bastian-Krause: Let's leave it as-is.

jluebbe (approved):

Bastian-Krause (approved):

PR #1320, opened 2024-01-19
Rawnetworkinterface stream

JoshuaWatt:
**Description**

Adds support for live streaming packets from a RawNetworkInterfaceDriver. This can be useful for writing tests that parse packets in real time looking for a specific pattern, for example:

```python
import dpkt
    
drv = target.get_driver("RawNetworkInterfaceDriver")
    
with drv.record("-", timeout=60) as p:
    pcap = dpkt.pcap.Reader(p.stdout)
    for timestamp, buf in pcap:
        eth = dpkt.ethernet.Ethernet(buf)

        if _some_criteria_:
            break
    else:
        assert False, "Packet not found!"
```

**Checklist**
- [ ] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/helpers/labgrid-raw-interface
+++ b/helpers/labgrid-raw-interface
@@ -55,13 +55,13 @@ def main(program, ifname, count):
 
     if program == "tcpreplay":
         args.append(f"--intf1={ifname}")
-        args.append('-')
+        args.append("-")
 
     if program == "tcpdump":
         args.append("-n")
         args.append(f"--interface={ifname}")
         args.append("-w")
-        args.append('-')
+        args.append("-")
 
         if count:
             args.append("-c")
@@ -75,22 +75,17 @@ def main(program, ifname, count):
 
 if __name__ == "__main__":
     parser = argparse.ArgumentParser()
-    parser.add_argument(
-        '-d',
-        '--debug',
-        action='store_true',
-        default=False,
-        help="enable debug mode"
-    )
-    parser.add_argument('program', type=str, help='program to run, either tcpreplay or tcpdump')
-    parser.add_argument('interface', type=str, help='interface name')
-    parser.add_argument('count', nargs="?", type=int, default=None, help='amount of frames to capture while recording')
+    parser.add_argument("-d", "--debug", action="store_true", default=False, help="enable debug mode")
+    parser.add_argument("program", type=str, help="program to run, either tcpreplay or tcpdump")
+    parser.add_argument("interface", type=str, help="interface name")
+    parser.add_argument("count", nargs="?", type=int, default=None, help="amount of frames to capture while recording")
     args = parser.parse_args()
     try:
         main(args.program, args.interface, args.count)
     except Exception as e:  # pylint: disable=broad-except
         if args.debug:
             import traceback
+
             traceback.print_exc(file=sys.stderr)
         print(f"ERROR: {e}", file=sys.stderr)
         exit(1)
--- a/helpers/labgrid-raw-interface
+++ b/helpers/labgrid-raw-interface
@@ -32,18 +32,18 @@ def get_denylist():
     return denylist
 
 
-def main(program, ifname, count):
-    if not ifname:
+def main(program, options):
+    if not options.ifname:
         raise ValueError("Empty interface name.")
-    if any((c == "/" or c.isspace()) for c in ifname):
-        raise ValueError(f"Interface name '{ifname}' contains invalid characters.")
-    if len(ifname) > 16:
-        raise ValueError(f"Interface name '{ifname}' is too long.")
+    if any((c == "/" or c.isspace()) for c in options.ifname):
+        raise ValueError(f"Interface name '{options.ifname}' contains invalid characters.")
+    if len(options.ifname) > 16:
+        raise ValueError(f"Interface name '{options.ifname}' is too long.")
 
     denylist = get_denylist()
 
-    if ifname in denylist:
-        raise ValueError(f"Interface name '{ifname}' is denied in denylist.")
+    if options.ifname in denylist:
+        raise ValueError(f"Interface name '{options.ifname}' is denied in denylist.")
 
     programs = ["tcpreplay", "tcpdump"]
     if program not in programs:
@@ -54,18 +54,18 @@ def main(program, ifname, count):
     ]
 
     if program == "tcpreplay":
-        args.append(f"--intf1={ifname}")
+        args.append(f"--intf1={options.ifname}")
         args.append("-")
 
     if program == "tcpdump":
         args.append("-n")
-        args.append(f"--interface={ifname}")
+        args.append(f"--interface={options.ifname}")
         args.append("-w")
         args.append("-")
 
-        if count:
+        if options.count:
             args.append("-c")
-            args.append(str(count))
+            args.append(str(options.count))
 
     try:
         os.execvp(args[0], args)
@@ -76,12 +76,20 @@ def main(program, ifname, count):
 if __name__ == "__main__":
     parser = argparse.ArgumentParser()
     parser.add_argument("-d", "--debug", action="store_true", default=False, help="enable debug mode")
-    parser.add_argument("program", type=str, help="program to run, either tcpreplay or tcpdump")
-    parser.add_argument("interface", type=str, help="interface name")
-    parser.add_argument("count", nargs="?", type=int, default=None, help="amount of frames to capture while recording")
+    subparsers = parser.add_subparsers(dest="program", help="program to run")
+
+    # tcpdump
+    tcpdump_parser = subparsers.add_parser("tcpdump")
+    tcpdump_parser.add_argument("ifname", type=str, help="interface name")
+    tcpdump_parser.add_argument("count", type=int, default=None, help="amount of frames to capture while recording")
+
+    # tcpreplay
+    tcpreplay_parser = subparsers.add_parser("tcpreplay")
+    tcpreplay_parser.add_argument("ifname", type=str, help="interface name")
+
     args = parser.parse_args()
     try:
-        main(args.program, args.interface, args.count)
+        main(args.program, args)
     except Exception as e:  # pylint: disable=broad-except
         if args.debug:
             import traceback
--- a/helpers/labgrid-raw-interface
+++ b/helpers/labgrid-raw-interface
@@ -67,6 +67,14 @@ def main(program, options):
             args.append("-c")
             args.append(str(options.count))
 
+        if options.timeout:
+            # The timeout is implemented by specifying the number of seconds before rotating the
+            # dump file, but limiting the number of files to 1
+            args.append("-G")
+            args.append(str(options.timeout))
+            args.append("-W")
+            args.append("1")
+
     try:
         os.execvp(args[0], args)
     except FileNotFoundError as e:
@@ -82,6 +90,9 @@ if __name__ == "__main__":
     tcpdump_parser = subparsers.add_parser("tcpdump")
     tcpdump_parser.add_argument("ifname", type=str, help="interface name")
     tcpdump_parser.add_argument("count", type=int, default=None, help="amount of frames to capture while recording")
+    tcpdump_parser.add_argument(
+        "--timeout", type=int, default=None, help="Amount of time to capture while recording. 0 means capture forever"
+    )
 
     # tcpreplay
     tcpreplay_parser = subparsers.add_parser("tcpreplay")
--- a/helpers/labgrid-raw-interface
+++ b/helpers/labgrid-raw-interface
@@ -60,6 +60,10 @@ def main(program, options):
     if program == "tcpdump":
         args.append("-n")
         args.append(f"--interface={options.ifname}")
+        # Write out each packet as it is received
+        args.append("--packet-buffered")
+        # Capture complete packets (for compatibility with older tcpdump versions)
+        args.append("--snapshot-length=0")
         args.append("-w")
         args.append("-")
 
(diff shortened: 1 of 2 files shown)


JoshuaWatt:
Of note, we've had a local driver that was very similar to `RawNetworkInterfaceDriver` for some time that was used to stream packets; the intention of this patch is to bring it in line with the functionality in that driver so we can eliminate it. For posterity, here is our driver:

```python
import attr
import subprocess
import dpkt
from contextlib import contextmanager

from labgrid import target_factory
from labgrid.driver import Driver


@target_factory.reg_driver
@attr.s(eq=False)
class InterfaceCaptureDriver(Driver):
    """Driver to obtain tcpdump output from device's network"""

    bindings = {
        "iface": {"RemoteNetworkInterface", "USBNetworkInterface", "NetworkInterface"},
    }

    @contextmanager
    def live_capture(self, timeout=None):
        """
        Perform live capture of traffic from device's network.

        :param int timeout: When specified, performs capture for ``timeout`` seconds. \
            Defaults to `None`.

        :yields: A `dpkt.pcap.Reader` object of the captured network traffic
        """
        extra = getattr(self.iface, "extra", {})
        command = ["sudo", "tcpdump", "-i", self.iface.ifname, "-U", "-s0", "-w", "-"]

        if timeout is not None:
            command.extend(["-G", str(timeout), "-W", "1"])

        if "proxy_required" in extra:
            command = ["ssh", extra["proxy"], "--"] + command

        with subprocess.Popen(command, stdout=subprocess.PIPE) as p:
            yield dpkt.pcap.Reader(p.stdout)

```

helpers/labgrid-raw-interface
```diff
@@ -60,6 +60,8 @@ def main(program, ifname, count, duration):
     if program == "tcpdump":
         args.append("-n")
         args.append(f"--interface={ifname}")
+        args.append("-U")
+        args.append("-s0")
```
Bastian-Krause, line 64: Why do we need this? If it is needed, please also use its longopt.

helpers/labgrid-raw-interface
```diff
@@ -60,6 +60,8 @@ def main(program, ifname, count, duration):
     if program == "tcpdump":
         args.append("-n")
         args.append(f"--interface={ifname}")
+        args.append("-U")
```
Bastian-Krause, line 63: Please use the longopt `--packet-buffered` instead.

helpers/labgrid-raw-interface
```diff
@@ -85,9 +96,10 @@ if __name__ == "__main__":
     parser.add_argument('program', type=str, help='program to run, either tcpreplay or tcpdump')
     parser.add_argument('interface', type=str, help='interface name')
     parser.add_argument('count', nargs="?", type=int, default=None, help='amount of frames to capture while recording')
+    parser.add_argument('--duration', type=int, default=None, help="Amount of time to capture while recording")
```
Bastian-Krause, line 99: Is there a reason for calling this "duration" instead of "timeout"?

labgrid/driver/rawnetworkinterfacedriver.py
```diff
@@ -69,20 +70,21 @@ def start_record(self, filename, *, count=None):
...
+        if filename == "-":
+            self._record_handle = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.DEVNULL)
+        else:
+            with open(filename, "wb") as outdata:
+                self._record_handle = subprocess.Popen(cmd, stdout=outdata, stderr=subprocess.PIPE)
         return self._record_handle
 
     @Driver.check_active
-    @step(args=["timeout"])
     def stop_record(self, *, timeout=None):
```
Bastian-Krause, line 85: Why do you keep timeout here? For backwards compatibility? I'm unsure whether we should do that.

Bastian-Krause (changes requested):
Thanks! Looks good overall.

In the commit message of d68686aa ("rawnetworkinterfacedriver: Implement live streaming"):
```diff
-with drv.record("-", timeout=60) as p
+with drv.record("-", timeout=60) as p:
```

helpers/labgrid-raw-interface
```diff
@@ -67,6 +67,15 @@ def main(program, ifname, count):
...
             args.append(str(count))
 
+        if duration:
+            # The duration is implemented by specifying the number of seconds
+            # before rotating the dump file, but limiting the number of files
+            # to 1
+            args.append("-G")
+            args.append(str(duration))
+            args.append("-W")
+            args.append("1")
```
jluebbe, line 79: I suspect this might fail in the future when used in combination with writing to stdout. Handling that in the helper would mean using a subprocess instead `os.execvp`, so that can wait util that actually is a problem.

labgrid/driver/rawnetworkinterfacedriver.py
```diff
@@ -59,7 +59,7 @@ def start_record(self, filename, *, count=None, timeout=None):
         Starts tcpdump on bound network interface resource.
 
         Args:
-            filename (str): name of a file to record to
+            filename (str): name of a file to record to, or "-" to record to stdout
```
jluebbe, line 62: Perhaps use filename=None for that case? "-" isn't really a convention in Python.

Bastian-Krause:
@JoshuaWatt I'm interested in using the changes in this PR in our network tests. Could you have a look at the review comments? If you don't have the time to integrate the changes, I'd be willing to take this over.

helpers/labgrid-raw-interface
JoshuaWatt: duration was chosen because it's distinct from the `timeout` argument used in `rawnetworkinterface.py`. It can be confusing have the same name meaning different things in that file, so I thought "duration" was a more descriptive name

labgrid/driver/rawnetworkinterfacedriver.py
```diff
@@ -74,8 +74,11 @@ def start_record(self, filename, *, count=None, timeout=None):
             cmd.append("--duration")
             cmd.append(str(timeout))
         cmd = self._wrap_command(cmd)
-        with open(filename, "wb") as outdata:
-            self._record_handle = subprocess.Popen(cmd, stdout=outdata, stderr=subprocess.PIPE)
+        if filename == "-":
+            self._record_handle = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.DEVNULL)
```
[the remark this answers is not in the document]
JoshuaWatt: That's not a good idea (IMHO). If you make `stderr=subprocess.PIPE`, then every consumer of the subprocess _must_ periodically read from the stderr pipe, otherwise if it fills up it will deadlock the process. Making it mandatory would make the code consuming this have to do that, which is annoying. Better would be to make it optional, although this still has problems because if you do `stdout=subprocess.PIPE, stderr=subprocess.PIPE`, a consumer would need to `poll()` on both stderr and stdout to prevent deadlock. Anyway, given how tricky this is to get correct, I think leaving it is the correct answer for now. If an error occurs, the tcpdump error output will be visible in the logs/stderr/etc. which is sufficient to debug the problem without a huge burden on the consuming code.

labgrid/driver/rawnetworkinterfacedriver.py
JoshuaWatt: Done

helpers/labgrid-raw-interface
```diff
@@ -67,6 +67,15 @@ def main(program, ifname, count):
             args.append("-c")
             args.append(str(count))
 
+        if duration:
```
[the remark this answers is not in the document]
JoshuaWatt: Done

labgrid/driver/rawnetworkinterfacedriver.py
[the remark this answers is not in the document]
Bastian-Krause: My expectation was that tcpdump doesn't produce too much output on stderr, but maybe that's too naive. I'm okay with keeping it as is.

Bastian-Krause (commented):
Thanks for addressing the feedback. The changes look good to me. I'll try to test this within the next weeks and add my approval then.

Bastian-Krause (approved):
Sorry for the delay. I've now tested this successfully.

I'm basically using it the same way as suggested, but with [scapy](https://scapy.net/) instead of [dpkt](https://github.com/kbandla/dpkt):
```python
from scapy.all import PcapReader

drv = target.get_driver("RawNetworkInterfaceDriver")

with drv.record(None, count=1, timeout=5) as record:
        pcap = PcapReader(record.stdout)
        packet = pcap.next()
        ...
```

helpers/labgrid-raw-interface
```diff
@@ -32,7 +32,7 @@ def get_denylist():
     return denylist
 
 
-def main(program, ifname, count):
+def main(program, ifname, count, timeout=0):
```
jluebbe, line 35: Don't specify defaults in two locations (here and in argparse as None). Just drop the default here, as this function is always called with the full arguments.

jluebbe (changes requested):
Looks good otherwise.

labgrid/driver/rawnetworkinterfacedriver.py
```diff
@@ -97,17 +104,26 @@ def record(self, filename, *, count=None, timeout=None):
...
-            self.stop_record(timeout=timeout)
+            # If live streaming packets, there is no reason to wait for tcpdump
+            # to finish, so terminate it as soon as the context is left
+            try:
+                self.stop_record(timeout=0 if filename is None else None)
+            except subprocess.TimeoutExpired:
+                # A timeout is expected if the context is exited early while
+                # piping to stdout
+                if filename is not None:
+                    raise
```
Bastian-Krause, line 126: I've tested the live streaming feature more: this unfortunately causes exceptions shown in the StepLogger output:

```
[...]
INFO         StepLogger:  → RawNetworkInterfaceDriver.start_record(filename=None count=1 timeout=5) 
INFO         StepLogger:  ← RawNetworkInterfaceDriver.start_record() 
INFO         StepLogger:  → SSHDriver.put(filename='/tmp/pytest-of-bst/pytest-426/test_port_bridge_from_internal1/replay-dut-br0.pcap' remotepath='/tmp/replay-br0.pcap') 
INFO         StepLogger:  ← SSHDriver.put() [0.050s]
INFO         StepLogger:  → SSHDriver.run_check(cmd='tcpreplay --intf1=br0 /tmp/replay-br0.pcap') 
INFO         StepLogger:  ← SSHDriver.run_check() result=['Actual: 1 packets (1523 bytes) sent in 0.000436 seconds', 'Rated: 3493119.2 Bps, 27.94 Mbps, 2293.… [0.169s]
INFO         StepLogger:  → RawNetworkInterfaceDriver.stop_record(timeout=0) 
INFO         StepLogger:  ⚠ RawNetworkInterfaceDriver.stop_record() [0.001s] exception=Command '['ssh', '-x', '-o', 'LogLevel=ERROR', '-o', 'PasswordAuthentication=no', 'myexporter', '--', 'sudo labgrid-raw-interface tcpdump enp2s0f1 1 --timeout 5']' timed out after 0 seconds
[...]
```

This is confusing. To fix this, let's drop the `TimeoutExpired` handling here:

```suggestion
            self.stop_record(timeout=0 if filename is None else None)
```

..and move it to `stop_record()` instead:

```python
@Driver.check_active
@step(args=["timeout"])
def stop_record(self, *, timeout=None):
    """
    Stops previously started tcpdump on bound network interface resource.

    Args:
        timeout (int): optional, maximum number of seconds to wait for the tcpdump process to
                       terminate
    """
    try:
        self._stop(self._record_handle, timeout=timeout)
        # If live streaming packets, there is no reason to wait for tcpdump                   
        # to finish, so expect a timeout if piping to stdout
        if self._record_handle.stdout is None:
            raise
    finally:
        self._record_handle = None
```

Bastian-Krause (changes requested):

Bastian-Krause:
@JoshuaWatt Are you still interested in this or should we take this over? I'm basing more and more network tests upon this PR, so I'd like to see this merged rather sooner than later.

JoshuaWatt:
> @JoshuaWatt Are you still interested in this or should we take this over? I'm basing more and more network tests upon this PR, so I'd like to see this merged rather sooner than later.

Ya, Sorry. I've been really busy with other stuff lately. Feel free to take it over if you need it

labgrid/driver/rawnetworkinterfacedriver.py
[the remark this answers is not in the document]
Bastian-Krause: I've now dropped the `stderr=subprocess.DEVNULL` leaving it to the default (`None`).

labgrid/driver/rawnetworkinterfacedriver.py
```diff
@@ -96,18 +99,25 @@ def record(self, filename, *, count=None, timeout=None):
...
         finally:
-            self.stop_record()
+            # If live streaming packets, there is no reason to wait for tcpdump
+            # to finish, so terminate it as soon as the context is left
+            try:
+                self.stop_record(timeout=0 if filename == "-" else None)
+            except subprocess.TimeoutExpired:
+                # A timeout is expected if the context is exited early while
+                # piping to stdout
+                pass
```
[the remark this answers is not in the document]
Bastian-Krause: Fixed.

helpers/labgrid-raw-interface
Bastian-Krause: Agreed, let's keep it like this until it becomes a problem.

labgrid/driver/rawnetworkinterfacedriver.py
Bastian-Krause: Integrated.

Bastian-Krause:
Will do, thanks for your work!

Bastian-Krause (approved):

Bastian-Krause:
Rebased on latest master.

jluebbe (approved):

PR #1322, opened 2024-01-26
doc/configuration: update/fix bindings, protocols and resource usage

Bastian-Krause:
**Description**
Initially, all driver bindings were documented with their respective binding key. This wasn't documented consistently since then.

Especially when using multiple instances of a single driver class, the bindings are needed to specify on which resource or other driver a driver should bind on. Until now, most binding keys had to be looked up in the source code. So improve this situation by fixing and adding binding keys as well as also adding all Remote/Network resource variants a driver is able to bind to.

Also update protocol implementations and resource usage. While at it, document drivers not expecting arguments, drop driver usage from network resouces, fix header underlines and remove unnecessary new lines.

**Checklist**
- [x] PR has been tested

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -8,8 +8,8 @@ have no dependencies.
 .. image:: res/config_graph.svg
    :width: 50%
 
-Here the resource `RawSerialPort` provides the information for the
-`SerialDriver`, which in turn is needed by the `ShellDriver`.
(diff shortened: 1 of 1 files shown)


Bastian-Krause:
CI fails due to #1327.

Bastian-Krause:
Rebased on current master.

Emantor (approved):

PR #1323, opened 2024-01-31
remote/exporter: hostname option is ignored

jeromebrunet:
Following the addition of the fqdn option, the hostname is ignored. Hostname is set with gethostname() regardless of args.hostname value because the condition is incorrect.

Fix this and get the hostname option working again.

Fixes #1290


Bastian-Krause (approved):
Nice find, thanks for the fix!

Emantor (approved):

PR #1324, opened 2024-02-01
driver/pe6216: add support for Aten PE6216 PDU

DanielKriz:
**Description**
We are using the [Aten PE6216 PDU](https://www.aten.com/global/en/products/power-distribution-&-racks/rack-pdu/pe6216/) on our testfleet and we want to use labgrid for test management. This PDU is not present in labgrid power drivers or in the `pdudaemon` project, hence we have to add it.

I did test it on our test fleet with few very simple tests:
```python
import time

def test_power_on(target):
    pwd = target.get_driver('NetworkPowerDriver')
    pwd.on()
    time.sleep(10.0)
    assert pwd.get() == True

def test_power_cycle(target):
    pwd = target.get_driver('NetworkPowerDriver')
    pwd.cycle()
    time.sleep(10.0)
    assert pwd.get() == True

def test_power_off(target):
    pwd = target.get_driver('NetworkPowerDriver')
    pwd.off()
    time.sleep(10.0)
    assert pwd.get() == False
```
(PDU takes some time to change states as it has built-in delay between changes (althrough they can be turned of with 'no-wait' option in the [HTTP request](https://assets.aten.com/product/manual/Restful-API-Guide-for-PDU_2022-11-18.pdf)))

**Checklist**
- [x] Tests for the feature 
- [x] PR has been tested

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -198,6 +198,9 @@ Currently available are:
 ``netio_kshell``
   Controls a NETIO 4C PDU via a Telnet interface.
 
+``pe6216``
+  Controls an Aten PE6216 PDU via a simple HTTP API.
+
 ``raritan``
   Controls Raritan PDUs via SNMP.
 
--- a/labgrid/driver/power/pe6216.py
+++ b/labgrid/driver/power/pe6216.py
@@ -1,0 +1,49 @@
+"""Tested with Aten PE6216.
+
+HTTP API is defined by in the aten PDU PE6216 specification:
+https://assets.aten.com/product/manual/Restful-API-Guide-for-PDU_2022-11-18.pdf
+"""
+import re
+import requests
+
+PORT = 80
+
+MIN_OUTLET_INDEX = 1
+MAX_OUTLET_INDEX = 16
+headers = {'Content-Type': 'application/x-www-form-urlencoded'}
+
+def power_set(host, port, index, value):
+    index = int(index)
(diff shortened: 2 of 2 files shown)


Bastian-Krause:
PR looks good to me.

> (PDU takes some time to change states as it has built-in delay between changes (althrough they can be turned of with 'no-wait' option in the [HTTP request](https://assets.aten.com/product/manual/Restful-API-Guide-for-PDU_2022-11-18.pdf)))

I can't find the 'no-wait' option in that pdf. Could you explain why you decided against using this option?

DanielKriz:
I am sorry for such a late reply. I made a mistake, it is not a `no-wait` (that is just our internal name), in the documentation it is called `on_immediate` and `off_immediate`.

I have decided against this option because it is safeguarding the DUTs against rapid switching of power, that might occur with immediate mode. I think that in more cases a lot of people would then implement such wait manually (with `time.sleep` or something like that) as the safety is from my point of view more important than speed. (And PE6216 this feature built-in).

Probably the best and most flexible solution would be some kind of option, that could be put inside the labgrid configuration, but I am not sure if there is any support in labgrid for that right now.

DanielKriz:
Hello, I would like to ask about the state of this pull request? What should I do to make it accepted?

Is the problem the coverage? Could you give me some pointers about how should I approach it? I  am not really sure how to test power drivers.

PR #1326, opened 2024-02-02
labgrid-client.rst: replace 'status' with 'cycle'

threexc:
**Description**

In the labgrid-client doc page, where the power option is detailed:

'get' seems to serve the same purpose as 'status', which isn't a valid option, and 'cycle' is not shown in the doc despite being in the CLI help output. This is a quick fix to address that.

**Checklist**
- [X] Tests for the feature - tested the actual power options with a local install from master branch
- [X] Man pages have been regenerated

### DIFF
--- a/man/labgrid-client.1
+++ b/man/labgrid-client.1
@@ -183,7 +183,7 @@ not at all.
 .sp
 \fBenv\fP                         Generate a labgrid environment file for a place
 .sp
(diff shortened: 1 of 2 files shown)


Bastian-Krause:
@threexc Thanks for taking the time to fix the man pages! Could you please add a `man: ` prefix to your commit subject?

threexc:
@Bastian-Krause done

Bastian-Krause (approved):
Thanks!

Issue #1185, opened 2023-05-19
Couldn't start crossbar using default "config-anonymous.yaml"


Sriram-JI:
Hi,
    I have been trying to run crossbar with default configuration file  - config-anonymous.yaml.
    
```
2023-05-19T17:04:12+0530 [Controller 369497] Ok, local node configuration ran successfully.
2023-05-19T17:04:12+0530 [Controller 369497] <crossbar.node.node.Node.boot>::NODE_BOOT_COMPLETE
2023-05-19T17:04:12+0530 [Guest      369511] Traceback (most recent call last):
2023-05-19T17:04:12+0530 [Guest      369511] File "<frozen runpy>", line 198, in _run_module_as_main
2023-05-19T17:04:12+0530 [Guest      369511] File "<frozen runpy>", line 88, in _run_code
2023-05-19T17:04:12+0530 [Guest      369511] File "/home/sriram/Documents/Tasks/8_CRATE/labgrid/crossbar-venv/lib/python3.11/site-packages/labgrid/remote/coordinator.py", line 14, in <module>
2023-05-19T17:04:12+0530 [Guest      369511] from autobahn.asyncio.wamp import ApplicationRunner, ApplicationSession
2023-05-19T17:04:12+0530 [Guest      369511] File "/home/sriram/Documents/Tasks/8_CRATE/labgrid/crossbar-venv/lib/python3.11/site-packages/autobahn/asyncio/__init__.py", line 32, in <module>
2023-05-19T17:04:12+0530 [Guest      369511] from autobahn.asyncio.websocket import \
2023-05-19T17:04:12+0530 [Guest      369511] File "/home/sriram/Documents/Tasks/8_CRATE/labgrid/crossbar-venv/lib/python3.11/site-packages/autobahn/asyncio/websocket.py", line 36, in <module>
2023-05-19T17:04:12+0530 [Guest      369511] from autobahn.wamp import websocket
2023-05-19T17:04:12+0530 [Guest      369511] File "/home/sriram/Documents/Tasks/8_CRATE/labgrid/crossbar-venv/lib/python3.11/site-packages/autobahn/wamp/websocket.py", line 32, in <module>
2023-05-19T17:04:12+0530 [Guest      369511] from autobahn.websocket import protocol
[...]

PR #1330, opened 2024-02-16
coordinator: Initialize txaio with asyncio to fix runtime error

arthursdo:
This commit fixes a RuntimeError by explicitly initializing txaio to use asyncio before its usage in the coordinator module. This ensures compatibility with the asyncio framework, addressing the issue where txaio required an explicit framework selection.

### DIFF
--- a/labgrid/remote/coordinator.py
+++ b/labgrid/remote/coordinator.py
@@ -10,6 +10,8 @@ from enum import Enum
 from functools import wraps
 
 import attr
+import txaio
+txaio.use_asyncio()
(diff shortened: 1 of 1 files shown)


Bastian-Krause:
> Sounds like #1185. Our CI tests don't stumble upon this. Please make sure you are not installing labgrid and crossbar into the same virtualenv.

Closing, because I'm pretty sure that's the case.

> If you still think this is an issue, please provide the information requested in #1185. Thanks!

Please comment if the above does not solve the issue.

PR #1346, opened 2024-04-02
driver/sigrok: various fixes

flxzt:
**Description**

This fixes some things in the sigrok driver implementation:

- explicitly set time field in csv conversion,
    apparently some versions of sigrok-cli on some distros (for example Ubuntu 22.04) default to false,
    resulting in invalid parsing
- analyze method checks/uses wrong (local) file path when the resource is remote
- capture file is transferred to wrong path when specified filename is not absolute and using a remote sigrok resource.
- avoid deadlock when sigrok-cli process failed
- unsuccessful sigrok-cli termination when using sigrok driver remotely

**Checklist**
- [x] PR has been tested

### DIFF
--- a/labgrid/driver/sigrokdriver.py
+++ b/labgrid/driver/sigrokdriver.py
@@ -212,7 +212,7 @@ class SigrokDriver(SigrokCommon):
     def analyze(self, args, filename=None):
         annotation_regex = re.compile(r'(?P<startnum>\d+)-(?P<endnum>\d+) (?P<decoder>[\w\-]+): (?P<annotation>[\w\-]+): (?P<data>".*)')  # pylint: disable=line-too-long
         if not filename and self._filename:
-            filename = self._filename
+            filename = os.path.join(self._tmpdir, self._basename)
         else:
             filename = os.path.abspath(filename)
         check_file(filename, command_prefix=self.sigrok.command_prefix)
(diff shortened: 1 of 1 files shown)


flxzt:
I pushed an additional fix:

When the `sigrok-cli` procress fails prematurely (which can happen when for example the device is used from another process, or the user does not have the permissions to access the device), the driver implementation waited forever for the creation of the capture file.

The implemented fix polls the subprocess while waiting for the capture file creation, and raises an Exception if it has failed. For easier debugging, it also logs the stdout and stderr of the failed `sigrok-cli` call.

flxzt:
Another issue I found is that the `SIGINT` signal that is supposed to terminate to the sigrok-cli process does not work when the command is executed through ssh. ([here](https://github.com/labgrid-project/labgrid/blob/b61bf551011f508d56a32415067d13c7931efd6e/labgrid/driver/sigrokdriver.py#L164)). I don't yet know how to fix it, but I want to document it.

flxzt:
Turns out `sigrok-cli` can be terminated through any keypress (see http://sigrok.org/gitweb/?p=sigrok-cli.git;a=blob;f=anykey.c;h=b1ac9ef430a80df681ee1442e99d8bbc6712fe11;hb=HEAD#l57 ).

So the fix actually is rather simple - instead of sending SIGINT, write any key to stdin. This works fine with ssh

Emantor (approved):

Issue #1357, opened 2024-04-11
No udisks2 device found for /dev/sdo1


sjg20:
Following along with video-tutorial number 4, I am trying to use the USBStorageDriver.write_files() with an SDwire mux and get the above error, on this line:

        mount_path = self.proxy.mount(self.devpath)

It works OK the second time I try, perhaps because the block device has appeared by then.

For my home-grown (labman) lab I have code that retries for up to 5 seconds for the device to appear. Perhaps we need something like that?

sjg20:
I found a call to _wait_for_medium in write_image() so added that to write_files() too.

Now I get this error:

```
 LG_CROSSBAR=ws://kea:20408/ws labgrid-client -vv -c env_rpi_try.cfg -s uboot console
Selected role rpi3 and place rpi3 from configuration file
Transitioning into state uboot
INFO         StepLogger:  → USBSDWireDriver.sdmux_set(mode='host') 
INFO    SSHConnection(h:   Created new SSH connection to kea
INFO         StepLogger:   → ProcessWrapper.check_output(command=['ssh', '-x', '-o', 'LogLevel=ERROR', '-o', 'PasswordAuthentication=no', '-o', 'ControlMaster=no', '-o', 'ControlPath=/tmp/lg-con-rpj8axax/control-kea', 'kea', '--', 'sd-mux-ctrl', '--ts', '-e', 'sdwire-18']) 
INFO            Process:    [60142] command: ssh -x -o LogLevel=ERROR -o PasswordAuthentication=no -o ControlMaster=no -o ControlPath=/tmp/lg-con-rpj8axax/control-kea kea -- sd-mux-ctrl --ts -e sdwire-18
INFO         StepLogger:   ← ProcessWrapper.check_output() result=b'' [0.235s]
INFO         StepLogger:  ← USBSDWireDriver.sdmux_set() [0.330s]
here /tmp/b/rpi_3_32b/u-boot.bin
INFO         StepLogger:  → USBStorageDriver.write_files(sources=['/tmp/b/rpi_3_32b/u-boot.bin'] target=PurePosixPath('/rpi3-u-boot.bin') partition=1 target_is_directory=False) 
INFO         StepLogger:   → USBStorageDriver.get_size(partition=1) 
cat: /sys/class/block/sdo1/size: No such file or directory
[...]

PR #1359, opened 2024-04-11
udisks2: Be more tolerant of device-startup time

sjg20:
When switching an SDWire to 'host' mode, it may take time for the partition to appear.

During that time the device (e.g. '/dev/sdo1') may not exist, so trying to cat e.g. '/sys/call/block/sdo1' fails. Handle this by returning a zero size in that case, as is done when the returned size is empty.

Even when the block device is present, udisks2 may take a short time to make the device available. Handle this by retrying a few times, until things settle.

Add the call to _wait_for_medium() in write_files() while we are here, to match what is done in write_image(). Also fix the parameter type for write_files() and mention the exception it may raise.

This change has been manually tested since there are no tests available for the code being modified.

**Checklist**
- [ ] Tests for the feature 
(I cannot find any tests for USBStorage)
(not needed, I think)
--->
- [X] PR has been tested
(it seems that these were out-of-sync before my PR)

Fixes: #1357

### DIFF
--- a/labgrid/util/agents/udisks2.py
+++ b/labgrid/util/agents/udisks2.py
@@ -15,8 +15,15 @@ class UDisks2Device:
     def __init__(self, devpath):
         self._logger = logging.getLogger("Device: ")
         self.devpath = devpath
-        client = UDisks.Client.new_sync(None)
+        self.fs = None
 
+    def _setup(self):
+        """Try to find the devpath
+
+        Raises:
+            ValueError: no udisks2 device or no filesystem found on devpath
+        """
+        client = UDisks.Client.new_sync(None)
         manager = client.get_object_manager()
         for obj in manager.get_objects():
             block = obj.get_block()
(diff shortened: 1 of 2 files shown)


Bastian-Krause:
@a3f Could you take a look at this?

labgrid/util/agents/udisks2.py
```diff
@@ -83,17 +90,41 @@ def unmount(self, lazy=False):
...
+        ValueError: Failed to obtain the device (e.g. does not exist)
+    """
     if devpath not in _devs:
-        _devs[devpath] = UDisks2Device(devpath=devpath)
+        dev = UDisks2Device(devpath=devpath)
+        done = False
+        while not done:
+            try:
+                dev._setup()
+                done = True
```
a3f, line 109: Can't you use break here?

labgrid/util/agents/udisks2.py
```diff
@@ -83,17 +90,41 @@ def unmount(self, lazy=False):
...
+            except ValueError as exc:
+                if 'No udisks2 device' not in str(exc) or not retries:
+                    raise
+                retries -= 1
```
a3f, line 113: retries needs to be compared to zero somewhere, doesn't it?

a3f (changes requested):
LGTM apart from the two comments.

labgrid/util/agents/udisks2.py
sjg20: Yes, in line 111

labgrid/util/agents/udisks2.py
sjg20: done

sjg20:
> LGTM apart from the two comments.

OK, I addressed one and explained the other

Emantor (dismissed):

Bastian-Krause (commented):
@Emantor Please don't merge this, yet. I'd like to look into this. I suspect the retry mechanism can be implemented more straight forward.

sjg20:
> @Emantor Please don't merge this, yet. I'd like to look into this. I suspect the retry mechanism can be implemented more straight forward.

Any word on when you will figure this out?  I am able to test things if it helps

Emantor (approved):

PR #1393, opened 2024-05-06
driver: add ability to customize the line separator in drivers using ConsoleProtocol

flxzt:
**Description**

This adds the ability to customize the line separator in drivers that implement `ConsoleProtocol` (through the `ConsoleExpectMixin` class).

These are:
- SerialDriver
- ExternalConsoleDriver
- QemuDriver

It is useful when certain systems expect a `\r\n` instead of the usual `\n` .

**Checklist**
- [x] Tests for the feature
- [x] PR has been tested

### DIFF
--- a/labgrid/driver/consoleexpectmixin.py
+++ b/labgrid/driver/consoleexpectmixin.py
@@ -11,12 +11,12 @@ class ConsoleExpectMixin:
     Console driver mixin to implement the read, write, expect and sendline methods. It uses
     the internal _read and _write methods.
 
-    The class using the ConsoleExpectMixin must provide a logger and a txdelay attribute.
+    The class using the ConsoleExpectMixin must provide a logger, txdelay and linesep attribute.
     """
 
     def __attrs_post_init__(self):
         super().__attrs_post_init__()
-        self._expect = PtxExpect(self)
+        self._expect = PtxExpect(self, self.linesep.encode("ASCII"))
 
     @Driver.check_active
     @step(result=True, tag='console')
--- a/labgrid/util/expect.py
+++ b/labgrid/util/expect.py
@@ -10,11 +10,12 @@ class PtxExpect(pexpect.spawn):
     driver: ConsoleProtocol object to be passed in
     """
 
-    def __init__(self, driver):
+    def __init__(self, driver, linesep=b"\n"):
         "Initializes a pexpect spawn instance with the required configuration"
         self.driver = driver
-        self.linesep = b"\n"
         pexpect.spawn.__init__(self, None, maxread=1)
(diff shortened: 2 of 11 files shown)


flxzt:
this is a relatively straight-forward change, I'd appreciate review @Emantor

labgrid/driver/consoleexpectmixin.py
```diff
@@ -14,9 +16,11 @@ class ConsoleExpectMixin:
     The class using the ConsoleExpectMixin must provide a logger and a txdelay attribute.
     """
 
+    linesep = attr.ib(default="\n", validator=attr.validators.instance_of(str), kw_only=True)
+
     def __attrs_post_init__(self):
         super().__attrs_post_init__()
-        self._expect = PtxExpect(self)
+        self._expect = PtxExpect(self, self.linesep.encode("ASCII"))
```
Bastian-Krause, line 19: Since this is a mixin, all the other changes in this file are not needed. Just mention the linesep attribute in the class's docstring. This makes sure all classes using the mixin have the attribute (and have it documented) or they will fail loudly during the tests.

The FakeConsoleDriver would also need the linesep attribute then.

Bastian-Krause (dismissed):
Looks okay to me, one comment though.

labgrid/driver/consoleexpectmixin.py
flxzt: fixed

labgrid/util/expect.py
```diff
@@ -10,11 +10,12 @@ class PtxExpect(pexpect.spawn):
...
     """
 
-    def __init__(self, driver):
+    def __init__(self, driver, linesep=b"\n"):
         "Initializes a pexpect spawn instance with the required configuration"
         self.driver = driver
-        self.linesep = b"\n"
         pexpect.spawn.__init__(self, None, maxread=1)
+        # the linesep needs to be set *after* pexpect is initialized, otherwise it would be overwritten
+        self.linesep = linesep
```
Emantor, line 18: Does this mean that our previous code never worked?

flxzt: In practice the linesep field in PExpect defaults to `os.linesep`, most likely resolving to `\n` as well in the majority of cases. But strictly speaking, yes

Emantor: And we never changed that anyway, so it was more of a useless line before. Than we don't need to backport since that wasn't changeable before either way.

Emantor (dismissed):

flxzt:
I rebased on main, no functional changes. Can this be merged? @Emantor

Emantor (approved):

Emantor:
I re-approved, I'll give the other maintainers some time to take a look if they want to, otherwise I'll merge on Friday.

Issue #1370, opened 2024-04-23
Add ModbusRTUDriver remote support


flxzt:
It is possible to use `SerialPort`'s and `NetworkSerialPort`'s in the ModbusRTU Driver (comparable to the `SerialDriver`).
Would it make sense to use them directly for adding support for remote Modbus RTU access or should there be an additional `NetworkModbusRTU` resource that gets exported?

The only difference to serial port currently is the additional timeout field, but it could also live in the driver.

Additionally, would it make more sense to also move the `address` field to `ModbusRTUDriver` ? The resource would be responsible for controlling the entire mod-bus serial interface, and the driver would control one individual slave.

I already have a (still WIP) branch for this, see: https://github.com/flxzt/labgrid/tree/remote-modbus-rtu . It already works, tested it with a Modbus RTU relay from Waveshare.

PR #1394, opened 2024-05-06
feat: add remote modbus rtu driver support, refactor ModbusRTUDriver to use SerialPort directly

flxzt:
**Description**

This PR adds support for using Modbus RTU serial ports remotely. It does that by removing the ModbusRTU resource, and let the driver use `SerialPort` and `NetworkSerialPort` resources directly.

In order to archieve this, I moved the fields `timeout` and `address` into the driver.

My reasoning is that it makes more sense to let the resource manage the entire bus, and let the driver only control one slave (= address). Same for the timeout, this can be slave-specific.

It seems to be working fine remotely with a Waveshare RTU relay, so far I only tested it with a low baudrate 9600 so I don't know how well the remote serial connection behaves with higher rates.

as mentioned in #1370 I am unsure if it is "okay" to replace the ModbusRTU resource, because this is obviously a breaking change.
Adding a exported NetworkModbusRTU resource would be the alternative, but it would mean a lot of repetition with NetworkSerialPort. But I might have missed intentional architectural decisions here, so I can implement whatever makes the most sense!

Once this is resolved/merged, I'll follow up with an added `WaveshareRTURelay` driver

**Checklist**
- [x] Tests for the feature 
- [x] PR has been tested

### DIFF
--- a/labgrid/driver/modbusrtudriver.py
+++ b/labgrid/driver/modbusrtudriver.py
@@ -1,33 +1,98 @@
 from importlib import import_module
+
 import attr
+import serial
+import serial.rfc2217
 
 from ..factory import target_factory
+from ..resource import SerialPort
+from ..util.proxy import proxymanager
 from .common import Driver
 
 
 @target_factory.reg_driver
 @attr.s(eq=False)
 class ModbusRTUDriver(Driver):
-    bindings = {"resource": "ModbusRTU"}
+    bindings = {"resource": {"SerialPort", "NetworkSerialPort"}}
+
+    timeout = attr.ib(default=0.25, validator=attr.validators.instance_of(float))
+    address = attr.ib(default=0, validator=attr.validators.instance_of(int))
 
     def __attrs_post_init__(self):
         super().__attrs_post_init__()
-        self._modbus = import_module('minimalmodbus')
+        self._modbus = import_module("minimalmodbus")
         self.instrument = None
 
     def on_activate(self):
-        self.instrument = self._modbus.Instrument(
-            self.resource.port,
-            self.resource.address,
-            debug=False)
+        if isinstance(self.resource, SerialPort):
+            self.instrument = self._modbus.Instrument(
+                self.resource.port, self.address, debug=False
+            )
+        else:
+            if self.resource.protocol == "rfc2217":
+                serial_if = serial.rfc2217.Serial()
+            elif self.resource.protocol == "raw":
+                serial_if = serial.serial_for_url("socket://", do_not_open=True)
+            else:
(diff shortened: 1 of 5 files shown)


labgrid/resource/modbusrtu.py
```diff
@@ -1,29 +0,0 @@
-import attr
-
-from ..factory import target_factory
-from .common import Resource
-from .base import SerialPort
-
-
-@target_factory.reg_resource
-@attr.s(eq=False)
-class ModbusRTU(SerialPort, Resource):
```
Emantor, line 10: This resource should still be available, issue a deprecation warning and use the new resource automatically.

flxzt: Of course, this should be fixed. I also adjusted the commit messages to (hopefully) match the style used in labgrid

jluebbe (changes requested):
Looking at the minimalmodbus documentation, it seems to support communicating with multiple instruments on the same bus by sharing a pyserial object, but with the ModbusRTUDriver we can bind only one device to the serial port resource.

Wouldn't it make more sense to have drivers for each instrument bind to the `SerialDriver` and pass the `SerialDriver.serial` attribute to the minimalmodbus `Instrument` constructor?

To allow multiple client drivers binding to a SerialDriver instance, you'd need to override the `resolve_conflicts` method from the `ConsoleExpectMixin` and allow multiple active clients **only if** they are all instrument drivers (and then call `super()`).

Going this route would also avoid the backwards compat problem, as the new driver would be a `ModbusRTUInstrumentDriver`.

jluebbe:
@b2vn As you contributed the original Modbus RTU support in https://github.com/labgrid-project/labgrid/pull/806, do you have an opinion regarding this refactoring and my suggested approach in https://github.com/labgrid-project/labgrid/pull/1394#pullrequestreview-2182947267?

flxzt:
what's the resolution, I haven't heard from the original author that introduced local Modbus RTU support. @jluebbe

labgrid/driver/modbusrtudriver.py
```diff
@@ -1,54 +1,106 @@
...
+                serial_if.port = f"socket://{host}:{port}/"
+            else:
+                raise Exception("ModbusRTUDriver: unknown protocol")
+            serial_if.baudrate = self.resource.speed
+            serial_if.open()
+
+            self.instrument = self._modbus.Instrument(
+                serial_if,
+                slaveaddress=self.address,
+                close_port_after_each_call=True,
```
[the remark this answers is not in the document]
flxzt: If I remember correctly: closing after each call leaves the port accessible for other users, otherwise it is locked.

PR #1405, opened 2024-05-13
labgrid/driver/power: Backend for tinycontrol.eu IP Power Socket 6G10A v2

ghost:
**Description**

This PR adds a power control backend for the IP Power Socket 6G10A manufactured by tinycontrol.eu. It is integrated in the current network power control framework with the `tinycontrol` model name.

No tests have been provided as it's not clear to me how this can be tested without real HW.

This code is being currently used in our deployment, running the labgrid test suite with this change does not result in any failed tests.

**Checklist**
- [X] PR has been tested

### DIFF
--- a/labgrid/driver/power/tinycontrol.py
+++ b/labgrid/driver/power/tinycontrol.py
@@ -1,0 +1,34 @@
+""" A driver to control the Tinycontrol IP Power Socket 6G10A v2
+    Reference: https://tinycontrol.pl/media/documents/manual_IP_Power_Socket__6G10A_v2_LANLIS-010-015_En-1.pdf
+
+    Example configuration to use port #3 on a device with URL 'http://172.17.180.53:9999/'
+
+    NetworkPowerPort:
+      model: tinycontrol
+      host: 'http://172.17.180.53:9999/'
+      index: 3
+"""
+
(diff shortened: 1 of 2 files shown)


labgrid/driver/power/tinycontrol.py
```diff
@@ -0,0 +1,34 @@
...
+    value = 1 if value else 0
+    r = requests.get(urljoin(host, f"/outs.cgi?out{index}={value}"))
+    r.raise_for_status()
+
+
+def power_get(host, port, index):
+    assert port is None
+
+    index = int(index)
+    r = requests.get(urljoin(host, "/st0.xml"))
```
Emantor, line 30: Shouldn't we use the xml module from python here instead of using a find?

mgrela: ElementTree is used here, not a plain string find(). ElementTree is an xml module in the python standard library. See here: https://docs.python.org/3/library/xml.etree.elementtree.html#xml.etree.ElementTree.Element.find

Emantor: Ah, I see, my bad!

jluebbe (approved):

Issue #1384, opened 2024-04-30
Add ability to jump in networkresource


flxzt:
It would be very nice to have the ability to specify SSH jumps for `NetworkResource`s.

The main use-case would be where DUT's only have exposed networking to another specific device and not reachable directly from the client or exporter.

Would adding a `jumps` field to `NetworkResouce` that contains a list of dicts specifying the jumps make sense?

Emantor:
Labgrid already supports jumping over the exporter to the DUT for NetworkServiceExports, this is done by using the `proxymanager` class inside of labgrid. This is done when the `address` includes a IPv6 link-local address with interface extension, i.e. `fe80::1%dut` where the interface is named `dut` and the IPv6 link local address is `fe80::1`. You can also force a different proxy by importing `proxymanager` and using the `force_proxy()` function.

The other choice is to ship a correctly configured SSH configuration which includes the correct `ProxyJump` depending on the IP address.

flxzt:
No, I meant specifying a network connection where you essentially have two targets, one is the DUT you want to connect to, but it might have a specific network connection and configuration where it is only available through another device which is part of the test setup. (like for example: devices with network connections only available through a PCI backplane)

I think this is currently possible by managing such a jump as a ssh configuration, but this would need to live on the client which I want to avoid.

For example it currently can work like this:

exporter config

```
NetworkService
    address: dut-through-controller
    username: 'root'
```

and the client-side ssh config

```
Host testcontroller
    HostName 10.0.1.23
    User root

Host dut-through-controller
    HostName 192.168.0.2
    User root
    ProxyJump controller
```

I'd like it to become something like this:

exporter config

```
NetworkService
    address: 192.168.0.2
    username: 'root'
    jumps:
    - address: 10.0.1.23
      username: 'root'
```

Emantor:
> No, I meant specifying a network connection where you essentially have two targets, one is the DUT you want to connect to, but it might have a specific network connection and configuration where it is only available through another device which is part of the test setup. (like for example: devices with network connections only available through a PCI backplane)

The problem here is that proxymanager only allows setting a proxy globally (since it's requirements arose from the client CLI proxy requirement).

> As you wrote this is currently possible by managing such a jump as a ssh configuration, but this would need to live on the client which I want to avoid because then each developer would need to specify this themselves.

You can ship the SSH configuration with your test files and provide a wrapper script to access the second device. However here the ssh `-F` option does not merge with the global configuration.

[...]

PR #1407, opened 2024-05-16
feat: add ability to provide password for networkresources

flxzt:
**Description**

This adds the ability to specify a connection ssh password for resources that derive `NetworkResource` in order to use them on hosts where adding a ssh key is not possible or where the host key is regenerated after a reboot.

The implementation uses `SSH_ASK_PASS` in the same way as it is already done in `SSHDriver`.

**Checklist**
- [x] Tests for the feature 
- [x] PR has been tested

### DIFF
--- a/labgrid/resource/common.py
+++ b/labgrid/resource/common.py
@@ -86,18 +86,21 @@ class NetworkResource(Resource):
 
     Args:
         host (str): remote host the resource is available on
+        sshpassword (str): remote host ssh password
     """
     host = attr.ib(validator=attr.validators.instance_of(str))
+    sshpassword = attr.ib(default=None, validator=attr.validators.optional(attr.validators.instance_of(str)), kw_only=True)
 
     @property
     def command_prefix(self):
         host = self.host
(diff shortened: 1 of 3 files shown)


Emantor:
This PR modifies the utility `SSHManager` which manages port forwarding, I don't think this is mergable since hosts used for ssh forwarding should be fundamentally under your control and the addition of SSH keys should not be a problem.

The other part I don't like is adding infrastructure passwords to the YAML, you should really be using public key authentication for this infrastructure.

flxzt:
Hm, I think you're seeing NetworkResources always in the context of the coordinator/exporter infrastructure. It's essentially the same discussion as in #1384 , I always think they could also be used in a more general way to access resources on any remotely accessible system..

Like : the exporter is just a service that happens to export local resources as network resources (and yes, should be easily connectable through passwordless ssh keypairs), but network resources themselves are not bound to only be created by the exporter.

This is a real use case here and works well for me already with this addition (and another one which implements ssh jumps)..

PR #1430, opened 2024-06-27
util/ssh: stop keepalive in disconnect() when existing connection is used

Bastian-Krause:
**Description**
SSHConnection does not open a new connection if an existing connection is open. It reuses the connection and starts a keepalive command (cat).

On `disconnect()`, the keepalive command is only stopped for newly initiated connections, not for existing ones. Fix that.

**Checklist**
- [x] PR has been tested

### DIFF
--- a/labgrid/util/ssh.py
+++ b/labgrid/util/ssh.py
@@ -521,9 +521,10 @@ class SSHConnection:
     def disconnect(self):
         assert self._connected
         try:
+            self._stop_keepalive()
+
             if self._socket:
                 self._logger.info("Closing SSH connection to %s", self.host)
-                self._stop_keepalive()
                 self._stop_own_master()
         finally:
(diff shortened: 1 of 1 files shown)


labgrid/util/ssh.py
```diff
@@ -525,6 +525,10 @@ def disconnect(self):
                 self._logger.info("Closing SSH connection to %s", self.host)
                 self._stop_keepalive()
                 self._stop_own_master()
+
+            # stop keepalive when existing connection is used
+            if self._keepalive:
+                self._stop_keepalive()
```
jluebbe, line 531: Why not stop the keepalive first? That would avoid the special case and duplicated `self._stop_keepalive()`?

jluebbe (changes requested):
Looks good otherwise.

PR #1431, opened 2024-06-27
driver/power/tplink: use IotStrip instead of deprecated SmartStrip

Bastian-Krause:
**Description**
Increase the minimum python-kasa version to 0.7.0. This allows us to use a newly introduced and not deprecated replacement for the kasa.SmartStrip class.

Since python-kasa 0.7.0, the SmartStrip class [is deprecated](https://python-kasa.readthedocs.io/en/stable/smartstrip.html).

The docs state:

  "If you want to continue to use the old API for older devices, you can use the classes in the iot module to avoid deprecation warnings."

https://python-kasa.readthedocs.io/en/stable/deprecated.html#deprecated-api-reference

Do that.

**Checklist**
- [ ] PR has been tested (untested, I don't have the required hardware)

### DIFF
--- a/labgrid/driver/power/tplink.py
+++ b/labgrid/driver/power/tplink.py
@@ -1,14 +1,14 @@
 """ Tested with TP Link KP303, and should be compatible with any strip supported by kasa """
 
 import asyncio
-from kasa import SmartStrip
+from kasa.iot import IotStrip
 
 
 async def _power_set(host, port, index, value):
     """We embed the coroutines in an `async` function to minimise calls to `asyncio.run`"""
     assert port is None
     index = int(index)
-    strip = SmartStrip(host)
+    strip = IotStrip(host)
     await strip.update()
     assert (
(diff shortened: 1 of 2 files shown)


Bastian-Krause:
@CardboardTurkey You contributed the original code. Could you take a look at this?

CardboardTurkey:
This looks reasonable! I'm afraid I no longer have access to the hardware required to test these changes though.

MarekSzczypinski:
Just my 3 cents in this - seems that yet another rebase is needed. In the meantime the structure of `pyproject.toml` has changed and support for Python >= 3.8 was dropped.
Last kasa version to support Python >= 3.9 was `0.7.7`. I guess the minimum python-kasa version can be increased as well.

Bastian-Krause:
@MarekSzczypinski Thanks for pinging me.

Emantor (approved):

PR #1440, opened 2024-07-07
driver/externalconsoledriver: use bufsize=0 for opening external process

Bastian-Krause:
**Description**
The ExternalConsoleDriver works on top of a command executed on the local computer. Until now the system default of `io.DEFAULT_BUFFER_SIZE` (typically 8192 bytes) was used for bufsize. This does not work well for this use case, we want unbuffered reads and writes. So set the bufsize to 0, see https://docs.python.org/3/library/subprocess.html#popen-constructor .

**Checklist**
- [x] PR has been tested (tested with microcom and barebox sandbox)


a3f (approved):
Tested working with barebox sandbox.Thanks!

jluebbe (approved):

PR #1442, opened 2024-07-15
[gRPC] remote: add ability to specify array attributes in exported resources

flxzt:
I was trying out the gRPC branch and noticed that certain attributes added to resources in my fork break when they are exported to the coordinator. The gRPC branch currently does not have support for array attributes.

This PR adds support for exactly that - (nested) array field attributes, by adding `MapValueArray` that contains repeated `MapValue`'s to the `.proto` definition file and the conversion between config dicts <-> grpc messages.

I know the gRPC branch is still very much in movement, but I hope the PR will be considered. It improves the expandability of the framework in forks and closes the feature gap to the old crossbar implementation


jluebbe:
Can you give more details on which attributes use arrays in your fork? For now, we've intentionally avoided nesting of `MapValues` to keep the complexity low.

Note that I've seen that the two conversion functions already take significant amounts of CPU time on slower exporters with many resources (with some ad-hoc profiling). So we should make sure that they don't become significantly slower.

flxzt:
sure: I have a branch that implements ssh jumps, these jumps can be specified on any resource that implements networkresource as a list like this:

```yaml
 NetworkResourceName:
   host: <remote.example.computer>
   sshpassword: <remote-password>
   jumps:
    - <user@jump-host-1:port>
    - <user@jump-host-2:port>
```

see here: https://github.com/flxzt/labgrid/tree/networkresource-jumps

flxzt:
Friendly ping :slightly_smiling_face:

flxzt:
is there anything I can do to get this merged?

flxzt:
@Emantor any chance to make progress here?

PR #1443, opened 2024-07-15
client: names in compact output when listing resources

flxzt:
**Description**

This adds that resources names are printed in the compact output of the `labgrid-client resources` command.

For example, instead of:

```
exporter-01/usbhub-p01/NetworkSerialPort
exporter-01/usbhub-p01/NetworkSerialPort
```

the output becomes

```
exporter-01/usbhub-p01/NetworkSerialPort[/modbus]
exporter-01/usbhub-p01/NetworkSerialPort[/serial-115200]
```

which makes it easier to differentiate between resources that have the same class.

The alternative would be to enable verbose printing with `labgrid-client -v resources`, but that adds a lot of other information and makes is quite hard to pick out the name quickly.

At least in my experience the resource name is the only info that I need 95% of the time, so I think it would be beneficial to add it to the default output.

What do you think?

- [x] PR has been tested

### DIFF
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -364,11 +364,11 @@ class ClientSession:
                         else:
                             places = None
 
-                        results.append((places, exporter, group_name, resource.cls))
+                        results.append((places, exporter, group_name, resource.cls, resource_name))
 
             results = sorted(results, key=lambda res: res[0][0].changed if res[0] else 0)
 
-            for places, exporter, group_name, resource_cls in results:
+            for places, exporter, group_name, resource_cls, resource_name in results:
                 if self.args.sort_by_matched_place_change:
                     places_strs = [f"{p.name}: {datetime.fromtimestamp(p.changed):%Y-%m-%d}" for p in places]
                     places_info = ", ".join(places_strs) if places_strs else "not used by any place"
@@ -376,7 +376,7 @@ class ClientSession:
                 else:
                     places_info = None
 
-                line = f"{exporter}/{group_name}/{resource_cls}"
+                line = f"{exporter}/{group_name}/{resource_cls}[/{resource_name}]"
                 if places_info is not None:
                     print(f"{line:<50s} {places_info}")
                 else:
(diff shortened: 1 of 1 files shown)


Bastian-Krause:
Sounds good. Is there a special reason for the square brackets? I would just drop them, so the match can be copy & pasted into `labgrid-client add-match <match>` directly.

flxzt:
I went with the brackets because it indicates that the name is optional when adding it as a match and for consistency to the verbose output.

I don't have a strong opinion about either, should that be changed as well?

flxzt:
I would appreciate feedback here. @Bastian-Krause

jluebbe:
In our lab, all exporters use resource [configuration](https://labgrid.readthedocs.io/en/latest/configuration.html#exporter-configuration) without names:
```
usb-hub-in-rack12:
  USBSerialPort:
    match:
      '@ID_PATH': 'pci-0000:05:00.0-usb-3-1.3'
```
instead of the named variant
```
usb-hub-in-rack12:
  console-main:
    cls: 'USBSerialPort'
    match:
      '@ID_PATH': 'pci-0000:05:00.0-usb-3-1.3'
```

We use the former, because the exporter configuration is relatively static and the "role" of a specific resource only becomes relevant when we add it to a specific place.

With this change, we'd get a lot of redundant information in the output, like this:
```
rlabQ-srv/q-usb-3-p1/NetworkSerialPort[/USBSerialPort]
rlabQ-srv/q-usb-3-p2/NetworkSerialPort[/USBSerialPort]
rlabQ-srv/q-usb-4-p10/NetworkSerialPort[/USBSerialPort]
rlabQ-srv/q-usb-4-p3/NetworkSerialPort[/USBSerialPort]
rlabQ-srv/q-usb-4-p3-1.1/NetworkSerialPort[/USBSerialPort]
rlabQ-srv/q-usb-4-p3-1.2/NetworkSerialPort[/USBSerialPort]
```
which could confuse users. I'm not sure how to handle both approaches equally well.

flxzt:
> I'm not sure how to handle both approaches equally well.

Would it make sense to introduce a `-n`/`--names` flag instead, only additionally showing the resource names when enabled?

PR #1451, opened 2024-07-26
util/yaml: warn on duplicate keys in mappings

Bastian-Krause:
**Description**
When adding drivers/resources to an environment config or resources to an exporter config, users can stumble upon redundant keys. The resulting errors are not obvious, the last key simply overrides any previous entries with the same key.

In order to make these config errors more obvious, emit a warning when duplicate keys are detected in mappings.

**Checklist**
- [ ] Tests for the feature 
- [ ] PR has been tested

### DIFF
--- a/labgrid/util/yaml.py
+++ b/labgrid/util/yaml.py
@@ -2,6 +2,8 @@
 This module contains the custom YAML load and dump functions and associated
 loader and dumper
 """
+
+import warnings
 from collections import OrderedDict, UserString
 from string import Template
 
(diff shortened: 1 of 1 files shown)


SmithChart (approved):
Tested on the setup where I ran into this issue recently.

I guess this warning would have made my problem really obvious:
```
/(...)/labgrid-lxatac/venv/lib/python3.11/site-packages/labgrid/util/yaml.py:26: UserWarning: lxatac-ptx.yaml: previous entry with duplicate YAML dictionary key 'LXAIOBusPIODriver' overwritten
```

`pyyaml` seems to had a discussion on if they want to solve this at least back tot 2018: https://github.com/yaml/pyyaml/issues/165
But it does not look like the project is willing to provide an upstream solution for this situation.

LGTM

Emantor (approved):

Bastian-Krause:
Before merging, let me check if this also works with templated exporter configs.

Bastian-Krause:
Works also with templated configs.

PR #1454, opened 2024-07-30
feat: support setting the pythonpath in configuration YAMLs

rpoisel:
**Description**

The ``pythonpath`` key allows specifying additional directories to be added to the Python search path (sys.path). This is particularly useful for including directories that contain Python modules and packages which are not installed in the standard library or not in the current directory.

Relative paths are resolved relative to the configuration file's location.

Specifying this option enables the environment to locate and import modules that are not in the default Python path.

**Checklist**
- [x] Tests for the feature 
- [ ] PR has been tested

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -48,6 +48,9 @@ New Features in 24.0
 - The pyproject.toml gained a config for `ruff <https://github.com/astral-sh/ruff>`_.
 - ``setuptools_scm`` is now used to generate a version file.
 - labgrid-client console will fallback to telnet if microcom is not available.
+- The ``pythonpath`` key allows specifying additional directories for the
+  Python search path (sys.path). This helps include modules/packages not
+  in the standard library or current directory.
 
 
 Bug fixes in 24.0
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -3555,6 +3555,8 @@ The skeleton for an environment consists of:
    imports:
      - <import.py>
      - <python module>
+   pythonpath:
+     - <absolute or relative path>
 
 If you have a single target in your environment, name it "main", as the
 ``get_target`` function defaults to "main".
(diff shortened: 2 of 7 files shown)


Emantor:
Isn't this covered by the `imports` property already?

rpoisel:
> Isn't this covered by the `imports` property already?

My use-case was to load modules and packages from relative paths or non-standard locations. In my situation, I want to use shared code that is has been added to my projects as git submodules.

Making these modules/packages available could also be done by setting the `PYTHONPATH` environment variable somewhere in the context of any labgrid-using python code. However, I wanted to make this part of the configuration, thus I introduced the `pythonpath` property.

Could you please have a look into the integration tests of this functionality?

- [test_env_pythonpath_yaml_module](https://github.com/honeytreelabs/labgrid/blob/dev-set-pythonpath-in-config/tests/test_environment.py#L85)
- [test_env_pythonpath_yaml_package](https://github.com/honeytreelabs/labgrid/blob/dev-set-pythonpath-in-config/tests/test_environment.py#L123)

rpoisel:
The `pythonpath` property is, in my opinion, similar to the `pythonpath` configuration option in [pytest](https://pytest.org). You can find more details here: [pytest pythonpath](https://docs.pytest.org/en/8.2.x/reference/reference.html#confval-pythonpath).

By setting this variable, tools like `labgrid-client` will also search the specified paths for imported modules and packages.

Emantor:
I am not convinced we should support two different mechanics for module imports. The `imports` functionality is narrower in scope and IMO suites the use case for in test case strategies quite well. The user can still set the `PYTHONPATH` explicitly outside of labgrid to make other modules available. @jluebbe, @Bastian-Krause whats your opinion on this?

Bastian-Krause:
I agree with @Emantor. I don't see why we would need anything else. Let's close this.

PR #1458, opened 2024-07-31
protocol/driver: add `DigitalInputProtocol`, `GpioDigitalInputDriver`

### LINKED ISSUE #1457: Add DigitalInputProtocol, GpioDigitalInputDriver
Currently it is only possible to specify GPIO's as outputs [see here](https://github.com/labgrid-project/labgrid/blob/64b995fb7a6ef32fb40d15a54ee7dde5603520a4/labgrid/util/agents/sysfsgpio.py#L36).

I am proposing to add a `DigitalInputProtocol` together with a GpioDigitalInputDriver that configures GPIO pins as input only through sysfs (or switch to the newer devfs/libgpiod based GPIO interface while at it, because the sysfs one is deprecated).
This ensures that there won't be any shorts when the pin is connected to an output of the DUT.

flxzt:
**Description**

this adds a new protocol `DigitalInputProtocol` and a GPIO driver that implements it.

The `DigitalOutputProtocol` ABC now also inherits it, ensuring it can be used in all existing driver implementations.

Additionally the client now automatically uses `DigitalInputProtocol` when the command is `io get`.

By seperating the input from output protocol it is possible to make sure the GPIO line is configured in direction `input` in sysfs, which avoids shorts when reading out digital states  from DUT's. Otherwise, one would have to rely on the circuitry of the GPIO line.

Tested on two Raspberry Pi's as exporter and DUT.

Note:

the `sysfsgpioin.py` part duplicates most of the `sysfsgpio.py` code but because of the way this code is executed by agentwrapper I thought it was cleaner to separate it into a new file.

**Checklist**
- [x] Tests for the feature 
- [x] PR has been tested

closes #1457

### DIFF
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -1017,14 +1017,13 @@ class ClientSession:
                 if drv:
                     break
 
-        if not drv:
-            raise UserError("target has no compatible resource available")
-        if action == "get":
-            print(f"digital IO{' ' + name if name else ''} for place {place.name} is {'high' if drv.get() else 'low'}")
-        elif action == "high":
-            drv.set(True)
-        elif action == "low":
-            drv.set(False)
+            if not drv:
+                raise UserError("target has no compatible resource available")
+
+            if action == "high":
+                drv.set(True)
+            elif action == "low":
+                drv.set(False)
 
     async def _console(self, place, target, timeout, *, logfile=None, loop=False, listen_only=False):
         name = self.args.name
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -2553,9 +2553,33 @@ Implements:
 Arguments:
   - delay (float, default=2.0): delay in seconds between off and on
 
+GpioDigitalInputDriver
+~~~~~~~~~~~~~~~~~~~~~~
+The :any:`GpioDigitalInputDriver` reads a digital signal from a GPIO line.
+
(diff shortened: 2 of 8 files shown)


flxzt:
can anyone review this?

labgrid/remote/client.py
```diff
@@ -1017,14 +1017,13 @@ def digital_io(self):
...
-            raise UserError("target has no compatible resource available")
-        if action == "get":
-            print(f"digital IO{' ' + name if name else ''} for place {place.name} is {'high' if drv.get() else 'low'}")
-        elif action == "high":
-            drv.set(True)
-        elif action == "low":
-            drv.set(False)
+            if not drv:
+                raise UserError("target has no compatible resource available")
+
```
Emantor, line 1022: Why is the `get()` action removed here?

Emantor (commented):
We should not need an additional agent that duplicates the code, the existing agent should be extended to support the direction setting as well.

If possible it would also be great if the new gpiochip based sysfs interface that is not deprecated could be used.

ozan956:
Hey @Emantor,

I took a look at this PR and would like to open a new PR based on it, with the review feedback addressed since its 2 years old and we need it.

I leave the gpiochip/libgpiod work as future work for now. I agree that it would be nice to support the newer non-deprecated GPIO interface, but that seems like a larger design topic and should not block a smaller sysfs-based digital input PR.

Emantor:
Superseded by #1935

PR #1460, opened 2024-08-01
docker: move to Docker compose v2

Bastian-Krause:
**Description**
Docker compose v1 has been deprecated since July 2023 [1]. Now the GitHub actions Ubuntu runner images removed it, too [2]. So move to v2.

labgrid is obviously not affected by any changes between v1 and v2 other than the compose call `docker-compose` -> `docker compose`.

While at it, drop the obsolete version element [3] from the docker compose config.

[1] https://docs.docker.com/compose/migrate/
[2] https://github.com/actions/runner-images/issues/9692
[3] https://docs.docker.com/compose/compose-file/04-version-and-name/#version-top-level-element-optional

**Checklist**
- [x] PR has been tested (CI only)

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -38,8 +38,8 @@ jobs:
     - name: Build amd64 docker image and validate
       run: |
         ./dockerfiles/build.sh --load
-        docker-compose -f dockerfiles/staging/docker-compose.yml up --exit-code-from client client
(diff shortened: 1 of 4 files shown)


jluebbe (approved):

Bastian-Krause:
@hoyes Since you've contributed #1307 before, could you take a look at this?

hoyes:
LGTM!

Bastian-Krause:
Thanks!

PR #1461, opened 2024-08-06
workflows/reusable-unit-tests: do not install ncurses-term, drop TERM setting

Bastian-Krause:
**Description**
ncurses-term and the TERM setting were added for the ColoredStepReporter functionality. This has been removed with the StepLogger introduction. So drop both.

**Checklist**
- [x] PR has been tested


Emantor:
Is the CI failure unrelated?

Bastian-Krause:
Yes, should be fixed with #1460. I've rebased this on top of latest master, should now succeed.

Emantor (approved):

PR #1260, opened 2023-08-24
Python 3.12 compatibility

Bastian-Krause:
**Description**
[Python 3.12 is expected to be released on 2023-10-02](https://peps.python.org/pep-0693/#release-schedule). Now that [3.12.0-rc1](https://www.python.org/downloads/release/python-3120rc1/) is released and the ABI settled, it's a good time to see what needs fixing. It turns out, labgrid itself works fine, but some dependencies are not ready, yet:

- [ ] crossbar release with:
  - [ ] https://github.com/crossbario/crossbar/pull/2091
  - [x] https://github.com/crossbario/crossbar/pull/2093
- [x] crossbar dependency: wsaccel https://github.com/methane/wsaccel/issues/30 (Python 3.12 compatibility)
- [x] crossbar dependency: aiohttp 3.9.0 release on pypi (Python 3.12 compatibility, see https://github.com/aio-libs/aiohttp/issues/7675)
- [x] crossbar dependency: numpy 1.26.0 release on pypi (Python 3.12 compatibility)
- [x] https://github.com/lextudio/pysnmp/issues/10 (`ModuleNotFoundError: No module named 'asyncore'`)
- [x] pylint 3.0.0 release on pypi (Python 3.12 compatibility, see [v3.0.0a7](https://github.com/pylint-dev/pylint/releases/tag/v3.0.0a7))

This draft PR works around these issues (by using pre-releases, switching from pypi to git URLs, switching to forks).

**Checklist**
- [ ] PR has been tested (currently CI only)

PR #1469, opened 2024-08-13
Move from crossbar/autobahn to gRPC

### LINKED ISSUE #1429: Assume port 20408 for GRPC exporter
We assume port 20408 already for coordinator and client, however not for the exporter. I suggest we check if the line mentioned below checks for a port, and if it does not exists, add port 20408.

https://github.com/labgrid-project/labgrid/blob/ae0907928feec6b4bbaba7658cf5095d241a848d/labgrid/remote/exporter.py#L1028

Bastian-Krause:
**Description**
As announced in #1417 and in the [v24.0 release](https://github.com/labgrid-project/labgrid/discussions/1467#discussioncomment-10314852), labgrid will move from crossbar/autobahn to gRPC for communication between client/exporter and coordinator.

Crossbar/autobahn are unfortunately not very well maintained anymore. The crossbar component was moved to its own virtualenv to cope with the high number of dependencies leading to conflicts. Support for Python 3.13 is still not
available in a crossbar release on PyPI.

That's why labgrid moves to gRPC with this release. gRPC is a well maintained RPC framework with a lot of users. As a side effect, the message transfer is more performant and the import times are shorter.

Maintaining support for both crossbar/autobahn as well as gRPC in labgrid would be a lot of effort due to the different architectures of those frameworks. Therefore, a hard migration to gRPC is deemed the lesser issue.

Due to the migration, this PR introduces the following breaking changes:                                                                    

- The labgrid environment config option `crossbar_url` was renamed to `coordinator_address`. The environment variable `LG_CROSSBAR` was renamed to `LG_COORDINATOR`.
- The labgrid environment config option `crossbar_realm` is now obsolete as well as the environment variable `LG_CROSSBAR_REALM`.
- The coordinator is available as `labgrid-coordinator` (instead of `crossbar start`). No additional configuration file is required.
- The systemd services in `contrib/systemd/` were updated.

Our plan is to merge this before the next release [v24.1](https://github.com/labgrid-project/labgrid/milestone/12).

**Checklist**
- [x] Tests for the feature 
- [x] PR has been tested (We've tested the grpc branch successfully during the last 7 weeks.)

Closes #1059
Closes #1260
Fixes #1429


Bastian-Krause:
Rebased on latest master (due to conflicts with #1470) .

jluebbe (approved):
Although there are probably still some issues in this PR, I think it's better to fix those with smaller PRs on master instead of iterating here. Merging now would also make testing easier and hopefully lead to more feedback.

sjg20 (approved):
Tested on my lab setup

Emantor (approved):

Emantor:
Let's go.

PR #1475, opened 2024-08-14
github/workflows/scheduled-unit-tests: drop scheduled docker tests on stable-23.0

Bastian-Krause:
**Description**
Adding the stable branches here is problematic: GitHub uses the workflow definitions from the default branch (master). The workflow might not match the stable branch. So drop it here.

As a replacement, the current stable branch is tested on a labgrid fork that has the stable branch set as its default branch, so the workflow is maintained in the same branch as the code.


jluebbe (approved):
I've not checked if there are better ways, but this change seems safe on it's own.

PR #1478, opened 2024-08-18
chore: add typing hints to pytest components

rpoisel:
**Description**

This PR introduces [Python typing hints](https://docs.python.org/3/library/typing.html) to the `pytest` specific components of the labgrid framework. These additions can be seen as initial steps towards enhancing the overall code navigation experience and improving the effectiveness of static code analysis tools.

**Checklist**
- [x] Tests for the feature 
- [x] PR has been tested

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,8 @@ New Features in 25.1
 
 - Guermok HDMI to USB 3.0 capture dongle supported
 
+- The ``pytest``-specific parts of the labgrid framework now have typing hints.
+
 Breaking changes in 25.1
 ~~~~~~~~~~~~~~~~~~~~~~~~
 - The deprecated pytest plugin option ``--env-config`` has been removed. Use
--- a/labgrid/config.py
+++ b/labgrid/config.py
@@ -262,7 +262,7 @@ class Config:
 
         return imports
 
-    def get_paths(self):
+    def get_paths(self) -> dict[str, str]:
         """Helper function that returns the subdict of all paths
 
         Returns:
@@ -275,7 +275,7 @@ class Config:
 
         return paths
 
-    def get_images(self):
+    def get_images(self) -> dict[str, str]:
         """Helper function that returns the subdict of all images
 
         Returns:
--- a/labgrid/logging.py
+++ b/labgrid/logging.py
@@ -17,9 +17,9 @@ def basicConfig(**kwargs):
     root.handlers[0].setFormatter(StepFormatter(indent=indent, parent=parent))
 
 
-logging.CONSOLE = logging.INFO - 5
-assert(logging.CONSOLE > logging.DEBUG)
-logging.addLevelName(logging.CONSOLE, "CONSOLE")
+CONSOLE = logging.INFO - 5
+assert(CONSOLE > logging.DEBUG)
+logging.addLevelName(CONSOLE, "CONSOLE")
 
 # Use composition instead of inheritance
 class StepFormatter:
@@ -106,11 +106,11 @@ class SerialLoggingReporter:
 
                 for part in parts:
                     data = self.vt100_replace_cr_nl(part)
-                    logger.log(logging.CONSOLE, self._create_message(event, data), extra=extra)
+                    logger.log(CONSOLE, self._create_message(event, data), extra=extra)
 
             elif state == "start" and step.args and "data" in step.args:
                 data = self.vt100_replace_cr_nl(step.args["data"])
-                logger.log(logging.CONSOLE, self._create_message(event, data), extra=extra)
+                logger.log(CONSOLE, self._create_message(event, data), extra=extra)
 
     def flush(self):
         if self.lastevent is None:
@@ -122,7 +122,7 @@ class SerialLoggingReporter:
         for source, logger in self.loggers.items():
             data = self.vt100_replace_cr_nl(self.bufs[source])
             if data:
-                logger.log(logging.CONSOLE, self._create_message(self.lastevent, data), extra=extra)
+                logger.log(CONSOLE, self._create_message(self.lastevent, data), extra=extra)
             self.bufs[source] = b""
 
 
--- a/labgrid/logging.py
+++ b/labgrid/logging.py
@@ -22,7 +22,7 @@ assert(CONSOLE > logging.DEBUG)
 logging.addLevelName(CONSOLE, "CONSOLE")
 
 # Use composition instead of inheritance
-class StepFormatter:
+class StepFormatter(logging.Formatter):
     def __init__(self, *args, indent=True, color=None, parent=None, **kwargs):
         self.formatter = parent or logging.Formatter(*args, **kwargs)
         self.indent = indent
--- a/labgrid/pytestplugin/__init__.py
+++ b/labgrid/pytestplugin/__init__.py
@@ -1,2 +1,13 @@
 from .fixtures import pytest_addoption, env, target, strategy
 from .hooks import pytest_configure, pytest_collection_modifyitems, pytest_cmdline_main, pytest_runtest_setup
+
+__all__ = [
+    "pytest_addoption",
+    "env",
+    "target",
+    "strategy",
+    "pytest_configure",
+    "pytest_collection_modifyitems",
+    "pytest_cmdline_main",
+    "pytest_runtest_setup",
+]
--- a/labgrid/pytestplugin/fixtures.py
+++ b/labgrid/pytestplugin/fixtures.py
@@ -1,8 +1,12 @@
 import os
 import subprocess
+from typing import Iterator
+
 import pytest
 
+from .. import Environment, Target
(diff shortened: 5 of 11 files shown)


labgrid/logging.py
sjg20, line 1: The commit message is missing the motivation for this change

labgrid/config.py
```diff
@@ -5,6 +5,7 @@
 """
 import os
 import warnings
+from typing import Dict
```
sjg20, line 8: Does Labgrid sort its imports?

labgrid/pytestplugin/__init__.py
```diff
@@ -1,2 +1,12 @@
...
+
+__all__ = [
+    "pytest_addoption",
+    "env",
+    "target",
+    "strategy",
+    "pytest_configure",
+    "pytest_collection_modifyitems",
+    "pytest_cmdline_main",
+]
```
sjg20, line 12: Should this be in a separate commit?

labgrid/pytestplugin/fixtures.py
```diff
@@ -74,7 +80,8 @@ def env(request, record_testsuite_property):
         try:
             target = env.get_target(target_name)
         except UserError as e:
-            pytest.exit(e)
+            pytest.exit(str(e))
```
sjg20, line 83: This change seems to be unrelated to typing? Or does it fix a bug? If so, best to put the bug-fixes first, in one or more separate commits.

tests/test_fixtures.py
```diff
@@ -17,73 +19,83 @@ def short_env(tmpdir):
...
 def test(env):
     assert True
 """
     )
     return t
 
+
 def test_config(short_test):
-    with pexpect.spawn(f'pytest --traceconfig {short_test}') as spawn:
+    with pexpect.spawn(f"pytest --traceconfig {short_test}") as spawn:
```
sjg20, line 37: Is Labgrid supposed to use " instead of ' ? I see both, I have been wondering

CHANGES.rst
```diff
@@ -17,6 +17,7 @@ more performant and the import times are shorter.
 New Features in 24.1
 ~~~~~~~~~~~~~~~~~~~~
 - All components can be installed into the same virtualenv again.
+- The ``pytest`` specific parts of the labgrid framework now have typing hints.
```
sjg20, line 20: ``pytest``-specific

sjg20 (changes requested):

labgrid/pytestplugin/fixtures.py
rpoisel: Actually, this fixes that the wrong type is passed to the `pytest.exit()` function: [pytest.exit](https://github.com/pytest-dev/pytest/blob/main/src/_pytest/outcomes.py#L107).

tests/test_fixtures.py
rpoisel: Good point, I'll make sure these things remain untouched in this PR.

Personally, I'd prefer them to be consistent (whether it's `'` or `"`).

sjg20: I agree. I much prefer ' since it is easier to distinguish as small font sizes, but Labgrid seems to use both

labgrid/config.py
sjg20: No, I suppose I just don't understand how the ordering works

Perhaps 'warnings' is consider an internal module and 'typing' is not?

labgrid/logging.py
sjg20: OK. I didn't even know that Labgrid had created its own debug level!

tests/test_fixtures.py
rpoisel: I just updated the `pyproject.toml` file to contain:

```toml
[tool.ruff.format]
quote-style = "single"
```

Actually, I'd suggest to re-format the whole code base. But this should go into a dedicated PR in my view.

In this PR, however, I already re-formatted the `tests/test_fixtures.py` file to this new formatting rule anyways.

rpoisel: Okay, this does not work yet. All sources would have to be re-formatted this way before passing CI checks.

As a quick workaround I applied this setting to the `tests/test_fixtures.py` file, but removed it from the `pyproject.toml` file.

rpoisel:
@sjg20, could you please review my PR again? All your comments should be implemented/resolved now.

To be honest, I'd love to see this PR merged into labgrid as it improves the code navigation and static code analysis experience quite a lot.

rpoisel:
@sjg20, could you please have a look? I just made sure this PR can be merged without conflicts.

labgrid/logging.py
```diff
@@ -22,7 +22,7 @@ def basicConfig(**kwargs):
 logging.addLevelName(CONSOLE, "CONSOLE")
 
 # Use composition instead of inheritance
-class StepFormatter:
+class StepFormatter(logging.Formatter):
```
dlech, line 25: This makes the comment above no longer true. This is not just adding a type hint.

Instead, we could solve this with:

```python
        root.handlers[0].setFormatter(cast(logging.Formatter, StepFormatter(indent=indent, parent=parent)))
```

labgrid/pytestplugin/fixtures.py
```diff
@@ -1,8 +1,12 @@
 import os
 import subprocess
+from typing import Iterator
```
dlech, line 3:
```suggestion
from collections.abc import Iterator
```

`collections.abc` was used in `labgrid/environment.py`, so would be nice to be consistent.

labgrid/pytestplugin/hooks.py
```diff
@@ -94,9 +100,11 @@ def pytest_configure(config):
     processwrapper.enable_logging()
 
 @pytest.hookimpl()
-def pytest_collection_modifyitems(config, items):
+def pytest_collection_modifyitems(session: pytest.Session, config: pytest.Config, items: List[pytest.Item]) -> None:
```
dlech, line 111:
```suggestion
def pytest_collection_modifyitems(session: pytest.Session, config: pytest.Config, items: list[pytest.Item]) -> None:
```

tests/test_fixtures.py
```diff
@@ -17,47 +19,61 @@ def short_env(tmpdir):
...
+
 def test_env_fixture_no_logging(short_env, short_test):
-    with pexpect.spawn(f'pytest -p no:logging --lg-env {short_env} {short_test}') as spawn:
+    with pexpect.spawn(f"pytest -p no:logging --lg-env {short_env} {short_test}") as spawn:
         spawn.expect(pexpect.EOF)
         spawn.close()
         assert spawn.exitstatus == 0, spawn.before
 
+
+def test_env_old_fixture(short_env, short_test):
```
dlech: This function is added in a commit that is just supposed to be linting.

tests/test_fixtures.py
```diff
@@ -1,84 +1,76 @@
...
-def test_env_fixture_no_logging(short_env, short_test):
-    with pexpect.spawn(f"pytest -p no:logging --lg-env {short_env} {short_test}") as spawn:
+def test_env_fixture_no_logging(short_env: pathlib.Path, short_test: pathlib.Path) -> None:
+    with pexpect.spawn(f'pytest -p no:logging --lg-env {short_env} {short_test}') as spawn:
         spawn.expect(pexpect.EOF)
         spawn.close()
         assert spawn.exitstatus == 0, spawn.before
 
 
-def test_env_old_fixture(short_env, short_test):
```
dlech: And now another commit is removing the same function that was just added.

tests/test_fixtures.py
```diff
@@ -1,84 +1,76 @@
...
 targets:
   test1:
     drivers: {}
   test2:
     role: foo
     resources: {}
-"""
-    )
+  main:
+    drivers: {}
```
dlech, line 21: This is being added in a commit that is supposed to just be adding type hints.

tests/test_fixtures.py
```diff
@@ -95,3 +97,157 @@ def test_log_without_capturing(short_env: pathlib.Path, short_test: pathlib.Path
...
         assert spawn.exitstatus == 0
+
+
+@dataclass
+class Scenario:
+    config: Optional[str]
+    test: str
+    exitcode: pytest.ExitCode
+    lines: List[str] = field(default_factory=list)
+    outcome: Dict[str, int] = field(default_factory=dict)
```
dlech, line 108:
```suggestion
    lines: list[str] = field(default_factory=list)
    outcome: dict[str, int] = field(default_factory=dict)
```

dlech (commented):
I'm new to this project, so I don't really know what expectations are but I'll make a few suggestions anyway.

Many commits in the PR are unrelated to type hints. I would split out all unrelated commits into separate pull requests.

And each commit message could use more of an explanation. Many of the commits are doing more than than what the subject line says. So some justification should be given why these changes are being made.

PR #1481, opened 2024-08-20
driver: remove `mktemp` tool dependency in `ShellDriver`

flxzt:
**Description**

This removes the use of the `mktemp` tool on targets when transferring files through the `ShellDriver`.
It is replaced by `touch /tmp/tmp-<random-uuid>` which should be always available since `touch` is part of the GNU core utilities.

**Checklist**
- [x] PR has been tested

### DIFF
--- a/labgrid/driver/shelldriver.py
+++ b/labgrid/driver/shelldriver.py
@@ -5,6 +5,7 @@ import io
 import re
 import shlex
 import ipaddress
+from uuid import uuid4
 
 import attr
 from pexpect import TIMEOUT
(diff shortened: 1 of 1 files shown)


jluebbe:
This is not equivalent to the old behavior, as touch may create files readable by other users. Also, touch doesn't fail if the file already exists.

At least, this change needs a good reasoning why this changes is an improvement and doesn't break existing usecases.

flxzt:
I can only speak for my use-case but it is the following: I came across a target that has only the most basic tools available, not much more than coreutils and does not have the ability to easily add additional tools when testing the production image.
I had hoped this dependency removal is something which benefits other users with similar targets as well.
I can't see how it breaks existing use-cases since the created tempfile really is only used for a short while until it's contents are `dd`'ed to the actual destination.

I didn't know that file permissions are a concern here, let me add a chmod 600 which should replicate the `mktemp`'s default behaviour (unless you have another suggestion).

Regarding not failing when a file with the same name already exists - If you want a check can be added before calling `touch`, but the possibility of collision of a UUID4 which the to-be-created file contains in it's path according to wikipedia : `number of random version-4 UUIDs which need to be generated in order to have a 50% probability of at least one collision is 2.71 quintillion` [source](https://en.wikipedia.org/wiki/Universally_unique_identifier#Collisions).

labgrid/driver/shelldriver.py
```diff
@@ -364,10 +365,8 @@ def _target_cleanup(tmpfile):
         stream = io.BytesIO(buf)
 
         # We first write to a temp file, which we'll `dd` onto the destination file later
-        tmpfile = self._run_check('mktemp')
-        if not tmpfile:
-            raise ExecutionError('Could not make temporary file on target')
-        tmpfile = tmpfile[0]
+        tmpfile = f'/tmp/tmp-{uuid4()}'
+        self._run_check(f'touch {tmpfile}')
```
sjg20, line 369: How about only doing this as a workaround when mktemp fails? Then we can keep using the normal approach for most cases?

sjg20 (changes requested):

labgrid/driver/shelldriver.py
flxzt: That would work for me - as a fallback. What are your thoughts about it @jluebbe ?

flxzt: refactored to do that.

Emantor (approved):

PR #1482, opened 2024-08-22
Add USB loaders for sunxi, tegra, samsung

sjg20:
This adds USB loaders for three different SoCs. The Tegra loader is fairly straightfoward and similar to iMX, so it can use the same protocol.

For Allwinner (sunxi) and Samsung, the loading happens in 2-3 phases, so a new 'phase' parameter is added to the protocol. This allows USB loading to work correctly on these boards, sending an initial BL1 image (in the case of Samsung), then U-Boot SPL and U-Boot proper.

This has been tested on:
- snow (exynos5250)
- Linksprite_pcDuino3 (sunxi)
- nyan-big (Tegra124-based Chromebook)

- [ ] Tests for the feature 
- [X] PR has been tested

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -823,6 +823,23 @@ Arguments:
 Used by:
   - `RKUSBDriver`_
 
+SamsungUSBLoader
+~~~~~~~~~~~~~~~~
+A :any:`SamsungUSBLoader` resource describes a USB device in the *Samsung
+loader state*.
+
+.. code-block:: yaml
+
+   SamsungUSBLoader:
+     match:
+       ID_PATH: pci-0000:03:00.2-usb-0:4.2.4
+
+Arguments:
+  - match (dict): key and value pairs for a udev match, see `udev Matching`_
+
+Used by:
+  - `SamsungUSBDriver`_
+
 SunxiUSBLoader
 ~~~~~~~~~~~~~~
 A :any:`SunxiUSBLoader` resource describes a USB device in the *Allwinner
@@ -870,6 +887,11 @@ a remote computer.
 NetworkRKUSBLoader
 ~~~~~~~~~~~~~~~~~~
 A :any:`NetworkRKUSBLoader` describes an `RKUSBLoader`_ resource available on a
+remote computer.
+
+NetworkSamsungUSBLoader
+~~~~~~~~~~~~~~~~~~~~~~~
+A :any:`NetworkSamsungUSBLoader` describes a `SamsungUSBLoader`_ available on a
 remote computer.
 
 NetworkSunxiUSBLoader
@@ -2975,6 +2997,52 @@ Arguments:
 
  Tools:
   - sunxi-fel: Path to the 'sunxi-fel' tool (default is 'sunxi-fel')
+
+SamsungUSBDriver
+~~~~~~~~~~~~~~~~
+A :any:`SamsungUSBDriver` is used to upload an image into a device in the
+*Samsung loader state*. This is useful to bootstrap a bootloader onto a device.
+
+The load happens in three stages: BL1, SPL and U-Boot proper.
+
+Binds to:
+  loader:
+    - `SamsungUSBLoader`_
+    - `NetworkSamsungUSBLoader`_
+
+Implements:
+  - :any:`BootstrapProtocol`
+
+.. code-block:: yaml
+
+   targets:
+     main:
+       drivers:
+         SamsungUSBDriver:
+           bl1_loadaddr: 0x02021400
+           spl_loadaddr: 0x02023400
+           loadaddr: 0x43e00000
+   images:
+     bl1: '/home/dev/exynos/snow/u-boot.bl1.bin'
+   tools:
+     smdk-usbdl: '/home/dev/bin/smdk-usbdl'
+
+Arguments:
+  - bl1_loadaddr (int): Load address of the BL1 file. This depends on the SoC
+    spl_load_addr (int): Load address of SPL. The easiest way to find this value
+    is to check the *CONFIG_SPL_TEXT_BASE* value in U-Boot
+  - loadaddr (int): address to use when loading U-Boot. This depends on the
+    SoC being used. The easiest way to find this value is to check the
+    *CONFIG_TEXT_BASE* value in U-Boot
+  - image (str): optional, reference to the images key for the image to load
+    when no filename is given
+  - bl1_image (str): optional, images key holding the BL1 file (default 'bl1')
+
+ Images:
+  - bl1: the BL1 file which sets up the SoC
+
+ Tools:
+  - smdk-usbdl: Path to the 'smdk-usbdl' tool (default is 'smdk-usbdl')
 
 TegraUSBDriver
 ~~~~~~~~~~~~~~
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -840,6 +840,23 @@ Arguments:
 Used by:
   - `SunxiUSBDriver`_
 
+TegraUSBLoader
+~~~~~~~~~~~~~~
+A :any:`TegraUSBLoader` resource describes a USB device in the *Tegra
+loader state*.
+
+.. code-block:: yaml
+
+   TegraUSBLoader:
+     match:
+       'ID_PATH': 'pci-0000:03:00.2-usb-0:3.4.1'
+
+Arguments:
+  - match (dict): key and value pairs for a udev match, see `udev Matching`_
+
+Used by:
+  - `TegraUSBDriver`_
+
 NetworkMXSUSBLoader
 ~~~~~~~~~~~~~~~~~~~
 A :any:`NetworkMXSUSBLoader` describes an `MXSUSBLoader`_ resource available on
@@ -858,6 +875,11 @@ remote computer.
 NetworkSunxiUSBLoader
 ~~~~~~~~~~~~~~~~~~~~~
 A :any:`NetworkSunxiUSBLoader` describes a `SunxiUSBLoader`_ available on a
+remote computer.
+
+NetworkTegraUSBLoader
+~~~~~~~~~~~~~~~~~~~~~
+A :any:`NetworkTegraUSBLoader` describes a `TegraUSBLoader`_ available on a
 remote computer.
 
 AndroidUSBFastboot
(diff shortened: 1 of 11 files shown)


doc/configuration.rst
```diff
@@ -733,6 +733,25 @@ Arguments:
...
+TegraUSBLoader
+~~~~~~~~~~~~~~
+A :any:`TegraUSBLoader` resource describes a USB device in the *Tegra
+loader state*.
+
+.. code-block:: yaml
+
+   TegraUSBLoader:
+     ## hub e9
+     ## 1-5.1.2.2
```
jluebbe, line 763: This doesn't match the ID_PATH below. Simply remove this?

doc/configuration.rst
```diff
@@ -2539,6 +2563,49 @@ Arguments:
...
 
+TegraUSBDriver
+~~~~~~~~~~~~~~
+A :any:`TegraUSBDriver` is used to upload an image into a device in the
+*Nvidia Tegra loader state*. This is useful to bootstrap a bootloader onto a
+device.
+
+The load happens in a single stage, with U-Boot proper sent along with a
+*BCT* (Binary Control Tool) configuration file which contains memory timings,
+etc.
```
jluebbe, line 2638: Can this driver be used with U-Boot only? Otherwise it should be worded in a generic way.

doc/configuration.rst
```diff
@@ -2539,6 +2563,49 @@ Arguments:
...
+           bct: '/home/dev/tegra124/nvidia/norrin/PM370_Hynix_2GB_H5TC4G63AFR_PBA_924MHz_01212014.bct'
+           usb_path: 1-3.4.1
+   tools:
+     tegrarcm: '/home/dev/bin/tegrarcm'
+
+Arguments:
+  - loadaddr (int): address to use when loading U-Boot. This depends on the
+    SoC being used. The easiest way to find this value is to check the
+    *CONFIG_TEXT_BASE* value in U-Boot
+  - bct (str): path to the BCT (configuration file)
```
jluebbe, line 2664: Similar to the `image` argument, this should refer to a key in the `images:` part of the target config in the environment. See `IMXUSBDriver`, `RKUSBDriver` or `UUUDriver`.

This also applies to the actual `image` key here (where it's actually supported in the code).

labgrid/driver/usbloader.py
```diff
@@ -286,3 +286,54 @@ def load(self, filename=None, phase=None):
...
+
+        processwrapper.check_output(
+            self.loader.command_prefix + args,
+            print_on_silent_log=True
+        )
+
+    @Driver.check_active
+    @step()
+    def execute(self):
+        pass
```
jluebbe, line 339: What is the purpose of this function? It seems unused.

labgrid/resource/udev.py
```diff
@@ -328,6 +328,17 @@ def filter_match(self, device):
...
         return super().filter_match(device)
 
+@target_factory.reg_resource
+@attr.s(eq=False)
+class TegraUSBLoader(USBResource):
+    def filter_match(self, device):
+        match = (device.properties.get('ID_VENDOR_ID'), device.properties.get('ID_MODEL_ID'))
+
+        if match not in [("0955", "7340")]:  # or 7740
+            return False
```
jluebbe, line 349: If `7740` is another supported device, it should be added to the list to match against. See `class IMXUSBLoader` for an example of how to allow multiple VID/PID pairs.

Otherwise, the comment's meaning is not clear to me.

labgrid/driver/usbloader.py
```diff
@@ -286,3 +286,54 @@ def load(self, filename=None, phase=None):
...
+    @Driver.check_active
+    @step(args=['filename'])
+    def load(self, filename=None, phase=None):
+        if filename is None and self.image is not None:
+            filename = self.target.env.config.get_image_path(self.image)
+        mf = ManagedFile(filename, self.loader)
+        mf.sync_to_resource()
+
+        pathname = mf.get_remote_path()
+        args = [self.tool, '--bct=' + self.bct,
```
jluebbe, line 326: Doesn't the bct need to be pushed to the exporter as well (using `ManagedFile`)?

labgrid/driver/usbloader.py
```diff
@@ -286,3 +286,54 @@ def load(self, filename=None, phase=None):
...
+               f'--bootloader={pathname}',
+               f'--loadaddr={self.loadaddr:#08x}',
+               '--usb-port-path', self.usb_path]
```
jluebbe, line 329: Hmm, why aren't you using `self.loader.path` here, which would be the USB path as supplied by the resource (potentially from the exporter).

Using that would also allow you to drop the manual `usb_path` property on the Driver.

doc/configuration.rst
```diff
@@ -715,6 +715,24 @@ Arguments:
...
 
+SunxiUSBLoader
+~~~~~~~~~~~~~~
+A :any:`SunxiUSBLoader` resource describes a USB device in the *Allwinner
+loader state*.
+
+.. code-block:: yaml
+
+   SunxiUSBLoader:
+     ## hub a12
```
jluebbe, line 744: This seems specific to your lab and isn't very useful in the docs.

labgrid/driver/usbloader.py
```diff
@@ -201,7 +201,7 @@ class BDIMXUSBDriver(Driver, BootstrapProtocol):
     must be specified explicitly.
     """
     bindings = {
-        "loader": {"IMXUSBLoader", "NetworkIMXUSBLoader"},
+        'loader': {'IMXUSBLoader', 'NetworkIMXUSBLoader'},
```
jluebbe, line 204: Unrelated change, drop from this PR.

labgrid/driver/usbloader.py
```diff
@@ -233,3 +233,56 @@ def load(self, filename):
...
+        pathname = mf.get_remote_path()
+        if phase == 'spl':
+            self.run_tool('spl', pathname)
+        else:
+            self.run_tool('write', '%#x' % self.loadaddr, pathname)
+
+    @Driver.check_active
+    @step()
+    def execute(self):
+        self.run_tool('exe', '%#x' % self.loadaddr)
```
jluebbe, line 288: What's the purpose of this method?

labgrid/driver/usbloader.py
```diff
@@ -233,3 +233,56 @@ def load(self, filename):
...
+        else:
+            self.tool = 'sunxi-fel'
+
+    def on_activate(self):
+        pass
+
+    def on_deactivate(self):
+        pass
+
+    def run_tool(self, *part):
```
jluebbe, line 262: Add a `_` prefix to indicate that this is an internal method.

doc/configuration.rst
```diff
@@ -715,6 +715,24 @@ Arguments:
...
 
+SamsungUSBLoader
+~~~~~~~~~~~~~~~~
+A :any:`SamsungUSBLoader` resource describes a USB device in the *Samsung
+loader state*.
+
+.. code-block:: yaml
+
+   SamsungUSBLoader:
+     ## hub d4
```
jluebbe, line 726: Remove lab-specific comment.

labgrid/driver/usbloader.py
```diff
@@ -337,3 +337,67 @@ def load(self, filename=None, phase=None):
...
+
+    @Driver.check_active
+    @step(args=['filename'])
+    def load(self, filename=None, phase=None):
+        if filename is None and phase == 'bl1':
+            filename = self.bl1
+        if filename is None and self.image is not None:
+            filename = self.target.env.config.get_image_path(self.image)
+        mf = ManagedFile(filename, self.loader)
+        mf.sync_to_resource()
```
jluebbe, line 377: Shouldn't bl1, spl and the main image be handle consistently, so with get_image_path() for defaults and a way to select the image key name in the driver arguments?

labgrid/driver/usbloader.py
```diff
@@ -337,3 +337,67 @@ def load(self, filename=None, phase=None):
...
+
+        processwrapper.check_output(
+            self.loader.command_prefix + args,
+            #print_on_silent_log=True
+        )
+
+    @Driver.check_active
+    @step()
+    def execute(self):
+        pass
```
jluebbe, line 403: This seems unused.

jluebbe (changes requested):

doc/configuration.rst
sjg20: I'm not sure there is any other loader, but I'll make the U-Boot mention an example.

doc/configuration.rst
sjg20: This is actually different for each board, so we cannot have a single bct file.

labgrid/driver/usbloader.py
sjg20: It tells the board to start execution. For Tegra this happens at the end of the load, so this method does nothing. I'll add a comment.

labgrid/resource/udev.py
sjg20: I'll drop the comment. I just saw that number in some code but I don't have a device which uses it.

labgrid/driver/usbloader.py
sjg20: It tells the board to start execution

labgrid/driver/usbloader.py
sjg20: This information is provided by the caller, since it differs depending on the board being used

doc/configuration.rst
sjg20: OK, much later, but coming back to this, as I understand the intent is to have a separate environment for each board and I've managed to make this work.

So...done

labgrid/driver/usbloader.py
sjg20: as above...done

sjg20:
I pushed a rework of this series: each board now uses its own env file, and the firmware filenames (`bct`, `bl1`) come from the `images:` section via `get_image_path()` rather than driver arguments, matching `IMXUSBDriver`/`RKUSBDriver`/`UUUDriver`.

Emantor:
There still seem to be open comments on this PR (lab specific comments in the examples, not using `self.loader.path` and rolling a `usb_path` property,…) and I think those will need to be resolved first.

labgrid/driver/usbloader.py
sjg20: Good catch - fixed: the bct is now wrapped in a ManagedFile and synced to the resource, the same as the bootloader image.

labgrid/driver/usbloader.py
sjg20: Fixed: the driver now uses self.loader.path and the usb_path argument is gone.

doc/configuration.rst
sjg20: Removed.

doc/configuration.rst
```diff
@@ -2478,6 +2501,44 @@ Arguments:
...
+         SunxiUSBDriver:
+           loadaddr: 0x4a000000
+   tools:
+     sunxi-fel: '/home/dev/bin/sunxi-fel'
+
+Arguments:
+  - loadaddr (int): address to use when loading U-Boot. This depends on the
+    SoC being used. The easiest way to find this value is to check the
+    *CONFIG_TEXT_BASE* value in U-Boot
+  - image (str): Optional image filename
```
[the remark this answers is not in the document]
sjg20: Fixed: the docs now describe 'image' as a reference to the images key, and the bct entry likewise.

labgrid/driver/usbloader.py
sjg20: Dropped.

sjg20:
All the review comments should be addressed now: the driver uses self.loader.path (usb_path is gone), the bct is pushed with ManagedFile, the lab-specific comments and the unrelated change are dropped, and the docs describe bct/image as images: keys. The branch is also rebased on current master and CI is green.

labgrid/driver/usbloader.py
sjg20: done

doc/configuration.rst
sjg20: done

sjg20:
> There still seem to be open comments on this PR (lab specific comments in the examples, not using `self.loader.path` and rolling a `usb_path` property,…) and I think those will need to be resolved first.

Ah yes, done now.

PR #1483, opened 2024-08-22
driver: Add driver for Dediprog SPI-flash emulator

sjg20:
The EM100-pro is a SPI-flash emulator which supports a variety of chips. It can be used to provide an image to a booting system.

Add a driver which allows images to be loaded ready for use. Provide a trace facility for use when the board does not boot.

This emulator is widely used on x86 machines since they boot using SPI flash. It is prohibitively expensive (in terms of time) to re-flash the SPI chip on test run, whereas the Dediprog emulator updates in about 5 seconds.

This has been tested on several x86 Chromebooks, including coral.

- [ ] Tests for the feature 
- [X] PR has been tested

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -1333,6 +1333,34 @@ NetworkDediprogFlasher
 A :any:`NetworkDediprogFlasher` describes a `DediprogFlasher`_ resource
 available on a remote computer.
 
+SFEmulator
+~~~~~~~~~~
+A :any:`SFEmulator` resource is used to configure the parameters for a
+SPI-flash emulator. This allows a DUT's SPI flash to be replaced or
+overridden with a USB-connected emulator, which is much faster to load than
+reprogramming the DUT's actual SPI flash.
+
+So far only the Dediprog EM100-PRO is supported, so there is no 'model' property
+defined.
+
+Arguments:
+  - serial (str): Serial number of the device
+  - chip (str): SPI-flash chip to emulate
+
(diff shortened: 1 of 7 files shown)


sjg20:
This one needs variables, so I've removed the trace function in this version

doc/configuration.rst
```diff
@@ -1168,6 +1168,34 @@ NetworkDediprogFlasher
...
 remote computer.
 
+SFEmulator
+~~~~~~~~~~
+A :any:`SFEmulator` resource is used to configure the parameters for a
+SPI-flash emulator. This allows a DUT's SPI flash to be replaced or
+overridden with a USB-connected emulator, which is much faster to load than
+reprogramming the DUT's actual SPI flash.
+
+So far only the Dediprog EM100-PRO is supported, so there is no 'model' property
```
obbardc, line 1178: Perhaps `model: Dediprog EM100-PRO` could be mandatory & checked in case of different models added in future?

sjg20: OK done

sjg20:
I've added a test and dropped the 'model' property (since the driver only supports em100 and I'm not sure there is anything else it might need to support).

Emantor (approved):

Bastian-Krause (changes requested):
@sjg20 Please add the tool including its upstream to `doc/man/device-config.rst`. The resource should be a managed USB resource and the tool should be called with `--device BUS:DEV`.

PR #1484, opened 2024-08-22
Make QEMU machine and cpu optional

sjg20:
This makes the QEMU driver's machine, cpu and extra_args parameters optional, since on x86 these are typically not specified. It allows a simpler configuration for x86 targets.

- [X] Tests for the feature 
- [X] PR has been tested

### DIFF
--- a/labgrid/driver/qemudriver.py
+++ b/labgrid/driver/qemudriver.py
@@ -33,9 +33,9 @@ class QEMUDriver(ConsoleExpectMixin, Driver, PowerProtocol, ConsoleProtocol):
 
     Args:
         qemu_bin (str): reference to the tools key for the QEMU binary
-        machine (str): QEMU machine type
-        cpu (str): QEMU cpu type
         memory (str): QEMU memory size (ends with M or G)
+        cpu (str): optional, QEMU cpu type
+        machine (str): optional, QEMU machine type
         extra_args (str): optional, extra QEMU arguments passed directly to the QEMU binary
         boot_args (str): optional, additional kernel boot argument
         kernel (str): optional, reference to the images key for the kernel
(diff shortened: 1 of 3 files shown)


labgrid/driver/qemudriver.py
```diff
@@ -192,6 +204,9 @@ def get_qemu_base_args(self):
             cmd.append("-bios")
             cmd.append(
                 self.target.env.config.get_image_path(self.bios))
+        elif self._bios_fname:
+            cmd.append("-bios")
+            cmd.append(self._bios_fname)
```
Emantor, line 209: Now we can provide `bios` via the client arguments and than set `_bios_fname` via the `set_bios()` function. Why don't you modify `bios` instead?

Emantor (changes requested):

labgrid/driver/qemudriver.py
sjg20: Well in this case 'bios' would not be provided.

The point here is that the strategy can provide the firmware, no matter what board type is chosen. We don't have to have a different 'bios' for every architecture / target.

Emantor: In case this driver is used by your strategy it is not provided, but it is still possible. Please modify this to use `self.bios` instead of the new hidden `self._bios_fname`.

sjg20: OK, I can check for an existing file and then that can work.

sjg20: done

sjg20:
OK I have slimmed this down a bit so PTAL

Emantor (dismissed):

Bastian-Krause (approved):
Force-pushed:
- drop extra_args handling as it is already optional
- drop redundant doc example
- improve docs

Emantor (approved):

PR #1486, opened 2024-08-26
USBVideoDriver: add support LogiLink UA0379

jluebbe:
This is pretty similar to the existing JPEG cameras and supports manual focus control.

Also do some minor clean-ups while we are here.

### DIFF
--- a/labgrid/driver/usbvideodriver.py
+++ b/labgrid/driver/usbvideodriver.py
@@ -63,6 +63,12 @@ class USBVideoDriver(Driver, VideoProtocol):
                 ("mid", "video/x-h264,width=1280,height=720,framerate=25/1"),
                 ("high", "video/x-h264,width=1920,height=1080,framerate=25/1"),
                 ])
+        elif match == (0x0c45, 0x636b): # LogiLink UA0379 / Microdia
+            return ("mid", [
+                ("low", "image/jpeg,width=640,height=480,pixel-aspect-ratio=1/1,framerate=30/1"),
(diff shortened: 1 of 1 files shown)


jonrebm:
Thank you.

Using this branch, I don't seem to have auto focus (with or without extra options) and I still can't seem to influence the focus using options such as:
```
-c 'focus_automatic_continuous=0,focus_absolute=0'
-c 'focus_automatic_continuous=0,focus_absolute=40'
-c 'focus_automatic_continuous=0,focus_absolute=100'
-c 'focus_auto=0,focus_absolute=100'
```

jluebbe:
The control names have changed in the kernel since https://github.com/torvalds/linux/commit/70fa906d6fceb07a49198d2f31cadecc76787419, which is included in v5.16. Thanks to @michaelolbrich for finding it.

It was accepted here: https://patchwork.linuxtv.org/project/linux-media/patch/20210618122923.385938-14-ribalda@chromium.org/

So, to re-enable autofocus, you need to use -c 'focus_automatic_continuous=1'. In a separate PR we'll likely need to detect new/old kernels (via `uname -r`?) and select the corresponding control name...

jonrebm:
I see, thank you. The value range for focus_absolute is larger than I thought. It works for me with fixed focus:

`labgrid-client video -c "focus_absolute=300"`

or automatic focus:

`labgrid-client video -c "focus_automatic_continuous=1"`

Emantor (approved):

Issue #1493, opened 2024-09-11
pyasn1 0.6.1 breaks eaton and poe_mib power backends


Bastian-Krause:
pysnmp depends on pyasn1. `pyasn1.compat.octets` was removed in [pyasn1 0.6.1](https://github.com/pyasn1/pyasn1/releases/tag/v0.6.1) leading to ModuleNotFoundErrors in labgrid's "eaton" and "poe_mib" power backends:

```
_______________ TestNetworkPowerDriver.test_import_backend_eaton _______________

self = <test_powerdriver.TestNetworkPowerDriver object at 0x7f84794bfdd0>

    def test_import_backend_eaton(self):
        pytest.importorskip("pysnmp")
>       import labgrid.driver.power.eaton

tests/test_powerdriver.py:295: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
labgrid/driver/power/eaton.py:2: in <module>
    from ...util.snmp import SimpleSNMP
labgrid/util/snmp.py:1: in <module>
    from pysnmp import hlapi
/opt/hostedtoolcache/Python/3.12.5/x64/lib/python3.12/site-packages/pysnmp/hlapi/__init__.py:7: in <module>
    from pysnmp.proto.rfc1902 import *
/opt/hostedtoolcache/Python/3.12.5/x64/lib/python3.12/site-packages/pysnmp/proto/rfc1902.py:8: in <module>
    from pysnmp.proto import rfc1155, error
/opt/hostedtoolcache/Python/3.12.5/x64/lib/python3.12/site-packages/pysnmp/proto/rfc1155.py:10: in <module>
    from pysnmp.proto import error
/opt/hostedtoolcache/Python/3.12.5/x64/lib/python3.12/site-packages/pysnmp/proto/error.py:9: in <module>
    from pysnmp import debug
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

    #
    # This file is part of pysnmp software.
    #
    # Copyright (c) 2005-2019, Ilya Etingof <etingof@gmail.com>
    # License: https://www.pysnmp.com/pysnmp/license.html
    #
    import logging
>   from pyasn1.compat.octets import octs2ints
[...]

PR #1494, opened 2024-09-11
driver/usbvideodriver: reduce latency of GStreamer pipeline

tretter:
The playbin3 uses a urisourcebin to open and read the uri. The urisourcebin detects the fd as a stream uri and internally crates a multiqueue with a default buffer duration of 5 s and a high watermark of 0.60 to buffer the stream. This adds 3 s latency to the video stream.

If the source signals that is it a live source, the urisourcebin would avoid buffering of the stream. Unfortunately, the fdsrc doesn't signal that the data is live, and doesn't have a property to change this.

Use buffer-duration to set the size of the queue for buffering to 0 to avoid this additional latency.

The queue could be completely removed by setting "buffering" in the flags property of playbin3 to 0, which would be equal to the behavior of a live source, but the changed flags are more difficult to understand than the buffer duration.

- [x] PR has been tested

Tested by running `labgrid-client video` on a place with a USB camera as resource, moving in front of the camera, and looking the delay of the movement on the screen.

### DIFF
--- a/labgrid/driver/usbvideodriver.py
+++ b/labgrid/driver/usbvideodriver.py
@@ -131,7 +131,7 @@ class USBVideoDriver(Driver, VideoProtocol):
         tx_cmd = self.video.command_prefix + ["gst-launch-1.0", "-q"]
         tx_cmd += pipeline.split()
         rx_cmd = ["gst-launch-1.0"]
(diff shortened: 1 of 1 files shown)


Bastian-Krause:
CI errors are unrelated, see #1493.

jluebbe (approved):

PR #1497, opened 2024-09-24
Use PySNMP >6 asyncio API

### LINKED ISSUE #1493: pyasn1 0.6.1 breaks eaton and poe_mib power backends

Emantor:
**Description**
Update pysnmp and it's users to use the new async API, this should fix CI and also ensures we can properly use pysnmp in the future.
Pushed additional fixes for no longer required pylint disables.

**Checklist**
- [x] PR has been tested

Fixes https://github.com/labgrid-project/labgrid/issues/1493
Closes #1456

### DIFF
--- a/labgrid/util/snmp.py
+++ b/labgrid/util/snmp.py
@@ -1,5 +1,6 @@
-from pysnmp import hlapi
+import pysnmp.hlapi.v3arch.asyncio as hlapi
 from ..driver.exception import ExecutionError
+from .loop import ensure_event_loop, is_new_loop
 
 
 class SimpleSNMP:
@@ -8,24 +9,42 @@ class SimpleSNMP:
         if port is None:
             port = 161
 
+        self.loop = ensure_event_loop()
+
         self.engine = hlapi.SnmpEngine()
-        self.transport = hlapi.UdpTransportTarget((host, port))
+        self.transport = self.loop.run_until_complete(hlapi.UdpTransportTarget.create((host, port)))
         self.community = hlapi.CommunityData(community, mpModel=0)
         self.context = hlapi.ContextData()
 
     def get(self, oid):
-        g = hlapi.getCmd(self.engine, self.community, self.transport,
-            self.context, hlapi.ObjectType(hlapi.ObjectIdentity(oid)),
-            lookupMib=False)
+        g = self.loop.run_until_complete(
+            hlapi.getCmd(
+                self.engine,
+                self.community,
+                self.transport,
+                self.context,
+                hlapi.ObjectType(hlapi.ObjectIdentity(oid)),
+                lookupMib=False,
+            )
+        )
 
-        error_indication, error_status, _, res = next(g)
+        error_indication, error_status, _, res = g
         if error_indication or error_status:
             raise ExecutionError("Failed to get SNMP value.")
         return res[0][1]
 
     def set(self, oid, value):
-        identify = hlapi.ObjectType(hlapi.ObjectIdentity(oid),
-                   hlapi.Integer(value))
-        g = hlapi.setCmd(self.engine, self.community, self.transport,
-            self.context, identify, lookupMib=False)
-        next(g)
+        identify = hlapi.ObjectType(hlapi.ObjectIdentity(oid), hlapi.Integer(value))
+        g = self.loop.run_until_complete(
+            hlapi.setCmd(self.engine, self.community, self.transport, self.context, identify, lookupMib=False)
+        )
+
+        error_indication, error_status, _, _ = g
+        if error_indication or error_status:
+            raise ExecutionError("Failed to set SNMP value.")
+
+    def cleanup(self):
+        self.engine.closeDispatcher()
+        if is_new_loop(self.loop):
+            self.loop.run_until_complete(self.loop.shutdown_asyncgens())
+            self.loop.close()
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -200,6 +200,7 @@ autodoc_mock_imports = ['onewire',
                         'gi.repository',
                         'vxi11',
                         'pysnmp',
+                        'pysnmp.hlapi.v3arch.asyncio',
                         'kasa',
                         'kasa.iot',
                         ]
--- a/labgrid/driver/power/eaton.py
+++ b/labgrid/driver/power/eaton.py
@@ -14,6 +14,7 @@ def power_set(host, port, index, value):
     outlet_control_oid = "{}.{}.0.{}".format(OID, cmd_id, index)
 
     _snmp.set(outlet_control_oid, 1)
+    _snmp.cleanup()
 
 
 def power_get(host, port, index):
@@ -24,6 +25,7 @@ def power_get(host, port, index):
 
     value = _snmp.get(output_status_oid)
 
+    _snmp.cleanup()
     if value == 1:  # On
         return True
     if value == 0:  # Off
--- a/labgrid/driver/power/poe_mib.py
+++ b/labgrid/driver/power/poe_mib.py
@@ -12,6 +12,7 @@ def power_set(host, port, index, value):
     oid_value = "1" if value else "2"
 
     _snmp.set(outlet_control_oid, oid_value)
+    _snmp.cleanup()
 
 def power_get(host, port, index):
     _snmp = SimpleSNMP(host, 'private', port=port)
@@ -19,6 +20,7 @@ def power_get(host, port, index):
 
     value = _snmp.get(output_status_oid)
 
+    _snmp.cleanup()
     if value == 1:  # On
         return True
     if value == 2:  # Off
--- a/labgrid/driver/power/raritan.py
+++ b/labgrid/driver/power/raritan.py
@@ -17,6 +17,7 @@ def power_set(host, port, index, value):
     outlet_control_oid = "{}.2.1.{}".format(OID, index)
 
     _snmp.set(outlet_control_oid, str(int(value)))
+    _snmp.cleanup()
 
 
 def power_get(host, port, index):
@@ -25,6 +26,7 @@ def power_get(host, port, index):
 
     value = _snmp.get(output_status_oid)
 
+    _snmp.cleanup()
     if value == 7:  # On
         return True
     if value == 8:  # Off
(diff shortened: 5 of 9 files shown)


labgrid/util/snmp.py
```diff
@@ -8,24 +10,39 @@ def __init__(self, host, community, port=161):
         if port is None:
             port = 161
 
+        try:
+            # if called from async code, try to get current's thread loop
+            self.loop = asyncio.get_running_loop()
+        except RuntimeError:
+            # no previous, external or running loop found, create a new one
+            self.loop = asyncio.new_event_loop()
```
Bastian-Krause, line 18: Client code does the same, but only in `start_session()` when no loop is provided otherwise (and stashes the loop for later use).

How would other code retrieve the newly created event loop from here?

I think this must rely on an existing event loop, similar to EthernetPortManager:

https://github.com/labgrid-project/labgrid/blob/6b541210c6063e97d81e257d2d6bda1444d9ec78/labgrid/resource/ethernetport.py#L251

labgrid/util/snmp.py
```diff
@@ -8,24 +10,39 @@ def __init__(self, host, community, port=161):
...
+            self.context, identify, lookupMib=False))
+
+        error_indication, error_status, _, _ = g
+        if error_indication or error_status:
+            raise ExecutionError("Failed to set SNMP value.")
+
+    def cleanup(self):
+        self.engine.closeDispatcher()
+        self.loop.run_until_complete(self.loop.shutdown_asyncgens())
+        self.loop.close()
```
Bastian-Krause, line 48: Doesn't this close a potential external loop?

labgrid/util/snmp.py
Emantor: But `get_event_loop` is deprecated, so I chose not to use this.

labgrid/util/snmp.py
Emantor: Ah, yes, we should probably stash the fact that we created a loop and only close this if we did.

Emantor: Implemented.

labgrid/util/snmp.py
Bastian-Krause: No, `get_event_loop()` is not deprecated per se: [only if there is no current event loop](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_event_loop). That's why I think we should simply expect a loop here. If there is none, that's an error that needs to be fixed outside of labgrid.

Emantor: Using labgrid as a library means that there is not necessarily an available loop and we need to support this (which is exactly the case where there is no loop).

Emantor: Testing this inside of labgrid client reveals that `get_running_loop` does not find the running client event loop and instead no loop is found and another is created for SNMP.

Bastian-Krause: > Using labgrid as a library means that there is not necessarily an available loop and we need to support this (which is exactly the case where there is no loop).

I disagree. There are certain "entry points" into labgrid where a loop can be passed or labgrid tries to find one or sets a new one up as a fallback. Exporter, coordinator and client are these "entry points". Client and libraries typically use `labgrid.remote.client.start_session()` to start a session.  So that makes sure there is a loop.

For all other cases, code that uses labgrid as a library needs to make sure a loop exists. Maybe that needs to be documented better.

Using `get_event_loop()` in EthernetPortManager was an active decision. Using that makes sure there is an event loop. Otherwise you get a warning (and a new event loop) or an error, depending on the Python version.

> Testing this inside of labgrid client reveals that `get_running_loop` does not find the running client event loop and instead no loop is found and another is created for SNMP.

That's because `get_running_loop()` does not work in sync contexts and there is no other way to achieve that.

Emantor: I guess this is fine for the SNMP switch handling, but isn't the correct requirement for the simple power backend handling we had beforehand. So how do we proceed with this? Requiring an event loop for sync library usage is not really feasible.

Bastian-Krause: How does the simple power backend handling work without a session?

Emantor: > How does the simple power backend handling work without a session?

What do you mean? You can totally describe all of the `NetworkPowerPort`s inside of the YAML configuration and not use a `RemotePlace` to retrieve the resources.

Bastian-Krause: Yes, I see.

See commit messages of da95c444, 0e1e4fba and 606af353 for context.

What about calling `labgrid.remote.client.ensure_event_loop()` whenever we need an event loop? That would restore the original behavior of using labgrid as a library without a coordinator session.

We could add a function that allows library users to pass their event loop into labgrid (stashing it) for use cases without `labgrid.remote.client.start_session()` (which already allows passing an external loop).

Bastian-Krause:
[Reworked](https://github.com/labgrid-project/labgrid/compare/871f4844e7d3b44a5d5a0c8559853e8df8990f82..9c7cb236b32a5298d85e9e278c7a79dee2420261):
- move `ensure_event_loop()` to `labgrid.util.loop` for usage outside client code
- add `is_new_loop()` to check whether `ensure_event_loop()` created a new loop
- use `ensure_event_loop()` and `is_new_loop()` in `labgrid.util.snmp` and `labgrid.resource.ethernetport`

Bastian-Krause:
@Emantor Did you have a look at the rework?

PR #1504, opened 2024-09-27
target: store features and options

Emantor:
**Description**
Instead of always having to revert to the configuration for options and features, store both directly in the target object. Also deprecate the setters and getters for both in the configuration object.

Reasoning is that setters are not required for this, the YAML configuration should be changed instead. The getters are not required if the options and features are bound to the target which is more intuitive anyway.

**Checklist**
- [ ] Tests for the feature 

Closes #704

### DIFF
--- a/labgrid/config.py
+++ b/labgrid/config.py
@@ -199,6 +199,12 @@ class Config:
                 configuration, or if the target can not be found in the
                 configuration.
         """
+        warn(
+            "get_target_option is deprecated, access the options on the target directly.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+
         if target not in self.data['targets']:
             raise KeyError(f"No target '{target}' found in configuration")
 
@@ -224,6 +230,11 @@ class Config:
             KeyError: if the requested target can not be found in the
                 configuration
         """
+        warn(
+            "set_target_option is deprecated, use the YAML configuration instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
         assert isinstance(target, str)
         assert isinstance(name, str)
 
--- a/labgrid/environment.py
+++ b/labgrid/environment.py
@@ -55,6 +55,14 @@ class Environment:
         return self.config.get_features()
 
     def get_target_features(self):
+        warn(
+            "use get_all_features instead. For target features, use target.features instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        return get_all_features()
+
+    def get_all_features(self):
         flags = set()
         for value in self.config.get_targets().values():
             flags = flags | set(value.get('features', {}))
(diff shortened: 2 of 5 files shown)


labgrid/config.py
```diff
@@ -199,6 +199,12 @@ def get_target_option(self, target, name, default=None):
                 configuration, or if the target can not be found in the
                 configuration.
         """
+        warn(
```
Bastian-Krause, line 202:
```suggestion
        warnings.warn(
```

labgrid/config.py
```diff
@@ -224,6 +230,11 @@ def set_target_option(self, target, name, value):
             KeyError: if the requested target can not be found in the
                 configuration
         """
+        warn(
```
Bastian-Krause, line 233:
```suggestion
        warnings.warn(
```

labgrid/environment.py
```diff
@@ -55,6 +55,14 @@ def get_features(self):
         return self.config.get_features()
 
     def get_target_features(self):
+        warn(
```
Bastian-Krause, line 58:
```suggestion
        warnings.warn(
```

labgrid/environment.py
```diff
@@ -55,6 +55,14 @@ def get_features(self):
...
+            "use get_all_features instead. For target features, use target.features instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        return get_all_features()
```
Bastian-Krause, line 63:
```suggestion
        return self.get_all_features()
```

labgrid/config.py
```diff
@@ -199,6 +199,12 @@ def get_target_option(self, target, name, default=None):
                 configuration, or if the target can not be found in the
                 configuration.
         """
+        warn(
+            "get_target_option is deprecated, access the options on the target directly.",
+            DeprecationWarning,
+            stacklevel=2,
```
Bastian-Krause, line 205: Why? This isn't a wrapper method.

labgrid/config.py
```diff
@@ -224,6 +230,11 @@ def set_target_option(self, target, name, value):
             KeyError: if the requested target can not be found in the
                 configuration
         """
+        warn(
+            "set_target_option is deprecated, use the YAML configuration instead.",
+            DeprecationWarning,
+            stacklevel=2,
```
Bastian-Krause, line 236: Why? This isn't a wrapper method.

labgrid/environment.py
```diff
@@ -55,6 +55,14 @@ def get_features(self):
         return self.config.get_features()
 
     def get_target_features(self):
+        warn(
+            "use get_all_features instead. For target features, use target.features instead.",
+            DeprecationWarning,
+            stacklevel=2,
```
Bastian-Krause, line 61: Why? This isn't a wrapper method.

Bastian-Krause (changes requested):

PR #1512, opened 2024-09-30
driver: digitalloggers_restapi: enable REST API

moto-timo:
The legacy HTTP API does not work on newer units without manually changing the configuration in Setup.

This driver is based on https://www.digital-loggers.com/restapi.pdf

Curl examples that were used for development are included as comments so that users can test their access outside of labgrid.

The REST API seems to only allow authenticated users, so the host: parameter is parsed to pass user and password to HTTPDigestAuth. CSRF is also required, so a valid (simple) header is provided. Non-authenticated URLs are supported, but most likely will not work.

HTTPS is recommended, but the units ship with self-signed certificates so SSL certificate verification warnings are intentionally ignored.

Example usage in lg-env.yaml (default as-shipped settings):

NetworkPowerPort:
    model: 'digitalloggers_restapi'
    host: 'http://admin:1234@192.168.0.100'
    index: 0

**Description**

**Checklist**
- [ ] Tests for the feature 

- [ ] PR has been tested


moto-timo:
Added to test_powerdriver.py in second push.

Emantor (dismissed):

Bastian-Krause:
Have you considered contributing this to pdudaemon? Please see our merge policy for network PDUs: #1031

moto-timo:
> Have you considered contributing this to pdudaemon? Please see our merge policy for network PDUs: #1031

There is no support for digitalloggers in pdudaemon but there was support in labgrid... I'm not currently using pdudaemon and the code will be significantly different so no sure when i will have time for that.

I remember conversations about pdudaemon in the past, not opposed to the policy... I almost feel like we should drop network power support in labgrid though to avoid confusion.

Emantor:
Rebased this on current master.

Emantor (approved):

PR #1525, opened 2024-10-15
driver/rawnetworkinterfacedriver: add interface up/down/(basic, EEE, pause) configuration

Bastian-Krause:
**Description**
Add methods to the RawNetworkInterfaceDriver to set interfaces up and down, as well as getting and waiting for the interface state. Also add basic/EEE/pause interface configuration and query support.

This allows the driver to take more control over the interface, preconfiguration is not needed anymore. Tests that expect the exporter interface to be down (such as ethernet selftests, cable tests) are now possible.

Note that the RawNetworkInterfaceDriver now brings the bound interface up on activate and down on deactivate.
Also note that ethtool gained the required `--json` support for the default sub command and for `--show-eee` only in v6.10.

**Checklist**
- [ ] Tests for the feature 
- [x] PR has been tested

### DIFF
--- a/helpers/labgrid-raw-interface
+++ b/helpers/labgrid-raw-interface
@@ -45,7 +45,7 @@ def main(program, options):
     if options.ifname in denylist:
         raise ValueError(f"Interface name '{options.ifname}' is denied in denylist.")
 
-    programs = ["tcpreplay", "tcpdump"]
+    programs = ["tcpreplay", "tcpdump", "ip"]
     if program not in programs:
         raise ValueError(f"Invalid program {program} called with wrapper, valid programs are: {programs}")
 
@@ -57,7 +57,7 @@ def main(program, options):
         args.append(f"--intf1={options.ifname}")
         args.append("-")
 
-    if program == "tcpdump":
+    elif program == "tcpdump":
         args.append("-n")
         args.append(f"--interface={options.ifname}")
         # Write out each packet as it is received
@@ -78,6 +78,13 @@ def main(program, options):
             args.append(str(options.timeout))
             args.append("-W")
             args.append("1")
+
+    elif program == "ip":
+        args.append("link")
+        args.append("set")
+        args.append("dev")
+        args.append(options.ifname)
+        args.append(options.action)
 
     try:
         os.execvp(args[0], args)
@@ -101,6 +108,11 @@ if __name__ == "__main__":
     # tcpreplay
     tcpreplay_parser = subparsers.add_parser("tcpreplay")
     tcpreplay_parser.add_argument("ifname", type=str, help="interface name")
+
+    # ip
+    ip_parser = subparsers.add_parser("ip")
+    ip_parser.add_argument("ifname", type=str, help="interface name")
+    ip_parser.add_argument("action", type=str, choices=["up", "down"], help="action, one of {%(choices)s}")
 
     args = parser.parse_args()
     try:
--- a/helpers/labgrid-raw-interface
+++ b/helpers/labgrid-raw-interface
@@ -9,6 +9,7 @@
 
 import argparse
 import os
+import string
 import sys
 
 import yaml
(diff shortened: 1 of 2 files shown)


labgrid/driver/rawnetworkinterfacedriver.py
```diff
@@ -97,6 +97,28 @@ def wait_state(self, expected_state, timeout=60):
...
+        """
+        Returns settings via ethtool of the bound network interface resource.
+        """
+        cmd = self.iface.command_prefix + ["ethtool", "--json", self.iface.ifname]
+        output = subprocess.check_output(cmd, encoding="utf-8")
+        return json.loads(output)[0]
+
+    @Driver.check_active
+    @step(args=["settings"])
+    def configure(self, **settings):
```
jluebbe, line 111: Perhaps these should be named after ethtool? Interfaces have many settings that are not configured via 'ethtool --change'.

helpers/labgrid-raw-interface
```diff
@@ -54,18 +55,67 @@ def main(program, ifname, count):
...
+        if options.subcommand == "change":
+            for arg in options.ethtool_change_args:
+                if arg.startswith("-") or not allowed_chars.issuperset(arg):
+                    raise ValueError(f"ethtool --change arg '{arg}' contains invalid characters")
+
+            args.append("--change")
+            args.append(options.ifname)
+            args.extend(options.ethtool_change_args)
+
+        if options.subcommand == "set-eee":
```
jluebbe, line 102: Shouldn't this use if/elif? Also at the top level?

labgrid/driver/rawnetworkinterfacedriver.py
```diff
@@ -35,6 +45,126 @@ def _wrap_command(self, args):
...
+            if timeout.expired:
+                raise TimeoutError(
+                    f"exported interface {self.iface.ifname} did not go {expected_state} within {timeout.timeout} seconds"
+                )
+            time.sleep(1)
+
+    @Driver.check_active
+    def wait_state(self, expected_state, timeout=60):
+        """Wait until the expected state is reached or the timeout expires."""
+        self._wait_state(expected_state, timeout=timeout)
```
jluebbe, line 98: Why do you add both a `wait_state()` and the hidden `_wait_state()`? It would be simpler to just merge them.

helpers/labgrid-raw-interface
```diff
@@ -75,22 +125,58 @@ def main(program, ifname, count):
...
-    parser.add_argument('count', nargs="?", type=int, default=None, help='amount of frames to capture while recording')
+
+    # tcpreplay
+    tcpreplay_parser = subparsers.add_parser("tcpreplay")
+    tcpreplay_parser.add_argument("ifname", type=str, help="interface name")
+
+    # ip
+    ip_parser = subparsers.add_parser("ip")
+    ip_parser.add_argument("ifname", type=str, help="interface name")
+    ip_parser.add_argument("action", type=str, choices=["up", "down"], help="action, one of {%(choices)s}")
```
jluebbe, line 146: This looks like it would be hard to extend with more options later.

labgrid/driver/rawnetworkinterfacedriver.py
Bastian-Krause: `_wait_state()` is used in `on_activate()`, meaning we can't decorate it with `@Driver.check_active`. That's why I've added both variants.

helpers/labgrid-raw-interface
Bastian-Krause: What should I do instead? `--up`/`--down`?

labgrid/driver/rawnetworkinterfacedriver.py
Bastian-Krause: I've now renamed the methods.

Bastian-Krause:
Rebased on latest master.

helpers/labgrid-raw-interface
jluebbe: Hmm, leave it as is. It can still be changed to a subparser or something custom if needed.

jluebbe (approved):

PR #1531, opened 2024-10-21
github/workflows: upload test results to codecov

jluebbe:


jluebbe:
The resulting output can be seen here: https://app.codecov.io/gh/jluebbe/labgrid/tests/codecov-test

This is a beta feature in Codecov, so this report should become more useful over time.

Emantor (approved):

PR #1533, opened 2024-10-28
add option to use local docker image

b2vn:
Prioritize existing local docker image, before pulling from a registry.

**Description**

Try to run a local docker image from the uri given. If the imaging is not found locally, docker will try to pull it from a registry. This will allow you to create and build an image locally, without having to push it to a registry first.

**Checklist**

- [X] Original documentation for the feature already covers this PR
- [X] Tests for the feature 
- [X] PR has been tested

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -3084,12 +3084,23 @@ Implements:
 
    DockerDriver:
      image_uri: 'rastasheep/ubuntu-sshd:16.04'
+     pull: 'always'
      container_name: 'ubuntu-lg-example'
      host_config: {'network_mode': 'bridge'}
      network_services: [{'port': 22, 'username': 'root', 'password': 'root'}]
 
 Arguments:
   - image_uri (str): identifier of the docker image to use (may have a tag suffix)
+  - pull (str): pull policy, supports "always", "missing", "never". Default is
+    "always"
+
+      - always: Always pull the image and throw an error if the pull fails.
+      - missing: Pull the image only when the image is not in the local
+        containers storage. Throw an error if no image is found and the pull
+        fails.
+      - never: Never pull the image but use the one from the local containers
+        storage. Throw a `docker.errors.ImageNotFound` if no image is found.
+
   - command (str): optional, command to run in the container (depends on image)
   - volumes (list): optional, list to configure volumes mounted inside the container
   - container_name (str): name of the container
--- a/labgrid/driver/dockerdriver.py
+++ b/labgrid/driver/dockerdriver.py
@@ -1,12 +1,41 @@
 """
 Class for connecting to a docker daemon running on the host machine.
 """
+from enum import Enum
+
 import attr
 
 from labgrid.factory import target_factory
 from labgrid.driver.common import Driver
 from labgrid.resource.docker import DockerConstants
 from labgrid.protocol.powerprotocol import PowerProtocol
+
+
+class PullPolicy(Enum):
+    """Pull policy for the `DockerDriver`.
+
+    Modelled after `podman run --pull` / `docker run --pull`.
+
+    * always: Always pull the image and throw an error if the pull fails.
+    * missing: Pull the image only when the image is not in the local
+      containers storage. Throw an error if no image is found and the pull
+      fails.
+    * never: Never pull the image but use the one from the local containers
+      storage. Throw an error if no image is found.
+    * newer: **Note** not supported by the driver, and therefore not
+      implemented.
+    """
+    Always = 'always'
+    Missing = 'missing'
+    Never = 'never'
+
+def pull_policy_converter(value):
+    if isinstance(value, PullPolicy):
+        return value
+    try:
+        return PullPolicy(value)
+    except ValueError:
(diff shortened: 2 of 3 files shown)


Emantor (approved):

b2vn:
@Emantor Could you merge this?

QSchulz:
Was wondering if it shouldn't be as simple as prefixing your container image name + tag/id with `localhost/` ?

e.g. `FROM localhost/rastasheep/ubuntu-sshd:16.04` instead of `FROM rastasheep/ubuntu-sshd:16.04` ?

Emantor:
@b2vn What do you think about @QSchulz question?

QSchulz:
> Was wondering if it shouldn't be as simple as prefixing your container image name + tag/id with `localhost/` ?
> 
> e.g. `FROM localhost/rastasheep/ubuntu-sshd:16.04` instead of `FROM rastasheep/ubuntu-sshd:16.04` ?

I would be interested to know if this works on Docker. It does seem to work fine on podman but talking with other people in another project where I suggested a similar change seems to have brought up that Docker doesn't seem to be making use of the `localhost/` prefix as full-qualified name for local images? [1]

Sooooo, while I'm curious if it would work on Docker, I don't want to be holding off merging this as I am not sure what I suggested actually works on Docker.

[1] https://lore.kernel.org/yocto-docs/20241121-docs-build-dockerfiles-v1-1-3b54e1237bf5@bootlin.com/T/#m1e15fc16bec9d5b3592099dda3b33759275a1b0d

b2vn:
I did try experimenting with `localhost` before I made the changes. To be honest, I don't remember exactly which combination I tried, but at least none of them worked, there for I made that proposed change. I can make another test later this week just to be sure. Still, if your suggestion works it could be a good approach, but it just seems a bit wrong to me, that the default action is to pull, before trying to use a local copy/cache. At least with the changes in the labgrid code, the behavior is deterministic, where the `localhost`-solution, as @QSchulz also says, might seem more like tool specific feature (thinking the minor differences in podman vs docker) rather than a generic solution.

QSchulz:
> I did try experimenting with `localhost` before I made the changes. To be honest, I don't remember exactly which combination I tried, but at least none of them worked, there for I made that proposed change. I can make another test later this week just to be sure. Still, if your suggestion works it could be a good approach, but it just seems a bit wrong to me, that the default action is to pull, before trying to use a local copy/cache. At least with the changes in the labgrid code, the behavior is deterministic, where the `localhost`-solution, as @QSchulz also says, might seem more like tool specific feature (thinking the minor differences in podman vs docker) rather than a generic solution.

OK I clearly got confused.

`podman pull localhost/some-image` will not work even if the image exists as it's going to try to reach a registry on localhost, which may not be running.

OK so, my understanding is you want to be able to use a local image, but that is prevented because a pull is done on the image, which cannot work because it's local.

The issue is that you're modifying the behavior of the driver for remote images I believe?

Indeed, I would expect right now the image to be updated to its latest hash if it's changed whenever `.pull()` is called. With this patch it would only called if the image doesn't exist yet.

I assume pulling the last hash associated with a tag is a pretty common usecase (e.g. using debian:bookworm, I may want to always use the last bookworm hash and not one I have locally already), so I think it wouldn't be a good idea to `.get()` the image and only if it fails `.pull()` it.

I couldn't find a pull policy in the docker python module so I guess the only option would be to call the `.pull()` function only when we want. I guess we could have some sort of `pull` property for the `DockerDriver` which states whether to pull the image at the `image_uri`? We could model it after the `pull` property from `podman run`/`docker run`.

What do you think?

b2vn:
I like that approach - it will make it really clear what the driver will do, while also ensuring that this mod will not affect any existing users.

I will work on that today.

doc/configuration.rst
```diff
@@ -3084,12 +3084,23 @@ Implements:
 
    DockerDriver:
      image_uri: 'rastasheep/ubuntu-sshd:16.04'
+     pull: 'newer'
```
QSchulz, line 3087: issue: not implemented

`newer` doesn't exist (not implemented in this change), but `never` does, did you mean `never` here?

b2vn: Ah yes, well spotted. Actually a bit worse - I meant to write `always` as it will then be the same behavior as the previous implementation.

QSchulz: but if it's already the default, why mention it in the example :) ?

b2vn: I think it is good practice to show the options. I would rather have the users using a function they might not need, than having them overlook it's existence, and therefore spend hours looking through documentation and source code (as I did).

Beside from that, I don't have a strong opinion whether it should be `always` or `never` - you get the final say in this :)

b2vn: So, you get to choose the value, `always`, `missing` or `never` :)?

QSchulz: `always` is fine so we don't need to have a new push just for that :)

b2vn:
I assume then the changes are then ready for review/merge again.

b2vn:
@Emantor, could you merge this if you also agree with our changes?

Emantor (approved):

PR #1538, opened 2024-11-06
labgrid/driver/power: Backend for Ubiquity mFi mPower

enkiusz:
**Description**

This module adds a backend driver for the Ubiquity mFi mPower controllable power strip. Only basic power on/off and status are supported, the web UI of the device itself has more information such as current power consumption. The code has been tested using local hardware. The simple REST driver could not be used for this device as it requires a login sequence in order to access the HTTP API.

**Checklist**
- [X] PR has been tested

### DIFF
--- a/labgrid/driver/power/mfi_mpower.py
+++ b/labgrid/driver/power/mfi_mpower.py
@@ -1,0 +1,99 @@
+"""
+    Controls the *Ubiquity mFi mPower* Power Strip with Ethernet and Wi-Fi connectivity via HTTP.
+    Reference: https://dl.ubnt.com/guides/mfi/mFi_mPower_PRO_US_QSG.pdf
+
+    Example configuration to use port #3 on a device with URL 'http://172.17.180.53/'
+    with the default credentials ('ubnt' for both username and password):
+
+    NetworkPowerPort:
+      model: mfi_mpower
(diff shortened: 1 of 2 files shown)


Emantor (approved):

labgrid/driver/power/mfi_mpower.py
```diff
@@ -0,0 +1,99 @@
...
+
+# Obtain credentials and repack base_url if needed
+def get_credentials(base_url: str) -> Tuple[str, dict]:
+    base_url = urlparse(base_url)
+
+    if base_url.username is None or base_url.password is None:
+        credentials = dict(username='ubnt', password='ubnt')
+    else:
+        credentials = dict(username=base_url.username, password=base_url.password)
+        base_url._replace(netloc=base_url.netloc.replace(f'{base_url.username}:{base_url.password}@', ''))
```
Bastian-Krause, line 46: @enkiusz See https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlsplit
> The _replace() method will return a new SplitResult object replacing specified fields with new values.

The return value is not used at all. Maybe you want to fix that?

enkiusz: Good catch, fixed in https://github.com/labgrid-project/labgrid/pull/1925

PR #1541, opened 2024-11-12
gRPC authentication proposal - using gRPC secure channels

JacekBartynowski:
**Description**

**Checklist**
- [ ] Tests for the feature 

- [x] PR has been tested


JacekBartynowski:
Here is a proposal for securing the gRPC channels (SSL encryption).
The authentication code will be reworked:
- possibly abstract interface would be added
- the concrete implementation would be done via additional package + certain 'entry points' as a way of the integration

ep1cman:
I imagine the private key (and thus the certificates?) would need to be generated on first start rather than included in the code, for anyone to see and copy?

JacekBartynowski:
These are self signed keys only for development and debugging purposes

Sent from Proton Mail Android

-------- Original Message --------
On 12/11/2024 11:49, Sebastian Goscik wrote:
I imagine the private key (and thus the certificates?) would need to be generated on first start rather than included in the code, for anyone to see and copy?

JacekBartynowski:
I update the Pull request, mainly added the instrumentation to select custom authentication plugins for the gRPC authentication.
These plugins would be delivered as a separate Python packages installed within the labgrid Python virtual environment and integrated/instantiated using Python _'entry-points'_ instrumentation.

JacekBartynowski:
I added instruction on generating SSL certificate and key + sample configuration file that could be re-used for this purpose. I updated the certificate + key files, they should contain now generic configuration.

Bastian-Krause:
See also #1885.

PR #1549, opened 2024-11-21
util/ssh: Enable StrictHostKeyChecking SSH option

onurcelep:
**Description**
Enables setting StrictHostKeyChecking option for SSH client by introducing LG_SSH_STRICT_HOST_KEY_CHECKING environment variable.

**Checklist**
- [ ] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/labgrid/util/ssh.py
+++ b/labgrid/util/ssh.py
@@ -18,6 +18,8 @@ __all__ = ['sshmanager', 'SSHConnection', 'ForwardError']
 def get_ssh_connect_timeout():
     return int(os.environ.get("LG_SSH_CONNECT_TIMEOUT", 30))
 
+def get_ssh_strict_host_key():
+    return os.environ.get("LG_SSH_STRICT_HOST_KEY_CHECKING", "yes").lower()
 
 @attr.s
 class SSHConnectionManager:
@@ -427,9 +429,8 @@ class SSHConnection:
     def _start_own_master(self):
         """Starts a controlmaster connection in a temporary directory."""
         control = os.path.join(self._tmpdir, f'control-{self.host}')
-
         connect_timeout = get_ssh_connect_timeout()
-
+        strict_host_key = get_ssh_strict_host_key()
         self._logger.debug("ControlSocket: %s", control)
         args = ["ssh"] + SSHConnection._get_ssh_base_args()
         args += [
@@ -439,7 +440,7 @@ class SSHConnection:
             "-o", "ControlMaster=yes",
             "-o", f"ControlPath={control}",
             # We don't want to ask the user to confirm host keys here.
-            "-o", "StrictHostKeyChecking=yes",
+            "-o", f"StrictHostKeyChecking={strict_host_key}",
             self.host,
         ]
 
--- a/man/labgrid-client.1
+++ b/man/labgrid-client.1
@@ -27,7 +27,7 @@ level margin: \\n[rst2man-indent\\n[rst2man-indent-level]]
 .\" new: \\n[rst2man-indent\\n[rst2man-indent-level]]
 .in \\n[rst2man-indent\\n[rst2man-indent-level]]u
 ..
-.TH "LABGRID-CLIENT" "1" "2017-04-15" "0.0.1" "embedded testing"
+.TH "LABGRID-CLIENT" 1 "2017-04-15" "0.0.1" "embedded testing"
 .SH NAME
 labgrid-client \- labgrid-client interface to control boards
 .SH SYNOPSIS
@@ -124,6 +124,10 @@ CI pipelines where the username may not be consistent between pipeline stages.
 .sp
 Set the connection timeout when using SSH (The \fBConnectTimeout\fP option). If
 unspecified, defaults to 30 seconds.
+.SS LG_SSH_STRICT_HOST_KEY_CHECKING
+.sp
+Set strict host key checking status for SSH client. (The \fBStrictHostKeyChecking\fP option). If
+unspecified, defaults to \fByes\fP
 .SH MATCHES
 .sp
 Match patterns are used to assign a resource to a specific place. The format is:
@@ -247,9 +251,11 @@ To retrieve a list of places run:
 .INDENT 0.0
 .INDENT 3.5
 .sp
-.EX
+.nf
+.ft C
 $ labgrid\-client places
-.EE
+.ft P
+.fi
 .UNINDENT
 .UNINDENT
 .sp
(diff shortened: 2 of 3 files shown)


onurcelep:
> `StrictHostKeyChecking` is already enabled for connections to exporters. Why would you want to disable that?

by "Enable" I mean being able to modify it :) In my setup exporter runs on a gateway device paired (1-on-1) with device under test (there is an isolated network between). When I try to reach video stream through dockerized exporter it throws strictHostKeyChecking error for exporter host, meaning I would need to add exporter hosts for each pair to my ssh config.

Emantor:
Yes, this is exactly the intention. For every exporter we expect it to be managed and accessible with your SSH key and expect the host key entry to be there already. Do you constantly add/remove containers?

onurcelep:
I am planning to implement a shared pool with a 'bring your own device' approach, allowing remote users to integrate their hardware into a centralized testing environment. In such a setup, users would need to manually add the host keys for each exporter. At scale, this could become a significant overhead for developers and may deter adoption.

Given that Labgrid is a development tool. I believe it should prioritize flexibility, allowing users to modify security features like strict host key checks at their own discretion and risk.In my case it's used in a controlled environment (VPN)

Regarding adding/removing containers, I view the testing infrastructure as inherently dynamic, with containers being spun up or down frequently. This approach aligns with the general principles of containerization, which are about flexibility and lifecycle independence. For example, I am considering using hooks for dynamic resource definitions, which would likely involve modifying or recreating containers (though I haven't explored this aspect in depth yet).

Emantor:
> I am planning to implement a shared pool with a 'bring your own device' approach, allowing remote users to integrate their hardware into a centralized testing environment. In such a setup, users would need to manually add the host keys for each exporter. At scale, this could become a significant overhead for developers and may deter adoption.

You will already need to manage individual SSH Keys for each developer on the exporters, expecting the developers to comply with SSH security mechanisms does not seem far fetched to me.

> Given that Labgrid is a development tool. I believe it should prioritize flexibility, allowing users to modify security features like strict host key checks at their own discretion and risk.In my case it's used in a controlled environment (VPN)

Even when used inside of a VPN I very much dislike disabling one of SSHs security options inside of labgrid. We enforce StrictHostKeyChecking via a `-o` option since we very much want to confirm that the developer knows which machine he is connecting to before running automated commands via labgrid.

Since you will probably want to host an internal DNS for your VPN infrastructure anyway, [host key validation via DNSSEC](https://www.redpill-linpro.com/techblog/2019/05/06/sshfp-and-dnssec.html) may be interesting to you.

> Regarding adding/removing containers, I view the testing infrastructure as inherently dynamic, with containers being spun up or down frequently. This approach aligns with the general principles of containerization, which are about flexibility and lifecycle independence. For example, I am considering using hooks for dynamic resource definitions, which would likely involve modifying or recreating containers (though I haven't explored this aspect in depth yet).

onurcelep:
> > I am planning to implement a shared pool with a 'bring your own device' approach, allowing remote users to integrate their hardware into a centralized testing environment. In such a setup, users would need to manually add the host keys for each exporter. At scale, this could become a significant overhead for developers and may deter adoption.
> 
> You will already need to manage individual SSH Keys for each developer on the exporters, expecting the developers to comply with SSH security mechanisms does not seem far fetched to me.
> 

In our setup, SSH key authentication for developers is handled via LDAP integration, so there's no need to manage individual SSH keys on each exporter. This streamlines access management and reduces administrative overhead.

> > Given that Labgrid is a development tool. I believe it should prioritize flexibility, allowing users to modify security features like strict host key checks at their own discretion and risk.In my case it's used in a controlled environment (VPN)
> 
> Even when used inside of a VPN I very much dislike disabling one of SSHs security options inside of labgrid. We enforce StrictHostKeyChecking via a `-o` option since we very much want to confirm that the developer knows which machine he is connecting to before running automated commands via labgrid.
> 
> Since you will probably want to host an internal DNS for your VPN infrastructure anyway, [host key validation via DNSSEC](https://www.redpill-linpro.com/techblog/2019/05/06/sshfp-and-dnssec.html) may be interesting to you.
> 
Thank you for your emphasis on security and for suggesting DNSSEC for host key validation—I will definitely look into integrating this into our infrastructure.

To clarify, my pull request does not disable StrictHostKeyChecking by default. It introduces an optional configuration that allows users to disable strict host key checking if they choose to, with the default behavior remaining unchanged. This means labgrid's security posture remains the same out of the box.

By making StrictHostKeyChecking configurable, we can reduce operational overhead in dynamic environments like ours without impacting users who rely on the default security settings. This flexibility would help accommodate different use cases while maintaining labgrid's commitment to security.

jluebbe:
> > > I am planning to implement a shared pool with a 'bring your own device' approach, allowing remote users to integrate their hardware into a centralized testing environment. In such a setup, users would need to manually add the host keys for each exporter. At scale, this could become a significant overhead for developers and may deter adoption.
> > 
> > 
> > You will already need to manage individual SSH Keys for each developer on the exporters, expecting the developers to comply with SSH security mechanisms does not seem far fetched to me.
> 
> In our setup, SSH key authentication for developers is handled via LDAP integration, so there's no need to manage individual SSH keys on each exporter. This streamlines access management and reduces administrative overhead.

As far as I know, you can manage SSH host keys ("known hosts") via LDAP as well.
 
> > > Given that Labgrid is a development tool. I believe it should prioritize flexibility, allowing users to modify security features like strict host key checks at their own discretion and risk.In my case it's used in a controlled environment (VPN)
> > 
> > Even when used inside of a VPN I very much dislike disabling one of SSHs security options inside of labgrid. We enforce StrictHostKeyChecking via a `-o` option since we very much want to confirm that the developer knows which machine he is connecting to before running automated commands via labgrid.
> > Since you will probably want to host an internal DNS for your VPN infrastructure anyway, [host key validation via DNSSEC](https://www.redpill-linpro.com/techblog/2019/05/06/sshfp-and-dnssec.html) may be interesting to you.
> 
> Thank you for your emphasis on security and for suggesting DNSSEC for host key validation—I will definitely look into integrating this into our infrastructure.

SSH also has a built-in mechanism to avoid the need for host key confirmations:
https://en.wikibooks.org/wiki/OpenSSH/Cookbook/Certificate-based_Authentication#SSH_Host_Certificates

> To clarify, my pull request does not disable StrictHostKeyChecking by default. It introduces an optional configuration that allows users to disable strict host key checking if they choose to, with the default behavior remaining unchanged. This means labgrid's security posture remains the same out of the box.
> 
> By making StrictHostKeyChecking configurable, we can reduce operational overhead in dynamic environments like ours without impacting users who rely on the default security settings. This flexibility would help accommodate different use cases while maintaining labgrid's commitment to security.

There are many different ways SSH may need to be configured differently, depending on how your lab is built. We don't want to add (and maintain) ad-hoc options to labgrid for each of them, so the proper way is to use OpenSSH's features directly, configured via the normal OpenSSH configuration.

onurcelep:
Thanks for clarification and leads for alternative solutions, closing the PR

PR #1552, opened 2024-11-28
Persist client connection session

gregory-bronstain:
**Description**
Added new feature - clean after failed session:  to client added new string argument --session <SESSION_NAME> in commands
        'acquire', 'reserve' and 'monitor'.
When those  commands being used without session argument, no changes should be seen.
When those commands used with session argument, we keep the session name with place 
        and/or reservation and in case of failed or killed monitor, the coordinator will initiate clean procedure 
        which will cancel all reservations created with same session and release (but not delete!) all places
        acquired (locked) with the same session.
This should allow the coordinator to automatically reveal event of failed client and clean the scene for next clients.
One of intentions here is to use in cloud CI procedures to continue to work properly in case of failure or disappearing of client  in virtual environment.

**Checklist**
- [ ] Documentation for the feature - Added to CHANGES.rst
- [ ] Tests for the feature - Added test_client.py::test_monitor_with_session()

- [ ] PR has been tested


gregory-bronstain:
Will be fixed and finished by arm team

PR #1557, opened 2024-12-10
setuptools_scm: support reading version from archive

alextercete:
**Description**
Currently, it's not possible to `pip install` the `labgrid` package by using the GitHub archive URL. However, `setuptools_scm` [supports](https://setuptools-scm.readthedocs.io/en/latest/usage/#git-archives) this with a few files added to the repo.

**Testing**
This fails:
```
pip install https://github.com/labgrid-project/labgrid/zipball/master
```
This succeeds:
```
pip install https://github.com/arm-software/labgrid/zipball/pip-install-from-archive
```

**Checklist**
- [x] Documentation for the feature: N/A
- [x] Tests for the feature: N/A
- [x] PR has been tested

### DIFF
--- a/.git_archival.txt
+++ b/.git_archival.txt
@@ -1,0 +1,3 @@
+node: $Format:%H$
+node-date: $Format:%cI$
+describe-name: $Format:%(describe:tags=true,match=v[0-9]*)$
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+.git_archival.txt  export-subst


.git_archival.txt
```diff
@@ -0,0 +1,3 @@
+node: $Format:%H$
+node-date: $Format:%cI$
+describe-name: $Format:%(describe:tags=true,match=*[0-9]*)$
```
Emantor, line 3: We should probably adjust this match to something like `v[0-9]+.[0-9]+.[0-9]+`?

Emantor (changes requested):

.git_archival.txt
alextercete: Sure, it makes sense to be more specific. Although it looks like not all [tags](https://github.com/labgrid-project/labgrid/tags) follow that pattern. For example, major releases omit the patch number (e.g. `v24.0`). But I think `v[0-9]+.[0-9]+` should do the trick.

alextercete: It turns out `match` takes a [`glob(7)`](https://www.man7.org/linux/man-pages/man7/glob.7.html) pattern, not a regular expression. So I think it would be best to go for `v[0-9]*` for simplicity.

alextercete:
@Emantor I've made the changes, would you mind taking another look?

alextercete:
@Emantor Just wanted to follow-up on this. Do you need me to do anything else here?

Emantor (approved):

jluebbe:
@alextercete Just out of interest, why would one use the tar archive instead of a git hash/branch?

alextercete:
@jluebbe The hash/branch URL requires `git` to be installed, whereas the archive URL works with just `pip`. This is particularly relevant when using something like Bazel, where [hermeticity](https://bazel.build/basics/hermeticity) is important.

PR #1559, opened 2024-12-15
doc: hostname can have a '-' character

vermaete:
rfc952 specifies:

  A "name" (Net, Host, Gateway, or Domain name) is a text string up
     to 24 characters drawn from the alphabet (A-Z), digits (0-9), minus
     sign (-), and period (.).  Note that periods are only allowed when
     they serve to delimit components of "domain style names".

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -1687,7 +1687,7 @@ Implements:
 .. code-block:: yaml
 
    ShellDriver:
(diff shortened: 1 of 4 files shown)


Emantor:
You'll need to regenerate the changed man pages.

vermaete:
I have updated the man pages and created a new PR (v2)
I'm closing this one

PR #1564, opened 2025-01-13
Add ADB (Android Debug Bridge) Support

ep1cman:
This pull request add support for ADB (Android Debug Bridge) devices in labgrid.

- It supports USB (remote and local) and TCP adb devices.
- It adds two new client commands
  - `adb` for interfacing with the device by the `adb` cli
  - `scrcpy` for remote screen/touch/keyboard/clipboard/etc. sharing
 

**Checklist**
- [ ] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/labgrid/driver/adb.py
+++ b/labgrid/driver/adb.py
@@ -1,0 +1,149 @@
+import subprocess
+from enum import Enum
+
+import attr
+
+from ..factory import target_factory
+from ..protocol import CommandProtocol, FileTransferProtocol, ResetProtocol
+from ..resource.adb import NetworkADBDevice, RemoteUSBADBDevice, USBADBDevice
+from ..step import step
+from ..util.proxy import proxymanager
+from .commandmixin import CommandMixin
+from .common import Driver
+
+# Default timeout for adb commands, in seconds
+ADB_TIMEOUT = 10
+
+
+class ADBRebootMode(Enum):
+    REBOOT = None
+    BOOTLOADER = "bootloader"
+    RECOVERY = "recovery"
+    SIDELOAD = "sideload"
+    SIDELOAD_AUTO_REBOOT = "sideload-auto-reboot"
+
+
+@target_factory.reg_driver
+@attr.s(eq=False)
+class ADBDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol, ResetProtocol):
+    """ADB driver to execute commands, transfer files and reset devices via ADB."""
+
+    bindings = {"device": {"USBADBDevice", "RemoteUSBADBDevice", "NetworkADBDevice"}}
+
+    def __attrs_post_init__(self):
+        super().__attrs_post_init__()
+        if self.target.env:
+            self.tool = self.target.env.config.get_tool("adb")
+        else:
+            self.tool = "adb"
+
+    def on_activate(self):
+        if isinstance(self.device, USBADBDevice):
+            self._base_command = [self.tool, "-s", self.device.serialno]
+
+        elif isinstance(self.device, RemoteUSBADBDevice):
+            self._host, self._port = proxymanager.get_host_and_port(self.device)
+            self._base_command = [self.tool, "-H", self._host, "-P", str(self._port), "-s", self.device.serialno]
+
+        elif isinstance(self.device, NetworkADBDevice):
+            self._host, self._port = proxymanager.get_host_and_port(self.device)
+            # ADB does not automatically remove a network device from its
+            # devices list when the connection is broken by the remote, so the
+            # adb connection may have gone "stale", resulting in adb blocking
+            # indefinitely when making calls to the device. To avoid this,
+            # always disconnect first.
+            # TODO: Replace subprocess.run() with process wrapper once it supports timeouts.
+            subprocess.run(
+                [self.tool, "disconnect", f"{self._host}:{str(self._port)}"],
+                stdout=subprocess.DEVNULL,
+                stderr=subprocess.DEVNULL,
+                timeout=ADB_TIMEOUT,
+                check=False,
+            )
+            subprocess.run(
+                [self.tool, "connect", f"{self._host}:{str(self._port)}"],
+                stdout=subprocess.DEVNULL,
+                stderr=subprocess.DEVNULL,
+                timeout=ADB_TIMEOUT,
+                check=True,
+            )  # Connect adb client to TCP adb device
+            self._base_command = [self.tool, "-s", f"{self._host}:{str(self._port)}"]
+
+    def on_deactivate(self):
+        if isinstance(self.device, NetworkADBDevice):
+            # Clean up TCP adb device once the driver is deactivated
+            subprocess.run(
+                [self.tool, "disconnect", f"{self._host}:{str(self._port)}"],
+                stdout=subprocess.DEVNULL,
+                stderr=subprocess.DEVNULL,
+                timeout=ADB_TIMEOUT,
+                check=True,
+            )
+
+    # Command Protocol
+
+    def _run(self, cmd, *, timeout=30.0, codec="utf-8", decodeerrors="strict"):
+        cmd = [*self._base_command, "shell", cmd]
+        result = subprocess.run(
+            cmd,
+            text=True,  # Automatically decode using default UTF-8
+            capture_output=True,
+            timeout=timeout,
+        )
+        return (
+            result.stdout.splitlines(),
+            result.stderr.splitlines(),
+            result.returncode,
+        )
+
+    @Driver.check_active
+    @step(args=["cmd"], result=True)
+    def run(self, cmd, timeout=30.0, codec="utf-8", decodeerrors="strict"):
+        return self._run(cmd, timeout=timeout, codec=codec, decodeerrors=decodeerrors)
+
+    @step()
+    def get_status(self):
+        return 1
+
+    # File Transfer Protocol
+
+    @Driver.check_active
+    @step(args=["filename", "remotepath", "timeout"])
+    def put(self, filename: str, remotepath: str, timeout: float | None = None):
+        subprocess.run(
+            [*self._base_command, "push", filename, remotepath],
+            stdout=subprocess.DEVNULL,
+            stderr=subprocess.DEVNULL,
+            timeout=timeout,
+            check=True,
+        )
+
+    @Driver.check_active
+    @step(args=["filename", "destination", "timeout"])
+    def get(self, filename: str, destination: str, timeout: float | None = None):
+        subprocess.run(
+            [*self._base_command, "pull", filename, destination],
+            stdout=subprocess.DEVNULL,
+            stderr=subprocess.DEVNULL,
+            timeout=timeout,
+            check=True,
+        )
+
+    # Reset Protocol
+
+    @Driver.check_active
+    @step(args=["mode"])
+    def reset(self, mode: ADBRebootMode | str | None = None):
+        cmd = [*self._base_command, "reboot"]
+
+        if mode is not None:
+            try:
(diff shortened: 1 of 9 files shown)


labgrid/remote/client.py
```diff
@@ -1530,6 +1531,85 @@ async def export(self, place, target):
...
+                    "-s",
+                    resource.serial,
+                ]
+
+                # If a proxy is required, we need to setup a ssh port forward for the port
+                # (27183) scrcpy will use to send data along side the adb port
+                if resource.extra.get("proxy_required") or self.args.proxy:
+                    proxy = resource.extra.get("proxy")
+                    scrcpy_cmd.append(f"--tunnel-host={ip_addr}")
+                    scrcpy_cmd.append(f"--tunnel-port={sshmanager.request_forward(proxy, host, 27183)}")
```
Emantor, line 1596: Is this always the same port? How will this interact with multiple devices connected over ADB to the same host?

labgrid/resource/adb.py
```diff
@@ -0,0 +1,23 @@
+import attr
+
+from ..factory import target_factory
+from .common import NetworkResource, Resource
+
+
+@target_factory.reg_resource
+@attr.s(eq=False)
+class ADBDevice(Resource):
+    serial = attr.ib(validator=attr.validators.instance_of(str))
```
Emantor, line 10: Maybe change this to `serialno` to avoid confusion with `serial` devices?

Emantor (changes requested):
Some minor comments in regard to subcommand handling and attribute naming.

labgrid/driver/adb.py
```diff
@@ -0,0 +1,113 @@
...
+    @step()
+    def get_status(self):
+        return 1
+
+    # File Transfer Protocol
+
+    @Driver.check_active
+    @step(args=["filename", "remotepath", "timeout"])
+    def put(self, filename: str, remotepath: str, timeout: float = ADB_TIMEOUT):
+        subprocess.run([*self._base_command, "push", filename, remotepath], timeout=timeout)
```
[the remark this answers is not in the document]
ep1cman: Do you mean running it with `check=True`? I am unable to find `run_checked`.

labgrid/driver/adb.py
```diff
@@ -0,0 +1,113 @@
...
+            )
+            subprocess.run(
+                ["adb", "connect", f"{self._host}:{str(self._port)}"], stdout=subprocess.DEVNULL, timeout=ADB_TIMEOUT
+            )  # Connect adb client to TCP adb device
+            self._base_command = [self.tool, "-s", f"{self._host}:{str(self._port)}"]
+
+    def on_deactivate(self):
+        if isinstance(self.device, RemoteADBDevice):
+            # Clean up TCP adb device once the driver is deactivated
+            subprocess.Popen(["adb", "disconnect", f"{self._host}:{str(self._port)}"], stderr=subprocess.DEVNULL).wait(
```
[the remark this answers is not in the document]
ep1cman: Good catch, I missed this one.

labgrid/driver/adb.py
```diff
@@ -0,0 +1,113 @@
...
+            self._base_command = [self.tool, "-H", self._host, "-P", str(self._port), "-s", self.device.serial]
+
+        elif isinstance(self.device, RemoteADBDevice):
+            self._host, self._port = proxymanager.get_host_and_port(self.device)
+            # ADB does not automatically remove a network device from its
+            # devices list when the connection is broken by the remote, so the
+            # adb connection may have gone "stale", resulting in adb blocking
+            # indefinitely when making calls to the device. To avoid this,
+            # always disconnect first.
+            subprocess.run(
```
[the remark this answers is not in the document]
ep1cman: I have added `check=True` to all calls to `subprocess.run`, which will raise an exception if they fail.

labgrid/remote/client.py
ep1cman: Good point, I had not tested multiple simultaneous screen sharing sessions. I have added some code to find a free port on the exporter and use that. It is based on the "get_free_port" code within labgrid that gets a free port on the client machine.

Is this worth moving into a more general location? Maybe within `SSHConnection`?

labgrid/resource/adb.py
ep1cman: Changed

labgrid/driver/adb.py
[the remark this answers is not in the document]
Emantor: It's `check_call()` I confused this with the labgrid implementations for the `CommandProtocol`. Sorry!

labgrid/remote/client.py
Emantor: `get_free_port` should already be in `util.helper`, so you should be able to import it directly.

ep1cman: That returns a free port on the machine running the labgrid client, I require a free port on the exporter. The way scrcpy works is that it runs a server on the android device and creates a port forward between it and the ADB host (the exporter in this case)

labgrid/driver/adb.py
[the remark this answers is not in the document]
ep1cman: I believe using the `check=True` argument for `subprocess.run` has the same result as running `subprocess.check_call` so this should now be resolved.

labgrid/remote/client.py
Emantor: You might be able to handle that within the remote resource for the exporter, the `SerialPortExport` does something similar to choose a free port for ser2net.

ep1cman: There is an additional complication to the `scrcpy` command. 

I already use `get_free_port` in the `ADBExport` to get a port for the ADB server, which is started by the `ADBExport`.

The `scrcpy` server is not setup at this time, it is started by the `scrcpy` client when it connects to the device. It first connects via ADB, transfers over the binaries, starts the server, then uses ADB to create a port forward between the device and the ADB host (the exporter). 

I am unable to know at the time of starting the exporter what ports will be available at the time `scrcpy` is launched, which could be weeks later.

alextercete:
Any updates on this? It would be great to get this merged.

labgrid/resource/adb.py
```diff
@@ -0,0 +1,23 @@
+import attr
+
+from ..factory import target_factory
+from .common import NetworkResource, Resource
+
+
+@target_factory.reg_resource
+@attr.s(eq=False)
+class ADBDevice(Resource):
```
Emantor: This should be named `USBADBDevice` to make it clear that it is a locally connected USB device.

labgrid/resource/adb.py
```diff
@@ -0,0 +1,23 @@
...
+@target_factory.reg_resource
+@attr.s(eq=False)
+class NetworkADBDevice(NetworkResource):
+    serialno = attr.ib(validator=attr.validators.instance_of(str))
+    port = attr.ib(converter=int, validator=attr.validators.instance_of(int))
+
+
+@target_factory.reg_resource
+@attr.s(eq=False)
+class RemoteADBDevice(NetworkResource):
```
Emantor: Similarly, this is going to be a `RemoteUSBADBDevice`.

labgrid/remote/client.py
```diff
@@ -2092,9 +2187,17 @@ def main():
     subparser = subparsers.add_parser("version", help="show version")
     subparser.set_defaults(func=ClientSession.print_version)
 
+    subparser = subparsers.add_parser("adb", help="Run Android Debug Bridge")
+    subparser.add_argument("--name", "-n", help="optional resource name")
+    subparser.set_defaults(func=ClientSession.adb)
+
+    subparser = subparsers.add_parser("scrcpy", help="Run scrcpy to remote control an android device")
```
Emantor: We should move this under the ADB subparser, so this becomes `labgrid-client adb scrcpy`.

labgrid/driver/adb.py
```diff
@@ -0,0 +1,120 @@
...
+            self._base_command = [self.tool, "-H", self._host, "-P", str(self._port), "-s", self.device.serialno]
+
+        elif isinstance(self.device, RemoteADBDevice):
+            self._host, self._port = proxymanager.get_host_and_port(self.device)
+            # ADB does not automatically remove a network device from its
+            # devices list when the connection is broken by the remote, so the
+            # adb connection may have gone "stale", resulting in adb blocking
+            # indefinitely when making calls to the device. To avoid this,
+            # always disconnect first.
+            subprocess.run(
```
Emantor, line 56: This should use `processwrapper` from `util.helper`, see for example the `FastbootDriver`.

labgrid/driver/adb.py
```diff
@@ -0,0 +1,120 @@
...
+    @step(args=["filename", "destination", "timeout"])
+    def get(self, filename: str, destination: str, timeout: float = ADB_TIMEOUT):
+        subprocess.run([*self._base_command, "pull", filename, destination], timeout=timeout, check=True)
+
+    # Reset Protocol
+
+    @Driver.check_active
+    @step(args=["mode"])
+    def reset(self, mode=None):
+        valid_modes = ["bootloader", "recovery", "sideload", "sideload-auto-reboot"]
```
Emantor: We should turn this into an enum.

Emantor (changes requested):

labgrid/driver/adb.py
ep1cman: Two questions about this:

 - `processwrapper` doesn't support a timeout, how should we handle this?
 - I don't think I can also do this in `_run` as I need to return a separate stdout, stderr and exit code.

Emantor: We should extend the processwrapper for timeout support, but IMO this is out of scope for this PR. Can you add a `TODO` entry to the first occasion in the driver? Than the subprocess usage should be fine.

I agree for `_run` since this needs explicit handling.

ep1cman: I have added the requested `TODO` comment

labgrid/driver/adb.py
```diff
@@ -0,0 +1,148 @@
...
+                check=False,
+            )
+            subprocess.run(
+                [self.tool, "connect", f"{self._host}:{str(self._port)}"],
+                stdout=subprocess.DEVNULL,
+                stderr=subprocess.DEVNULL,
+                timeout=ADB_TIMEOUT,
+                check=True,
+            )  # Connect adb client to TCP adb device
+            self._base_command = [self.tool, "-s", f"{self._host}:{str(self._port)}"]
```
luke-hackwell, line 70: I'm having an issue due to the host, port being determined and initial connection being done in `__attrs_post_init__`.

In our use case we have a `NetworkADBDevice` whose host and port are dynamically determined, meaning that when `__attrs_post_init__` runs they have the wrong values. Would it make sense to shift this logic to `on_activate`? This way we can assure that we have the `NetworkADBDevice` resource set correctly before we activate the `ADBDriver`.

labgrid/resource/__init__.py
```diff
@@ -52,3 +52,4 @@
 from .laa import LAASerialPort, LAAPowerPort, LAAUSBGadgetMassStorage, \
                  LAAUSBPort, LAAButtonPort, \
                  LAALed, LAATempSensor, LAAWattMeter, LAAProvider
+from .adb import RemoteUSBADBDevice, USBADBDevice
```
luke-hackwell: Can we add `NetworkADBDevice` here

Emantor (approved):

alextercete:
@Emantor Is this waiting on anything or can it be merged?

Emantor:
I usually try to give @jluebbe and @Bastian-Krause at least an opportunity to chime in. But to me this looks good, so I'll go ahead with the merge. This is not in a release yet, so we could still revert/fix any upcoming issues.

PR #1568, opened 2025-01-16
add support for KMTronic usb relay controller

benjamin1313:
This adds support for KMTronic usb relay controller. 

Tests has been done using a KMTronic four channel usb relay controller: https://kmtronic.com/product/2786/usb-relay-controller-four-channel-box.html
Small tests for the resource and driver has been written and run with pytest to see in the driver works. 
A small setup with coordinator and exporter has also been setup to make sure network resources are available and work with the driver,

**Checklist**

- [x] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/labgrid/remote/exporter.py
+++ b/labgrid/remote/exporter.py
@@ -525,6 +525,24 @@ class USBHIDRelayExport(USBGenericExport):
             "index": self.local.index,
         }
 
+@attr.s(eq=False)
+class KMTronicRelayExport(USBGenericExport):
+    """ResourceExport for outputs on KMTronic relays"""
+
+    def __attrs_post_init__(self):
+        super().__attrs_post_init__()
+
+    def _get_params(self):
+        """Helper function to return parameters"""
+        return {
+            "host": self.host,
+            "busnum": self.local.busnum,
+            "devnum": self.local.devnum,
+            "path": self.local.path,
+            "vendor_id": self.local.vendor_id,
+            "model_id": self.local.model_id,
+            "index": self.local.index,
+        }
 
 @attr.s(eq=False)
 class USBFlashableExport(USBGenericExport):
@@ -567,6 +585,7 @@ exports["SiSPMPowerPort"] = SiSPMPowerPortExport
 exports["USBPowerPort"] = USBPowerPortExport
 exports["DeditecRelais8"] = USBDeditecRelaisExport
 exports["HIDRelay"] = USBHIDRelayExport
+exports["KMTronicRelay"] = KMTronicRelayExport
 exports["USBFlashableDevice"] = USBFlashableExport
 exports["LXAUSBMux"] = USBGenericExport
 
--- a/labgrid/remote/exporter.py
+++ b/labgrid/remote/exporter.py
@@ -542,6 +542,7 @@ class KMTronicRelayExport(USBGenericExport):
             "vendor_id": self.local.vendor_id,
             "model_id": self.local.model_id,
             "index": self.local.index,
+            "ports": self.local.ports,
         }
 
 @attr.s(eq=False)
--- a/labgrid/util/agents/kmtronic_relay.py
+++ b/labgrid/util/agents/kmtronic_relay.py
@@ -1,0 +1,38 @@
+import serial
+
+class USBKMTronicRelay:
+    def set_output(self, path, index, status):
+        # second and third is index and status on/off
+        # \xFF\x01\x00 = turn relay 1 off
+        # \xFF\x01\x01 = turn relay 1 on
+        cmd = bytes([255, index, int(status == True)])
+        with serial.Serial(path, 9600) as ser:
+            ser.write(cmd)
+
+    def get_output(self, path, index):
+        # \xFF\x01\x03 will read relay 1 status
+        cmd = bytes([255, index, 3])
+        with serial.Serial(path, 9600) as ser:
+            ser.write(cmd)
+            data = ser.read(3)
+        return data[2]
+
+_relays = {}
+
+def _get_relay(path):
+    if (path) not in _relays:
+        _relays[(path)] = USBKMTronicRelay()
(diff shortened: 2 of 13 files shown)


benjamin1313:
I will note i needed to install pyserial(outside of venv) on my exporter in order to be able to use the driver and resource over the network.

tests/test_kmtronicrelay.py
```diff
@@ -0,0 +1,22 @@
+from labgrid.resource.udev import KMTronicRelay
+from labgrid.driver.kmtronicrelay import KMTronicRelayDriver
+
+
+def test_kmtronicrelay_resource(target):
+    r = KMTronicRelay(target, name=None, match={"ID_SERIAL_SHORT": "AB0LBF2U"})
```
cidlik, line 6: I catch the errors on this test:
```
$ pytest tests/test_kmtronicrelay.py::test_kmtronicrelay_resource
...
E           TypeError: ("'ports' must be <class 'int'> (got None that is a <class 'NoneType'>).", Attribute(name='ports', default=None, validator=<instance_of validator for type <class 'int'>>, repr=True, eq=True, eq_key=None, order=True, order_key=None, hash=None, init=True, metadata=mappingproxy({}), type=None, converter=None, kw_only=False, inherited=False, on_setattr=None, alias='ports'), <class 'int'>, None)
```
The reason is
```
class KMTronicRelay(USBResource):
    ...
    ports = attr.ib(default=None, validator=attr.validators.instance_of(int))
```

If to be honest, I don't understand this definition. In my sense, `default=None` and `attr.validators.instance_of(int)` are incompatible.

Arguably should set `default=1` for `ports`?

labgrid/util/agents/kmtronic_relay.py
```diff
@@ -9,13 +9,14 @@ def set_output(self, path, index, status):
         with serial.Serial(path, 9600) as ser:
             ser.write(cmd)
 
-    def get_output(self, path, index):
+    def get_output(self, path, index, ports):
```
cidlik, line 12: It breaks the working with USB One Relay https://forums.raspberrypi.com/viewtopic.php?t=125621
I don't have four channel relay for testing, and link from PR description is broken, thus I don't know the commands.
If commands differ for one and four channel relays, I suggest to add branch:
```
        if ports == 1:
            # Branch for KMTronic USB One Relay
            cmd = bytes([255, 1, 3])
            with serial.Serial(path, 9600) as ser:
                ser.write(cmd)
                data = ser.read(3)
            return data[2]
```

tests/test_agent.py
```diff
@@ -110,6 +110,11 @@ def test_all_modules():
     methods = aw.list()
     assert 'usb_hid_relay.set' in methods
     assert 'usb_hid_relay.get' in methods
+    aw.load('kmtronix_relay')
```
cidlik, line 113: Should be `kmtronic_relay`?
```
>       with open(filename, 'r') as source_fd:
E       FileNotFoundError: [Errno 2] No such file or directory: '/home/rkuznecov/3dpty/labgrid/.venv/lib/python3.12/site-packages/labgrid/util/agents/kmtronix_relay.py'
```

labgrid/remote/exporter.py
```diff
@@ -517,6 +517,24 @@ def _get_params(self):
             "index": self.local.index,
         }
 
+@attr.s(eq=False)
+class KMTronicRelayExport(USBGenericExport):
+    """ResourceExport for outputs on KMtronic relays"""
```
cidlik, line 522: Should be `KMTronic`?

tests/test_kmtronicrelay.py
benjamin1313: Can see i forgot to set ports in the test. 
The reason for the `default=None` is to force the user to select the amount of ports their kmtronic supports since in this way i will know how many bytes to read when using the read all command. 

Assuming 8 ports on the kmtronic will result in the driver hanging while waiting for more bytes if a user plugs in a kmtronic with less that 8 ports. 

And while assuming 1 does seem fair and I'm not that much against it. It does mean the driver only reads 1 byte from the kmtronic relay and if it only reads one byte at a time we might get problems where it reads the byte value from another relay next time the driver tries reading. A plausible fix for this is clearing the buffer before reading. 
That just leaves the fact the user might be a bit confused if the don't set the port. can set the state of a relay but not read it if it is more that 1. 
So to keep it simple I made the default none. this creates an error so a user will have to set ports.

labgrid/util/agents/kmtronic_relay.py
benjamin1313: Okay, there's not much to be done about that. Thanks for letting me know. 
Do you know if the two port has support for reading from a single relay and or reading from all?

cidlik: Unfortunately, no, I don't know. The only thing that could have helped me, it's a page [1]. But it doesn't have any information about reading.
I would offer you to add extra branch and to print warning or to raise an exception like "we don't know how it works with 2-channel relay".

[1] https://kmtronic.com/product/2783/usb-relay-controller-two-channels.html

benjamin1313: I wrote an email to KMTronic Technical support. And they tell me only the `FF 0x 03` read command is supported on the usb 2 relay controller. Seems going back to `FF 0x 03` being the default is the way to go. Then if port is set to something other that 1 it tries to read from that number of ports using `FF 09 00`

benjamin1313: maybe with a warnning on 2 port since it is not supported on that one.

benjamin1313: @cidlik would you be able to check if the new changes work on the 1 relay controller?

cidlik: I did it. Verified, it works.

PR #1569, opened 2025-01-19
Add support for GPIO Control using libgpiod (as sysfsgpio is deprecated)

pmelange:
**Description**
* Adds a new protocol ButtonProtocol and the drivers ManualButtonDriver, ExternalButtonDriver, and DigitalOutputButtonDriver to simulate pushing a button (via GPIO lines).  This is needed for the tests I will be doing with the logic behind it.
* Adds new resources LibGPIO / NetworkLibGPIO / MatchedLibGPIO which uses the libgpiod python library.  These new resources also support the active_low attribute
* Adds a new field to the existing SysfsGPIO / NetworkSysfsGPIO / MatchedSysfsGPIO resource classes to support the active_low option of a GPIO line, used to invert the logical value of an IO line.
* Adds a new LibGPIODigitalOutputDriver which implements the DigitalOutputProtocol, ResetProtocol, PowerProtocol, and the new ButtonProtocol.
* Adds support to GPIODigitalOutputDriver to support ResetProtocol, PowerProtocol, and the new ButtonProtocol.
* Adds a new LibGPIOExport class
* Updates the GPIOSysFSExport to support the active_low attribute
* Adds support the labgrid-client to support the new NetworkLibGPIO class and support for the Button actions.
* Marks the Sysfs GPIO classes as deprecated
* Updates corresponding documentation
* Updates the examples for sysfsgpio
* Adds examples for libgpio.

As far as new features go...

ButtonProtocol is to be used in my test environment to control pressing the reset button of DUTs. For many of the DUTs, pressing the reset button when powering on provides (via uboot) a TFTP client or server or simply access to a web socket to be able to upload a new firmware image.  A logical button is better for the test environment I am creating since it encapsulates the high/low logic levels of IO ports into something logical.

LibGPIO classes are added to get away from the deprecated sysfsgpio subsystem (see https://www.kernel.org/doc/Documentation/gpio/sysfs.txt).  

Extending the GPIO based drivers (both libgpio and sysfs) to the Reset, Power, and Button protocols was very straight forward and I did wonder why the Reset and Power protocols had not jet been implemented.  They seem like logical extensions to general purpose IO lines.

Adding the active_low attribute was necessary because the IO lines used in the testbed are connected to relays which are active_low.  

---

All development and testing was done on a raspberry pi 4 with opto-coupler relays connected to the GPIO lines to control power and reset buttons.

---

**Checklist**
- [X] Tests for the feature can be found in the examples/sysfsgpio and examples/libgpio directories 
- [X] PR has been tested by pulling high and low IO lines using various calls from the DigitalOutput, Reset, Power and Button protocols, both using the exporter/client and as a standalone python script.


pmelange:
Rebased so that the commits are no longer behind the main branch :)

doc/configuration.rst
```diff
@@ -2213,11 +2267,102 @@ Implements:
...
+.. code-block:: yaml
+
+   ExternalButtonDriver:
+     cmd_press: 'example_command press'
+     cmd_release: 'example_command release'
+     cmd_toggle: 'example_command toggle'
+     cmd_toggle_for: 'example_command toggle_for'
+
+Arguments:
+  - cmd_press (str): command to press the button on the board
```
PaulVittorino, line 2312:
```suggestion
  - cmd_press (str): command to press and hold the button on the board
```

I suggest clarifying that this means the button must be released.

doc/configuration.rst
```diff
@@ -2213,11 +2267,102 @@ Implements:
...
+  - cmd_release (str): command to release the button on the board
+  - cmd_toogle (str): command to toggle the button on the board
```
PaulVittorino, line 2314:
```suggestion
  - cmd_toogle (str): command to press and release  the button on the board
```

labgrid/driver/buttondriver.py
```diff
@@ -0,0 +1,128 @@
...
+@target_factory.reg_driver
+@attr.s(eq=False)
+class ManualButtonDriver(Driver, ButtonProtocol):
+    """ManualButtonDriver - Driver to tell the user to control a target's button"""
+
+    @Driver.check_active
+    @step()
+    def press(self):
+        self.target.interact(
+            f"Press the button on target {self.target.name} and press enter"
```
PaulVittorino, line 26:
```suggestion
            f"Press and hold the button on target {self.target.name} and press enter"
```

labgrid/driver/buttondriver.py
```diff
@@ -0,0 +1,128 @@
...
+    def release(self):
+        self.target.interact(
+            f"Release the button on the target {self.target.name} press enter"
+        )
+
+    @Driver.check_active
+    @step()
+    def toggle(self):
+        self.target.interact(
+            f"Toggle the bottuon on the target {self.target.name} and press enter"
```
PaulVittorino, line 40:
```suggestion
            f"Toggle the button on the target {self.target.name} and press enter"
```

doc/configuration.rst
```diff
@@ -2213,11 +2267,102 @@ Implements:
...
+  - cmd_toggle_for (str): options command to toggle, pause, and toggle again
```
PaulVittorino, line 2315: The current description reads to me as button down, immediately button up, wait, then button down, immediately button up. Is that correct?

Or did you mean button down, wait, button up? If so, would `press_and_hold_for` be a better name?

pmelange: I must admit that it really isn't clear.  

Actually, the toggle methods don't really apply to a button.  Maybe it would be better to add a new method to DigitialOutputProtocol called "invert".  An "invert_for" method is probably not necessary.

About the button methods, I think of press as "press and hold".  I can understand how people would think that pressing a button is press then release as one movement.  I, on the other hand, think of buttons as being pressed and then released as two movements..  I would like to have it as clear as possible.  I am open to suggestions.  

Currently my thoughts go more towards "press" for press and hold, "release" to release, and "press_for" as press, wait a delay, and release.  I would like to keep the names of the commands short for labgrid-client.

Thanks for catching the spelling errors.  :)

I might have time this evening to make the changes, otherwise they will be made tomorrow.

doc/configuration.rst
pmelange: fixed

labgrid/driver/buttondriver.py
pmelange: fixed

doc/configuration.rst
pmelange: updated as described

labgrid/driver/buttondriver.py
pmelange: fixed

doc/configuration.rst
pmelange: fixed

labgrid/remote/client.py
```diff
@@ -906,6 +938,8 @@ def digital_io(self):
                     drv = self._get_driver_or_new(target, "HttpDigitalOutputDriver", name=name)
                 elif isinstance(resource, NetworkDeditecRelais8):
                     drv = self._get_driver_or_new(target, "DeditecRelaisDriver", name=name)
+                elif isinstance(resource, NetworkLibGPIO):
```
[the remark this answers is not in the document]
pmelange: With all the tests I have done, the line state stays the same.  Also, I am not calling command ```gpioset```

After completion of a command via labgrid-client, such as ```power on```, the line stays on.  Afterwards, I can get with ```power get`` and call ```power cycle``` and the GPIO lines act as expected.  Do you have a specific test scenario that I could use to test with?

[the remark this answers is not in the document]
a3f: > With all the tests I have done, the line state stays the same.

I haven't tested on a Raspberry Pi, but you can try:

 - boot kernel with `pinctrl_bcm2835.persist_gpio_outputs=n`
 - remove optocoupler/latches and work on the raw GPIO pin directly
 - attach weak pull-down
 - set GPIO high
 - terminate GPIO setting program
 - measure GPIO line

Raspberry Pi users have a way out using above module parameter, but in general, it's up to the driver what happens when a line is released.

If this happens to work, one would need to check what is preventing the kernel from calling `bcm2835_pmx_gpio_disable_free()`, which would mux the pin as input.

> Also, I am not calling command gpioset

This behavior is inherent to the GPIOD API, `gpioset` is only one example of a tool affected by it.

> After completion of a command via labgrid-client, such as power on, the line stays on. Afterwards, I can get with power get and call power cycle and the GPIO lines act as expected. Do you have a specific test scenario that I could use to test with?

Undefined can mean that it works for some, because they have the lucky combination of hardware, kernel module parameter, other GPIOs requested on the same controller, ... etc, but it's not universal.

The GPIOD DBus API is one solution for this problem. I don't know if it's a suitable solution for Labgrid, but replacing SysFS usage with libgpiod without ensuring the lines remain requested won't cut it, I think.

a3f:
There have been past discussion about this fyi, e.g. https://github.com/labgrid-project/labgrid/pull/468

This shouldn't mean that libgpiod is not wanted (I don't make maintenance decisions anyway), but just so you are aware of past arguments for and against.

labgrid/remote/client.py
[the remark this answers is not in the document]
jluebbe: An agent is just a way to execute some Python code on the exporter. The main issue with the gpiochip/libgpiod API in some cases state is lost as soon as `labgrid-client io high` exits (which also terminates any agents it started on the exporter).

I'm not sure if that has been solved with the daemon and D-Bus API. If so, using that instead of the sysfs could be an option.

pmelange:
I have created a new PR ( #1583) which does not include the LibGPIO stuff.  The new PR adds the new Button Protocol, the "invert" command to the IO calls, and the new active_low attribute.

Working with dbus and libgpio will have to wait until I have more time to work on it.  Using dbus requires using the "testing" repos from Debian and I figured that it might just be a bit too early to integrate it into labgrid.

jubkncht:
Regarding the LibGPIO discussion, this week I tried to use SysfsGPIO with an exporter v24.0.2 on a Pi 3+ with the current bookworm release, which was unsuccessful. I discovered that Bookworm dropped the sysfsgpio support altogether. So I think in the Future it will become unavoidable to switch from the deprecated sysfs to a supported alternative for some plattforms.

pmelange:
> Regarding the LibGPIO discussion, this week I tried to use SysfsGPIO with an exporter v24.0.2 on a Pi 3+ with the current bookworm release, which was unsuccessful. I discovered that Bookworm dropped the sysfsgpio support altogether. So I think in the Future it will become unavoidable to switch from the deprecated sysfs to a supported alternative for some plattforms.

That seems quite strange to me. I don't have a pi 3, but it's using bookworm with sysfsgpio.  Could you please post your export.yaml file?  

The problem with getting libgpio up and running with dbus is that it's really quite new and the librarie versions that libgpiod are based upon are in debian/testing.  I don't think that having debian/testing as minimal requirements for labgrid will be accepted.

jubkncht:
OK, after your comment, i went back and set up different Bookworm versions for testing. I tried the sysfs manually on fresh Boards with no additional software installed, and it seems that sysfs support is only dropped in the latest release, in older Pi OS Versions with Bookworm, everything works fine. So it's no problem for me to stay with an older Version at the moment, but I think it's good to know that Labgrid SysfsGPIO driver will not work with the current Pi OS release and assumable newer versions from now on. 
Yeah, debian/testing also seems not acceptable. At the moment, I also don't know how to solve this problem.

PR #1571, opened 2025-01-20
Various minor enhancements for the 'console' command

sjg20:
Provides an internal console and a way to automatically select a final state

The internal console avoids the problem of losing output when labgrid closes the serial port and re-opens it in the switch from to microcom.

The final state can be used to ensure that the board is off after use, without needing to run labgrid-client again. It is similar to the initial state in that sense.

Together, this PR allows 'labgrid client' to function as a console device suitable for use with U-Boot's test.py, for example

See here for how this is used in a real lab:

https://ci.u-boot.org/u-boot/u-boot/-/jobs/19119

**Checklist**
- [] Documentation for the feature
- [ ] Tests for the feature 
- [x] PR has been tested

### DIFF
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -786,6 +786,8 @@ class ClientSession:
         """Release a previously acquired place"""
         place = self.get_place(place)
         if not place.acquired:
+            if self.args.auto:
+                return
             raise UserError(f"place {place.name} is not acquired")
         _, user = place.acquired.split("/")
         if user != self.getuser():
@@ -1969,6 +1971,9 @@ def get_parser(auto_doc_mode=False) -> "argparse.ArgumentParser | AutoProgramArg
     subparser.set_defaults(func=ClientSession.acquire)
 
     subparser = subparsers.add_parser("release", aliases=("unlock",), help="release a place")
+    subparser.add_argument(
+        "-f", "--force", action="store_true", help="don't raise an error if the place is not acquired"
+    )
     subparser.add_argument(
         "-k", "--kick", action="store_true", help="release a place even if it is acquired by a different user"
     )
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -875,6 +875,15 @@ class ClientSession:
         manager.session = self
         manager.loop = self.loop
 
+    def set_initial_state(self, target):
+        if self.args.state:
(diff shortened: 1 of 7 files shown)


labgrid/remote/client.py
```diff
@@ -988,7 +988,9 @@ async def _console(self, place, target, timeout, *, logfile=None, loop=False, li
             if logfile:
                 logging.warning("--logfile option not supported by telnet, ignoring")
 
-        print(f"connecting to {resource} calling {' '.join(call)}")
+        if logfile:
+            call.append(f"--logfile={logfile}")
```
aparcar: Maybe this should be added to the commit message?

labgrid/remote/client.py
```diff
@@ -1807,6 +1809,9 @@ def main():
     subparser.set_defaults(func=ClientSession.acquire)
 
     subparser = subparsers.add_parser("release", aliases=("unlock",), help="release a place")
+    subparser.add_argument(
+        "-a", "--auto", action="store_true", help="don't raise an error if the place is not acquired"
+    )
```
aparcar, line 1961: Earlier `-a` is used as *acquire*, wouldn't be a `-f` *force* option here be less confusing?

aparcar:
Thanks for these patches, I applied most of them to my staging tree for testing in my lab. Some patches no longer apply against the master branch, could you kindly rebase this?

Regarding the internal console, does it make sense to make it more configurable so it's used for i.e. `pytest` runs, too?

labgrid/remote/client.py
sjg20: OK done

sjg20:
> Thanks for these patches, I applied most of them to my staging tree for testing in my lab. Some patches no longer apply against the master branch, could you kindly rebase this?

Yes, done. Thanks for looking at it.

> 
> Regarding the internal console, does it make sense to make it more configurable so it's used for i.e. `pytest` runs, too?

There is a LG_CONSOLE variable which I set in the pytest setup. Do you have any suggestions on how to make this work for you?

labgrid/remote/client.py
sjg20: OK done, thanks

sjg20:
This has now been pending for over a year, so probably best to close it.

PR #1583, opened 2025-01-27
Gpio invert (acitve-low) and button protocol

pmelange:
**Description**
* Adds a new protocol ButtonProtocol and the drivers ManualButtonDriver, ExternalButtonDriver, and DigitalOutputButtonDriver to simulate pushing a button (via GPIO lines).  This is needed for the tests I will be doing with the logic behind it.
* Adds a new field to the existing SysfsGPIO / NetworkSysfsGPIO / MatchedSysfsGPIO resource classes to support the invert (active-low) option of a GPIO line, used to invert the logical value of an IO line.
* Adds support to GPIODigitalOutputDriver to support ResetProtocol, PowerProtocol, the new ButtonProtocol and implements the new "invert" command for the DigitalOutputProtocol.
* Updates the GPIOSysFSExport to support the invert (active-low) attribute
* Adds support the labgrid-client to support the Button actions and the "invert" action for IO.
* Updates corresponding documentation
* Updates the examples for sysfsgpio

As far as new features go...

ButtonProtocol is to be used in my test environment to control pressing the reset button of DUTs. For many of the DUTs, pressing the reset button when powering on provides (via uboot) a TFTP client or server or simply access to a web socket to be able to upload a new firmware image.  A logical button is better for the test environment I am creating since it encapsulates the high/low logic levels of IO ports into something logical.

Extending the GPIO based drivers to the Reset, Power, and Button protocols was very straight forward and I did wonder why the Reset and Power protocols had not jet been implemented.  They seem like logical extensions to general purpose IO lines.

Adding the invert (active-low) attribute was necessary because the IO lines used in the testbed are connected to relays which are active_low.  

---

All development and testing was done on a raspberry pi 4 with opto-coupler relays connected to the GPIO lines to control power and reset buttons.

---

**Checklist**
- [X] Tests for the feature can be found in the examples/sysfsgpio directory 
- [X] PR has been tested by pulling high and low IO lines using various calls from the DigitalOutput, Reset, Power and Button protocols, both using the exporter/client and as a standalone python script.


pmelange:
This PR is a reduced version of #1569 without the libgpiod code.

pmelange:
Rebased to main development branch.

Emantor:
A few notes:
- The drivers implementing the `DigitalOutputProtocol` don't need to implement the `Power`or `ResetProtocol`, the repective adapter drivers `DigitalOutputPowerDriver` and `DigitalOutputResetDriver` should be used on top of the initial driver.
- The other `DigitalOutputDriver`s implement an `invert` property on the resource, the `GpioDigitalOutputDriver` should implement this pattern as well

I don't think the button abstraction is needed, from the labgrid POV we are pulling a GPIO line either high or low, that this corresponds to holding a button should be irrelevant and is best abstracted in the board strategy.

pmelange:
Modified to use the attribute "invert" instead of "active_low".  I kept in the ButtonProtocol and the implementation of the Power, Reset, and Button protocols in the GpioDigitalOutputDriver.  

If there is some way for the DigitalOutputPowerDriver to control power via SysfsGPIO from labgrid-client, then I don't seem to be able to get it to work.  If someone can provide an example where such a configuration works, please show it.

Also, rebased to the main branch.

pmelange:
PR #1609 created with a subset of these changes without the ButtonProtocol.

PR #1584, opened 2025-01-28
agentwrapper: Allow specifying agent prefix

benjamin1313:
Allows client to specify where on the exporter .labgrid_agent_{agent_hash}.py should be copied to. Default is the user homedir.

**Description**

**Checklist**
- [ ] Tests for the feature 

- [ ] PR has been tested

### DIFF
--- a/labgrid/util/agentwrapper.py
+++ b/labgrid/util/agentwrapper.py
@@ -46,12 +46,13 @@ class AgentWrapper:
         agent = os.path.join(
             os.path.abspath(os.path.dirname(__file__)),
             'agent.py')
+        agent_prefix = os.environ.get("LG_AGENT_PREFIX", "")
         if host:
             # copy agent.py and run via ssh
             with open(agent, 'rb') as agent_fd:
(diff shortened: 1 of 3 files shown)


Emantor (commented):
Can you elaborate a bit more on why you need this? Uploading these files to other directories than the users home directory could mean that they can be modified by a different user and this may provide a security risk.

benjamin1313:
> Can you elaborate a bit more on why you need this? Uploading these files to other directories than the users home directory could mean that they can be modified by a different user and this may provide a security risk.

We are running the labgrid exporter as a systemd portable service. The homedir for the exporter user is read-only (this is on purpose). We have tmpfs dirs available for the service we use for this purpose. They are local to the portable service, so not available to other users.
With a agent prefix we get more control over where the agent is saved allowing us to place it in a tmpfs dir instead of the homedir.

Emantor (approved):

Bastian-Krause (approved):

PR #1587, opened 2025-01-30
Update UUUDriver

tafilz:
**Description**
Changes default binary name from `uuu-loader` to `uuu`. The driver will use the usb-path to pass it to `-m` parameter to specifiy the usb-device to monitor.

With this multiple i.MX devices can be attached to one exporter device.
This has been tested on a Raspberry Pi with 3 attached i.MX8M Nano SoCs in SDP-Mode.
The uuu-tool used the specified usb-device.

**Checklist**

   - not needed
- [x] PR has been tested

### DIFF
--- a/labgrid/driver/usbloader.py
+++ b/labgrid/driver/usbloader.py
@@ -157,6 +157,14 @@ class RKUSBDriver(Driver, BootstrapProtocol):
 @target_factory.reg_driver
 @attr.s(eq=False)
 class UUUDriver(Driver, BootstrapProtocol):
+    """
+    This is a driver for NXPs uuu tool.
+    https://github.com/nxp-imx/mfgtools
+
+    It uses the -m parameter of uuu to lock in the specific usb device.
+    Everything specified in the script attribute will be executed with '-b'
+    parameter.
+    """
     bindings = {
         "loader": {"IMXUSBLoader", "NetworkIMXUSBLoader", "MXSUSBLoader", "NetworkMXSUSBLoader"},
     }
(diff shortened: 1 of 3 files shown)


Emantor (commented):
Since we are changing the binary name, is that something that NXP has done since a specific uuu release? If so can you link to the release? It's probably better to add a new driver based on uuu instead of reusing the older driver to not break existing users.

Emantor (commented):
Since we are changing the binary name, is that something that NXP has done since a specific uuu release? If so can you link to the release? It's probably better to add a new driver based on uuu instead of reusing the older driver to not break existing users.

tafilz:
> Since we are changing the binary name, is that something that NXP has done since a specific uuu release? If so can you link to the release? It's probably better to add a new driver based on uuu instead of reusing the older driver to not break existing users.

I never found in the documentation mentioned `uuu-loader`. The documentation also links to mfgtools repo.
I just skimmed through the tags of mfgtools, but didn't find any reference to `uuu-loader`.

Emantor:
@jremmet what it your opinion on this?

labgrid/driver/usbloader.py
```diff
@@ -181,10 +189,18 @@ def load(self, filename=None):
...
         mf.sync_to_resource()
 
-        cmd = ['-b', self.script] if self.script else []
+        # Convert usb path to the format the uuu-tool expects
+        # 1-1.2.1 -> 1:121
+        usb_device = str(self.loader.path).replace('.', '').replace('-', ':')
+
+        cmd = (
+            [self.tool, '-v', '-m', usb_device]
+            + ['-b', self.script] if self.script else []
```
tafilz, line 198:
```suggestion
            + (['-b', self.script] if self.script else [])
```
I forgot to retest after refactoring the `cmd`. I'll force push the change

hundeboll (approved):

Issue #1585, opened 2025-01-29
docker: build.sh: has_buildx returns false positive


sgross-emlix:
# Reproduction
```
dockerfiles/build.sh
```

# Observed Behaviour (with -x)

```
+ export DOCKER_BUILDKIT=1
+ DOCKER_BUILDKIT=1
+ : docker.io/labgrid/
+ : latest
+ main
+ local script_dir version
+ has_docker
+ command -v docker
+++ realpath ./build.sh
++ dirname /home/sgross/workspace/emlix/labgrid/dockerfiles/build.sh
+ script_dir=/home/sgross/workspace/emlix/labgrid/dockerfiles
++ python -m setuptools_scm
/usr/bin/python: No module named setuptools_scm
+ version=
++ get_docker_cmd ''
++ local docker_cmd
++ docker_cmd=
++ '[' -n '' ']'
++ has_docker
++ command -v docker
++ echo docker
+ docker_cmd=docker
+ cd /home/sgross/workspace/emlix/labgrid/dockerfiles/..
+ has_buildx docker
+ local docker_cmd
+ docker_cmd=docker
+ docker buildx --help
+ perform_docker_buildx_build docker /home/sgross/workspace/emlix/labgrid/dockerfiles ''
+ local docker_cmd script_dir version
+ docker_cmd=docker
+ script_dir=/home/sgross/workspace/emlix/labgrid/dockerfiles
+ version=
+ extra_args=("${@:4}")
+ for t in client exporter coordinator
+ docker buildx build --build-arg VERSION= --target labgrid-client -t docker.io/labgrid/client:latest -f /home/sgross/workspace/emlix/labgrid/dockerfiles/Dockerfile .
unknown flag: --build-arg
See 'docker --help'.
```

# Expected Behaviour

labgrid docker images should be build as stated in `dockerfiles/README.rst`

# System

## Linux
```
Linux phoebe 6.12.8-arch1-1 #1 SMP PREEMPT_DYNAMIC Thu, 02 Jan 2025 22:52:26 +0000 x86_64 GNU/Linux
```
## Distro
```
NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
[...]

PR #1597, opened 2025-02-04
dockerfiles/build.sh: fix buildx detection

Bastian-Krause:
**Description**
`docker <subcommand> --help` cannot be used to detect whether a sub command is available because it exits with 0 no matter if the sub command exists or not.

Use `docker buildx version` as a test for buildx support instead.

**Checklist**
- [ ] PR has been tested

Fixes: #1585


sgross-emlix:
LGTM

Emantor (approved):

PR #1600, opened 2025-02-09
doc: PDU, hostname regex improvements

vermaete:
Added the full "Power Distribution Unit" next to the first time the abbreviation 'PDU' comes in the configuration documentation.


Bastian-Krause (approved):
Thanks for your contribution!

PR #1603, opened 2025-02-12
Add powerdrivers for AMT and IPMI powered PC's

sessl3r:
**Description**

Adds resources/drivers to power on/off/get PC's which can be accessed via AMT/IPMI remotely.

  Just by running `labgrid-client` and Pytest powering such devices on/off, no direct tests implemented.
- if hardware is needed for the feature, which hardware is supported and which
  hardware did you test with?
  Tests were executed using HPz620/HPz640 workstations for AMT and for  example ASRockRack EPYCD8-2T for IPMI.

**Checklist**
- [ ] Tests for the feature 
- [x] PR has been tested

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -10,7 +10,7 @@ Package: labgrid
 Architecture: any
 Pre-Depends: dpkg (>= 1.16.1), python3, ${misc:Pre-Depends}
 Depends: ${python3:Depends}, ${misc:Depends}, ${shlibs:Depends}
-Recommends: openssh-client, microcom, socat, sshfs, rsync, bash-completion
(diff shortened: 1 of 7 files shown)


Emantor:
Has this PR been tested?

sessl3r:
Testing is done. Missed the args attribute we added as for some machines we need to customize the executed IPMI command further.

ep1cman:
This PR has some overlap with: https://github.com/labgrid-project/labgrid/pull/1567

sessl3r:
> This PR has some overlap with: #1567

I see. Not sure if using ìpmi-power` or ìpmitool` makes more sense. Quickly tested `ipmitool` against our machines and it also works for all of them. How would you like to proceed?

PR #1608, opened 2025-02-13
remote/client: consider acquired places on search

gumulka:
When requesting an acquired place, match the pattern only to the acquired places and not to every place.
This also allows to skip the place on the command line if only one place is acquired by the current user.

**Description**

**Checklist**
- [ ] Tests for the feature 
- [x] PR has been tested

I skipped documentation and automatic tests for this feature and I'm not sure about the documentation, because it is a change on how the client behaves from a user perspective.

On the one hand, the behavior before was not documented and only know through usage, the behavior afterwards will also only be known through usage and is not documented. Not sure on where to document it either.

Test case locally was that I created three boards with an serial port:
* board-1
* board-2
* other-1
Afterwards, I acquired board-1 and other-1, then called `labgrid-client console`, which resulted in an error, that there are two places acquired.
`labgrid-client -p board console` resulted in the console and after releasing `board-1`, then `labgrid-client console` resulted in a console.

### DIFF
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -456,13 +456,24 @@ class ClientSession:
                     f"place {place.name} is not acquired on this computer, acquired on {host}. To allow this host, use labgrid-client -p {place.name} allow {self.gethostname()}/{self.getuser()} on the other host"
                 )
 
-    def get_place(self, place=None):
-        pattern = place or self.args.place
+    def get_place(self, place=None, only_acquired: bool = False):
+        pattern = place or self.args.place or ("" if only_acquired else None)
         if pattern is None:
             raise UserError("place pattern not specified")
         places = self._match_places(pattern)
         if not places:
             raise UserError(f"place pattern {pattern} matches nothing")
+        if only_acquired:
+            acquired_places = []
+            for name in places:
+                try:
+                    self._check_allowed(self.places[name])
+                    acquired_places.append(name)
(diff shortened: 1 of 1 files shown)


Emantor:
This change is missing a DCO. Additionally this change uses `""` to retrieve all places which currently works since the remote client synchronizes all place information from the coordinator. The longterm plan for the client is to move the blanket `*` subscription of all place information to a more specific subscription to places that match the pattern specified by the client.

This change would make this harder to implement, since to enumerate all possible acquired places for a user we have to request all places or change the subscription mechanism.

Note that the place pattern can not only be specified with the `-p` command line option, but also via the `LG_PLACE` environment variable. This makes client usage on the CLI more straightforward.

gumulka:
"" is used, because it matches every pattern and is what is working with the current codebase without rewriting more parts about getting places. I could not look into the future of the project when writing it.

I'm well aware of the environment variable. but since I mostly only have one place acquired, but I'm working with multiple terminals, it is easier to not think about what I have acquired and just accept it.
Currently, I have `export LG_PLACE=$(labgrid-client who | grep $(hostname) | grep $USER | xargs | cut -d " " -f 3)` in my history to accomplish what this PR does for me with one command more, but it would have been nice to get rid of one command. (also this fails, if there are two or more place acquired and returns the first place, this safetycheck is nice in the PR for me)

If it is in the way for future development and is hindering it, I'm okay with not accepting it.

PR #1609, opened 2025-02-14
SysfsGPIO: add invert attribute (active-low) and invert method

pmelange:
**Description**

Adds the invert (active-low) attribute to SysfsGPIO, NetworkSysfsGPIO, MatchedSysfsGPIO, GPIOSysFSExport, GpioDigitalOutputDriver and the GpioDigitalOutput agent.

Adds a new method "invert" to the DigitalOutputProtocol which switches the GPIO output line from True to False or False to True.  This is useful for cases where a sensor attached to the gpio line reacts to any change of state (both rising edge and falling edge).  The new invert method is implemented in all the drivers implementing the DigitalOutputProtocol (DeditecRelaisDriver, FileDigitalOutputDriver, GpioDigitalOutputDriver, HttpDigitalOutputDriver, LXAIOBusPIODriver, ManualSwitchDriver, ModbusCoilDriver, OneWirePIODriver, SerialPortDigitalOutputDriver, HIDRelayDriver) and the GpioDigitalOutput agent

Adds the new "invert" method to the "io" calls in the client.

Updates the documentation and the sysfsgpio examples

**Checklist**
- [x] Tests for the feature done on a Raspberry Pi 4 running Debian 11 via the examples/sysfsgpio examples and the labgrid-client command line interface.
- [x] PR has been tested

***Note***
This is a reduced version of PR #1583 without the new ButtonProtocol.


Emantor (commented):
Why do we need an invert method for the drivers when we already have the invert attribute for the resources? If there are devices where this is a problem, this should be encapsulated inside of a strategy since only switching the io-line from the client side may not meed the timing requirements?

Emantor:
Please remove the invert functions, IMO this specific behaviour of your device can be either encapsulated in the device strategy (for i.e. something like a boot mode reset), or in case of enabling/disabling wifi in your test case the test case itself retrieve the current polarity and than set the correct direction.

pmelange:
> Please remove the invert functions, IMO this specific behaviour of your device can be either encapsulated in the device strategy (for i.e. something like a boot mode reset), or in case of enabling/disabling wifi in your test case the test case itself retrieve the current polarity and than set the correct direction.

Done

pmelange:
Added pylint to disable=unused-variable to labgrid/remove/exporter.py line 657

labgrid/util/agents/sysfsgpio.py
```diff
@@ -2,6 +2,7 @@
 This module implements switching GPIOs via sysfs GPIO kernel interface.
 
 Takes an integer property 'index' which refers to the already exported GPIO device.
+Takes a boolean property 'invert' which inverts logical values if set to True (active-low)
```
Emantor, line 5: Why do we have to pass this down to the agent? Isn't it enough to do the invert handling in the driver as for, i.e. `ModbusCoilDriver`?

pmelange: As the export of the gpio line is happening in the ```__init__``` method and setting active_low happs directly afterwards, then knowing the value of invert is needed (see lines 44-46)

Emantor: But we don't have to use the `active_low` property in the sysfs if we invert in the driver or am I missing something?
`set()/get()` can perform the invert without touching the other sysfs property.

Emantor: I am not sure whether changing the start params is not a breaking change in the coordinator <-> exporter interface.

pmelange: Unless there is a way to explicitly test your concern, then I don't really know what to say.  

But I can say that creating a SysfsGpio resource either with the invert flag or without out works fine.  Acquiring those resources works, having the resources as named resources also works.  Accessing the resource either locally or remotely works.

What kind of conditions are you thinking of which are different from the ones described above?

pmelange: Is there some way to test the coordinarot<->exporter interface to see if anything is broken?

pmelange: > I am not sure whether changing the start params is not a breaking change in the coordinator <-> exporter interface.

Since version 25.0 will not be compatible with earlier version, would it make sense to incorporate these changes before the release?

PR #1612, opened 2025-02-17
Video: add screenshot function

Emantor:
**Description**

Add screenshot support to the `VideoProtocol` and implement it for both drivers.
**Checklist**
- [x] PR has been tested (USBVideo part)


jluebbe:
This should probably be exposed in labgrid client as well, perhaps similar to `tmc screen`?

PR #1616, opened 2025-02-19
contrib/systemd/labgrid-exporter: make service restart always and indefinitely

Bastian-Krause:
**Description**
The exporter sets `reexec=True` for known errors only. If other errors occur (e.g. network issues), the exporter exits with 0.

So set `Restart=always` to make sure even these exit conditions lead to a restarted exporter. Drop the now redundant `RestartForceExitStatus`. Add `StartLimitIntervalSec=0` to make it restart indefinitely.

**Checklist**
- [ ] PR has been tested

### DIFF
--- a/contrib/systemd/labgrid-exporter.service
+++ b/contrib/systemd/labgrid-exporter.service
@@ -8,9 +8,9 @@ Environment="PYTHONUNBUFFERED=1"
 # Should contain LG_COORDINATOR configuration
 EnvironmentFile=-/etc/environment
 ExecStart=/path/to/labgrid/venv/bin/labgrid-exporter /etc/labgrid/exporter.yaml
-Restart=on-failure
-RestartForceExitStatus=100
+Restart=always
 RestartSec=30
+StartLimitIntervalSec=0
 DynamicUser=yes
 # Adjust to your distribution (most often "dialout" or "tty")
 SupplementaryGroups=dialout plugdev


Bastian-Krause:
> We also need to configure `StartLimitIntervalSec=0`, otherwise it will just get stuck if the coordinator is unreachable for some time.

True.

> Perhaps also use `RestartSec=60s` instead of the `100ms` default to avoid burning a lot of CPU time in a tight loop.

The service already contains `RestartSec=30`.

Bastian-Krause:
Added `StartLimitIntervalSec=0` and rebased.

jluebbe (approved):

contrib/systemd/labgrid-exporter.service
```diff
@@ -8,9 +8,9 @@ Environment="PYTHONUNBUFFERED=1"
 # Should contain LG_COORDINATOR configuration
 EnvironmentFile=-/etc/environment
 ExecStart=/path/to/labgrid/venv/bin/labgrid-exporter /etc/labgrid/exporter.yaml
-Restart=on-failure
-RestartForceExitStatus=100
+Restart=always
 RestartSec=30
+StartLimitIntervalSec=0
```
pmelange, line 13: Is this a valid option?  I am getting the following in my log files:
```
Feb 25 13:35:52 testlab02 systemd[1]: /etc/systemd/system/labgrid-exporter.service:13: Unknown key 'StartLimitIntervalSec' in section [Service], ignoring.
```

```
(labgrid-venv) pi@testlab02:~/labgrid $ uname -a
Linux testlab02 6.6.74+rpt-rpi-v8 #1 SMP PREEMPT Debian 1:6.6.74-1+rpt1 (2025-01-27) aarch64 GNU/Linux
(labgrid-venv) pi@testlab02:~/labgrid $ lsb_release -a
No LSB modules are available.
Distributor ID:	Debian
Description:	Debian GNU/Linux 12 (bookworm)
Release:	12
Codename:	bookworm
```

jluebbe: Hmm, the StartLimit options need to go in the [Unit] section.

Bastian-Krause: I'll create a PR.

Bastian-Krause: See #1620.

PR #1623, opened 2025-02-28
Barebox/UBoot/SmallUBootDriver: handle hex arguements properly

### LINKED ISSUE #1622: BareboxDriver and (Small)UBootDriver: does not handle hex interrupt arguments (such as Ctrl-C) properly
In the BareboxDriver, UBootDriver, and SmallUBootDriver, the option sent to the bootloader to interrupt boot loading is done with some form of:
```
self.console.write(interrupstring.encode('ASCII'))
```

According to the documentation  
> use \\x03 for CTRL-C
or
> use “\x03” for CTRL-C

Neither of these work.

After using labgrid-client -vv I see the following is written to the console
```
CONSOLE SerialLogger.ma:   SerialDriver(main) > \x03␍␤
DEBUG   SerialDriver(ma:   Write 4 bytes: b'\\x03'
````
or
```
CONSOLE SerialLogger.ma:   SerialDriver(main) > \\x03␍␤
DEBUG   SerialDriver(ma:   Write 5 bytes: b'\\\\x03'
```

Neither of which is correct and causing the boot loader not to be interrupted.

### PR DESCRIPTION
**Pull request by pmelange**
Checks to see if the interrupt string starts with \x and when that is the case, converts the remained of the string to a bytearray before sending the interrupt to the boot loader.

fixes #1622

### DIFF
--- a/labgrid/driver/bareboxdriver.py
+++ b/labgrid/driver/bareboxdriver.py
@@ -166,7 +166,13 @@ class BareboxDriver(CommandMixin, Driver, CommandProtocol, LinuxBootProtocol):
 
             elif index == 1:
                 # we need to interrupt autoboot
-                self.console.write(self.interrupt.encode('ASCII'))
+                interrupt = self.interrupt.encode('ASCII')
+                if self.interrupt.startswith('\\x'):
+                    try:
+                        interrupt = bytearray.fromhex(self.interrupt[2:])
+                    except ValueError:
+                        pass
+                self.console.write(interrupt)
 
             elif index == 2:
                 # we need to enter the password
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: labgrid/driver/bareboxdriver.py [hunk 1]
```diff
@@ -166,7 +166,13 @@ def _await_prompt(self):
 
             elif index == 1:
                 # we need to interrupt autoboot
-                self.console.write(self.interrupt.encode('ASCII'))
+                interrupt = self.interrupt.encode('ASCII')
```
Emantor, line 169: Wondering if we should do the encoding after the hex string detection?

Emantor (changes requested):
We definitely need documentation for this feature.

labgrid/driver/bareboxdriver.py
pmelange: the call to encode transforms the interrupt string from UTF-8 to ASCII.  And since the code was originally converting these strings to ASCII, I left the conversion to ASCII at the top.

If the hex detection will still be working without converting to ASCII is not 100% clear to me.  Also, as @aparcar wrote, this MR is quite dry, but as it is it is correctly sending the bytes.  I would be happy if the labgrid crew were to create their own competing commit.  Then I would happily withdraw this one.

PR #1630, opened 2025-03-09
driver/power: Add support for TAPO devices

MarekSzczypinski:
**Description**
This pull request adds support for `TAPO/TP-Link` programmable power strips and single socket devices using python-kasa.

This feature is used to manage `TAPO` power strip or single socket from Labgrid. It adds `tapo` model to `NetworkPowerPort`. As all `Tapo` devices require credentials this feature requires user to set `KASA_LOGIN` and `KASA_PASSWORD` env variables to work (see [python-kasa website](https://github.com/python-kasa/python-kasa?tab=readme-ov-file#supported-tapo1-devices) for details).

I have tested it manually with Labgrid (client build and run wth Python3.9 and 3.12). Additionally I wrote unit tests for it and also I have run the `tox -r` command (Python: 3.9, 3.10, 3.11 and 3.12). All worked without errors and all the tests passed.

I have used this driver with `Tapo P300` power strip and `Tapo P100` socket. So it should work with all `Tapo` strips and sockets. 

While there already is support for `KASA/TP-Link` products in Labgrid using python-kasa, it doesn't work with TAPO devices as they apparently use some other protocol and require password and login to work. In order to not complicate the implementation of `tplink` model I decided to implement support for `Tapo` devices as a separate one. Additionally KASA products are not available in Europe so I didn't have means to test current (`tplink`) solution, which was one more reason to separate this into another model.

Because of the necessity to create credentials this solution is a bit more complicated than the `tplink` model.

This solution is based on the `tplink` model. Two key differences are:
* Support for single sockets like `P100` that ignores value of `index`
* Requires `Tapo` cloud credentials in order to communicate with devices

**Checklist**
- [x] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/labgrid/driver/power/tapo.py
+++ b/labgrid/driver/power/tapo.py
@@ -73,7 +73,7 @@ async def _power_set(host: str, port: str, index: str, value: bool) -> None:
     await device.update()
 
     if device.children:
-        assert len(device.children) > index, "Trying to access non-existant plug socket on device"
+        assert len(device.children) > index, "Trying to access non-existent plug socket on device"
 
     target = device if not device.children else device.children[index]
     if value:
@@ -98,7 +98,7 @@ async def _power_get(host: str, port: str, index: str) -> bool:
     if not device.children:
         pwr_state = device.is_on
     else:
-        assert len(device.children) > index, "Trying to access non-existant plug socket on device"
+        assert len(device.children) > index, "Trying to access non-existent plug socket on device"
         pwr_state = device.children[index].is_on
     await device.disconnect()
     return pwr_state
--- a/labgrid/driver/power/tapo.py
+++ b/labgrid/driver/power/tapo.py
@@ -14,9 +14,16 @@ Requirements:
 
 import asyncio
 import os
-import sys
 
+import kasa
 from kasa import Credentials, Device, DeviceConfig, DeviceConnectionParameters, DeviceEncryptionType, DeviceFamily
+from packaging.version import parse
+
+
+def _using_old_kasa_api() -> bool:
+    target_kasa_version = parse("0.10.0")
+    current_kasa_version = parse(kasa.__version__)
+    return current_kasa_version < target_kasa_version
 
 
 def _get_credentials() -> Credentials:
@@ -28,10 +35,8 @@ def _get_credentials() -> Credentials:
 
 
 def _get_connection_type() -> DeviceConnectionParameters:
-    # Somewhere between python-kasa 0.7.7 and 0.10.2 the API changed
-    # Labgrid on Python <= 3.10 uses python-kasa 0.7.7
-    # Labgrid on Python >= 3.11 uses python-kasa 0.10.2
-    if sys.version_info < (3, 11):
+    # API changed between versions 0.9.1 and 0.10.0 of python-kasa
+    if _using_old_kasa_api():
         return DeviceConnectionParameters(
             device_family=DeviceFamily.SmartTapoPlug,
             encryption_type=DeviceEncryptionType.Klap,
@@ -49,10 +54,10 @@ def _get_connection_type() -> DeviceConnectionParameters:
 
 def _get_device_config(host: str) -> DeviceConfig:
     # Same as with `_get_connection_type` - python-kasa API changed
-    if sys.version_info < (3, 11):
+    if _using_old_kasa_api():
         return DeviceConfig(
             host=host, credentials=_get_credentials(), connection_type=_get_connection_type(), uses_http=True
-        )
+        )  # type: ignore[call-arg]
     return DeviceConfig(
         host=host, credentials=_get_credentials(), connection_type=_get_connection_type()
     )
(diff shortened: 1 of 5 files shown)


labgrid/driver/power/tapo.py
```diff
@@ -0,0 +1,101 @@
...
+    if username is None or password is None:
+        raise EnvironmentError("KASA_USERNAME or KASA_PASSWORD environment variable not set")
+    return Credentials(username=username, password=password)
+
+
+def _get_connection_type() -> DeviceConnectionParameters:
+    # Somewhere between python-kasa 0.7.7 and 0.10.2 the API changed
+    # Labgrid on Python <= 3.10 uses python-kasa 0.7.7
+    # Labgrid on Python >= 3.11 uses python-kasa 0.10.2
+    if sys.version_info < (3, 11):
```
Emantor: We should instead check the python-kasa version instead of relying on the indirect python dependency here.

Emantor (changes requested):

labgrid/driver/power/tapo.py
MarekSzczypinski: Done. All tests as written in the description of this PR were re-done.

MarekSzczypinski: Hey, @Emantor. Any update on this one? :-)

MarekSzczypinski: @Emantor just a friendly reminder that I'm still waiting for review ;-)

MarekSzczypinski: Hi @Emantor, still waiting for any comment ;)

MarekSzczypinski:
Yet another update to master, nothing changed in my MR

labgrid/driver/power/tapo.py
MarekSzczypinski: @Emantor I've fixed problem with pylint (see my comment below). Can you give it another try? :)

MarekSzczypinski: Hey @Emantor, just a friendly reminder (checked with pylint and tox -r)

MarekSzczypinski:
Yet another update to master, nothing changed in my MR.
tox -r as been re-run and code has been checked with pylint on python 3.9 through 3.14

labgrid/driver/power/tapo.py
MarekSzczypinski: @Emantor maybe someone else would have time to take a look?

MarekSzczypinski: @Emantor just a friendly reminder that I'm still waiting for review ;-)

MarekSzczypinski: @Emantor ping :)

MarekSzczypinski: @Emantor friendly reminder that I'm still here waiting.

threexc:
Hi,

Following the community call, I've been looking this over and trying to compare it to my own PR to extend the `tplink` driver support: https://github.com/labgrid-project/labgrid/pull/1693

I don't think that the functionality added here overlaps with #1693, as despite the devices being similar I believe there are different protocols at play. 

To test, I used the same [HS103](https://www.kasasmart.com/us/products/smart-plugs/kasa-smart-wifi-mini-plug-hs103) I'd used previously with my PR, and the following exporter config:

```
smartplug-1:
  location: shelf
  kasa-smartplug-1:
    cls: 'NetworkPowerPort'
    model: 'tplink'
    host: '192.168.40.5'
```

and I see this error after changing to `model: tapo` with your branch, re-exporting, and running `labgrid-client -p smartplug-test pw on`:

```
labgrid-client: error: failed to create NetworkPowerPort for target 'Target(name='smartplug-test', env=None)' using {'host': '192.168.40.5', 'model': 'tapo'}
This is likely caused by an error in the environment configuration or invalid
resource information provided by the coordinator.
```

We may be able to combine some of the logic in both PRs, but it is worth noting too that they don't follow exactly the same flow - for example, the existing `tplink` driver makes async `Discover.discover_single()` calls as part of the set/get methods, while I don't see that here. 

Maybe we need to align the two sets of code and implement a `tplink_common` module? Alternatively, we combine them into a single `kasa` module to avoid duplication at all? There is some appeal to making them distinct so that users can find them by product name, but maybe not for maintenance :)

@jluebbe @Emantor thoughts?

labgrid/driver/power/tapo.py
```diff
@@ -0,0 +1,108 @@
...
+
+    if device.children:
+        assert len(device.children) > index, "Trying to access non-existent plug socket on device"
+
+    target = device if not device.children else device.children[index]
+    if value:
+        await target.turn_on()
+    else:
+        await target.turn_off()
+    await device.disconnect()
```
ozan956: What if we get an exception on turn on/off ?

labgrid/driver/power/tapo.py
```diff
@@ -0,0 +1,108 @@
...
+            https=False,
+            login_version=2,
+        )
+    # http_port parameter exists in new kasa (>= 0.10.0) API but not in old versions (<= 0.9.1)
+    # pylint: disable-next=unexpected-keyword-arg
+    return DeviceConnectionParameters(
+        device_family=DeviceFamily.SmartTapoPlug,
+        encryption_type=DeviceEncryptionType.Klap,
+        https=False,
+        login_version=2,
```
ozan956: Can you inform us more on why we hard-code these options?

What is login_version=2?

This configuration is expected to work on all devices?

PR #1631, opened 2025-03-10
ShellDriver: add optional arg dest_authorized_keys

### LINKED ISSUE #1626: Request: ShellDriver: put_ssh_key would benefit from a destination file attribute
The call ```put_ssh_key``` assumes a destination file of ```~/.ssh/authroized_keys``` on the DUT.  Yet when an embedded system (such as OpenWrt) uses dropbear as the ssh server daemon, the authorized_keys file may be located elsewhere.  For example, with OpenWrt, the authorized_keys file is found in ```/etc/dropbear/authorized_keys```

An optional attribute such as ```dest_authorized_keys``` with a default value set to ```~/.ssh/authorized_keys``` which can be overridden with an alternative filename for the DUT would be a practical solution.

pmelange:
ShellDriver: add optional arg dest_authorized_keys
    
Add an addition optional arg "dest_authorized_keys" with the default value of ```~/.ssh/authorized_keys```,

An example where this might be used is when an embedded system uses dropbear as it's ssh server daemon.  In such a case, the authorized_keys file is ```/etc/dropbear/authorized_keys```

Additionally, the ```put_key_file``` takes an optional ```dest_authorized_keys``` function parameter which defaults to the above arg.

Update docs

**Checklist**
- [x] Tests for the feature: runtime test with the arg set and without the arg set. 
- [x] PR has been tested locally

Fixes: #1626

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -1699,6 +1699,8 @@ Arguments:
     Can be an empty string.
   - keyfile (str): optional, keyfile to upload after login, making the
     `SSHDriver`_ usable
+  - dest_authorized_keys (str): optional, default="~/.ssh/authorized_keys", 
+    filename of the authorized_keys file
   - login_timeout (int, default=60): timeout for login prompt detection in
     seconds
   - await_login_timeout (int, default=2): time in seconds of silence that needs
--- a/labgrid/driver/shelldriver.py
+++ b/labgrid/driver/shelldriver.py
@@ -1,6 +1,7 @@
 # pylint: disable=unused-argument
 """The ShellDriver provides the CommandProtocol, ConsoleProtocol and
  InfoProtocol on top of a SerialPort."""
+import os
 import io
 import re
 import shlex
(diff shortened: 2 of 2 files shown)


Emantor:
Isn't a keyword argument with default for the `put_ssh_key()` function not enought to achieve this? Do we need the additional driver argument?

pmelange:
> Isn't a keyword argument with default for the `put_ssh_key()` function not enought to achieve this? Do we need the additional driver argument?

put_ssh_key() is called from on_activate().  So, no.  I don't think it would work as you suggested.

https://github.com/labgrid-project/labgrid/blob/0b9dd26940103cde1fc9c72384ee01ef5cdc933d/labgrid/driver/shelldriver.py#L70-L75

pmelange:
> Isn't a keyword argument with default for the `put_ssh_key()` function not enought to achieve this? Do we need the additional driver argument?

put_ssh_key() is called from on_activate().  So, no.  I don't think it would work as you suggested.

https://github.com/labgrid-project/labgrid/blob/0b9dd26940103cde1fc9c72384ee01ef5cdc933d/labgrid/driver/shelldriver.py#L70-L75

> Making this an automatic upload on `keyfile` availability on activation is a design decision I'd like to revise with the original author. I'll have stern talking to with myself, your implementation looks fine and changing this is not something I want to do at this point in time. Is there a specific reason for switching to `"""` for the f-strings? The logging functions will also need to be fixed to not use f-strings.

I can imagine that removing the put_ssh_key() from on_activate() will break a lot of people's code.  So making that change, which is totally the decision of the maintainers, is totally fine with me as I would be aware of it and can modify my code.

I used the ```"""``` f-strings because it seems (from the code that I have looked at within labgrid) that it is the most-often style used.  I personally also find that style quite readable.  Could you please explain what you mean by "The logging functions will also need to be fixed to not use f-strings"?

PR #1632, opened 2025-03-10
Support extra files in UUUDriver, and pass usb device path

hundeboll:
Teach the `UUUDriver` to syncrhonize extra files to the exporter before invoking `uuu`. This is useful the the "image" passed to `uuu` is a script file that references such extra files. In our case, we load the u-boot image twice (first SPL, then U-Boot proper). After U-Boot starts, we load a fitimage using `uuu`'s fastboot support.

While working with `UUUDriver`, we might as well tell it what USB device path to operate, so we can attach multiple targets to a single exporter. Also, avoid messing up the pytest output by add `-v` to the `uuu` command.

This is tested using with and without a coordinator.

**Checklist**
- [ ] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -2541,6 +2541,8 @@ Implements:
          UUUDriver:
            image: 'mybootloaderkey'
            script: 'spl'
+           extra_files:
+             - 'fitImage'
 
    images:
      mybootloaderkey: 'path/to/mybootloader.img'
@@ -2549,6 +2551,7 @@ Arguments:
   - image (str): optional, key in :ref:`images <labgrid-device-config-images>` containing the path
     of an image to bootstrap onto the target
   - script (str): optional, run built-in script with ``uuu -b``, called with image as arg0
+  - extra_files (list): optional, paths to synchronize to exporter before invoking ``uuu``
 
 USBStorageDriver
 ~~~~~~~~~~~~~~~~
--- a/labgrid/driver/usbloader.py
+++ b/labgrid/driver/usbloader.py
@@ -192,8 +192,9 @@ class UUUDriver(Driver, BootstrapProtocol):
             file_mf.hash = self.get_hash(filename)
             file_mf.sync_to_resource()
 
-        # Enable verbose mode to avoid messing up the terminal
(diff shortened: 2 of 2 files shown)


hundeboll:
Hmm... I see now that two of my commits duplicate this (not so much) older PR:
https://github.com/labgrid-project/labgrid/pull/1587

I'm happy to rebase mine, should you merge the other one.

jremmet:
We run into the same issue using `snagrecover`. Here we need multiple files with descripting names. This doesn't fit into the `BootstrapProtocol`.
Maybe we can add multi file support there? We would prefer a dict as input to  be able to use the key as name of the file to be used in `snagrecover`

hundeboll:
> We run into the same issue using `snagrecover`. Here we need multiple files with descripting names. This doesn't fit into the `BootstrapProtocol`. Maybe we can add multi file support there? We would prefer a dict as input to be able to use the key as name of the file to be used in `snagrecover`

Hmm.. As far as I can tell, protocols cannot specify functionality like that.

Instead, we can add a `ManagedFiles` util, that can synchronize multiple files to a single folder. That util can then take either a simple list of sources to transfer, or a dictionary of `src: dst` pairs to transfer instead?

jremmet:
It think the interface `load` with a single files as parameter is not sufficient for this use cases. 
Also in your case, you can only use the `extra_files` from the config. You can't call `load` with the extra files as parameter.
Not sure if others see the need of this too.

hundeboll:
> It think the interface `load` with a single files as parameter is not sufficient for this use cases.
> Also in your case, you can only use the `extra_files` from the config. You can't call `load` with the extra files as parameter.
> Not sure if others see the need of this too.

I see. Yeah, we can expand the `load()` method in the protocol to take an optional dictionay of extra files as you suggest. That parameter can then take precedence over any configured extra files...

But we need the labgrid maintainers to weigh in here...

lauritsriple:
It would be nice if the `load()` method was expanded. That way i could easily expand or make my own driver to use a bmap file together with a wic file for proper sparse flashing.

PR #1633, opened 2025-03-11
helpers: install as setuptools scripts

hundeboll:
Allow using e.g. labgrid-bound-connect also when installing labgrid using pip.

**Description**
We install labgrid using `pip` in a `Containerfile`, but need to use `labgrid-bound-connect` on the exporter. This isn't currently possible, as the two scripts in `./helpers/` aren't installed unless building a debian package.

To make it possible to configure the two helpers as scripts in `pyproject.toml` they must reside in the python package itself, so move the two files from `./helpers/` to `./labgrid/helpers`.

Both scripts are also changed to configure/parse arguments in a function instead of the global scope, as the setuptools scripts configuration needs to call a function from the specified files.

**Checklist**
- [ ] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/debian/labgrid-bound-connect
+++ b/debian/labgrid-bound-connect
@@ -1,0 +1,3 @@
+#!/bin/sh
+
+exec /opt/venvs/labgrid/bin/labgrid-bound-connect "$@"
--- a/debian/labgrid-raw-interface
+++ b/debian/labgrid-raw-interface
@@ -1,0 +1,3 @@
+#!/bin/sh
+
+exec /opt/venvs/labgrid/bin/labgrid-raw-interface "$@"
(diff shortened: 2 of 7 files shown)


jluebbe:
The commit message contains a typo.

Are there any security implications of calling the helpers via the setuptools wrappers?

hundeboll:
> Are there any security implications of calling the helpers via the setuptools wrappers?

I can see a few different scenarios:

1. labgrid checked out as root, installed as non-root (e.g. `pip install --editable .` in python venv)
    Then a malicious user can change the wrapper script, which is then called as root
2. labgrid checked out as non-root, installed as root (`pip install --editable .`)
    Then a malicious user can change the helper script, which is then called as root
3. labgrid installed as root or non-root (e.g. `pip install labgrid`)
    Only the installing user can change both the wrapper and the helper script
4. labgrid installed as debian package (e.g. `dpkg -i labgrid.deb`)
     Both wrapper and helper script are editable by root only.

Case 1 and 2 should be avoided as both helper scripts are called as root, but I don't see how that i different from today. Case 3 should take care if installing as non-root, but that is no different than calling the script directly. Case 4 dosn't change.

More or other concerns?

PR #1637, opened 2025-03-20
doc/configuration: recommend CTRL-d as BareboxDriver interrupt

Bastian-Krause:
**Description**
Since barebox v2025.03.0 CTRL-d is the recommended way to interrupt barebox autoboot for testing [1].

[1] https://git.pengutronix.de/cgit/barebox/commit/?id=ae482477b60e1e308327658a899358746ff2f5b

/cc @a3f


ejoerns (approved):
Not sure if the requiered barebox version should be part of the documentation, too. But since at least the required commit is referenced, fine from my side.

@Bastian-Krause thanks!

Emantor (approved):

PR #1638, opened 2025-03-25
target: poll needed resources only when activating

hundeboll:
No need to wait for resources that aren't used when activating a specific resource.

This reduces invocation times of various operations significantly (at least in our case):

- `labgrid-client power get`
  3.2 -> 1.2 s
- `labrid-client io get ign`
  2.7 -> 1.1 s
- `labgrid-client ssh true`
  4.2 -> 3.1 s

**Description**

Waiting needlessly when interacting with a target can be frustrating, so after some good old `print()` debugging, it turned out to be the polling of resources that took up (at least some of) the time.

**Checklist**
- [ ] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/labgrid/target.py
+++ b/labgrid/target.py
@@ -41,7 +41,7 @@ class Target:
         else:
             input(msg)
 
-    def update_resources(self):
+    def update_resources(self, resources=None):
         """
         Iterate over all relevant resources and deactivate any active but
         unavailable resources.
@@ -49,7 +49,7 @@ class Target:
         if (monotonic() - self.last_update) < 0.1:
             return
         self.last_update = monotonic()
-        for resource in self.resources:
+        for resource in self.resources if resources is None else resources:
             resource.poll()
             if not resource.avail and resource.state is BindingState.active:
                 deactivated = self.deactivate(resource)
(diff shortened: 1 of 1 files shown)


Emantor (approved):
Missing a DCO, otherwise looks good to me.

jluebbe:
@Emantor @hundeboll With this change, how do we notice resources that become unavailable? For example, an IMXUSBLoader that is replaced by a AndroidUSBFastboot?

Emantor:
> @Emantor @hundeboll With this change, how do we notice resources that become unavailable? For example, an IMXUSBLoader that is replaced by a AndroidUSBFastboot?

AFAICS this is still handled by the `RemotePlaceManager.poll()` which does not use `update_resources` at all and instead receives notifications independent from the `await_resources` logic. Or did I understand your question wrong?

hundeboll:
> @Emantor @Bastian-Krause I'd like to revert this for the 25.0 release, if you agree.

Fine by me. I'm not that well into labgrid internals, so a proper fix would be very welcome.

Emantor:
> @Emantor @Bastian-Krause I'd like to revert this for the 25.0 release, if you agree.

Fine by me too.

jluebbe:
> > @Emantor @Bastian-Krause I'd like to revert this for the 25.0 release, if you agree.
> 
> Fine by me. I'm not that well into labgrid internals, so a proper fix would be very welcome.

Could you try how much reducing the sleep in https://github.com/labgrid-project/labgrid/blob/master/labgrid/resource/remote.py#L66 has in your case? How many resources to you have configured on that place?

PR #1645, opened 2025-04-11
driver/power: Add support for tinycontrol tcPDU

lsa-cs:
**Description**

This module adds a power driver for the tinycontrol tcPDU.

tcPDU is a network-enabled PDU: <https://tinycontrol.pl/en/tcpdu/>

Unfortunately, the existing 'tinycontrol'-driver cannot be reused because the API is quite different.
API-Documentation: <https://docs.tinycontrol.pl/en/tcpdu/api/commands/>

I have tested this code with real hardware with these versions: 
- HW: "v1.1" 
- SW: "1.28-tcpdu"

**Checklist**
- [x] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -248,6 +248,10 @@ Currently available are:
   It was tested on the *6G10A v2* model.
   `Manual <https://tinycontrol.pl/media/documents/manual_IP_Power_Socket__6G10A_v2_LANLIS-010-015_En-1.pdf>`__
 
+``tinycontrol_tcpdu``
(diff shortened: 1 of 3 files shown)


Bastian-Krause (approved):
Thanks for the contribution. The power backend looks good.

Since this adds support for another network PDU, see our merge policy for those: #1031. I'm gonna merge this into labgrid, but maybe it makes sense in the long run to add support for it into [pdudaemon](https://github.com/pdudaemon/pdudaemon) and use that.

PR #1653, opened 2025-05-07
fix inconsistent driver states by reverting #1638

jluebbe:
This reverts commit d81eb868c445943f332764b428e8111344db47ba.

The change could cause inconsistent driver states (e.g. drivers active
for resources that have become unavailable), as not all resources are
processed by update_resources. See the original PR
https://github.com/labgrid-project/labgrid/pull/1638 for more details.

### DIFF
--- a/labgrid/target.py
+++ b/labgrid/target.py
@@ -41,7 +41,7 @@ class Target:
         else:
             input(msg)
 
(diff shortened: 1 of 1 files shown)


Emantor (approved):

Emantor:
I like the updated comment :+1:

PR #1655, opened 2025-05-10
Fix/docker conditional import

Emantor:
**Description**
This way we can correctly skip the tests if the module is not available,
otherwise we gat a ModuleNotFoundError because of the unconditional
docker import.

**Checklist**
- [ ] PR has been tested


Emantor:
@b2vn could you give this a spin? I currently do not have a docker setup.

PR #1249, opened 2023-07-28
Provider Improvements and Fixes

Bastian-Krause:
**Description**
Collection of various provider fixes and improvements:
- fix typo in the NFSProviderDriver class name
- make the NFSProviderDriver usable
- slim down NFSProvider resources (local and remote)
- fix exporting host information for local provider resources
- extend documentation

Contains breaking changes mentioned in CHANGES.rst.

**Checklist**
- [x] Documentation for the feature
- [.] Tests for the feature 
- [x] The arguments and description in doc/configuration.rst have been updated
- [x] PR has been tested

Fixes #762

PR #1657, opened 2025-05-19
Nfs provider export

jrichardsen:
**Description**

Since #1249, the `NFSProvider` no longer inherits from `BaseProvider` and no longer has the `internal` and `external` properties. This leads to an error with `ProviderGenericExport` which is meant for subclasses of `BaseProvider` and, thus, expects these properties to be present.

This PR removes `NFSProvider` from the `ProviderGenericExport` and implements a new `NFSProviderExport` instead.

**Checklist**
- [x] Tests for the feature 

- [x] PR has been tested


Bastian-Krause:
The tests introduced here succeed even without the other changes. Why is that?

Bastian-Krause:
Wouldn't it be sufficient to try starting an exporter exporting an NFSProvider in the test?

Maybe..
```python
def test_export_nfsprovider(tmpdir, coordinator):
    config = "exports.yaml"
    p = tmpdir.join(config)
    p.write(
        """
    nfs:
        NFSProvider: {}
    """
    )

    with pexpect.spawn(f"python -m labgrid.remote.exporter {config}", cwd=tmpdir) as spawn:
        spawn.expect("connected to coordinator", timeout=10)
```

Bastian-Krause:
@jrichardsen Are you still interested in this?

PR #1662, opened 2025-06-04
util/managedfile: ensure directory containing symlink exists for sync_to_resource()

### LINKED ISSUE #1661: Managedfile doesn't create a symlink for local tftp server
Hi,

When ManagedFile.sync_to_resource synces a file to a remote for use with tftp, a symlink to that file is created on the remote in the tftp server's internal directory.

When a local tftp server ('TFTPProvider' instead of 'RemoteTFTPProvider') is used, the file sync doesn't have to take place, but the symlink creation is also skipped, which was unexpected.

rvdgracht:
When the internal dir contains a subdir that does not yet exist (i.e. '/srv/tftp/board-23/') the symlink creation will fail.

### DIFF
--- a/labgrid/util/managedfile.py
+++ b/labgrid/util/managedfile.py
@@ -77,6 +77,7 @@ class ManagedFile:
                     conn.run_check(f"test ! -e {symlink} -o -L {symlink}")
                 except ExecutionError:
                     raise ManagedFileError(f"Path {symlink} exists but is not a symlink.")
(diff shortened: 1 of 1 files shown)


Bastian-Krause:
After looking at this again, I think this makes sense. I've rebased this and added the local use case.

Bastian-Krause:
@Emantor Could you take a look at this?

Emantor (approved):

PR #1670, opened 2025-06-24
driver/bareboxdriver: change default interrupt key to CTRL+D

a3f:
For developer convenience, barebox v2022.11.0 onwards brings up network interfaces automatically in the background when the boot is aborted by user input. By the time a user then runs boot net for example, the seconds it takes for an Ethernet link to be established will have passed.

The effects of this have already been seen in Labgrid; Commit 6f7bf6614ffd ("driver/bareboxdriver: silence barebox in _await_prompt() and unsilence on boot()") changes the log level to suppress "ethX: N Mbps ... link detected" messages, but it leaves another problem unaddressed: A system that Labgrid aborts to run the boot command now behaves differently to a system that was left to boot normally, because network adapters are reconfigured that normally aren't.

This bit me while trying to use Labgrid to debug a network issue in Linux as barebox ifup changed the Linux behavior.

barebox v2025.03.0 introduces a new halt state that can be entered by ctrl-d that skips these automatic hooks. Using it should not introduce regressions for existing users, regardless of barebox version:

- Prior to barebox v2025.01.0, the only options for an abort key were ctrl-c or any and both line feed and CTRL+D behave the same with either option.

- After v2025.01.0, global.autoboot_abort_key can be set to an arbitrary single printable ASCII key and it seems not possible to set a new line this way.

**Description**

**Checklist**
- [ ] Tests for the feature 

- [ ] PR has been tested

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,14 @@
+Release 25.1 (Unreleased)
+-------------------------
+
+New Features in 25.1
+~~~~~~~~~~~~~~~~~~~~
+
+- The default interrupt key for the `BareboxDriver` is now CTRL+D (``\x04``)
+  instead of a line feed (``\n``). barebox v2025.03.0 onwards handles
+  CTRL+D specially to halt autoboot countdown without running interactive
+  hooks like bringing up network interfaces automatically.
+
 Release 25.0 (Released May 7, 2025)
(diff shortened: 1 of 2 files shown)


Bastian-Krause (approved):
Thanks!

PR #1678, opened 2025-07-02
tftp: add external_ip setting

aparcar:
Just as with the `external` (path) setting, it's helpful for the DUT to know from which IP the server offers firmware.

**Description**

**Checklist**
- [x] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -1289,10 +1289,12 @@ A :any:`TFTPProvider` resource describes a TFTP server.
    TFTPProvider:
      internal: '/srv/tftp/board-23/'
      external: 'board-23/'
+     external_ip: '192.168.1.100'  # optional
 
 Arguments:
   - internal (str): path prefix to the local directory accessible by the target
   - external (str): corresponding path prefix for use by the target
+  - external_ip (str, optional): IP address that the target should use to access the TFTP server. If not specified, defaults to empty string.
 
 Used by:
   - `TFTPProviderDriver`_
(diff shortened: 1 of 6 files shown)


aparcar:
Here is an example how to use it https://github.com/aparcar/openwrt-tests/blob/main/strategies/tftpstrategy.py#L65

Bastian-Krause:
Why can't the DUT simply use the host attribute instead?

aparcar:
Using a proxied setup, the host value is something like hostname-foobar, which doesn't help the tftp DUT at all. The specified IP is instead handed over via a setenv call in a custom strategy

aparcar:
@Bastian-Krause any further comments? My code looks like this right now and works just fine

```python
elif status == Status.uboot:
            self.transition(Status.off)
            self.target.activate(self.tftp)
            self.target.activate(self.console)

            staged_file = self.tftp.stage(self.target.env.config.get_image_path("root"))
            tftp_server_ip = self.target.get_resource(
                RemoteTFTPProvider, wait_avail=False
            ).external_ip

            self.power.cycle()
            # interrupt uboot

            self.uboot.init_commands = (
                f"setenv bootfile {staged_file}",
            ) + self.uboot.init_commands

            if tftp_server_ip:
                tftp_dut_ip = ipaddress.ip_address(tftp_server_ip) + 1
                self.uboot.init_commands = (
                    f"setenv serverip {tftp_server_ip}",
                    f"setenv ipaddr {tftp_dut_ip}",
                ) + self.uboot.init_commands

            self.target.activate(self.uboot)
```

jluebbe:
> ```python
>             staged_file = self.tftp.stage(self.target.env.config.get_image_path("root"))
>             tftp_server_ip = self.target.get_resource(
>                 RemoteTFTPProvider, wait_avail=False
>             ).external_ip
> ```

I don't think would be a good to have to use **both** the driver and resource.

My suggestion would be to use `tftp://<host?>/<path>` URLs in the external property. Then, you can translate that to what your on-target software needs by using https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse.

@aparcar If that covers your use-case, we should probably just document this approach.

aparcar:
What is a clean way to access that from a strategy?

Issue #1123, opened 2023-03-30
Strategy timeouts cause inconsistent state


jpuderer:
When running pytest tests using the strategies shipped with labgrid, the state can get confused if a timeout occurs while bringing up the target.

I ran into this issue when doing some simple endurance testing.

By way of an example, consider what happens if run the following simple tests using the UBootStrategy (although others have the same problem).

```
@pymark.test.parameterize("iteration", [1, 2, 3, 4, 5])
def test_boot(target, strategy, iteration):
   strategy.transition("shell")
```

Now suppose that a timeout occurs on iteration 1 while waiting for a login prompt.  This causes the first test to fail, and leaves our strategy in the "u-boot" state.

At this point, we're in an inconsistent state.  The target is almost certainly not at the u-boot prompt... It could be hung, or crashed, or it could have simply took too long to boot and is now at the login prompt.

Unfortunately, this will cause test 2 and all following tests to fail.

First because the u-boot driver was deactivated when the shell driver was activated in test #1, so when it tries to run:

```
        elif status == Status.shell:
            # transition to uboot
            self.transition(Status.uboot)
            self.uboot.boot("")
            self.uboot.await_boot()
```

It skips over the transistion to uboot (since it is already there), but fails with an StateError when it tries to run uboot.boot("")

```
E           labgrid.binding.StateError: UBootDriver(target=Target(name='main', env=Environment(config_file='test.yaml')), name=None, state=<BindingState.bound: 1>, prompt='=> ', autoboot='stop autoboot', password='', interrupt='\n', init_commands=(), password_prompt='enter Password:', boot_expression='', bootstring='Linux version \\d', boot_command='run bootcmd', boot_commands={}, login_timeout=30, boot_timeout=30) has not been activated, UBootDriver.boot cannot be called in state "bound"
```

Even if that weren't the case and we reactivated the uboot driver, our strategy would get a tiny bit farther, but will timeout waiting for the kernel to start booting in uboot.await_boot().

I think what needs to happen is for the Strategy to catch (and later re-reraise) any exceptions when it performs a transition, and set the state to "unknown".

Bastian-Krause:
[...]

PR #1681, opened 2025-07-03
Introduce Broken Strategy State, Prevent Further Use and Skip Subsequent Tests

Bastian-Krause:
**Description**
Introduce a decorator for the Strategy's `transition()` and `force()` methods. It stores the original exception in the Strategy's new `broken` attribute and raises a `StrategyError` from it. All subsequent calls to the decorated method will lead to the original exception.

Now that a Strategy can be marked broken easily, check for that in labgrid's pytest plugin before executing a test and its fixtures. If a broken Strategy is found in one of the targets, skip the test. It's not worth continuing with it. It only leads to cascading and confusing errors.

Having a strategy prepared with the new decorator, a strategy exception now leads to either a test error (when triggered in a fixture) or a test fail (when triggered in the test itself). All subsequent tests requesting the broken strategy (via the `strategy` fixture) will be skipped.

In order to get the word out about the `@raise_if_broken` decorator, decorate the Strategies in labgrid with it.

Previously this kind of workaround was often used:
```python
@pytest.fixture
def shell(strategy, target):
    try:
        strategy.transition("shell")
        return target.get_driver("ShellDriver")
    except Exception:
        pytest.exit("Strategy failed")
```
This is now obsolete.

**Checklist**
- [x] Tests for the feature (covered in `tests/test_docker.py::test_docker_without_daemon`)
- [x] PR has been tested

Related to #1123.

### DIFF
--- a/labgrid/pytestplugin/hooks.py
+++ b/labgrid/pytestplugin/hooks.py
@@ -7,6 +7,7 @@ from .. import Environment
 from ..consoleloggingreporter import ConsoleLoggingReporter
 from ..util.helper import processwrapper
 from ..logging import StepFormatter, StepLogger
+from ..exceptions import NoStrategyFoundError
 
 LABGRID_ENV_KEY = pytest.StashKey[Environment]()
 
@@ -130,3 +131,23 @@ def pytest_collection_modifyitems(config, items):
                     reason=f'Skipping because features "{missing_feature}" are not supported'
                 )
             item.add_marker(skip)
+
+@pytest.hookimpl(tryfirst=True)
+def pytest_runtest_setup(item):
+    """
+    Skip test if one of the targets uses a strategy considered broken.
+    """
+    # Before any fixtures run for the test, check if the session-scoped strategy fixture was
+    # requested (might have been executed already for a prior test). If that's the case and the
+    # strategy is broken, skip the test.
+    if "strategy" in item.fixturenames:
+        env = item.config.stash[LABGRID_ENV_KEY]
+        # skip test even if only one of the targets in the env has a broken strategy
+        for target_name in env.config.get_targets():
+            target = env.get_target(target_name)
+            try:
+                strategy = target.get_strategy()
+                if strategy.broken:
+                    pytest.skip(f"{strategy.__class__.__name__} is in broken state")
+            except NoStrategyFoundError:
+                pass
--- a/doc/development.rst
+++ b/doc/development.rst
@@ -226,7 +226,7 @@ Start by creating a strategy skeleton:
     import attr
 
     from labgrid.step import step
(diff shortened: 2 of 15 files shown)


Bastian-Krause:
Rebased on latest master.

Bastian-Krause:
After discussions with @jluebbe, rename decorator `@raise_if_broken` to `@never_retry`.

Emantor (approved):

labgrid/pytestplugin/hooks.py
```diff
@@ -130,3 +130,22 @@ def pytest_collection_modifyitems(config, items):
                     reason=f'Skipping because features "{missing_feature}" are not supported'
                 )
             item.add_marker(skip)
+
+@pytest.hookimpl(wrapper=True)
```
jluebbe: Why is this a [hook wrapper](https://docs.pytest.org/en/stable/how-to/writing_hook_functions.html#hook-wrappers-executing-around-other-hooks)? It doesn't process/return the `yield`.

[`tryfirst=True`](https://docs.pytest.org/en/stable/how-to/writing_hook_functions.html#hook-function-ordering-call-example) should be enough.

labgrid/pytestplugin/hooks.py
```diff
@@ -130,3 +130,22 @@ def pytest_collection_modifyitems(config, items):
...
+    # requested (might have been executed already for a prior test). If that's the case and the
+    # strategy is broken, skip the test.
+    if "strategy" in item.fixturenames:
+        env = item.config.stash[LABGRID_ENV_KEY]
+        # skip test even if only one of the targets in the env has a broken strategy
+        for target_name in env.config.get_targets():
+            target = env.get_target(target_name)
+            strategy = target.get_driver("Strategy")
+            if strategy and strategy.broken:
+                pytest.skip(f"{strategy.__class__.__name__} is in broken state")
```
jluebbe: Use `target.get_strategy()` and handle `NoStrategyFoundError` (by ignoring it).

jluebbe (changes requested):

labgrid/pytestplugin/hooks.py
Bastian-Krause: Sounds good, applied.

labgrid/pytestplugin/hooks.py
Bastian-Krause: Yes, that's better. Done.

Bastian-Krause:
> @Bastian-Krause [Emantor@ad78a2f](https://github.com/Emantor/labgrid/commit/ad78a2f0594b1da66558ced8e8e85a972d4bf6fb)

Applied in combination with the suggestions from @jluebbe.

jluebbe (approved):

PR #1683, opened 2025-07-04
util/managedfile: create symlink when staging files for local providers

Bastian-Krause:
**Description**
When a local TFTPProviderDriver/HTTPProviderDriver is used, the file sync can be skipped but the symlink creation in the server internal directory shouldn't.

Commit by @rvdgracht taken from #1662, leaving the maybe controversial part of creating the directory there for further discussion.

**Checklist**
- [ ] PR has been tested

### DIFF
--- a/labgrid/util/managedfile.py
+++ b/labgrid/util/managedfile.py
@@ -67,17 +67,19 @@ class ManagedFile:
                     self.local_path,
                     f"{self.rpath}{os.path.basename(self.local_path)}"
                 )
+        else:
+            conn = sshmanager.open("localhost")
+            self.rpath = os.path.dirname(self.local_path) + "/"
 
-            if symlink is not None:
-                self.logger.info("Linking")
-                try:
-                    conn.run_check(f"test ! -e {symlink} -o -L {symlink}")
-                except ExecutionError:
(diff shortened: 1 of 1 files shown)


labgrid/util/managedfile.py
```diff
@@ -67,17 +67,19 @@ def sync_to_resource(self, symlink=None):
...
-                try:
-                    conn.run_check(f"test ! -e {symlink} -o -L {symlink}")
-                except ExecutionError:
-                    raise ManagedFileError(f"Path {symlink} exists but is not a symlink.")
-                # use short options to be compatible with busybox
-                # --symbolic --force --no-dereference
-                conn.run_check(f"ln -sfn {self.rpath}{os.path.basename(self.local_path)} {symlink}")
-
+        else:
+            conn = sshmanager.open("localhost")
```
Emantor, line 71: This looks broken, why do we need to open an SSH connection when operating on a local machine? This breaks the sandboxed tests on nixpkgs since we now open SSH connections for tests that never did this before.

Bastian-Krause: Hm okay, I thought it would be a good idea to not duplicate the symlinking logic locally in Python, but you're right, it would actually be better.

PR #1692, opened 2025-07-11
contrib/completion: fix dfu sub command file completion

Bastian-Krause:
**Description**
The dfu completion has no idea about the "altsetting" required for download and stops completing after that.

Since `labgrid-client dfu` needs to complete only download/detach/list after dfu, simply complete files/directories in all other cases.

**Checklist**
- [x] PR has been tested

Fixes #1055


jluebbe:
Thanks, this fixes the problem I saw.

Emantor (approved):

PR #1695, opened 2025-07-14
DenkoviRelay: add resource, driver and export

DiogoSilva14:
**Description**
Add support for denkovi usb serial relay boards (https://denkovi.com/).
This driver works the same way as HIDRelay, but with a different agent code.

This driver was used over a long course of time on a local testing setup.

**Checklist**
- [x] PR has been tested - by integrating it into a local testing setup

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -58,6 +58,7 @@ doc = [
     "Sphinx>=2.0.0",
 ]
 docker = ["docker>=5.0.2"]
+libftdi = ["pylibftdi>=0.23.0"]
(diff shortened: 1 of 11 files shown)


pyproject.toml
```diff
@@ -45,6 +45,7 @@ dependencies = [
     "PyYAML>=6.0.1",
     "requests>=2.26.0",
     "xmodem>=0.4.6",
+    "pylibftdi>=0.23.0",
```
Emantor: Please move this to an `optional-dependencies` entry.

PR #1701, opened 2025-07-23
udev: Enable RKUSBLoader for rk3568 and rk3588

tretter:
**Description**

Add the `VENDOR_ID` and `MODEL_ID` for the Maskrom mode of the Rockchip rk3568 and rk3588 to the possible matches for RKUSBLoader to allow using the RKUSBLoader with these SoCs.

With this change, `labgrid-suggest` recognizes Rockchip rk3568 and rk3588 SoCs if booted into Maskrom mode.

**Checklist**
- [x] PR has been tested:


sjg20 (changes requested):
Can we use this one instead?

https://github.com/labgrid-project/labgrid/pull/1702

PR #1712, opened 2025-08-04
ConsoleProtocol: Update Signatures to match ConsoleExpectMixin

SmithChart:
**Description**

The ConsoleExpectMixin has learned some new tricks over time. But the ConsoleProtocol was not updated accordingly.

Every class implementing the ConsoleProtocol also uses the ConsoleExpectMixin - so it's safe to update the signatures according to the ConsoleExpectMixin.

For ConsoleProtocol.expect() I've used the type definition of the underlying pexpect library.

I have not updated the return types, since it's not clear to me what the ConsoleExpectMixin should return.

This helps users with IDEs to get better type hinting when using the ConsoleProtocol.

**Checklist**
- [ ] Tests for the feature 

- [ ] PR has been tested

### DIFF
--- a/labgrid/protocol/consoleprotocol.py
+++ b/labgrid/protocol/consoleprotocol.py
@@ -1,11 +1,13 @@
 import abc
+from typing import Optional, Union
+from pexpect import EOF, TIMEOUT
 
 
 class ConsoleProtocol(abc.ABC):
     """Abstract class for the ConsoleProtocol"""
 
     @abc.abstractmethod
-    def read(self):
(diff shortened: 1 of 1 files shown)


labgrid/protocol/consoleprotocol.py
```diff
@@ -1,11 +1,14 @@
 import abc
+from typing import Optional
+
+from pexpext import EOF, TIMEOUT
```
jluebbe: This seems to be a typo breaking the CI?

```suggestion
from pexpect import EOF, TIMEOUT
```

Emantor (approved):

PR #1714, opened 2025-08-11
IMXUSBDriver: conditionally check image

gumulka:
**Description**

Some devices do not allow USB Serial Download read commands, which will result in a failure, when trying to verify the correctness of the image. Make the verify step optional but on by default.

**Checklist**
- [ ] ~~Tests for the feature~~

- [ ] ~~Add a section on how to use the feature to doc/usage.rst~~

- [ ] ~~Add a section on how to use the feature to doc/development.rst~~

- [x] PR has been tested

- [ ] ~~Man pages have been regenerated~~

Test has been done by creating a minimal place and setting verify to True, False and leaving it empty.

```
targets:
  main:
    resources:
    - RemotePlace:
        name: "myboard"
    drivers:
    - IMXUSBDriver:
        verify: False
```

### DIFF
--- a/labgrid/driver/usbloader.py
+++ b/labgrid/driver/usbloader.py
@@ -55,6 +55,7 @@ class IMXUSBDriver(Driver, BootstrapProtocol):
     }
 
     image = attr.ib(default=None)
+    verify = attr.ib(default=True, validator=attr.validators.instance_of(bool))
 
     def __attrs_post_init__(self):
         super().__attrs_post_init__()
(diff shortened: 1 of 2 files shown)


Emantor:
The documentation under `doc/configuration.rst` needs to be updated, also please provide an example of a device where the read command no longer succeeds for the configuration option in the documentation so users have a breadcrumb to find when to disable the verification.

gumulka:
@Emantor Is this enough of a hint, or should it be more elaborate?

Emantor (approved):

labgrid/driver/usbloader.py
```diff
@@ -55,6 +55,7 @@ class IMXUSBDriver(Driver, BootstrapProtocol):
     }
 
     image = attr.ib(default=None)
+    verify = attr.ib(default=True)
```
Bastian-Krause:
```suggestion
    verify = attr.ib(default=True, validator=attr.validators.instance_of(bool))
```
We should validate the verify argument.

Bastian-Krause (approved):

PR #1717, opened 2025-08-25
driver: Add TenmaSerial power driver and integration support

dooraim:
This patch introduces a new power driver, TenmaSerial, providing control support for Tenma power devices. Integration includes:

- Implementation of TenmaSerial in powerdriver.py
- Exporter and remote client support for the new driver
- Configuration documentation in configuration.rst
- New unit tests added in test_tenmaserial.py
- Minor updates to pyproject.toml and man pages

This initial draft enables basic functionality and lays the groundwork for full Tenma device intration

**Description**

This pull request aims to provide initial support for the Tenma power supply, which is controlled via USB. To accomplish this, we created the `TenmaSerialDriver` in `labgrid/driver/powerdriver.py`, as well as other necessary components. Our goal is to create an interface between the Tenma Serial Python package and Labgrid. We used the `Tenma 72-2540` device and created the tests/test_tenmaserial.py test to check the functionality.

**Checklist**
- [x] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -97,6 +97,7 @@ dev = [
     # GRPC Channelz support
     "grpcio-channelz>=1.64.1, <2.0.0",
 ]
+tenma = ["tenma-serial~=1.1.3"]
 
 [project.scripts]
 labgrid-autoinstall = "labgrid.autoinstall.main:main"
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -422,6 +422,28 @@ Arguments:
 Used by:
   - `TasmotaPowerDriver`_
 
+TenmaSerialPort
+++++++++++++++++
+A :any:`TenmaSerialPort` describes a *Tenma* as supported by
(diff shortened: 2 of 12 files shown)


pyproject.toml
```diff
@@ -45,6 +45,7 @@ dependencies = [
     "PyYAML>=6.0.1",
     "requests>=2.26.0",
     "xmodem>=0.4.6",
+    "tenma-serial~=1.1.3",
```
Emantor: Please add this as an optional dependency (tenma?) See further down in the same file.

Emantor (changes requested):

pyproject.toml
dooraim: @Emantor I have implemented your suggestion. I have also re-run `tests/test_tenmaserial.py` and it works correctly

OP01406:
Sorry for the 'ping', again.
But is there something we can do to help this PR to be accepted.
Having a rather common Bench Power Supply in Labgrid would be nice.

Issue #1289, opened 2023-10-23
Add remote CAN resource


stoupa-cz:
With local can resources in place (see #1287 ), we can add remote access with NetworkCANPort. We discussed the approach at EOSS2023. Pengutronix guys suggest bridging by one of these tools:

- [cannelloni](https://github.com/mguentner/cannelloni)
- [socketcand](https://github.com/linux-can/socketcand)

The idea of bridging is similar to NetworkSerialPort. We will start a daemon on an exporter. A client will create a local `vcan<x>` interface by connecting to the daemon.

I don't know about these projects, so if there is any recommendation which is better to start with, please let me know.

jluebbe:
@marckleinebudde Can you give some guidance here?

stoupa-cz:
We are again on that topic. I briefly checked both projects above, and both still seem fine. I'd be happy to hear your input on the selection and suitability for labgrid (@marckleinebudde ?).

CC @ca-ja

stoupa-cz:
We discussed this topic (and specifically #1719) on the Labgrid Community Call.

@JoshuaWatt briefly checked netns (merged into master) for CAN tunnelling. I haven't tested it, but `vxcan` should do the job.

Does anyone have experience with `vxcan`? Does it make sense to switch to it instead of cannelloni?

hundeboll:
Looks interesting. I might be able to give at try before summer...

hundeboll:
> [@JoshuaWatt](https://github.com/JoshuaWatt) briefly checked netns (merged into master) for CAN tunnelling. I haven't tested it, but `vxcan` should do the job.
> 
> Does anyone have experience with `vxcan`? Does it make sense to switch to it instead of cannelloni?

I looked at the `netns` code and `vxcan` today, and I fail to understand how the latter is supposed to help here?

As far as I can tell, `vxcan` allows us to create a virtual can interface pair with either end in different network namespaces. But it isn't useful for tunneling between different hosts?

We can use `netns` to create a `vcan` interface on the labgrid client without needing privileges. This would be useful to run e.g. `cannelloni` on both the client and the agent/exporter, and let test cases work on a "regular" vcan interface.

hundeboll:
Okay, I started working a bit on this. My approach is to use the `tapfwd.py` script to pipe CAN frames to/from the exporter.

[...]

Issue #1287, opened 2023-10-19
Add local CAN resource


stoupa-cz:
The CAN devices are common on our devices. The support for them is limited in labgrid, so let's start with the crucial part -- local can resources:

- RawCANPort - device specified by the interface name
- USBCANPort - udev matching (useful for USB-CAN adapters)

The goal is to have NetworkCANPort. It will be discussed in a separate issue.

I don't plan any Driver now, as I don't see a need (and I hope that it is possible to use it this way). Our test infrastructure needs just the interface name at the moment.

The checklist:

- [x] Initial dirty implementation.
- [ ] Polish and cleanup.
- [ ] Add documenation.
- [ ] Start review process.

I'm playing with it in our fork at https://gitlab.com/vzlu/tools/labgrid/-/tree/tom/add_can_resource.

Bastian-Krause:
I'm wondering if you could use the [NetworkInterface](https://labgrid.readthedocs.io/en/latest/configuration.html#networkinterface)/[RemoteNetworkInterface](https://labgrid.readthedocs.io/en/latest/configuration.html#remotenetworkinterface) resources for this?

stoupa-cz:
> I'm wondering if you could use the [NetworkInterface](https://labgrid.readthedocs.io/en/latest/configuration.html#networkinterface)/[RemoteNetworkInterface](https://labgrid.readthedocs.io/en/latest/configuration.html#remotenetworkinterface) resources for this?

I was also thinking about it. I'm proposing the separate Resource because:

There will probably be new attributes ('speed' at the moment), and the NetworkCANPort will have different bridging similar to `ser2net` (I will create a new issue to summarize the discussion we had at EOSS2023). There might be a CAN Driver in the future, which should be bound only to the respective resource. I like the SocketCAN abstraction, but on the other hand, I see CAN as a different type of interface with different approaches under the hood.

PR #1719, opened 2025-08-27
RFC: Add basic support for local and remote CAN

stoupa-cz:
**Description**

This is a draft of basic CAN support. It is in RFC, as there are open questions, missing docs and tests, and moreover, it needs more testing. ~~We plan to use it to remotely access the satellite's telemetry buses during development.~~ It is tested on our local branch to remotely access the satellite's telemetry buses during development.

Local and remote CAN resources are added. Additionally, a simple CAN driver is added. The driver has very limited capabilities. For local use, it can set up the interface and tell you the interface name. For remote use, it establishes a transparent tunnel using cannelloni. With it, you can use the local virtual CAN interface. All the traffic is routed to the real interface on the other end.

Be aware, as cannelloni states in its doc:

> cannelloni is not suited for production deployments. Use it only in environments where packet loss is tolerable. There is no guarantee that CAN frames will reach their destination at all and/or in the right order.

~~I tested in the lab only. We will test it in production during September/October.~~

Draft version (without `labgrid-can-interface`) is being tested in production for ~two months.

~~I see the main blocker: it requires root privileges on both sides (exporter and client; even for local use). What is the preferred way to handle privileged access? May we use `labgrid-raw-interface`?~~ Helper `labgrid-can-interface` has been added.

**Checklist**
- [x] Solve the need for root privileges
- [x] Test initial implementation
- [ ] More robust handling of cannelloni processes
- [ ] CAN configuration via labgrid or e.g., systemd-networkd, [see the discission](https://github.com/labgrid-project/labgrid/pull/1719#issuecomment-3668743001)
- [ ] Fix remaining TODOs in the code
- [ ] Rebase on master
- [ ] Documentation
- [ ] Write tests
- [ ] Test in production

Reference issues: #1287 and #1289 

This is a part of my work at VZLU AEROSPACE.

### DIFF
--- a/debian/labgrid.install
+++ b/debian/labgrid.install
@@ -6,5 +6,6 @@ debian/labgrid-exporter /usr/bin
 debian/labgrid-pytest /usr/bin
 debian/labgrid-suggest /usr/bin
 helpers/labgrid-bound-connect /usr/sbin
+helpers/labgrid-can-interface /usr/sbin
 helpers/labgrid-raw-interface /usr/sbin
 contrib/completion/labgrid-client.bash => /usr/share/bash-completion/completions/labgrid-client
--- a/helpers/labgrid-can-interface
+++ b/helpers/labgrid-can-interface
@@ -1,0 +1,146 @@
+#!/usr/bin/env python3
+#
+# Wrapper script to be deployed on machines that should allow controlled
+# manipulation of CAN network interfaces. A /etc/labgrid/helpers.yaml
+# can deny access to CAN interfaces. See below.
+#
+# This is intended to be used via sudo. For example, add via visudo:
+# %developers ALL = NOPASSWD: /usr/sbin/labgrid-can-interface
+
+import argparse
+import os
+import sys
+
+import yaml
+
+# Linux CAN interface type
+ARPHRD_CAN = 280
+
+def get_denylist():
+    denylist_file = "/etc/labgrid/helpers.yaml"
+    try:
+        with open(denylist_file) as stream:
+            data = yaml.load(stream, Loader=yaml.SafeLoader)
+    except (PermissionError, FileNotFoundError, AttributeError) as e:
+        raise Exception(f"No configuration file ({denylist_file}), inaccessable or invalid yaml") from e
+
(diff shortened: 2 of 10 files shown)


hundeboll:
Hi @stoupa-cz 

Thanks for working on this. I've been experimenting with `[socketcand](https://github.com/linux-can/socketcand)`. It has the same limitation as `cannelloni` (root on both ends for `vcan` setup), but the protocol is pretty simple, so the client side can be implemented as python functions in the labgrid driver. This should be enough for tests with limited CAN traffic.

stoupa-cz:
> I tried running an exporter with your PR, where a `USBCANPort` is exported as an `NetworkCANPort`. On the client, I implemented the `cannelloni` protocol in the driver ([can.py](https://github.com/user-attachments/files/24166243/can.py)).
> 
> It works quite well for our purpose.
> 
> One thing I miss from `socketcand` is the ability to subscribe to certain CAN IDs, so I added (very) simple filtering in the driver.

It seems we have different needs. For us, a client-side "real" `vcan` interface is necessary. Fortunately, we have a configurable driver, so any client (e.g., your cannelloni protocol reimplementation) can be selected. For the sake of simplicity, I’d prefer to continue using the "native" cannelloni approach.

BTW, what is your use case? Do you want to replay some traffic?
 
> And again: It might make sense to remove the interface configuration, and leave that to e.g. `systemd-networkd`?

On exporter side, the interface is controlled (e.g., down, set-bitrate, up) when the place is acquired. The bitrate configuration comes from the expoter configuration. It is similar on client when the driver is activated, although the bitrate is loaded from the exporter's configuration. Which interface configuration managed by `systemd-networkd` are you referring to?

hundeboll:
I agree. Just let the CANPort resource default to no bitrate setting, and only run `ip link ...` commands if a bitrate is configured. Having a default is probably bad anyways, as we cannot guess the bitrate reliably.

I think we should focus on getting the resource merged, as it's easy to write custom drivers in the client context. Then we can discuss the details of the driver later.

stoupa-cz:
`vxcan` in netns might be a replacement, see https://github.com/labgrid-project/labgrid/issues/1289#issuecomment-4213516277 for more details.

PR #1721, opened 2025-08-27
driver: add RKUSBMaskromDriver bootstrap driver

Kwiboo:
**Description**

Add a `RKUSBMaskromDriver` bootstrap driver with an accompanying `rkusbmaskrom` agent to support bootstrapping targets with Rockchip SoCs.

The `rkusbmaskrom` agent expects the target to be in MASKROM mode and send images to BootROM using vendor specific `0x471` and `0x472` control transfers, in 4 KiB chunks.

The first image is loaded to SRAM using `0x471` and is expected to initialize DRAM and then return back to BootROM.
A second image can then be loaded to start of DRAM using `0x472`.

The `rkusbmaskrom` agent also support extracting the needed `0x471` and `0x472` images from a vendor loader image, typically created with the vendor `boot_merger` tool from the rkbin repo, or idblock boot images, typically created with the U-Boot `mkimage` or the Barebox `rkimage` tools.

**Related**

- [[PATCH v3 0/4] rockchip: Add support for RAM boot from maskrom mode](https://lore.kernel.org/r/20250802220738.3018910-1-jonas@kwiboo.se)
- #1542
- #1703

**Checklist**
- [x] PR has been tested

**Example usage**

```
$ labgrid-client -v -c remote.yaml -p rk3528-radxa-e20c bootstrap u-boot-rockchip-usb472.bin
INFO               root: Selected role rk3528-radxa-e20c from configuration file
INFO         StepLogger:  → RKUSBMaskromDriver.load(filename='<path>/u-boot-rockchip-usb472.bin')
INFO    SSHConnection(h:   Created new SSH connection to noble
INFO    ManagedFile(loc:   Synchronizing <path>/u-boot-rockchip-usb471.bin to noble
INFO         StepLogger:   → ProcessWrapper.check_output(command=['rsync', ...])
INFO         StepLogger:   ← ProcessWrapper.check_output() result=b'' [0.073s]
INFO    ManagedFile(loc:   Synchronizing <path>/u-boot-rockchip-usb472.bin to noble
INFO         StepLogger:   → ProcessWrapper.check_output(command=['rsync', ...])
INFO         StepLogger:   ← ProcessWrapper.check_output() result=b'' [0.072s]
INFO         StepLogger:  ← RKUSBMaskromDriver.load() [3.001s]
INFO    SSHConnection(h:  Closing SSH connection to noble
```

```
$ labgrid-client -v -c remote.yaml -p rk3528-radxa-e20c console
INFO               root: Selected role rk3528-radxa-e20c from configuration file
INFO               root: microcom not available, using telnet instead
INFO               root: connecting to NetworkSerialPort(target=Target(name='rk3528-radxa-e20c', env=Environment(config_file='remote.yaml')), name='USBSerialPort', state=<BindingState.bound: 1>, avail=True, host='noble', port=57985, speed=1500000, protocol='rfc2217') calling telnet noble 57985
Connected to noble.
Escape character is '^]'.
DDR 56f70fd2ad huan.he 25/02/26-09:55:30,fwver: v1.11
LP4/4x derate e, other dram:1x trefi
ddrconfig:0
MID:0x6
LPDDR4, 324MHz
BW=32 Co=10 Bk=8 CS0 Row=16 CS=1 Die BW=16 Size=2048MB
...
out

U-Boot SPL 2025.10-rc2 (Aug 25 2025 - 22:27:40 +0000)
Trying to boot from RAM
## Checking hash(es) for config config-1 ... OK
## Checking hash(es) for Image atf-1 ... sha256+ OK
## Checking hash(es) for Image u-boot ... sha256+ OK
## Checking hash(es) for Image fdt-1 ... sha256+ OK
## Checking hash(es) for Image atf-2 ... sha256+ OK
## Checking hash(es) for Image atf-3 ... sha256+ OK
NOTICE:  BL31: v2.3():v2.3-912-gfe961b77f:derrick.huang, fwver: v1.20
NOTICE:  BL31: Built : 16:59:19, Mar 31 2025
INFO:    rk_otp_init finish!
INFO:    code: 0x28
INFO:    ARM GICv2 driver initialized
INFO:    nonboot_cpus_off: clst_st=0xc0e, core_st=0xe1e0 boot_cpu=0
INFO:    idle_st=0x0, pd_st=0x0
INFO:    Using opteed sec cpu_context!
INFO:    boot cpu mask: 1
INFO:    RK3528 SoC (0x101)
INFO:    BL31: Initializing runtime services
WARNING: No OPTEE provided by BL2 boot loader, Booting device without OPTEE initialization. SMC`s destined for OPTEE will return SMC_UNK
ERROR:   Error initializing runtime service opteed_fast
INFO:    BL31: Preparing for EL3 exit to normal world
INFO:    Entry point address = 0x800000
INFO:    SPSR = 0x3c9


U-Boot 2025.10-rc2 (Aug 25 2025 - 22:27:40 +0000)

Model: Radxa E20C
SoC:   RK3528A
DRAM:  2 GiB
Core:  197 devices, 32 uclasses, devicetree: separate
MMC:   mmc@ffbf0000: 0, mmc@ffc30000: 1
Loading Environment from nowhere... OK
In:    serial@ff9f0000
Out:   serial@ff9f0000
Err:   serial@ff9f0000
Net:   eth0: ethernet@ffbe0000
Hit any key to stop autoboot: 0
=>
telnet> quit
Connection closed.
```


doc/configuration.rst
```diff
@@ -2638,6 +2639,39 @@ Arguments:
...
+
+   images:
+     rkmaskrom471: 'path/to/u-boot-rockchip-usb471.bin'
+     rkmaskrom472: 'path/to/u-boot-rockchip-usb472.bin'
+
+Arguments:
+  - initial (str): optional, key in :ref:`images <labgrid-device-config-images>` containing the path
+    of an image that typically initializes DRAM of the target
+  - image (str): optional, key in :ref:`images <labgrid-device-config-images>` containing the path
+    of a bootloader image to load to start of DRAM, or to SRAM when an initial image is unused
```
QSchulz: issue: missing `delay`

labgrid/driver/usbloader.py
```diff
@@ -149,6 +151,53 @@ def load(self, filename=None):
...
+        self.proxy = None
+        if self.wrapper:
+            self.wrapper.close()
+        self.wrapper = None
+
+    @Driver.check_active
+    @step(args=['filename'])
+    def load(self, filename=None):
+        images = []
+        if self.initial is not None:
```
QSchulz: suggestion (non-blocking): simply check "if self.initial"

This allows to catch the case where `self.initial` is the empty string for example

labgrid/driver/usbloader.py
```diff
@@ -149,6 +151,53 @@ def load(self, filename=None):
...
+            images.append(self.target.env.config.get_image_path(self.initial))
+        if not filename and self.image is not None:
```
QSchulz: suggestion (non-blocking): simply check "if self.image"

This allows to catch the case where `self.image` is the empty string for example

labgrid/util/agents/rkusbmaskrom.py
QSchulz, line 1: question: use ``snagboot`` instead?

https://github.com/bootlin/snagboot/pull/56 brings Rockchip support to snagboot.

The benefit is that other SoC vendors are supported by snagboot so we could have a snagboot driver to support multiple SoC vendors at once in labgrid.

This would also stop reimplementing the rockusb protocol in yet another project (hello rkdeveloptool, rockusb from rockchiprs, rkflashtool, snagboot, ...).

This was briefly discussed with @a3f @jluebbe at OSSEU/ELCE this year. Talking with snagboot's maintainer, there are plans to merge the Rockchip support for snagboot very soon and do a release soon after.

doc/configuration.rst
```diff
@@ -2638,6 +2639,39 @@ Arguments:
...
+  - :any:`BootstrapProtocol`
+
+.. code-block:: yaml
+
+   targets:
+     main:
+       drivers:
+         RKUSBMaskromDriver:
+           initial: 'rkmaskrom471'
+           image: 'rkmaskrom472'
```
QSchulz, line 2982: question: is this compatible with Barebox?

Barebox seems to be generating only one binary which contains both "files". I think it'd make sense to support Barebox file format as well, which likely differs from U-Boot's (currently two separate files IIRC) or the binary generated by Rockchip's boot_merger?

@a3f any opinion on that maybe?

EDIT: I see that the next patch adds the implementation for reading the file from ``boot_merger``, so I guess if we have a unique way of identifying Barebox's implementation that would work too. Still, wondering if we shouldn't simply use snagboot and implement stuff there if necessary?

labgrid/util/agents/rkusbmaskrom.py
```diff
@@ -79,6 +124,36 @@ def rc4_prga(S):
         yield K
 
 
+def prase_rkboot_header(data):
```
QSchulz: typo: parse

labgrid/util/agents/rkusbmaskrom.py
Kwiboo: > question: use `snagboot` instead?

I will be using this agent/driver and not `snagboot` in my personal lab ;-)

> [bootlin/snagboot#56](https://github.com/bootlin/snagboot/pull/56) brings Rockchip support to snagboot.

I will take a closer look little bit later and add some review comments, I can already see some issues in that implementation.

> This would also stop reimplementing the rockusb protocol in yet another project (hello rkdeveloptool, rockusb from rockchiprs, rkflashtool, snagboot, ...).

All of those seem to have their own special issues, are not fully featured or can get stuck etc.
This driver works and only requires pyusb installed on the exporter, for my personal lab I do not want to deal with lots of third party tools ;-)

doc/configuration.rst
Kwiboo: > Barebox seems to be generating only one binary which contains both "files". I think it'd make sense to support Barebox file format as well, which likely differs from U-Boot's (currently two separate files IIRC)

Barebox uses the IDBlock v2 format, i.e. same format that `mkimage` produce for `TPL` and `SPL` in U-Boot for RK35xx.
Will add an implementation for extracting images from the IDBlock v1 and v2 formats, thanks!

>  or the binary generated by Rockchip's boot_merger?

This format is supported with the second commit, one special thing that `boot_merger` does differently for `RK35xx` is that it adds an IDBlock v2 header for the 471 and 472 files referenced in the rkboot .ini-files.
However, this extra header does not seem to be required on all RK35xx SoCs I have tested.

Also the BootROM seem to use a much slower crc16 implementation, so validating a huge blob for i.e. U-Boot RAM boot can take several seconds.
Will send an updated Rockchip U-Boot RAM-boot series that take advantage of FIT compression along with relocation the FIT to the 2 MiB offset to help speed up RAM-boot.

> Still, wondering if we shouldn't simply use snagboot and implement stuff there if necessary?

Feel free to implement and use a `snagboot` driver, I will be using this for my personal lab ;-)

doc/configuration.rst
Kwiboo: Fixed, thanks!

labgrid/driver/usbloader.py
Kwiboo: Fixed, thanks!

labgrid/driver/usbloader.py
Kwiboo: Fixed, thanks!

labgrid/util/agents/rkusbmaskrom.py
Kwiboo: Hehe, fixed, thanks!

doc/configuration.rst
Kwiboo: I have now added a new commit with support for decoding the IDBlock v2 image format (both unsigned and signed) in addition to the loader image format.

With that added we should be able to use the Barebox output image or `u-boot-rockchip.bin` to at least boot into SPL on newer Rockchip SoCs.

Will run some more tests and add one more commit to add support for the legacy IDBlock image format to make this fully featured.

This will need to be tested some more on e.g. RK3576 where more than two idblock images is typically used to e.g. fixup booting from SD-card.

Kwiboo: I have now pushed one more commit with support for decoding the legacy IDBlock image format, the boot image format used for Rockchip SoCs prior to the RK35xx line.

With this I can now use something like the following `labgrid-client` command on my Rockchip boards:

```
labgrid-client -v -c remote.yaml -p rk3399-rock-4 bootstrap u-boot-rockchip.bin
```

And as expected, without a proper FIT available or SPL DFU working, U-Boot SPL will fail trying to bootstrap a plain U-Boot `u-boot-rockchip.bin` image:

```
U-Boot TPL 2025.10-rc5 (Sep 29 2025 - 19:04:10)
lpddr4_set_rate: change freq to 400MHz 0, 1
Channel 0: LPDDR4, 400MHz
BW=32 Col=10 Bk=8 CS0 Row=15 CS=1 Die BW=16 Size=1024MB
Channel 1: LPDDR4, 400MHz
BW=32 Col=10 Bk=8 CS0 Row=15 CS=1 Die BW=16 Size=1024MB
256B stride
lpddr4_set_rate: change freq to 800MHz 1, 0
Trying to boot from BOOTROM
Returning to boot ROM...

spl:init dram

U-Boot SPL 2025.10-rc5 (Sep 29 2025 - 19:04:10 +0000)
board_spl_was_booted_from: failed to resolve brom_bootdevice_id a
Trying to boot from MMC1
mmc_load_image_raw_sector: mmc block read error
Error: -38
Trying to boot from MMC2
Card did not respond to voltage select! : -110
spl: mmc init failed with error: -95
Error: -95
SPL: failed to boot from all boot devices
### ERROR ### Please RESET the board ###
```

labgrid/util/agents/rkusbmaskrom.py
jluebbe: @Kwiboo Are you aware of a specific issue why snagboot couldn't work as well as this driver? They also use pyusb and no vendor-specific tool.

Kwiboo: > Are you aware of a specific issue why snagboot couldn't work as well as this driver? They also use pyusb and no vendor-specific tool.

@jluebbe snagboot does not support Rockchip, I did send review comments on the snagboot Rockchip PR, but there has not been any progress. 

This is adding optimized and full featured support to ramboot all Rockchip boards in my labgrid boardfarm using a simple template:

```
targets:
  main:
    resources:
      RemotePlace:
        name: !template $LG_PLACE
    drivers:
      SerialDriver: {}
      LinkPiSmartHUBPowerDriver: {}
      RKUSBMaskromDriver:
        initial: rkmaskrom471
        image: rkmaskrom472
      BootstrapStrategy: {}

images:
  rkmaskrom471: /srv/u-boot/$LG_PLACE/u-boot-rockchip-usb471.bin
  rkmaskrom472: /srv/u-boot/$LG_PLACE/u-boot-rockchip-usb472.bin

options:
  coordinator_address: 'labgrid'
```

This works fine for my limited need and allows me to not have to deal with trying to bend yet another third party tools or system like snagboot and pdudaemon to work for my need.

QSchulz: @jluebbe The original contributor for Rockchip support in snagboot doesn't seem to be active anymore since June 2nd last year, so I'm of the opinion we shouldn't wait for this hypothetical support in snagboot and go ahead with this here. If support comes along to snagboot in the future, then we can always deprecate this implementation here and switch to snagboot if we wanted to. What do you think?

doc/configuration.rst
```diff
@@ -2687,6 +2688,40 @@ Arguments:
...
+
+   images:
+     rkmaskrom471: 'path/to/u-boot-rockchip-usb471.bin'
+     rkmaskrom472: 'path/to/u-boot-rockchip-usb472.bin'
+
+Arguments:
+  - initial (str): optional, key in :ref:`images <labgrid-device-config-images>` containing the path
+    of an image that typically initializes DRAM of the target
+  - image (str): optional, key in :ref:`images <labgrid-device-config-images>` containing the path
+    of a bootloader image to load to start of DRAM, or to SRAM when an initial image is unused
```
QSchulz: suggestion (non-blocking): specify this can be overridden at runtime

The user can actually override (or provide if none is provided in the conf) the image via the ``load(filename: str)`` method of the ``BootstrapProtocol`` class.

But we cannot change ``initial``.

doc/configuration.rst
```diff
@@ -2687,6 +2688,40 @@ Arguments:
   - usb_loader (str): optional, key in :ref:`images <labgrid-device-config-images>` containing the path
     of a first-stage bootloader image to write
 
+RKUSBMaskromDriver
+~~~~~~~~~~~~~~~~~~
+An :any:`RKUSBMaskromDriver` is used to upload an image into a device in the
+*Rockchip USB Maskrom state*.
```
QSchulz: issue: confusing documentation

This states that we're expecting the device to be in the *USB Maskrom state* while we bind to an `RKUSBLoader` which

> describes a USB device in the *Rockchip loader state*.

and we even check here: https://github.com/labgrid-project/labgrid/pull/1721/changes/c091cde961a8e02f970bc492ff76942aaa43234f#diff-4717e427bdffde63a29a83ebf945dd18f8e5caf30896ef89f71f303891a66cecR89 that we are NOT in the LOADER state. So which is which?

doc/configuration.rst
QSchulz: > And as expected, without a proper FIT available or SPL DFU working, U-Boot SPL will fail trying to bootstrap a plain U-Boot

In u-boot-rockchip.bin, there's a FIT already. But I'm assuming since its information is not stored in the idblock header, we cannot know where to look for it in the file? Do you think it'd be possible to find where in the file the FIT binary is (without elaborate and error-prone guessing)? We do have binman symbols available in SPL so maybe there's something to do with that.

I'm wondering if another option isn't to generate a u-boot-rockchip-usb.bin in that format (or maybe rather the one from `boot_merger`?) in U-Boot. I'm assuming this is why you wanted to have the ability to upload to 0x472 directly with the first patch in this PR, so you can directly use `u-boot-rockchip-usb472.bin` generated by U-Boot without having to go through `boot_merger` or `mkimage`?

QSchulz (changes requested):

doc/configuration.rst
Kwiboo: Added a `RKUSBLoader: use correct loader state terminology in docs` commit to change the documentation to use `maskrom state` terminology instead of current ambiguous `loader state`, to matche the state that e.g. the `rkdeveloptool db` command possible executed by `RKUSBDriver` expects the device to be in.

Kwiboo:
Updated with few smaller changes/improvements:
- Use 30 second timeout for last packet to fix a timeout issue on RV1106
- Pad payload to be 64 bytes aligned to fix an issue initializing DRAM on RK3288 (cache line issue?)
- Update documentation to use `maskrom state` terminology consistently
- Transfer the header of IDBlock v2 images, matches the `UsbHead.bin` part of `boot_merger` loader images

doc/configuration.rst
Kwiboo: > I'm wondering if another option isn't to generate a u-boot-rockchip-usb.bin in that format (or maybe rather the one from `boot_merger`?) in U-Boot. I'm assuming this is why you wanted to have the ability to upload to 0x472 directly with the first patch in this PR, so you can directly use `u-boot-rockchip-usb472.bin` generated by U-Boot without having to go through `boot_merger` or `mkimage`?

With the added support for IDBlock v1 and v2 image format we could possible use `mkimage` to to generate a `u-boot-rockchip-usb.bin` in IDBlock format. Adding something like following to `rockchip-u-boot.dtsi` seem to work:

```
	simple-bin-usb {
		filename = "u-boot-rockchip-usb.bin";
		pad-byte = <0x00>;

		mkimage {
			args = "-n", CONFIG_SYS_SOC, "-T", "rksd";
			multiple-data-files;

			usb471 {
				type = "blob";
				filename = "u-boot-rockchip-usb471.bin";
			};
			usb472 {
				type = "blob";
				filename = "u-boot-rockchip-usb472.bin";
			};
		};
	};
```

I do not think using FIT format is ideal, mostly because no existing RK tooling support this format. A different option could also be to add basic RK loader image support using a `binman` etype.

doc/configuration.rst
Kwiboo: The text used closely matches current documentation of `RKUSBDriver` and other `BootstrapProtocol` drivers.
Instead of `usb_loader` for `RKUSBDriver` there is now an `initial` option for `RKUSBMaskromDriver` for similar first stage usb-loader/dram-init.

`BootstrapProtocol.load(filename: str)` and/or `labgrid-client bootstrap` is probably a better target for documenting how a user can override (or provide if none is provided in the conf) the image to bootstrap, possible something for  a different PR.

Kwiboo:
> Considering you're parsing the file generated by `boot_merger`... How difficult would it be to implement an `rkusb` type in `mkimage` such that we generate an `rkdeveloptool`-compatible image? It would help get rid of the awkwardness around handling two binaries (471 and 427) in labgrid (and encourage people using it outside of labgrid too). I'm not asking you to do it, I'm just wondering if I'm missing something. What stands out are the `_PARAM` INI sections (`BOOT0_PARAM`, `BOOT1_PARAM`, `LOADER2_PARAM`, etc...). I'm assuming the LOADER ones aren't necessary, only for the LOADER_OPTION section which I'm assuming is only needed with `rkdeveloptool ul`? I'm assuming we can carry something similar to the RKBOOT ini files but in DTS for binman?

Adding a new `binman` etype is probably a better option, should likely make it easier to add and handle `_PARAM` options. Source for `boot_merger` v1.31 can be found in git history at https://github.com/rockchip-linux/rkbin/blob/b851481d4f6438a766a8f8f4a01fdc32c748a399/tools/lib/boot_merger.py

> My understanding is that you need the environment (the one with `targets` like above) on the client-side. Considering this is where you define the images to use, you necessarily need the binary client-side as well (or is there some trick I'm unaware of?)... So what's the use-case of passing an argument to the bootstrap command if we need an environment file which defines the files to bootstrap anyway?

My understanding is also that managed files must/should be available on client, and will be rsync to exporter when needed. And I guess it all depends on how you make use of `labgrid` e.g. using cli, custom strategies, scripting and/or pytest.

> I'm asking because I frankly see as a blocker only being able to override `RKUSBMaskromDriver:image` from the CLI if `RKUSBMaskromDriver:initial` is set. This is the kind of awkwardness I was talking about earlier.

It all depend on how you configure and use `labgrid` for your specific need and/or environment. How the image/images you use with `RKUSBMaskromDriver` is created/generate is fully up to the user, the driver was designed to be as flexible as possible to not limit the user to specific use-cases.

Other existing RK tooling have different shortcomings and/or only support specific scenarios/use-cases, e.g. `rkdeveloptool` and `rockusb` use RK loader image format and typically require closed-source tools `boot_merger` to create such image, `rk-usb-loader` only support the IDBlock v2 image format and is limited to usb product-id of rk3568/rk3588. Instead with `RKUSBMaskromDriver` all common RK image formats or use of stand-alone blobs is fully supported, i.e. the user can make their own choice on how it is best used for their specific need.

saschahauer:
For what it's worth: I gave it a try and it works for me for the case I actually care about: Uploading a barebox image.

labgrid/util/agents/rkusbmaskrom.py
saschahauer: Fast forward a few months the Snagboot Rockchip support still hasn't made it into snagboot. Even then we would still need snagboot support for labgrid. A nice longterm goal would be to replace the several USB/xmodem/whatever Recovery mechanisms we have in labgrid with a snagboot integration and let the Snagboot maintainers handle the SoC specific hassles. That being a longterm goal I don't see that happening anytime soon, so I am also in favor for having working Rockchip Support in form of this Pull Request

a3f: > That being a longterm goal I don't see that happening anytime soon, so I am also in favor for having working Rockchip Support in form of this Pull Request

Agreed. Also the implementation here being licensed as LGPL-2.1 like the rest of Labgrid would allow its incorporation into snagrecover at a future point of time.

PR #1727, opened 2025-09-09
strategy: add arguments wait_for_systemd and systemd_timeout

joergho:
**Description**

```
Add the following arguments to the strategies BareboxStrategy, ShellStrategy and UBootStrategy:

- wait_for_systemd: If True, wait for all systemd services to be up and running when transitioning to the "shell" state default: True
- systemd_timeout: Timeout to wait for systemd services in seconds default: 30
```

Before the change, the `systemctl is-system-running --wait` command was called unconditionally with default timeout (30 seconds).

This does not cover the following scenarios:
1. Maybe the user does not want to execute the check at all (maybe systemd is not even used on the device)?
    - this is what the `wait_for_systemd` attribute is for
2. Or maybe the user wants to have a higher timeout because a certain service takes very long to start?
    - this is what the `systemd_timeout` attribute is for

**Checklist**
- [ ] Tests for the feature 

- [ ] PR has been tested

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -3458,6 +3458,11 @@ the "shell" state:
 This command would transition from the bootloader into a Linux shell and
 activate the `ShellDriver`_.
 
+Arguments:
+  - wait_for_systemd (bool, default=True): If True, wait for all systemd services
+    to be up and running when transitioning to the "shell" state
+  - systemd_timeout (int, default=30): Timeout to wait for systemd services in seconds
+
 ShellStrategy
 ~~~~~~~~~~~~~
 A :any:`ShellStrategy` has three states:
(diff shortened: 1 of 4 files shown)


joergho:
> This seems useful. How did you test it?

@threexc I tested it manually.

Environment for testing:

1. wrote an environment config file which works with my device and set the `UBootStrategy` arguments accordingly (see below)
2. created a dummy test similar to the `test_shell` test using the `shell_command` fixture from the [documentation](https://labgrid.readthedocs.io/en/latest/usage.html#strategy-fixture-example)
3. Ran pytest with `-vv` argument (e.g. `pytest --lg-env myconfig.yml test_shell.py -vv`) to see the steplogger output.

Things I checked:

1. Set `wait_for_systemd=False` in myconfig.yml and checked that `systemctl is-system-running --wait` is not executed
2. Set `wait_for_systemd=True` in myconfig.yml and checked that `systemctl is-system-running --wait` is executed
3. Set `systemd_timeout=30` (some value which is known to be working) and checked that the transition to shell works as expected
4. Set `systemd_timeout=1` (some value which is known to be working) and checked that executing `systemctl is-system-running --wait` fails

joergho:
BTW, now that I think of it: Wouldn't it make sense to move this functionality out of the strategies completely and instead move it into the `ShellDriver`?

PR #1731, opened 2025-09-11
Add the concept of a Button (i.e. reset) controlled by a DigitalOutput

pmelange:
**Description**

Adds support for a button (i.e. reset) controlled by a DigitalOutput.  Support for labgrid-client, similar to power, is also included.  Methods are press, release, press_for, and get.  With press_for a time period can be set.

This feature is needed for devices which need to have the reset button manipulated in order to do any of the following:
* reset the configuration
* enter into uboot
* start in tftpboot/bootp mode

Devices under test which require the reset button include at least
* mikrotik RB750GR3
* mikrotik SXTsq 5 AC
* Cudy WR3000

**Checklist**
- [x] Tests for the feature 
- [x] PR has been tested

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -2313,6 +2313,72 @@ Implements:
 Arguments:
   - delay (float, default=2.0): delay in seconds between off and on
 
+ManualButtonDriver
+~~~~~~~~~~~~~~~~~~
+A :any:`ManualButtonDriver` requires the user to control the taget button.
+This is required if a strategy is used with the target, but no automatic
+button control is available.
+
+The driver's name will be displayed during interaction.
+
+Binds to:
+  - None
+
+Implements:
+  - :any:`ButtonProtocol`
+
+.. code-block:: yaml
+
+   ManualButtonDriver:
+     name: 'example-board'
+
+Arguments:
+  - None
+
+ExternalButtonDriver
+~~~~~~~~~~~~~~~~~~~~
+An :any:`ExternalButtonDriver` is used to control a target button via an
+external command.
+
+Binds to:
+  - None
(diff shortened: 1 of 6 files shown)


Emantor:
I see the value on the user interface side to support buttons as a concept, however I am hesitant to take this into Labgrid core when the digital output driver can already model a button well enough and we have the `ManualSwitchDriver` which competes with the `ManualButtonDriver`.

pmelange:
> I see the value on the user interface side to support buttons as a concept, however I am hesitant to take this into Labgrid core when the digital output driver can already model a button well enough and we have the `ManualSwitchDriver` which competes with the `ManualButtonDriver`.

I added the ManualButtonDriver for completeness, even though I do not use it in my code.  But if you feel that a console message like "Set $BUTTONNAME for target $TARGET to 'ON' and press enter" means to push a button and hold it, and "Set $BUTTONNAME for target $TARGET to 'OFF' and press enter" means to stop holding the button, then fine.  But to a some users that might no be clear.

Some examples of how the button driver is used in my code can be seen here...

https://github.com/pmelange/openwrt-testlab/blob/50b65d2a7f01c6b83fb30a0b24b599f73d64e134/labgrid/driver/ubootinteraction.py#L182-L186
https://github.com/pmelange/openwrt-testlab/blob/50b65d2a7f01c6b83fb30a0b24b599f73d64e134/labgrid/driver/openwrtubootdriver.py#L84-L129
https://github.com/pmelange/openwrt-testlab/blob/50b65d2a7f01c6b83fb30a0b24b599f73d64e134/labgrid/strategy/openwrtstrategy.py#L176-L184

PR #1732, opened 2025-09-11
Shelldriver: make destination authorized keys file an arguement

### LINKED ISSUE #1626: Request: ShellDriver: put_ssh_key would benefit from a destination file attribute

pmelange:
**Description**

Change the ShellDriver so that the authorized_keys file can be set to something other than ```~/.ssh```.  This is necessary if the DUT is running dropbear, where the authorized_keys file is located (on OpenWrt) in ```/etc/dropbear/authorized_keys```.

The default value for the new arguement is ```~/.ssh``` and therefore any old code should not need to be updated.  In addition, the call to put_ssh_key() has a new optional arg where the user can set the destination authorized_keys file.  This could also be useful for setting up a multi-user system and to put the authorized_keys file for each user in their own home directory.

**Checklist**
- [x] PR has been tested

Fixes: #1626

### DIFF
--- a/labgrid/driver/shelldriver.py
+++ b/labgrid/driver/shelldriver.py
@@ -222,7 +222,6 @@ class ShellDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
 
     @step(args=['keyfile_path', 'dest_authorized_keys'])
     def _put_ssh_key(self, keyfile_path, dest_authorized_keys):
-        """Upload an SSH Key to a target"""
         regex = re.compile(
             r"""ssh-(rsa|ed25519)
             \s+(?P<key>[a-zA-Z0-9/+=]+) # Match Keystring
@@ -290,6 +289,13 @@ class ShellDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
 
     @Driver.check_active
     def put_ssh_key(self, keyfile_path, dest_authorized_keys = None):
+        """
+        Upload an SSH Key to a target
+
+        Arguments:
+        keyfile_path (str): Path to a list of authorized_keys for ssh access
+        dest_authorized_keys (str|None): Target location for the authorized_keys file
+        """
         if dest_authorized_keys is None:
             dest_authorized_keys = self.dest_authorized_keys
         self._put_ssh_key(keyfile_path, dest_authorized_keys)
--- a/labgrid/driver/shelldriver.py
+++ b/labgrid/driver/shelldriver.py
@@ -1,6 +1,7 @@
 # pylint: disable=unused-argument
 """The ShellDriver provides the CommandProtocol, ConsoleProtocol and
  InfoProtocol on top of a SerialPort."""
+import os
 import io
 import re
 import shlex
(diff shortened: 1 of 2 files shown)


labgrid/driver/shelldriver.py
```diff
@@ -217,8 +220,8 @@ def _write_key(self, keyline, dest):
             self._run_check(f'echo -n "{part}" >> {dest}')
         self._run_check(f'echo "" >> {dest}')
 
-    @step(args=['keyfile_path'])
-    def _put_ssh_key(self, keyfile_path):
+    @step(args=['keyfile_path', 'dest_authorized_keys'])
+    def _put_ssh_key(self, keyfile_path, dest_authorized_keys):
         """Upload an SSH Key to a target"""
```
Emantor: This documentation comment should be moved with the function.

labgrid/driver/shelldriver.py
```diff
@@ -258,35 +262,37 @@ def _put_ssh_key(self, keyfile_path):
...
+        out, err, exitcode = self._run(f"""mount --bind /tmp/labgrid-ssh/ {dest_authorized_keys_dir}""")
         if exitcode != 0:
-            self.logger.warning("Could not bind mount ~/.ssh directory: %s %s", out, err)
+            self.logger.warning("Could not bind mount %s directory: %s %s", 
+                                dest_authorized_keys_dir, out, err)
 
     @Driver.check_active
-    def put_ssh_key(self, keyfile_path):
-        self._put_ssh_key(keyfile_path)
+    def put_ssh_key(self, keyfile_path, dest_authorized_keys = None):
```
Emantor, line 291: While moving the documentation comment, please extend it for all arguments of the function.

Emantor (changes requested):

labgrid/driver/shelldriver.py
pmelange: updated

labgrid/driver/shelldriver.py
pmelange: updated

labgrid/driver/shelldriver.py
```diff
@@ -258,35 +261,44 @@ def _put_ssh_key(self, keyfile_path):
...
+            self._run_check(f"""chmod 700 {dest_authorized_keys_dir}""")
+            self.logger.debug("Creating %s", dest_authorized_keys)
+            self._write_key(keyline, dest_authorized_keys)
             self._run_check("rm ~/.test")
             return
 
         self.logger.debug("Key not on target and not writeable, using bind mount...")
         self._run_check('mkdir -p -m 700 /tmp/labgrid-ssh/')
-        self._run("cp -aL ~/.ssh/* /tmp/labgrid-ssh/")
+        self._run(f"""cp -aL {dest_authorized_keys_dir}/* /tmp/labgrid-ssh/""")
```
Emantor: Why do we need to use the triple `"` syntax here?

PR #1743, opened 2025-09-24
exporter: limit ser2net port range

benjamin1313:
This allows for specifying a start port and a range up from that to be used when looking for a free port to use with ser2net when creating a networkserialport resource. This can be helpful on networks where access to ports on the network is limited.

Currently this feature is planed for use in a labgrid setup where the exporter and duts are placed on a network separate from the main network the developers are on and there are restrictions on the tcp ports. Being able to specify what port(s) labgrid can use makes it easier to manage and keep track of what ports are open between the two networks.

**Checklist**
- [ ] Tests for the feature 
- [ ] PR has been tested

### DIFF
--- a/labgrid/remote/exporter.py
+++ b/labgrid/remote/exporter.py
@@ -235,7 +235,7 @@ class SerialPortExport(ResourceExport):
         assert self.local.avail
         assert self.child is None
         assert start_params["path"].startswith("/dev/")
-        self.port = get_free_port()
+        self.port = get_free_port(os.environ.get("LG_PREFERRED_NETWORKSERIAL_PORT", None))
 
         # Ser2net has switched to using YAML format at version 4.0.0.
         result = subprocess.run([self.ser2net_bin, "-v"], capture_output=True, text=True)
--- a/labgrid/util/helper.py
+++ b/labgrid/util/helper.py
@@ -15,10 +15,23 @@ from ..step import step
 
 re_vt100 = re.compile(r"(\x1b\[|\x9b)[^@-_a-z]*[@-_a-z]|\x1b[@-_a-z]")
 
-def get_free_port():
+def get_free_port(prefered_port=None):
     """Helper function to always return an unused port."""
     with closing(socket(AF_INET, SOCK_STREAM)) as s:
-        s.bind(('', 0))
+        if prefered_port is not None:
+            port_range = [int(i) for i in prefered_port.split(':')]
+            port = port_range[0]
+            max_range = port_range[1] if len(port_range) > 1 else 10
+            max_no_of_ports = port + max_range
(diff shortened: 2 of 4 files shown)


Emantor (commented):
I think it's usually more common to have a `STARTPORT:ENDPORT` definition instead of the `PORT:RANGE` used here, e.g. `4000:5000` instead of `4000:1000`. This might also simplify the code, what do you think?

Bastian-Krause:
I think this should not be limited to `ser2net`. If the exporter needs to bind anything else in the future, it could use that range as well. So a more generic name would be good.

I'm not a huge fan of adding more environment variables, a CLI argument for `labgrid-exporter` would be a better fit. Especially since `--isolated` exists for similar cases where the exporter/DUT network setup is configured in a special way. So maybe `--port-range=4000-5000`?

benjamin1313:
> I think it's usually more common to have a `STARTPORT:ENDPORT` definition instead of the `PORT:RANGE` used here, e.g. `4000:5000` instead of `4000:1000`. This might also simplify the code, what do you think?

Yeah I see what you mean and will look into it.


> I think this should not be limited to `ser2net`. If the exporter needs to bind anything else in the future, it could use that range as well. So a more generic name would be good.

Do you mean a global range for everything else that also needs a port?
Where I'm currently working on a Labgrid setup I'm not in charge of the network and I think giving Labgrid a range of ports to use for whatever is going to be a hard sell to the IT department in charge of the network.
Being able specify which ports are needed for what tool, service etc. is preferred.

jluebbe:
Please keep in mind that without allocating a new port on acquire, you have the risk that an out of date client connects to the port it knows even if it no longer has a lock.

benjamin1313:
@jluebbe yes i will look into it.

Is is currently understood what needs to be done is.
use `--port-range=4000-5000` instead of environment variable.
use start and stop port instead of start and number of ports to be used.
other resources calling `get_free_port()` should also use these ports.
and ´get_free_port()´ should choose a port at random in the range given in order to avoid the risk an out of date client connects to the port?

AiyionPrime:
@benjamin1313 I'm interested in a solution as well and have a few hours to spare on this topic.

I think I understand the requirements and will open a new PR this weekend if you don't object?

AiyionPrime:
> I don't think we should add a ser2net-specific override. [...]

_Originally posted by @jluebbe in https://github.com/labgrid-project/labgrid/issues/1743#issuecomment-3665413061_
            

Looking at it, the strictest approach to not be (ser2net-)specific would be to not implement this feature at all and instead reduce the ephemeral port range to the desired limits in order for the regular `bind(('', 0))` to pick it up.

I'll give it a shot tomorrow and find whether this more radical approach limits the systems usability.

PR #1746, opened 2025-09-25
driver/power/gude: Make gude.py generic for all Gude PDUs

### LINKED ISSUE #1263: Gude power strip access failed
Hi Everyone,

I'm using labgrid in its release v23.0.3. 
I bought a EPC 1202 from Gude (firmware v1.11) and want to use this driver to control my device power cycle.

I try all gude driver without success.
Here is the output of the ov.html:
```html
<!DOCTYPE html>
<html><head><!--
<meta http-equiv="powerstate" content="Power Port,1">
<meta http-equiv="powerstate" content="Power Port,1">
<meta http-equiv="powerstate" content="Power Port,1">
<meta http-equiv="powerstate" content="Power Port,1">
-->
</head><body></body></html>
```

And this is not compatible with available power drivers.
- gude.py => driver does not search for ov.html
- gude24.py => pattern is not compatible with ov.hml
- gude8031.py => driver does not search for ov.html
- gude8225.py => driver does not search for ov.html
- gude8316.py => pattern is not compatible with ov.hml

I'm quite confused with this and don't know how to handle it.
I modified gude24.py to fit the ov.html output, but don't know if it's the correct way.

Moreover, we can use http://192.168.0.2/statusjsn.js?components=1 to get statuses of outputs and parse using json.

BenM-github:
- changed the assert upper limit to a dynamic value.
- changed api endpoint to status.json, as it should be supported for all Gude PDUs

**Description**
We had a couple of issues with the current driver situation for Gude PDUs, so this is a suggestion of a single driver that might just work with all of them:

- `gude.py`, `gude24.py` and `gude8316.py` do not work for named ports, they depend on "Power Port <n>" in the HTML content, but that can be renamed in the Web UI:
  ```
  <meta http-equiv="powerstate" content="Power Port 1,0">
  <meta http-equiv="powerstate" content="Some Device,0">
  ```
- `gude8316.py` would return the wrong port with our Gude 8316-1, as there are two lines before the `<meta>` section and it relies on the line number. That might have been the firmware difference that's referred to in the driver?
  ```
  <!DOCTYPE html>
  <html><head><!--
  <meta http-equiv="powerstate" content="-,1">
  ...
  ```
- `gude8225.py` uses diffent APIs for `power_set` and `power_get`, which is a bit weird but works.
- `gude8031.py` seems to be the most consistent one and works well for us, so that's what we went for as a basis, but it just checks against a constant port limit of 20 (so it breaks with 24-port PDUs) and suggests only being compatible with certain models whilst it is the most generic one.

According to the manuals, the `status.json` approach should work with almost all of the PDUs listed as supported in the drivers, just with the 8225 and 8080 we couldn't find any reference to status.json, so it would be nice to test this out with the other PDUs and possibly just replace all of the drivers with a single one if it also works with those two.

What would you suggest there, is that approach feasible or are there more specific behaviour differences that we are missing here? Wizkd 

In general, are the power drivers for PDUs still supported or is pdudaemon now officially the replacement as mentioned in #1263 and #1031?

**Checklist**

- [x] PR has been tested

Fixes #1263

### DIFF
--- a/labgrid/driver/power/gude8031.py
+++ b/labgrid/driver/power/gude8031.py
@@ -1,11 +1,12 @@
 import requests
 
 # Driver has been tested with:
-# * Gude Expert Power Control 8031()
+# * Gude Expert Power Control 8031-1
 # * Gude Expert Power Control 87-1210-18
-#   This device needs to be used in 'Basic Compatible' mode for HTTP GET
-#   to be usable. Do not turn on session authentication.
-
+#
+# This device needs to be used in 'Basic Compatible' mode for HTTP GET
+# to be usable. Do not turn on session authentication.
+#
 # HTTP-GET API is defined in the Gude EPC-HTTP-Interface specification:
 # http://wiki.gude.info/EPC_HTTP_Interface
 #
@@ -20,7 +21,8 @@ PORT = 80
 
 def power_set(host, port, index, value):
     index = int(index)
-    assert 1 <= index <= 20
+    upper_limit = count_ports(host, port)
+    assert 1 <= index <= upper_limit, f'index ({index}) out of port range (1-{upper_limit})'
     # access the web interface...
     value = 1 if value else 0
     r = requests.get(f"http://{host}:{port}/status.json?components=0&cmd=1&p={index}&s={value}")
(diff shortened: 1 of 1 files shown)


Bastian-Krause:
> According to the manuals, the `status.json` approach should work with almost all of the PDUs listed as supported in the drivers, just with the 8225 and 8080 we couldn't find any reference to status.json, so it would be nice to test this out with the other PDUs and possibly just replace all of the drivers with a single one if it also works with those two.

I've just tested the 8012 and 8080 variants and these unfortunately do not support the json API.


> In general, are the power drivers for PDUs still supported or is pdudaemon now officially the replacement as mentioned in #1263 and #1031?

We won't deprecate any power backends soon, but I'd like to refrain from merging new drivers that are already supported by pdudaemon.

PR #1748, opened 2025-09-25
remote/client: make acquire/release operate on all RemotePlaces in env

Bastian-Krause:
**Description**
If an environment config with multiple `RemotePlace`s is given to `labgrid-client` via `-c`/`--config`, only the first `RemotePlace` is used on `acquire`/`lock` and `release`/`unlock`.

Having multiple `RemotePlace`s is meant for multi place testing, so all `RemotePlace`s need to be acquired/released.

This is now implemented in a best effort way: If an error occurs, the remaining places are still tried to be acquired/released. A combined error message is shown if the operation failed on multiple places.
When called with `-d`, tracebacks of all errors are shown. If any of the operations failed, the process exits with 1.

CI pipelines should use this pattern:

```shell
$ export LG_ENV=env.yaml
$ labgrid-client release || true
$ labgrid-client acquire
```

**Checklist**
- [x] Tests for the feature 
- [x] PR has been tested

### DIFF
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -724,10 +724,10 @@ class ClientSession:
             allowhelp = f"'labgrid-client -p {place.name} allow {self.gethostname()}/{self.getuser()}' on {host}."
             if self.getuser() == user:
                 if self.gethostname() == host:
-                    raise UserError("You have already acquired this place.")
+                    raise UserError(f"You have already acquired place {place.name}.")
                 else:
                     raise UserError(
-                        f"You have already acquired this place on {host}. To work simultaneously, execute {allowhelp}"
+                        f"You have already acquired place {place.name} on {host}. To work simultaneously, execute {allowhelp}"
                     )
             else:
                 raise UserError(


Bastian-Krause:
> Cool! Great work, Basti!

Thanks!

> > This is now implemented in a best effort way: If an error occurs, the remaining places are still tried to be acquired/released
> 
> You could make a case for failing early if you fail to acquire just 1 place. If an env file has multiple places, it is likely that all of them are needed for what you're trying to test.

I'd like `acquire` and `release` work the same way. If `release` fails early, the operation will leave places acquired. To overcome this, we would probably need some kind of rollback mechanism for `acquire` if any of the places fail (but that doesn't really make sense for `release`, does it?),  – but I'm hesitant to implement this. That's why I chose the best effort way.

> I'm not quite sure how other labgrid users would want this to be implemented. If there are any strong voices from within pengutronix or other users, I'd await their review.

Yes, input welcome! Since the whole multi place acquire/release is not atomic in any way, we should not over-engineer this, though.

PR #1751, opened 2025-09-30
pytestplugin: introduce `@pytest.mark.lg_xfail_feature`

Bastian-Krause:
**Description**
`@pytest.mark.lg_xfail_feature()` is meant to mark a test as ["xfail"](https://docs.pytest.org/en/stable/how-to/skipping.html#xfail) if the target or global environment contains the specified feature flag. There is no other sane way of testing for features in pytest's collection phase. Marking a test as xfail later is unsupported in pytest. So a `pytest.mark` decorator is necessary.

Imagine two targets contain the ``camera`` feature flag. One of them has the additional ``special-camera-2000`` feature flag. he other has the additional ``special-camera-3000`` feature flag. Due to a known bug on ``special-camera-3000``, the test is expected to fail. The test can be marked as ``xfail`` for that feature:                                     
                                                                                          
```python
import pytest

@pytest.mark.lg_feature("camera"])
@pytest.mark.lg_xfail_feature(
    "special-camera-3000",
    reason="known bug xy on special-camera-3000",
    raises=AssertionError,
    strict=True,
)
def test_camera(target):
    pass
```

While at it, improve error messages, skip reason and documentation of `@pytest.mark.lg_feature()` and feature flags. Especially the documentation changes make it easier to add documentation for the new `@pytest.mark.lg_xfail_feature()`.

**Checklist**
- [x] Tests for the feature 
- [x] PR has been tested


labgrid/pytestplugin/hooks.py
```diff
@@ -117,6 +122,24 @@ def pytest_collection_modifyitems(config, items):
             reason = f'unsupported feature(s): {", ".join(missing_feature)}'
             item.add_marker(pytest.mark.skip(reason=reason))
 
+        # pytest.mark.lg_xfail_feature
+        for marker in item.iter_markers("lg_xfail_feature"):
+            [feature] = marker.args
```
SmithChart: Is `markers.args` guaranteed to only have one element? Or could a user by accident add a second element (e.g. by adding a second parameter to the `@pytest.mark.lg_xfail_feature()` decorator)?

If it can contain more than one element I would prefer to report a meaningful exception to the user.
I would consider a `ValueError` somewhere in library code a bad error message.

Bastian-Krause: Yes, that can happen and it should be a proper `pytest.UsageError`. The same applies to `@pytest.mark.lg_feature` currently ignoring any extra args.

Emantor (commented):
Overall looks good, some smaller suggestions.

tests/test_flags.py
```diff
@@ -179,3 +179,55 @@ def test(self, env):
...
+@pytest.mark.lg_xfail_feature("test")
+def test(env):
+    assert False
+"""
+    )
+
+    with pexpect.spawn(f'pytest --lg-env {conf} {test}') as spawn:
+        spawn.expect("1 failed")
+        spawn.expect(pexpect.EOF)
+    assert spawn.exitstatus == 1
```
Emantor, line 210: We should test the `UsageError` as well.

labgrid/pytestplugin/hooks.py
```diff
@@ -110,18 +115,30 @@ def pytest_collection_modifyitems(config, items):
...
-                skip = pytest.mark.skip(
-                    reason=f'Skipping because features "{missing_feature}" are not supported'
+            reason = f'unsupported feature(s): {", ".join(missing_feature)}'
+            item.add_marker(pytest.mark.skip(reason=reason))
+
+        # pytest.mark.lg_xfail_feature
+        for marker in item.iter_markers("lg_xfail_feature"):
+            [feature] = marker.args
+            if not isinstance(feature, str):
+                raise pytest.UsageError("Unsupported feature argument type in pytest.mark.lg_xfail_feature")
```
Emantor: Please add the wrong and expected type here, this will make it easier for the user to change the code without going to the source code.

labgrid/pytestplugin/hooks.py
Bastian-Krause: Done.

tests/test_flags.py
Bastian-Krause: Done.

labgrid/pytestplugin/hooks.py
Bastian-Krause: Done.

Bastian-Krause:
Changes:
- raise `pytest.UsageError` for unexpected marker args and add expected function signatures to error message (for both `@pytest.mark.lg_feature` and `@pytest.mark.lg_xfail_feature`)
- add/use `env_feature_config` factory fixture for env config creation to `tests/test_flags.py`
- add tests for `pytest.UsageError`s (for both `@pytest.mark.lg_feature` and `@pytest.mark.lg_xfail_feature`)

SmithChart (approved):

PR #1752, opened 2025-10-01
driver/fastbootdriver: decode output to str before split

tretter:
processwrapper.check_output returns a byte-like object, which is not compatible with split() with an str. Decode the output to a str to fix this.

This fixes direct calls of the getvar function from a test suite. Calling `fastboot getvar` from the command line doesn't use the filtering mechanism and is successful without this fix.

The alternatives would have been to change the callers of _filter_fastboot_output to convert the byte-like object to a string, or to change '\n' to a byte-like object. The latter would also need a change of the startswith call.


jluebbe (approved):
Seems file to me if @Bastian-Krause is fine with this approach.

PR #1753, opened 2025-10-06
qemudriver: Allow machine-specific options and pre-start QMP commands

JSydll:
**Description**

Two minor features are added to the `QemuDriver`:

_Support machine-specific options_ (e.g. `-machine virt,secure=on`)

> While raising an exception for unsupported machines/boards is reasonable, adding machine-specific options shall not break the parsing. Therefore, the check is adjusted to only account for the leading board name.

_Support interactions with the QEMU instance via QMP before releasing CPU(s)_

> Especially for bad case testing, it can be useful to be able to interact with the prepared but not yet spun off instance - e.g. to corrupt the boot medium without the need to build a corrupted image only for that purpose.

For both features, straight forward unit tests are provided.
Lastly, the changeset includes minor documentation updates and the introduction of type hints for the `qemudriver.py` file.

**Checklist**
- [x] Tests for the feature 
- [x] PR has been tested

### DIFF
--- a/labgrid/driver/qemudriver.py
+++ b/labgrid/driver/qemudriver.py
@@ -156,19 +156,20 @@ class QEMUDriver(ConsoleExpectMixin, Driver, PowerProtocol, ConsoleProtocol):
             disk_opts = ""
             if self.disk_opts:
                 disk_opts = f",{self.disk_opts}"
-            if self.machine == "vexpress-a9":
+            machine_base = self.machine.split(',')[0]
+            if machine_base == "vexpress-a9":
                 cmd.append("-drive")
                 cmd.append(
                     f"if=sd,format={disk_format},file={disk_path},id=mmc0{disk_opts}")
                 boot_args.append("root=/dev/mmcblk0p1 rootfstype=ext4 rootwait")
-            elif self.machine in ["pc", "q35", "virt"]:
+            elif machine_base in ["pc", "q35", "virt"]:
                 cmd.append("-drive")
                 cmd.append(
                     f"if=virtio,format={disk_format},file={disk_path}{disk_opts}")
                 boot_args.append("root=/dev/vda rootwait")
             else:
                 raise NotImplementedError(
-                    f"QEMU disk image support not implemented for machine '{self.machine}'"
+                    f"QEMU disk image support not implemented for machine '{machine_base}'"
                 )
         if self.rootfs is not None:
             cmd.append("-fsdev")
--- a/labgrid/driver/qemudriver.py
+++ b/labgrid/driver/qemudriver.py
@@ -259,9 +259,9 @@ class QEMUDriver(ConsoleExpectMixin, Driver, PowerProtocol, ConsoleProtocol):
         shutil.rmtree(self._tempdir)
 
     @step()
-    def on(self):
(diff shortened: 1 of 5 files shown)


Emantor (commented):
Rather than adding a pre-start hook, I think it is more clear to expose somthing like a `prepare()` function which sets up the qemu process, and than have `on()` only do the `cont` monitor command. For backwards compatibility, the class can check whether the `self._child` attribute exists and call `prepare()` in `on()` if that was not done yet.

Emantor (dismissed):
I think we can take the other fixes in this PR as well.

Emantor:
It's all good and thanks for keeping up with the PR even if it takes some time on our side :+1:

JSydll:
I shortly glimpsed into the failing pipeline and now I'm wondering whether some environment changes happened? The test_qemudriver is not able to locate qemu-system-arm...

Emantor (dismissed):

Bastian-Krause:
Prior to this, we've mocked qemu. Why do we need to use the real binary now?

If we decide we need the real qemu, we should at least add a marker so the tests get skipped on systems without it. Something along the lines of:

https://github.com/labgrid-project/labgrid/blob/f63dfd70a5da2c7007ee8c320fa969df2c0bf16e/tests/test_sigrok.py#L11-L12

JSydll:
Hmm, I'm actually not sure where this dependency is now coming from. It's not like the tests I want add depend on it. Maybe I forgot to inject the qemu_mock in one of the test cases? Could that be the root cause? I'll try this as soon as time permits.

labgrid/driver/qemudriver.py
```diff
@@ -286,12 +288,19 @@ def on(self):
 
         # Restore port forwards
         for v in self._forwarded_ports.values():
-            self._add_port_forward(*v)
+            self._add_port_forward(*v)
```
Bastian-Krause:
```suggestion
            self._add_port_forward(*v)
```

Bastian-Krause:
After looking at the QEMUDriver again, I'm not too happy about starting the qemu process in `on()` (or with this PR in `prepare()` and stopping it in `off()`. This is not the fault of this PR, though. This PR only makes it even more visible.

I'll create an alternative PR starting qemu during `on_activate()` and stopping it in `on_deactivate()`. `on()` and `off()` will interact with it via QMP only. This should cover your use case because you can call `monitor_command()` right after activation. If we agree that this is a good solution, you can rebase your other changes on top.

Bastian-Krause:
> After looking at the QEMUDriver again, I'm not too happy about starting the qemu process in `on()` (or with this PR in `prepare()` and stopping it in `off()`. This is not the fault of this PR, though. This PR only makes it even more visible.
> 
> I'll create an alternative PR starting qemu during `on_activate()` and stopping it in `on_deactivate()`. `on()` and `off()` will interact with it via QMP only. This should cover your use case because you can call `monitor_command()` right after activation. If we agree that this is a good solution, you can rebase your other changes on top.

See #1771.

JSydll:
Makes sense. I haven't expected a proposal so early, thanks!

Bastian-Krause:
#1771 was merged, please rebase.

PR #1763, opened 2025-10-29
Fix _check_prompt with interjected logs

LukasLendvorsky:
**Description**
ShellDriver._check_prompt uses regexp that expects you get marker immediately followed by prompt, only separated by whitespace.
But sometimes, you can get some unrelated message on the shell console in between the marker and prompt, something like this:
```
root@user:~# echo 'SVVZ''SYBQPQ'
SVVZSYBQPQ
[   37.369462] systemd-journald[159]: Time jumped backwards, rotating.
root@user:~#

```
This is enough to break _check_prompt, it will timeout as the regexp never maches. This fixes the issue by waiting separately for the marker and prompt.

### DIFF
--- a/labgrid/driver/shelldriver.py
+++ b/labgrid/driver/shelldriver.py
@@ -197,8 +197,12 @@ class ShellDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
         self.console.sendline(f"echo '{marker[:4]}''{marker[4:]}'")
         try:
             self.console.expect(
-                rf"{marker}\s+{self.prompt}",
+                rf"{marker}\s+",
                 timeout=30
+            )
+            self.console.expect(
+                rf"{self.prompt}",
+                timeout=1
             )
             self._status = 1
         except TIMEOUT:


jluebbe:
Nothing guarantees that the kernel log is on its own line. To really fix this, you need to disable console logging via dmesg (which the ShellDriver tries to do very early).

LukasLendvorsky:
I don't have any newline expectations in my regexp, it should still pass even if its not on its own line. 
I found the code that is supposed to disable the logs and have to agree that it would be better solution. Unfortunately it doesn't work for me. The code looks like this:
```
if did_login and not did_silence_kernel:
                    # Silence the kernel and wait for another prompt
                    self.console.sendline("dmesg -n 1")|
```

But my device has auto-login on serial console, so did_login is false so it never disables the logs. I would then propose to set it everytime, regardless of did_login, WDYT?

LukasLendvorsky:
Closing this in favor of https://github.com/labgrid-project/labgrid/pull/1765, which should be better fix for the issue.

PR #1766, opened 2025-11-05
Fix Shelly Gen 2 power_get()

vasileio:
**Description**
Fix KeyError in shelly_gen2 power_get() function due to incorrect JSON response parsing.  The current implementation attempts to access `response["output"]` directly, but the Shelly Gen 2 RPC API returns the output field nested inside a `result` object.

API Response Structure:

```json
{
  "id": 1,
  "result": {
    "output": true,
    "apower": 0.0,
    "voltage": 245.3
  }
}
```

Fix:
Change r.json()["output"] to r.json()["result"]["output"] in line 30.

Testing:
* Verified with Shelly Plus Plug UK device
* All power operations (get/set/cycle) now work correctly
* Compatible with existing NetworkPowerPort configurations

**Checklist**
- [x] Tests for the feature 
- [x] PR has been tested


jiez:
I tested it using Shelly Plug US Gen4. This PR works for me.

Emantor (approved):

PR #1707, opened 2025-07-25
Generate labgrid-client subcommand and option documentation from argparse

jonrebm:
Manage the documentation with sphinx and auto-document labgrid-client arguments and subcommands from source using sphinx-argparse.

Sections from the original man/labgrid-client.rst which are not obsoleted by the autogenerated reference are migrated into sphinx. Sphinx generates both the manpages as well as their html documentation counterpart.

I believe the result of this is much more useful than the current manpages we have that just don't sufficiently document usage of the subcommands.

See the readthedocs builds for results.

PR #1767, opened 2025-11-05
Generate CLI documentation using sphinx with autoprogram

jonrebm:
This is a second try after the formatting in https://github.com/labgrid-project/labgrid/pull/1707 using sphinx-argparse did not satisfy.

Manage the documentation with sphinx and auto-document labgrid-client arguments and subcommands from source using sphinx autoprogram.

Sections from the original man/labgrid-client.rst which are not obsoleted by the autogenerated reference are migrated into the argparse structure. Sphinx generates both the manpages as well as their html documentation counterpart.

I believe the result of this is much more useful than the current manpages we have that just don't sufficiently document usage of the subcommands.

See the [readthedocs builds](https://labgrid--1767.org.readthedocs.build/en/1767/man.html) for results.

### DIFF
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -1884,7 +1884,10 @@ def get_parser(auto_doc_mode=False) -> "argparse.ArgumentParser | AutoProgramArg
     subparser = subparsers.add_parser("show", help="show a place and related resources")
     subparser.set_defaults(func=ClientSession.print_place)
 
-    subparser = subparsers.add_parser("create", help="add a new place")
+    subparser = subparsers.add_parser(
+        "create",
+        help="add a new place with the name specified via --place or the LG_PLACE environment variable",
+    )
     subparser.set_defaults(func=ClientSession.add_place)
 
     subparser = subparsers.add_parser("delete", help="delete an existing place")
@@ -1932,7 +1935,10 @@ def get_parser(auto_doc_mode=False) -> "argparse.ArgumentParser | AutoProgramArg
     subparser.set_defaults(func=ClientSession.release)
 
     subparser = subparsers.add_parser(
-        "release-from", help="atomically release a place, but only if locked by a specific user"
+        "release-from",
+        help="atomically release a place, but only if locked by a specific user",
+        epilog="Note that this command returns success as long as the specified user no longer owns the place, "
+        "meaning it may be acquired by another user or not at all.",
     )
     subparser.add_argument("acquired", metavar="HOST/USER", help="User and host to match against when releasing")
     subparser.set_defaults(func=ClientSession.release_from)
@@ -1982,7 +1988,7 @@ def get_parser(auto_doc_mode=False) -> "argparse.ArgumentParser | AutoProgramArg
     subparser.add_argument("--name", "-n", help="optional resource name")
     subparser.set_defaults(func=ClientSession.fastboot)
 
-    subparser = subparsers.add_parser("flashscript", help="run flash script")
+    subparser = subparsers.add_parser("flashscript", help="Run arbitrary script with arguments to flash device")
     subparser.add_argument("script", help="Flashing script")
     subparser.add_argument("script_args", metavar="ARG", nargs=argparse.REMAINDER, help="script arguments")
     subparser.add_argument("--name", "-n", help="optional resource name")
--- a/man/labgrid-client.1
+++ b/man/labgrid-client.1
@@ -259,7 +259,7 @@ Log output to FILE
 .UNINDENT
 .SS labgrid\-client create
 .sp
-add a new place
+add a new place with the name specified via \-\-place or the LG_PLACE environment variable
 .INDENT 0.0
 .INDENT 3.5
 .sp
@@ -404,7 +404,7 @@ optional resource name
 .UNINDENT
 .SS labgrid\-client flashscript
 .sp
-run flash script
+Run arbitrary script with arguments to flash device
 .INDENT 0.0
 .INDENT 3.5
 .sp
@@ -552,6 +552,8 @@ usage: labgrid\-client release\-from HOST/USER
 .B host/user
 User and host to match against when releasing
 .UNINDENT
+.sp
+Note that this command returns success as long as the specified user no longer owns the place, meaning it may be acquired by another user or not at all.
 .SS labgrid\-client release|unlock
 .sp
 release a place


jonrebm:
- Rebased onto master integrating the manpage changes that happened in the meantime
- Updated to keep tracking generated troff manpages as suggested by @Bastian-Krause in https://github.com/labgrid-project/labgrid/pull/1707#issuecomment-3333218098
- Formatted to pass "Format with ruff" in CI.

Ready to merge as far as I can tell

jonrebm:
Okay, thanks.

Not ready for merge just yet. CI is using a different version of sphinx than what I used to generate them. This is most likely the reason why it outputs manpages with a slightly different formattingin some of the pipleines (and CI is configured to guarantee manpages are generated as tracked).

jonrebm:
> Nice, this is so much better!

Happy to hear!

> I haven't looked through this thoroughly, but I noticed that positional arguments are currently indistinguishable from sub commands in "usage", e.g.
> 
>     * https://labgrid--1767.org.readthedocs.build/en/1767/man/client.html#labgrid-client-allow
> 
>     * https://labgrid--1767.org.readthedocs.build/en/1767/man/client.html#labgrid-client-add-alias

autoprogram will exactly print the usage line as generated by argparse, we have (and had) the same ambiguity when usage is shown via e.g. `labgrid-client allow -h`. I'm not sure why argparse displays it this way but it's out of scope for this pr.

labgrid/remote/client.py
```diff
@@ -1708,8 +1708,35 @@ def __str__(self):
...
 
-def get_parser() -> argparse.ArgumentParser:
-    parser = argparse.ArgumentParser()
+def get_parser(include_undocumented=False) -> argparse.ArgumentParser:
+
+    # A fixed-width formatter to make usage string formatting predictable for manpage generation
+    def formatter(prog):
+        return argparse.HelpFormatter(prog, width=(None if include_undocumented else 180))
+
+    parser = argparse.ArgumentParser(add_help=include_undocumented, formatter_class=formatter)
```
Bastian-Krause: `HelpFormatter` is not public API: https://github.com/python/cpython/blob/35528fccdcaa0890e959eb7884332d1a426819ac/Lib/argparse.py#L62

`ArgumentParser` expects `formatter_class` to be a class, you're passing an instance. No idea how that works.

Couldn't we just set the `COLUMNS` environment variable to "80" here if `include_undocumented`?

labgrid/remote/client.py
```diff
@@ -1708,8 +1708,35 @@ def __str__(self):
...
+            if aliases:
+                name = "|".join([name] + list(aliases))
+            # the "help" message is ignore by autoprogram. Use "description" instead.
+            if "description" not in kwargs and "help" in kwargs:
+                kwargs["description"] = kwargs.pop("help")
+            # make usage string formatting predictable
+            kwargs["formatter_class"] = formatter
+            return argparse_add_parser(self, name, **kwargs)
+
+        argparse._SubParsersAction.add_parser = man_add_parser
```
Bastian-Krause: `argparse._SubParsersAction.add_parser` is not public API. I think it would be better to subclass `argparse.ArgumentParser` and override `add_subparsers` with the above. Then use the subclassed variant if `include_undocumented`, otherwise `argparse.ArgumentParser`.

labgrid/remote/client.py
```diff
@@ -1708,8 +1708,35 @@ def __str__(self):
         return self.value
 
 
-def get_parser() -> argparse.ArgumentParser:
-    parser = argparse.ArgumentParser()
+def get_parser(include_undocumented=False) -> argparse.ArgumentParser:
```
Bastian-Krause: `include_undocumented` does a lot more than including undocumented things. This needs a better name.

labgrid/remote/client.py
```diff
@@ -1747,11 +1774,12 @@ def get_parser() -> argparse.ArgumentParser:
...
-    subparser = subparsers.add_parser("help")
+    if include_undocumented:
+        subparser = subparsers.add_parser("help")
 
-    subparser = subparsers.add_parser("complete")
-    subparser.add_argument("type", choices=["resources", "places", "matches", "match-names"])
-    subparser.set_defaults(func=ClientSession.complete)
+        subparser = subparsers.add_parser("complete")
+        subparser.add_argument("type", choices=["resources", "places", "matches", "match-names"])
+        subparser.set_defaults(func=ClientSession.complete)
```
Bastian-Krause, line 1856: Could also happen in the subclassed `argparse.ArgumentParser`.

labgrid/remote/client.py
```diff
@@ -2000,7 +2028,7 @@ def get_parser() -> argparse.ArgumentParser:
         "-p",
         "--partition",
         type=int,
-        choices=range(0, 256),
+        choices=(range(0, 256) if include_undocumented else None),
```
Bastian-Krause: Could also happen in the subclassed `argparse.ArgumentParser`.

labgrid/remote/client.py
```diff
@@ -1810,7 +1810,10 @@ def man_add_parser(self, name, **kwargs):
     subparser = subparsers.add_parser("show", help="show a place and related resources")
     subparser.set_defaults(func=ClientSession.print_place)
 
-    subparser = subparsers.add_parser("create", help="add a new place")
+    subparser = subparsers.add_parser(
+        "create",
+        help="add a new place with the name specified via ``--place`` or the ``LG_PLACE`` environment variable",
```
Bastian-Krause: I don't think we should use rst syntax in help texts.

doc/conf.py
```diff
@@ -21,7 +21,7 @@
 import sys
 sys.path.insert(0, os.path.abspath('..'))
 
-from importlib.metadata import version
+from importlib import metadata
```
Bastian-Krause, line 24: What's the reason for this? Seems unrelated.

doc/conf.py
```diff
@@ -65,7 +65,7 @@
 # built documents.
 #
 # The full version, including alpha/beta/rc tags.
-release = version('labgrid')
+release = metadata.version('labgrid')
```
Bastian-Krause, line 69: What's the reason for this? Seems unrelated.

doc/man/client.rst
```diff
@@ -1,55 +1,23 @@
 .. _labgrid-client:
 
-================
- labgrid-client
-================
+labgrid-client CLI
```
Bastian-Krause, line 3: Looks strange. Maybe rename the labgrid-client section below to "CLI"?

doc/man/coordinator.rst
```diff
@@ -1,2 +1,34 @@
-.. _labgrid-coordinator:
-.. include:: ../../man/labgrid-coordinator.rst
+=====================
+ labgrid-coordinator
+=====================
+
+labgrid-coordinator managing labgrid resources and places
+=========================================================
+
```
Bastian-Krause, line 7: I think we should drop this. Makes the man page and HTML menu look strange.

doc/man/device-config.rst
```diff
@@ -1,2 +1,241 @@
 .. _labgrid-device-config:
-.. include:: ../../man/labgrid-device-config.rst
+
+=======================
+ labgrid-device-config
+=======================
+
+labgrid test configuration files
+================================
```
Bastian-Krause, line 8: Same here: we should drop this.

doc/man/exporter.rst
```diff
@@ -1,2 +1,105 @@
 .. _labgrid-exporter:
-.. include:: ../../man/labgrid-exporter.rst
+
+==================
+ labgrid-exporter
+==================
+
+labgrid-exporter interface to control boards
+============================================
+
```
Bastian-Krause, line 9: Same here: we should drop this.

doc/man/suggest.rst
```diff
@@ -5,19 +5,6 @@
 labgrid-suggest generator for YAML config files
 ===============================================
```
Bastian-Krause, line 6: Same here: we should drop this.

doc/conf.py
jonrebm: It's just a minor unrelated change.
I was looking at these lines and found them problematic due to the naming conflict of sphinx `version` configuration and the `version` import:
```python
release = version('labgrid')
version = '.'.join(release.split('.')[:2])
```
Should I move this change to a separate commit?

doc/conf.py
jonrebm: see above

doc/man/client.rst
jonrebm: Sadly, unlike as with sphinx-argparse it is not possible to rename that below section without modifying the autoprogram-plugin.
I also have tried and not succeeded in finding a way to leave the page titled "labgrid-client" while autoprogram generates a "labgrid-client" that would not generate a warning about duplicate autosection labels. Those warnings are promoted to errors in CI. We'd have to disable `SPHINXOPTS=-W` or disable auto labels (which I didn't way a way to do for only select pages).
I've been pouring an amount of time into working around quirks of autoprogram that I've come to believe we need to either live with some quirks for now or just drop the external dependency and do the manpage generation ourselves in-tree.

Would that be a quirk you could live with?

doc/man/coordinator.rst
jonrebm: Can we treat improvements on the non-generated manpages as out of scope for this changeset? I only integrated them with sphinx without change in content for consistency.

doc/man/device-config.rst
jonrebm: see above

doc/man/exporter.rst
jonrebm: see above

doc/man/suggest.rst
jonrebm: see above

doc/conf.py
Bastian-Krause: Yes please, just add that detail to the commit message.

doc/man/client.rst
Bastian-Krause: Let's keep it as it currently is. We can try improving this in a separate PR.

doc/man/coordinator.rst
Bastian-Krause: Sounds okay to me.

labgrid/remote/client.py
jonrebm: I'm moving the other places into the proposed subclass. In this case however I'd like to leave it as-is. Putting a conditional at option/subparser addition to me seems cleaner than keeping in the subclass a list of subparsers/options to retroactively skip.

labgrid/remote/client.py
jonrebm: I don't see how. This is a rather specific cosmetic for the `-p` option to `write-files`. As above I would not like to maintain a parallel data structure of fixes inside the subclass.

labgrid/remote/client.py
Bastian-Krause: True.

labgrid/remote/client.py
Bastian-Krause: I've added a generic way of hiding choices with ranges in them.

jonrebm:
Thank you and looks good to me

PR #1769, opened 2025-11-10
Feature: client: Add scp multifile support

sgross-emlix:
**Description**

Implement `-r` and `[SOURCES...] [DESTINATION]` for `scp`
Collegues requested this feature since it matches the expectation on how to use the widely known `scp` command.

**Checklist**
- [x] PR has been tested

### to Target
```
labgrid@phoebe:~/client$ lgc scp dummy/bar dummy/foo :/tmp
bar                                                                                                                                                                                100%    3     1.2KB/s 00:00
foo                                                                                                                                                                                100%    5     3.3KB/s   00:00
labgrid@phoebe:~/client$ lgc scp -r dummy :/tmp
foo                                                                                                                                                                                100%    5     2.6KB/s   00:00
bar                                                                                                                                                                                100%    3     1.7KB/s   00:00
labgrid@phoebe:~/client$ lgc ssh


BusyBox v1.34.1 () built-in shell (ash)

# cd /tmp
# ls
WebKitCache          bar                  env.yml              foo                  mesa_shader_cache    weston
avahi-daemon.socket  dummy                fontconfig           hsts-storage.sqlite  update-tool.lock
# cd dummy/
# ls
bar  foo
```
### from Target

```
labgrid@phoebe:~/client$ mkdir rev
labgrid@phoebe:~/client$ lgc scp :/tmp/dummy/foo :/tmp/dummy/bar rev/
foo                                                                                                                                                                                100%    5     1.0KB/s   00:00
bar                                                                                                                                                                                100%    3     0.8KB/s   00:00
labgrid@phoebe:~/client$ lgc scp -r :/tmp/dummy rev
bar                                                                                                                                                                                100%    3     0.4KB/s   00:00
foo    
labgrid@phoebe:~/client$ find rev
rev
rev/dummy
rev/dummy/foo
rev/dummy/bar
rev/foo
rev/bar
```

### DIFF
--- a/labgrid/driver/sshdriver.py
+++ b/labgrid/driver/sshdriver.py
@@ -356,25 +356,38 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
             yield localport
 
     @Driver.check_active
-    @step(args=['src', 'dst'])
-    def scp(self, *, src, dst):
+    @step(args=['src', 'dst', 'recursive'])
+    def scp(self, *, src, dst, recursive=False):
         if not self._check_keepalive():
             raise ExecutionError("Keepalive no longer running")
 
-        if src.startswith(':') == dst.startswith(':'):
+        if not isinstance(src, list):
+            src = [src]
+
+        # take Path like objects into account
+        src = [str(f) for f in src]
+        dst = str(dst)
+
+        remote_src = [f.startswith(':') for f in src]
+        if any(remote_src) != all(remote_src):
+            raise ValueError("All sources must be consistently local or remote (start with :)")
+
+        if all(remote_src) == dst.startswith(':'):
             raise ValueError("Either source or destination must be remote (start with :)")
-        if src.startswith(':'):
-            src = '_' + src
-        if dst.startswith(':'):
-            dst = '_' + dst
+
+        src = [s.replace(':', '_:') for s in src]
+        dst = dst.replace(':', '_:')
 
         complete_cmd = [self._scp,
                 "-S", self._ssh,
                 "-F", "none",
                 "-o", f"ControlPath={self.control.replace('%', '%%')}",
-                src, dst,
+                *src,
+                dst,
         ]
-        
+
+        if recursive:
+            complete_cmd.insert(1, "-r")
         if self.explicit_sftp_mode and self._scp_supports_explicit_sftp_mode():
             complete_cmd.insert(1, "-s")
         if self.explicit_scp_mode and self._scp_supports_explicit_scp_mode():
@@ -594,3 +607,4 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
         if stdout:
             for line in stdout.splitlines():
                 self.logger.warning("Keepalive %s: %s", self.networkservice.address, line)
+
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -1306,7 +1306,7 @@ class ClientSession:
     def scp(self):
         drv = self._get_ssh()
 
-        res = drv.scp(src=self.args.src, dst=self.args.dst)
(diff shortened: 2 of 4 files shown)


sgross-emlix:
I would like to update tests, but I have no idea where to start. `test_sshdriver` looks confusing to me

labgrid/remote/client.py
```diff
@@ -1922,8 +1922,10 @@ def main():
 
     subparser = subparsers.add_parser("scp", help="transfer file via scp")
     subparser.add_argument("--name", "-n", help="optional resource name")
-    subparser.add_argument("src", help="source path (use :dir/file for remote side)")
-    subparser.add_argument("dst", help="destination path (use :dir/file for remote side)")
+    subparser.add_argument("--recursive", "-r", action="store_true", help="copy recursive")
+    subparser.add_argument(
+        "files", nargs="+", metavar="SRC/DST", help="source and destination path (use :dir/file for remote side)"
+    )
```
jluebbe: You should be able to just use `nargs="+"` for `src` and keep the separate `dst` argument.

labgrid/driver/sshdriver.py
```diff
@@ -356,24 +356,31 @@ def forward_unix_socket(self, unixsocket, localport=None):
...
+        else:
+            if not dst.startswith(':'):
+                raise ValueError("Either source or destination must be remote (start with :)")
             dst = '_' + dst
 
         complete_cmd = [self._scp,
                 "-S", self._ssh,
                 "-F", "none",
                 "-o", f"ControlPath={self.control.replace('%', '%%')}",
-                src, dst,
```
jluebbe, line 375: You can use `*src` to avoid the insert loop below.

labgrid/driver/sshdriver.py
```diff
@@ -356,24 +356,31 @@ def forward_unix_socket(self, unixsocket, localport=None):
             yield localport
 
     @Driver.check_active
-    @step(args=['src', 'dst'])
-    def scp(self, *, src, dst):
+    @step(args=['src', 'dst', 'recursive'])
+    def scp(self, *, src, dst, recursive=False):
```
jluebbe, line 360: We need to consider backwards compatibility here. So, if `src` is a string, wrap it in a list.

labgrid/driver/sshdriver.py
```diff
@@ -356,24 +356,31 @@ def forward_unix_socket(self, unixsocket, localport=None):
...
         if not self._check_keepalive():
             raise ExecutionError("Keepalive no longer running")
 
-        if src.startswith(':') == dst.startswith(':'):
-            raise ValueError("Either source or destination must be remote (start with :)")
-        if src.startswith(':'):
-            src = '_' + src
-        if dst.startswith(':'):
+        if all([f.startswith(':') for f in src]):
```
jluebbe: You need to check that either all `src` elements start with ':' or none of them do.

So perhaps first something like
``` python
remote_src = [f.startswith(':') for f in src]
if any(remote_src) != all(remote_src):
   raise ValueError("All sources must be consistently local or remote (start with :)")
```
before the other checks?

jluebbe (changes requested):
You can pull out bb72bc0158019aaa16bfbf2794c6500e6ddfbf79 into a separate PR so we can merge that early.

sgross-emlix:
@jluebbe I addressed all issues and implemented tests. Is there anything else you would need from my side?

Bastian-Krause:
Similar to #1904, we should decide what to merge.

sgross-emlix:
> Similar to #1904, we should decide what to merge.

IMHO this PR deals more with the scp driver side and seems a bit more mature ie. adaptet tests and documentation.
Besides globbing is different from `-r`.
I would suggest cherry-picking the globbing part in `client.py` from #1904 so the other PR just deals with `sshfs`

Would this be ok with you?

PR #1782, opened 2025-11-25
agentwrapper: Allow specyfing Python executable

steelman:
**Description**

Enable users to choose which Python executable to use to run the agent by setting LG_AGENT_PYTHON environment variable. This is necessary when a remote labgrid runs in a virtual environment which provide packages, or their versions, not available in the system installation of Python.

**Checklist**
- [ ] Tests for the feature 

- [x] PR has been tested


steelman:
I haven't thought about connecting to many different exporters at once, indeed. In such case, I suppose, each exporter should advertise the interpreter it uses and clients need to discover it.

Bastian-Krause:
Solved via #1894.

PR #1783, opened 2025-11-28
remote/exporter: Export invert parameter on relays

DiogoSilva14:
The invert parameter is currently not being exported on Network HID and Deditec relays. Since the client is the one responsible for inverting the signal, and it does not recieve this information from the exporter (which causes it to default to False), the invert is only working on local setups and not on network exported devices.


steelman:
I've got this one on my branch too (-:

jluebbe (approved):

jluebbe (changes requested):
Looks good. Please add the DCO sign-off: https://labgrid.readthedocs.io/en/stable/development.html#developer-s-certificate-of-origin

jluebbe (approved):

PR #1785, opened 2025-12-08
remote/config: support drop-in files

steelman:
Pass "dropins" variable to a template being rendered containing a list of `*.yaml` files in a directory named after the template being rendered with ".d" appended. For example when rendering /etc/labgrid/configuration.yaml a list of files matching `/etc/labgrid/configuration.yaml.d/*.yaml` will be available. Then `configuration.yaml` may be as simple as

{% for f in dropins %}
{% include f %}
{% endfor %}

And adding a target is possible by simply creating a file in `/etc/labgrid/configuration.yaml.d` directory.

### DIFF
--- a/labgrid/remote/config.py
+++ b/labgrid/remote/config.py
@@ -1,5 +1,6 @@
 from pprint import pprint
 
+import glob
 import os
 
 import attr
@@ -15,8 +16,9 @@ class ResourceConfig:
     template_env = attr.ib(default=attr.Factory(dict), validator=attr.validators.instance_of(dict))
 
     def __attrs_post_init__(self):
+        _dirname = os.path.dirname(self.filename)
         env = jinja2.Environment(
-            loader=jinja2.FileSystemLoader(os.path.dirname(self.filename)),
+            loader=jinja2.FileSystemLoader(_dirname),
             line_statement_prefix="#",
             line_comment_prefix="##",
         )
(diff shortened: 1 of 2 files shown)


labgrid/remote/config.py
```diff
@@ -3,6 +3,7 @@
 import os
 
 import attr
+import glob
```
obbardc: nitpick: in commit message & github PR description you've got `{% for f in dorpins %}` but it should be `{% for f in dropins %}`

labgrid/remote/config.py
```diff
@@ -24,6 +25,8 @@ def __attrs_post_init__(self):
             template = env.get_template(os.path.basename(self.filename))
         except jinja2.TemplateNotFound:
             raise NoConfigFoundError(f"{self.filename} could not be found")
+        drop_ins = sorted(glob.glob(os.path.join(self.filename + '.d','*.yaml')))
+        self.template_env["dropins"] = drop_ins
```
obbardc:
```suggestion
        self.template_env["drop_ins"] = drop_ins
```

nitpick: perhaps `drop_ins` is a better name ?

labgrid/remote/config.py
steelman: DONE

labgrid/remote/config.py
steelman: DONE

PR #1787, opened 2025-12-11
util/helper: add stderr param to ProcessWrapper

ImHoppy:
**Description**

Add a `stderr` parameter to the `ProcessWrapper`, allowing the separation of `stderr` from the output. In my driver, I use this feature to ignore `stderr` by redirecting it to `subprocess.DEVNULL`

**what do you use the feature for?**
  It allows calling `ProcessWrapper` without having `stderr` polluting the standard output.

**how does labgrid benefit as a testing library from the feature?**
  This feature is used in my custom `IpmiPowerDriver` to check the status without having debug messages mixed.

**how did you verify the feature works?**
  I tested the feature with a DUT in my labs.

**if hardware is needed for the feature, which hardware is supported and which
  hardware did you test with?**
  No specific hardware is required

---

**Checklist**

* [ ] Documentation for the feature
* [ ] Tests for the feature
* [ ] The arguments and descriptions in `doc/configuration.rst` have been updated
* [ ] Add a section in `doc/usage.rst` to explain how to use the feature
* [ ] Add a section in `doc/development.rst` to explain how developers can use the feature
* [x] PR has been tested locally
* [ ] Man pages have been regenerated

### DIFF
--- a/labgrid/util/helper.py
+++ b/labgrid/util/helper.py
@@ -41,8 +41,14 @@ class ProcessWrapper:
     loglevel = logging.INFO
 
     @step(args=['command'], result=True, tag='process')
-    def check_output(self, command, *, print_on_silent_log=False, input=None, stdin=None, timeout=None): # pylint: disable=redefined-builtin
+    def check_output(
+        self, command, *, print_on_silent_log=False, input=None, stdin=None, stderr=subprocess.STDOUT, timeout=None
+    ):  # pylint: disable=redefined-builtin
         """Run a command and supply the output to callback functions"""
+
+        if stderr not in (None, subprocess.STDOUT, subprocess.DEVNULL):
+            raise NotImplementedError(f"{stderr=} not supported")
+
         logger = logging.getLogger("Process")
         res = []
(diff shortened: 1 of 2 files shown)


Emantor (dismissed):

Bastian-Krause (changes requested):
Passing `stderr=subprocess.DEVNULL` or `stderr=None` is fine. But `stderr=subprocess.PIPE` would never drain the pipe nor close it. That's a problem. Maybe raise `NotImplementedError` if stderr is something else than `None` or `subprocess.DEVNULL`?

Bastian-Krause:
Force-pushed: I've added the requested changes myself:
- use default similar to subprocess: subprocess.STDOUT means merging stdout and stderr, None means no capturing
-  raise on unsupported stderr values
- adjusted the commit message accordingly

Attention: Since you're not using a real name, I've created #1940 to address that. Do not merge until #1940 is merged.

jluebbe (dismissed):

Bastian-Krause:
Tests fixed, let's merge this.

PR #1788, opened 2025-12-18
remote/client: cleanup InteractiveCommandError

sgross-emlix:
When handling InteractiveCommandError there is an implicit assumption that it has a field `exitcode`.
This assumption holds since in all places it is raised there is the same repeated code block.

Make this assumption more explicit by providing a proper constructor to InteractiveCommandError class.


sgross-emlix:
Moved out of #1769 as suggested by @jluebbe

PR #1798, opened 2026-01-14
doc/man/pytest: fix inline interpreted text without a role

Bastian-Krause:
**Description**
sphinx became stricter when handling interpreted text without a role, e.g. "`foo`" since [1], see [2]. This results in errors such as:

```
  sphinx.errors.SphinxError: cannot determine default role!
```

In the past, heuristics were used to determine how to format the text (inline code, italic, ..).

Fix another occurrence that should have been marked as inline code instead.

[1] https://github.com/sphinx-doc/sphinx/commit/114093cff01562a6a50f8c88b59c7b2fed52a39a
[2] https://github.com/sphinx-doc/sphinx/issues/13904

**Checklist**
- [ ] Tests for the feature 

- [ ] PR has been tested

See also #1755

### DIFF
--- a/doc/man/pytest.rst
+++ b/doc/man/pytest.rst
@@ -15,7 +15,7 @@ SYNOPSIS
 DESCRIPTION
 -----------
 Labgrid ships a pytest plugin to integrate with the pytest infrastructure. It is
-activated if the `--lg-env` parameter is supplied to the pytest command.
+activated if the ``--lg-env`` parameter is supplied to the pytest command.
 
 The labgrid plugin parses the supplied configuration yaml file as described in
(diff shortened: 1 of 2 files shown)


MarekSzczypinski:
For the `default_role` fix we could just add such entry to the `conf.py` file. The problem is just which setting to use: `any`, `code`, something else? 

Another approach is just to pin the Sphinx version ("Sphinx>=2.0.0,<9.0.0") which also pulls in "docutils<0.22" and thus prevents forcing the `default_role` and rendering man pages differently. This way we have working builds in versions 3.9 through to 3.10. After dependency on python 3.9 and 3.10 will be dropped from project we can "unpin" the Sphinx version.

Emantor (approved):

PR #1808, opened 2026-01-27
labgrid-raw-interface: Add permissions

JoshuaWatt:
Adds support for the helper.yaml file to specify which subcommands users are allowed to execute using labgrid-raw-interface. Denied commands will fail with an error.

The RawNetworkInterfaceDriver is updated to query if the `ip` command is allowed before attemping to bring the interface up and down on activate/deactivate. This prevents it from erroring out if this behavior is not allowed

**Description**

**Checklist**
- [ ] Tests for the feature 

- [x] PR has been tested

### DIFF
--- a/helpers/labgrid-raw-interface
+++ b/helpers/labgrid-raw-interface
@@ -11,19 +11,27 @@ import argparse
 import os
 import string
 import sys
+import json
 
 import yaml
 
+DENYLIST_FILE = "/etc/labgrid/helpers.yaml"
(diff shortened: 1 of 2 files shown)


Bastian-Krause:
Looks okay to me. Do we really need to check the allowed commands in the driver? I think I would drop that and just let it fail if the permissions do not match.

Bastian-Krause:
@JoshuaWatt Would that be okay for you?

PR #1812, opened 2026-02-02
fix: quoting of QEMU driver boot args

flxzt:
**Description**
This fixes the issue that boot arguments were passed unquoted to QEMU.

- [x] PR has been tested

### DIFF
--- a/labgrid/driver/qemudriver.py
+++ b/labgrid/driver/qemudriver.py
@@ -225,7 +225,7 @@ class QEMUDriver(ConsoleExpectMixin, Driver, PowerProtocol, ConsoleProtocol):
             boot_args.append(self.boot_args)
         if self.kernel is not None and boot_args:
             cmd.append("-append")
-            cmd.append(" ".join(boot_args))
+            cmd.append("\'" + " ".join(boot_args) + "\'")
 
         return cmd
 
--- a/labgrid/driver/qemudriver.py
+++ b/labgrid/driver/qemudriver.py
@@ -225,7 +225,7 @@ class QEMUDriver(ConsoleExpectMixin, Driver, PowerProtocol, ConsoleProtocol):
             boot_args.append(self.boot_args)
         if self.kernel is not None and boot_args:
             cmd.append("-append")
-            cmd.append("\'" + " ".join(boot_args) + "\'")
+            cmd.append(shlex.quote(" ".join(boot_args)))
 
         return cmd
 


labgrid/driver/qemudriver.py
```diff
@@ -225,7 +225,7 @@ def get_qemu_base_args(self):
             boot_args.append(self.boot_args)
         if self.kernel is not None and boot_args:
             cmd.append("-append")
-            cmd.append(" ".join(boot_args))
+            cmd.append("\'" + " ".join(boot_args) + "\'")
```
Emantor: This should probably use `shlex.quote()`

Emantor (changes requested):

labgrid/driver/qemudriver.py
flxzt: fixed!

Emantor (approved):

Bastian-Krause:
@flxzt What problem does this PR fix at your end? Are you using `get_qemu_base_args()` externally? If so, how?

The QEMUDriver passes the argument list to `subprocess.Popen()` (in `on()`), there is no shell involved. Meaning, this PR introduces a regression:
```python
subprocess.Popen([
    '/usr/bin/qemu-system-x86_64',
    '-kernel', linuximage',
    '-drive', 'if=virtio,format=raw,file=root.ext2',
    '-accel',
    'kvm',
    '-smp', '2',
    '-rtc', 'base=2000-01-01',
    '-machine', 'pc',
    '-cpu', 'host',
    '-m', '1G',
    '-nographic',
    '-append', "'root=/dev/vda rootwait console=ttyS0,115200 rw console=ttyS0,115200 loglevel=5 systemd.log_level=warning systemd.show_status=auto'",
    '-S',
    '-qmp', 'stdio',
    '-chardev', 'socket,id=serialsocket,path=/tmp/labgrid-qemu-tmp-qbqgve19/serialrw',
    '-serial', 'chardev:serialsocket'
], stdin=subprocess.PIPE, stdout=subprocess.PIPE)
```

The unexpected quoted argument is not processed correctly. This results in Kernel panics due to the missing parameters.

Emantor:
@Bastian-Krause Let us revert this change for now on master, we can still discuss with an unbroken master. See #1876 and merge it if you agree.

flxzt:
I have to check what was the original issue for which this fix provided a solution - I'll get back to you.

PR #1818, opened 2026-02-13
gpio-manager D-Bus service support

dlech:
**Description**

Add support for controlling GPIOs via the [gpio-manager](https://libgpiod.readthedocs.io/en/stable/gpio-manager.html) D-Bus service.

There have been many attempts to use the gpio character device interface in labgrid since the sysfs interface is deprecated and is turned off in some kernel builds. However, these were rejected because the state of the gpio can change when the character device closes. 

The `gpio-manager` D-Bus daemon is supposed to be the "new" way to address this. It has been around for a few years, but isn't widely available yet. I suspect that will change though. And it is fairly simple to install from source if needed ([some notes](https://gist.github.com/dlech/8455950e7fd0dbcf3bba1a3c74b35504)).

I have tested this locally using the examples for both local and remote usage (slightly customized to my gpiochip path and pin number). I haven't implemented the lookup by name part yet (I'll have to do that on a Rapsberry Pi, which will be slower). I'm just putting this out there to see what people think first before I polish it up the rest of the way.

Similar to the way the sysfs gpio implementation will export gpios that are not already exported, this will request gpios that are not already requested. Since there isn't a process managing this, it means the line will stay requested until the user manually releases the request, e.g using the `gpiocli` command line too.

**Checklist**
- [ ] Tests for the feature (There are a couple of tests already, but they don't cover everything. It is hard to have tests for something that requires real hardware. I opened https://github.com/labgrid-project/labgrid/issues/1820 to discuss a solution, but I consider that out of scope for this PR.)

- [x] PR has been tested

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -608,6 +608,34 @@ NetworkHIDRelay
 A :any:`NetworkHIDRelay` describes an `HIDRelay`_ resource available on a
 remote computer.
 
+ManagedGPIO
++++++++++++
+
+A :any:`ManagedGPIO` resource describes a GPIO line that is managed by the
+`gpio-manager <https://libgpiod.readthedocs.io/en/stable/gpio-manager.html>`__
+daemon.
+
+.. note::
+   The ``gpio-manager`` daemon needs to be running on the same system as the
+   gpiochip and configured to manage the specified GPIO line. See the
+   `gpiocli request <https://libgpiod.readthedocs.io/en/stable/gpiocli-request.html>`__
+   documentation for details.
+
+.. code-block:: yaml
+
+   ManagedGPIO:
+     chip: /dev/gpiochip0
+     pin: 0
+
+Arguments:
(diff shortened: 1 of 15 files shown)


dlech:
> Which benefits do you see compared to adding support for the [new per-chip sysfs GPIO interface](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=2028f854b3f5b3816cd5d5dd83057a873eddc4d6) (which is not deprecated

I was not aware of this. So perhaps there is not much benefit.

It could still be interesting to people who are already using `gpio-manager`/`gpiocli` though.

Although, I suppose one could easily create an HTTP server that talks to `gpio-manager` and use `HttpDigitalOutput` instead, which would be more efficient than the agent way of doing things.

Emantor:
In our opinion, the existing sysfsgpio agent should be extended to support the new per gpiochip paths instead of relying on the deprecated interface. This should be significantly less code and complexity than a new agent or HTTP server interface.

PR #1835, opened 2026-03-04
driver/pyvisa: extend Test and Measurement equipment support

Sabolik:
**Description**

Main purpose:

- Extend generic support for T&M instruments (SCPI based test & measurement equipment) using already existing pyvisa driver with appropriate backend (e.g. pyvisa-py)
- Add pyvisa remote access by introducing an agent wrapper module.
- Extend pyvisa driver with the option to select specific resource type (e.g. INSTR/SOCKET).
- Add remote client support.
- Introduce the tminstrument driver as high-level wrapper for low-level drivers, (pyvisa, usbtmc), with PowerProtocol support. Potentially replacing the standalone usbtmc driver.
- Add backend support for various laboratory instruments, e.g. convenient command wrapper for SCPI commands.

The implementation was tested only by using it with devices on our lab environment, both locally and remotely for following interfaces:

- TCPIP INSTR
- TCPIP SOCKET
- ASRL INSTR
- USB INSTR

**Checklist**
- [ ] Tests for the feature 

- [ ] PR has been tested

### DIFF
--- a/labgrid/driver/pyvisadriver.py
+++ b/labgrid/driver/pyvisadriver.py
@@ -21,7 +21,7 @@ class PyVISADriver(Driver):
 
     def on_activate(self):
         url = "" if self.pyvisa_resource.url == "" else f"::{self.pyvisa_resource.url}"
-        self.device_identifier = f"{self.pyvisa_resource.type}{url}::INSTR"
+        self.device_identifier = f"{self.pyvisa_resource.type}{url}::{self.pyvisa_resource.resource}"
         if isinstance(self.pyvisa_resource, NetworkPyVISADevice):
             host = self.pyvisa_resource.host
         else:
--- a/labgrid/driver/pyvisadriver.py
+++ b/labgrid/driver/pyvisadriver.py
@@ -63,6 +63,28 @@ class PyVISADriver(Driver):
         return self.proxy.query(self.device_identifier, self.pyvisa_resource.backend, cmd, timeout).rstrip()
 
     @Driver.check_active
+    def query_iterable(self, cmd, timeout=2000, **kwargs):
+        """
+        Sent the specified command to the instrument and read the response.
+
+        Args:
+            cmd (str): command to be executed
+            timeout (int): optional, I/O operation timeout in ms, default is 2000ms
+
+        Returns:
+            List[float]: response from the instrument, retunring list of float values (readings).
+        """
+        return self.proxy.query(
+            self.device_identifier,
+            self.pyvisa_resource.backend,
+            cmd,
+            timeout,
+            iterable=True,
+            is_ascii=True,
+            **kwargs
+        )
+
+    @Driver.check_active
     def identify(self):
         """
         Sent the identify command to the instrument and read the response.
--- a/labgrid/driver/pyvisadriver.py
+++ b/labgrid/driver/pyvisadriver.py
@@ -2,6 +2,8 @@ from importlib import import_module
 import attr
 
 from ..factory import target_factory
+from ..resource.pyvisa import NetworkPyVISADevice
+from ..util.agentwrapper import AgentWrapper
 from .common import Driver
 
 
@@ -13,22 +15,38 @@ class PyVISADriver(Driver):
     Args:
         bindings (dict): driver to use with PyVISA
     """
-
-    bindings = {"pyvisa_resource": "PyVISADevice"}
+    bindings = {"pyvisa_resource": {"PyVISADevice", "NetworkPyVISADevice"}}
 
     def __attrs_post_init__(self):
         super().__attrs_post_init__()
-        _py_pyvisa_module = import_module("pyvisa")
-        self._pyvisa_resource_manager = _py_pyvisa_module.ResourceManager(self.pyvisa_resource.backend)
-        self.pyvisa_device = None
+        self.wrapper = None
 
     def on_activate(self):
-        device_identifier = f"{self.pyvisa_resource.type}::{self.pyvisa_resource.url}::INSTR"
-        self.pyvisa_device = self._pyvisa_resource_manager.open_resource(device_identifier)
+        url = "" if self.pyvisa_resource.url == "" else f"::{self.pyvisa_resource.url}"
+        self.device_identifier = f"{self.pyvisa_resource.type}{url}::INSTR"
+        if isinstance(self.pyvisa_resource, NetworkPyVISADevice):
+            host = self.pyvisa_resource.host
+        else:
+            host = None
+        self.wrapper = AgentWrapper(host)
+        self.proxy = self.wrapper.load("visa_instrument")
 
     def on_deactivate(self):
-        self.pyvisa_device = None
+        self.wrapper.close()
+        self.wrapper = None
 
     @Driver.check_active
     def get_session(self):
-        return self.pyvisa_device
+        raise NotImplementedError('Deprecated, use command or query instead')
+
+    @Driver.check_active
+    def command(self, cmd):
+        self.proxy.write(self.device_identifier, self.pyvisa_resource.backend, cmd)
+
+    @Driver.check_active
+    def query(self, cmd):
+        return self.proxy.query(self.device_identifier, self.pyvisa_resource.backend, cmd)
+
+    @Driver.check_active
+    def identify(self):
+        return self.query("*IDN?")
--- a/labgrid/driver/pyvisadriver.py
+++ b/labgrid/driver/pyvisadriver.py
@@ -1,11 +1,9 @@
-from importlib import import_module
 import attr
 
 from ..factory import target_factory
 from ..resource.pyvisa import NetworkPyVISADevice
 from ..util.agentwrapper import AgentWrapper
 from .common import Driver
-
 
 @target_factory.reg_driver
 @attr.s(eq=False)
@@ -40,13 +38,36 @@ class PyVISADriver(Driver):
         raise NotImplementedError('Deprecated, use command or query instead')
 
     @Driver.check_active
-    def command(self, cmd):
-        self.proxy.write(self.device_identifier, self.pyvisa_resource.backend, cmd)
+    def command(self, cmd, timeout=2000):
+        """
+        Sent the specified command to the instrument.
+
+        Args:
+            cmd (str): command to be executed
+            timeout (int): optional, I/O operation timeout in ms, default is 2000ms
+        """
+        self.proxy.write(self.device_identifier, self.pyvisa_resource.backend, cmd, timeout)
 
     @Driver.check_active
-    def query(self, cmd):
-        return self.proxy.query(self.device_identifier, self.pyvisa_resource.backend, cmd)
+    def query(self, cmd, timeout=2000):
+        """
+        Sent the specified command to the instrument and read the response.
+
+        Args:
+            cmd (str): command to be executed
+            timeout (int): optional, I/O operation timeout in ms, default is 2000ms
+
+        Returns:
+            str: response from the instrument, with possible termination removed.
+        """
+        return self.proxy.query(self.device_identifier, self.pyvisa_resource.backend, cmd, timeout).rstrip()
 
     @Driver.check_active
     def identify(self):
+        """
+        Sent the identify command to the instrument and read the response.
+
+        Returns:
+            str: returns a comma-separated string with the information about the instrument (manufacturer, model, serial number, firmware version).
+        """
         return self.query("*IDN?")
--- a/labgrid/resource/pyvisa.py
+++ b/labgrid/resource/pyvisa.py
@@ -1,7 +1,7 @@
 import attr
 
 from ..factory import target_factory
-from .common import Resource
+from .common import Resource, NetworkResource
 
 
 @target_factory.reg_resource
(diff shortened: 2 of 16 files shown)


labgrid/driver/pyvisadriver.py
```diff
@@ -40,12 +40,12 @@ def get_session(self):
...
-    def command(self, cmd):
-        self.proxy.write(self.device_identifier, self.pyvisa_resource.backend, cmd)
+    def command(self, cmd, timeout=TIMEOUT_DEFAULT):
+        self.proxy.write(self.device_identifier, self.pyvisa_resource.backend, cmd, timeout)
 
     @Driver.check_active
-    def query(self, cmd):
-        return self.proxy.query(self.device_identifier, self.pyvisa_resource.backend, cmd)
+    def query(self, cmd, timeout=TIMEOUT_DEFAULT):
+        return self.proxy.query(self.device_identifier, self.pyvisa_resource.backend, cmd, timeout).rstrip()
```
Emantor, line 63: This `rstrip()` is added here, can you add a description to the commit why it was added?

labgrid/driver/pyvisadriver.py
```diff
@@ -1,11 +1,11 @@
-from importlib import import_module
 import attr
 
 from ..factory import target_factory
 from ..resource.pyvisa import NetworkPyVISADevice
 from ..util.agentwrapper import AgentWrapper
 from .common import Driver
 
+TIMEOUT_DEFAULT = 2000  # Default timeout for visa operations in ms
```
Emantor: I think I prefer having the default in the keyword arguments i.e.  `def command(self, cmd, timeout=2000)` and we also want a docstring which documents this to be in miliseconds.

labgrid/resource/pyvisa.py
```diff
@@ -11,10 +11,24 @@ class PyVISADevice(Resource):
 
     Args:
         type (str): device resource type following the pyVISA resource syntax, e.g. ASRL, TCPIP...
-        url (str): device identifier on selected resource, e.g. <ip> for TCPIP resource
+        url (str, default=''): optional device identifier on selected resource, e.g. <ip> for TCPIP resource
```
Emantor, line 14: Why can the default be changed to an empty string here? Is this not mandatory?

labgrid/resource/pyvisa.py
```diff
@@ -12,11 +12,13 @@ class PyVISADevice(Resource):
     Args:
         type (str): device resource type following the pyVISA resource syntax, e.g. ASRL, TCPIP...
         url (str, default=''): optional device identifier on selected resource, e.g. <ip> for TCPIP resource
+        resource (str, default='INSTR'): resource type, e.g. INSTR, SOCKET...
         backend (str, default=''): Visa library backend, e.g. '@sim' for pyvisa-sim backend
     """
 
     type = attr.ib(validator=attr.validators.instance_of(str))
     url = attr.ib(default="", validator=attr.validators.instance_of(str))
+    resource = attr.ib(default="INSTR", validator=attr.validators.instance_of(str))
```
Emantor: If there is a finite amount of resource type definitions we might want to extend the validator, but this can be added later.

Emantor (changes requested):
`remote: client: add remote pyvisa access` needs a commit message explaining why the change is done.

I wonder if the current `backend()` calling mechanic can't be simplified to be more explicit instead of inspecting the imported module backend and calling functions on it again. It looks like we are cramming a lot of calls through the `backend()` function, while we have the explicit module with function definitions for each instrument. Why not call the functions on the module directly?

labgrid/driver/pyvisadriver.py
Sabolik: The termination might differ among instruments. Another option might be to properly set the `read_termination `parameter of the resource, but I would rather keep `rstrip` for simplicity.

labgrid/driver/pyvisadriver.py
Sabolik: Agree. Thanks

labgrid/resource/pyvisa.py
Sabolik: No, it is optional. Some interfaces are fully specified by the `type` parameter, e.g. serial devices like `ASRL/dev/ttyACM0`.

labgrid/resource/pyvisa.py
Sabolik: It seems to be fixed based on the [VISA documantation](https://www.ni.com/docs/en-US/bundle/ni-visa/page/visa-resource-types.html?srsltid=AfmBOoppHIPT7WYekdZuaDpQIlThk6pFuw16zNUURTWOcXxi-rNO-xyq), but honestly I am not familiar with all of them, so I am not sure what are the consequences. But the most common are `INSTR` and `SOCKET` and `RAW`, might be good idea to create a validator for them.

labgrid/driver/usbtmcdriver.py
```diff
@@ -49,13 +49,13 @@ def on_deactivate(self):
...
         cmd = b2s(cmd.encode("ASCII") + b"\n")
-        self.wrapper.usbtmc(self.index, cmd, read=False)
+        self.wrapper.usbtmc(self.dev_index, cmd, read=False)
 
     @Driver.check_active
     def query(self, cmd, binary=False, raw=False):
         assert isinstance(cmd, str)
         cmd = b2s(cmd.encode("ASCII") + b"\n")
-        res = s2b(self.wrapper.usbtmc(self.index, cmd, read=True))
+        res = s2b(self.wrapper.usbtmc(self.dev_index, cmd, read=True))
```
gastmaier: is any change in this file really needed?

labgrid/util/agents/visa_instrument.py
```diff
@@ -0,0 +1,53 @@
...
+            _py_pyvisa_module = import_module('pyvisa')
+        except ModuleNotFoundError as e:
+            raise ModuleNotFoundError("pyvisa module not found, please install it") from e
+        else:
+            _pyvisa_resource_manager = _py_pyvisa_module.ResourceManager(backend)
+            if _pyvisa_resource_manager is None:
+                raise ValueError("pyVISA backend not found")
+            self._pyvisa_device = _pyvisa_resource_manager.open_resource(device_identifier)
+            if self._pyvisa_device is None:
+                raise ValueError("pyVISA device not found")
```
gastmaier: If any exception but ModuleNotFoundError, execution stops.
`else` only occurs if no exception, so the nested else seems unnecessary and gives the impression of another behavior.
If you are not suppressing ModuleNotFoundError anyway, but providing a more useful user message, maybe use `importlib`
Still, `import_module('pyvisa')` will raise a very similar message anyway, so the first two files, in this pattern, also look unnessary

```suggestion
        if not importlib.util.find_spec("pyvisa"):
            raise ModuleNotFoundError("pyvisa module not found, please install it")
		
        _py_pyvisa_module = import_module('pyvisa')
        _pyvisa_resource_manager = _py_pyvisa_module.ResourceManager(backend)
        if _pyvisa_resource_manager is None:
            raise ValueError("pyVISA backend not found")
        self._pyvisa_device = _pyvisa_resource_manager.open_resource(device_identifier)
        if self._pyvisa_device is None:
            raise ValueError("pyVISA device not found
```
I would say only catch ModuleNotFoundError/use importlib.util.find_spec if the dependency is truly optional and the execution continues even without the lib (e.g., has a fallback).

labgrid/util/agents/visa_instrument.py
```diff
@@ -0,0 +1,72 @@
...
+
+- Command, Query, Identify
+"""
+
+from importlib import import_module
+
+
+class VISAInstrument:
+    def __init__(self, device_identifier, backend, timeout):
+        try:
```
gastmaier: Same as last comment

labgrid/util/agents/visa_instrument.py
Sabolik: Thanks for proposal, adopted with some minor fixing.

labgrid/driver/usbtmcdriver.py
Sabolik: It seems to be needed. The reason is that the same attribute is used by `power` resource as power port identification. Power resource is directly supported by `tminstrument` driver while conflicting with low-level `usbtmc` driver (usbtmc device identification)

Sabolik: actually not needed. There was conflict between power driver implementation using USBTMC resource used locally in past in our lab, but this is not part of this PR.

Sabolik:
> > @threexc Did you find some time to test this?
> 
> I've been looking at it. I was able to do a basic test on my side with my SDG1032X and confirm it can communicate:
> 
> ```
> (venv) tgamblin@outpost:~/workspace/git/labgrid-power-up$ labgrid-client -c workbench/wb-client.yaml inst identify --name 'tm-inst-sdg1032x'
> Selected role main and place workbench from configuration file
> /usr/lib/python3/dist-packages/pyvisa_py/protocols/usbtmc.py:116: UserWarning: Unexpected MsgID format. Consider updating the device's firmware. See https://github.com/pyvisa/pyvisa-py/issues/20Expected message id was 2, got 46.
>   warnings.warn(
> Siglent Technologies,SDG1032X,SDG1XDFQ6R5627,1.01.01.33R
> ```
> 
> This is with the following exporter config for the device:
> 
> ```
>   wb-generator-1:
>     cls: 'PyVISADevice'
>     type: 'USB0'
>     url: '0xF4EC::0x1103::SDG1XDFQ6R5627::0'
>     backend: '@py'
> ```
> 
> and a client config like:
> 
> ```
>       - PyVISADriver:
>           name: 'sdg1032x-driver'
>           bindings:
>             pyvisa_resource: 'wb-generator-1'
>       - TMInstrument:
>           name: 'tm-inst-sdg1032x'
>           bindings:
>             inst: 'sdg1032x-driver'
> ```
> 
> Seems reasonable enough to me. My only major issue with this PR is that there are examples in the commits but not really any updates to the docs? It would be good to have decent examples of the exporter and client configs, along with example commands to run (the ones in some of the commit messages that I've been pointed at are probably enough).
> 
> @Sabolik, can you add another commit to reflect that?

- doc/configuration.rst: has been updated
- examples/tminstrument: added example with two instruments, tested
- labgrid-client usage example, I am unfortunatelly not sure where to put such example. There is generic `labgrid-client CLI` usage well docummented on `Manual Pages`.

Bastian-Krause:
Please open a separate PR for the USBVideoDriver change, it has nothing to do with the pyvisa changes.

Sabolik:
> Please open a separate PR for the USBVideoDriver change, it has nothing to do with the pyvisa changes.

accidentally pushed by me, but seems to be already [merged](https://github.com/labgrid-project/labgrid/pull/1827). Related commit removed from this PR.

jluebbe:
> ```yaml
> targets:
>   main:
>     resources:
>     - PyVISADevice:
>         name: dmm-rigol
>         type: "USB0"
>         url: "6833::3220::DM3O224500792::0"
>         backend: "@py"
>     drivers:
>     - PyVISADriver:
>         name: driver-visa-rigol-dmm
>         bindings: { pyvisa_resource : 'dmm-rigol'}
>     - TMInstrument: 
>           bindings: { inst : 'driver-visa-rigol-dmm' }
>           name: 'tm-inst-visa-rigol-dmm'
> ```

Why is PyVISADriver separate from TMInstrument? Do you expect cases where you have a single PyVISADriver instance but multiple TMInstrument drivers which are used alternatively, like SerialDriver+BareboxDriver/ShellDriver (where you switch the upper driver based on the software state of the target)?

There should be a good justification for the additional configuration complexity. :)

Sabolik:
> > ```yaml
> > targets:
> >   main:
> >     resources:
> >     - PyVISADevice:
> >         name: dmm-rigol
> >         type: "USB0"
> >         url: "6833::3220::DM3O224500792::0"
> >         backend: "@py"
> >     drivers:
> >     - PyVISADriver:
> >         name: driver-visa-rigol-dmm
> >         bindings: { pyvisa_resource : 'dmm-rigol'}
> >     - TMInstrument: 
> >           bindings: { inst : 'driver-visa-rigol-dmm' }
> >           name: 'tm-inst-visa-rigol-dmm'
> > ```
> 
> Why is PyVISADriver separate from TMInstrument? Do you expect cases where you have a single PyVISADriver instance but multiple TMInstrument drivers which are used alternatively, like SerialDriver+BareboxDriver/ShellDriver (where you switch the upper driver based on the software state of the target)?
> 
> There should be a good justification for the additional configuration complexity. :)

The Idea is slightly different. `TMInstrument` is high-level driver, with more low-level drivers supported and sharing the same API. Currently, [PyVISA and USBTMC driver](https://github.com/labgrid-project/labgrid/pull/1835/changes#diff-5e3dd36d509111b80e62344e1e91e35be10f1a23bb5b7184aa29305e513a4b89R27). On the other hand, `USBTMC` should be fully supported by `PyVISA`, so `UCBTMC` driver might be potentially deprecated.

Sabolik:
> > > ```yaml
> > > targets:
> > >   main:
> > >     resources:
> > >     - PyVISADevice:
> > >         name: dmm-rigol
> > >         type: "USB0"
> > >         url: "6833::3220::DM3O224500792::0"
> > >         backend: "@py"
> > >     drivers:
> > >     - PyVISADriver:
> > >         name: driver-visa-rigol-dmm
> > >         bindings: { pyvisa_resource : 'dmm-rigol'}
> > >     - TMInstrument: 
> > >           bindings: { inst : 'driver-visa-rigol-dmm' }
> > >           name: 'tm-inst-visa-rigol-dmm'
> > > ```
> > 
> > 
> > Why is PyVISADriver separate from TMInstrument? Do you expect cases where you have a single PyVISADriver instance but multiple TMInstrument drivers which are used alternatively, like SerialDriver+BareboxDriver/ShellDriver (where you switch the upper driver based on the software state of the target)?
> > There should be a good justification for the additional configuration complexity. :)
> 
> The Idea is slightly different. `TMInstrument` is high-level driver, with more low-level drivers supported and sharing the same API. Currently, [PyVISA and USBTMC driver](https://github.com/labgrid-project/labgrid/pull/1835/changes#diff-5e3dd36d509111b80e62344e1e91e35be10f1a23bb5b7184aa29305e513a4b89R27). On the other hand, `USBTMC` should be fully supported by `PyVISA`, so `UCBTMC` driver might be potentially deprecated.

Also, `TMInstrument` is meant to support some convenient functionality, like [Instruments set of commands](https://github.com/vzlu/labgrid_msa/tree/vzlu/pyvisa/labgrid/driver/tminstr). While `PyVISA` is kept as pure low-level read/write driver.

PR #1840, opened 2026-03-24
dockerfiles: Add statsd docker image

nick-potenski:
**Description**
Adds a new docker image for a coordinator-statsd service, which can be used to easily run a statsd reporter for the coordinator leveraging the `contrib/coordinator-statsd.py` script.

**Checklist**
- [x] PR has been tested

Testing included running the example in the `dockerfiles/staging` directory as well as running a locally built image in our production environment and verifying that statsd reporting still functioned as expected.

### DIFF
--- a/dockerfiles/Dockerfile
+++ b/dockerfiles/Dockerfile
@@ -57,3 +57,13 @@ RUN apt update -q=2 && \
 VOLUME /opt/conf
 
 CMD ["/entrypoint.sh"]
+
+#
+# Coordinator statsd
+#
+FROM labgrid-base AS labgrid-coordinator-statsd
+ARG VERSION
+
+RUN pip3 install "statsd>=4.0.1"
+
+ENTRYPOINT ["/opt/labgrid/contrib/coordinator-statsd.py"]
--- a/dockerfiles/README.rst
+++ b/dockerfiles/README.rst
@@ -7,6 +7,8 @@ for the 3 different components of a Labgrid distributed infrastructure.
 - **labgrid-coordinator**
   An image with the Labgrid coordinator.
   a Labgrid coordinator instance.
+- **labgrid-coordinator-statsd**
+  An image with a statsd reporter for the Labgrid coordinator.
 - **labgrid-client**
   An image with the Labgrid client tools and pytest integration.
 - **labgrid-exporter**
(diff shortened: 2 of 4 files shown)


dockerfiles/Dockerfile
```diff
@@ -57,3 +57,13 @@ RUN apt update -q=2 && \
...
 CMD ["/entrypoint.sh"]
+
+#
+# Coordinator statsd
+#
+FROM labgrid-base AS labgrid-coordinator-statsd
+ARG VERSION
+
+RUN pip3 install statsd
+
```
krevsbech, line 68: Suggestion: could the `statsd` be a dependency in the pyproject.toml somehow?
As there is no version boundary here, it can be a bit fragile in statsd makes breaking changes?

dockerfiles/README.rst
```diff
@@ -79,6 +80,19 @@ so you can restart the service without losing state.
...
+
+The labgrid-coordinator-statsd image can be used to run a statsd reporter for the coordinator.
+
+Addresses of the coordinator and statsd server and other options can be configured using
+parameters as passed to the ``contrib/coordinator-statsd.py`` script.
+
+.. code-block:: bash
+
+   $ docker run docker.io/labgrid/coordinator-statsd
+
```
krevsbech, line 95: Include the `Addresses of the coordinator and statsd servers` parameters in the example?

dockerfiles/Dockerfile
jluebbe: labgrid doesn't need or use statsd, so it shouldn't be in labgrid's pyproject.toml.

krevsbech: That's fair.
I would then suggest to just use a version constraint here, perhaps just a lower bound.

dockerfiles/README.rst
nick-potenski: Done.

dockerfiles/Dockerfile
nick-potenski: The statsd Python package hasn't changed in a while. 😄 I added the lower bound, as suggested.

krevsbech (approved):

Emantor (approved):

jluebbe (approved):

nick-potenski:
As a note to future users, we have our statsd server connected to a Grafana dashboard to display information about available places and reservations, e.g.
image
image

We pass `--tag=board` to the docker container invocation (and hence on to `coordinator-statsd.py`), so that we can group results based on each place's board tag.  Then we can monitor queues for each like shown below.  Each color represents a unique board type (names are redacted).
image

PR #1843, opened 2026-03-26
labgrid/qemudriver: add USB network device hotplug support

krevsbech:
**Description**
Use the QMPMonitor to add and remove a USB network device from a running QEMU instance.

Assumes the QEMU instance is already configured with a USB bus, e.g. `-device qemu-xhci,id=xhci`

**Checklist**
- [x] Documentation for the feature (pydocs)
- [x] PR has been tested


Emantor (approved):

Bastian-Krause:
@krevsbech Could you give a little more motivation for this? Why do you need to hotplug the device? Seems rather exotic and it's already possible to call the few `QEMUDriver.monitor_command()` without adding them to the driver.

PR #1848, opened 2026-04-05
add secomp power backend

Rubusch:
Add driver for the SECOMP PDU and similar generic devices.

---
Please first check if you already have support! If so, pease leave me a note! ;)

This is a driver for a pretty generic device. It's sold under various names/brands (or they're using the same/similar hardware). So, chances are you already have support for it, I did not find. I'm controlling three of such devices, so it's "tested" here in different setups. AFAIR we bought this one here.
https://www.galaxus.ch/de/s1/product/value-ip-steckdosenleiste-usv-zubehoer-13359261
I've heard that the hardware goes under the name/brand SECOMP.

If labgrid really still lacks support for such device, feel free to merge if you find it useful or let me know what to improve.

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -229,6 +229,11 @@ Currently available are:
   <https://github.com/labgrid-project/labgrid/blob/master/labgrid/driver/power/rest.py>`__
   for details.
 
+``secomp``
+  Controls *Secomp* PDUs via HTTP. This is a 19" Rackmount PDU sold and
+  rebranded under various names e.g. VALUE IP, Argus, Intellinet, or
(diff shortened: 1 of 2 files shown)


ozan956 (commented):
Hi, thank you for the PR. Few things to think about:

I believe this driver can be simplified by matching the tinycontrol.py's style (same idea: GET request to set, GET + XML parse to read status). 

That tinycontrol driver uses urljoin and lets requests handle embedded URL credentials natively. So, no manual urlparse/netloc.split('@') needed. Adopting that pattern would drop custom URL surgery.

You can also add "assert port is None", as in tinycontrol.

PR #1664, opened 2025-06-13
NetworkPowerPort: add support for Shelly Gen2+

aparcar:
**Description**

The API changed and no longer excepts POST for the legacy API. This commit adds support for the "new" RPC API.

Ref: https://shelly-api-docs.shelly.cloud/gen2/General/RPCProtocol

Tested with a Shelly Plug S MTR Gen3

**Checklist**
- [x] Documentation for the feature
  - Added in docstring
- [x] Tests for the feature 
  - Tested with real hardware

- [ ] The arguments and description in doc/configuration.rst have been updated

- [ ] Add a section on how to use the feature to doc/usage.rst

- [ ] Add a section on how to use the feature to doc/development.rst

- [x] PR has been tested

- [ ] Man pages have been regenerated

PR #1128, opened 2023-04-06
NetworkPowerPort: add support for Shelly plugs/switches using the Gen 1 Device API

rpoisel:
**Description**

As we have some Shelly switches built into our demo/test wall, we decided to add support for these devices.

In order to achieve this goal, the generic `NetworkPowerDriver` has been provided with an additional backend called `shelly_gen1`.

**Checklist**
- [x] Documentation for the feature
- [x] Tests for the feature 

- [x] The arguments and description in doc/configuration.rst have been updated

- [ ] Add a section on how to use the feature to doc/usage.rst

- [ ] Add a section on how to use the feature to doc/development.rst

- [x] PR has been tested: both manually as well as with unit-tests

- [ ] Man pages have been regenerated

PR #1852, opened 2026-04-14
driver: shelly_gen1: fix for Shelly Plug US Gen4

jiez:
POST is not supported by Shelly Plug US Gen4. Use GET instead.

I tried shelly_gen1 driver on a Shelly Plug US Gen4 and found that this Shelly plug does not support POST. Replacing POST with GET seems working. But I don't have other Shelly plugs to test for.

- [x] PR has been tested

### DIFF
--- a/labgrid/driver/power/shelly_gen1.py
+++ b/labgrid/driver/power/shelly_gen1.py
@@ -6,8 +6,7 @@ NetworkPowerPort:
     index: 0
 
 Will do a GET request to http://192.168.0.42/relay/0 to get the current
-relay state, and a POST request to http://192.168.0.42/relay/0 with request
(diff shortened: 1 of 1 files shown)


gastmaier:
Hi @rpoisel (from #1128 Gen1) @aparcar (from #1664 Gen2)! Are you able to test if with GET request still works for Shelly Plug Gen 1 and 2?

I see the file edits Gen 1 only, @jiez have you tried `driver/power/shelly_gen2.py` ?

jiez:
I'm using the labgrid packages from Debian forky. They were created based on the latest release, which has no shelly_gen2 driver yet. I copied shelly_gen2 python file into the installation location and tried it. It has a bug. And there is already a PR for it: https://github.com/labgrid-project/labgrid/pull/1766

PR #1853, opened 2026-04-14
flashscriptdriver: Add optional timeout

mcdanielcolin:
**Description**
Adds an optional timeout for flashscriptdriver to be passed to ProcessWrapper.check_output().  This can be used to prevent processes from hanging and requiring external input to stop the session.  Testing included running flashscriptdriver on our local setup and DUT with various timeouts, including no timeout.

**Checklist**

- [x] PR has been tested

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -3405,6 +3405,7 @@ Arguments:
   - script (str): optional, key in :ref:`images <labgrid-device-config-images>`
     containing the script to execute for writing of the flashable device
   - args (list of str): optional, list of arguments for flash script execution
+  - timeout (float): optional, timeout for script execution in seconds
 
 The FlashScriptDriver allows running arbitrary programs to flash a device.
 Some SoC or devices may require custom, one-off, or proprietary programs to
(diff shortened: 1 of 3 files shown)


same-code-different-day (commented):
👍

nick-potenski (approved):
Looks good to me! 👍

Emantor (approved):

Bastian-Krause (approved):

Bastian-Krause:
@mcdanielcolin After looking at this again: if the timeout expired, the fd is closed (what if the process continues to write to stdout/stderr now?) and we still wait for it to finish. Is this really the desired behavior? Maybe we should kill the process right after we found the timeout to be expired?

mcdanielcolin:
That's a good point, and it is probably best to kill the process if the timeout expires.  I can verify on our cluster and then I can make a new request

mcdanielcolin:
Created a change to add killing the process on timeout here: https://github.com/labgrid-project/labgrid/pull/1970

PR #1854, opened 2026-04-14
exporter: Add USB hub abstraction for simpler device matching

sjg20:
The exporter config requires full @ID_PATH strings for every USB resource, which are long and hard to maintain, particular for labs with multiple hubs.

This PR adds a 'hubs' section to the exporter config that defines USB hubs by name, base path, and a mapping of logical port numbers to USB path suffixes. Resources can then use 'hub' and 'port' in their match dict instead of a raw @ID_PATH string.

For example, instead of:
```
  board1:
    USBSerialPort:
      match:
        '@ID_PATH': 'pci-0000:04:00.0-usb-0:2.2.3'
```
the config can be written as:
```
  hubs:
    a:
      base: 'pci-0000:04:00.0-usb-0:2'
      ports:
        7: '2.3'

  board1:
    USBSerialPort:
      match:
        hub: a
        port: 7
```
The expansion happens at config load time before any resource matching, so it works with any USB resource type.

This feature is used to help diagnose problems - e.g. if a device's UART stops working, which USB port do I need to replug? It saves a lot of confusion mapping a string of numbers to a physical port.

This feature has been manually verified on my lab (about 30 boards) with 7 16-port hubs connected to a single machine. Operation appears to be the same before and after the change. I think modified the lab to actually use this feature and checked that things still work.

**Checklist**
- [x] Tests for the feature 
If you add a feature other drivers/resources can benefit from:
- [x] PR has been tested

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -4402,6 +4402,77 @@ to achieve the same effect:
        match:
          '@ID_PATH': 'pci-0000:05:00.0-usb-3-1.4'
 
+.. _exporter-hub-abstraction:
+
+USB Hub Abstraction
+~~~~~~~~~~~~~~~~~~~
+When a lab uses USB hubs with many ports, the raw ``ID_PATH`` strings in
+match entries become long and hard to maintain.  The exporter supports a
+``hubs`` section that defines USB hubs by name, with a base path and a
+mapping from logical port numbers to USB path suffixes.  Resources can
+then use ``hub`` and ``port`` in their match dict instead of a raw
+``ID_PATH``.
+
+Define hubs at the top level of the exporter configuration:
+
+.. code-block:: yaml
+
+   hubs:
+     a:
+       base: 'pci-0000:00:14.0-usb-0:10'
+       ports:
+         1: '2.1'
+         2: '2.2'
+         3: '2.3'
+         # ...
+     b:
+       base: 'pci-0000:04:00.0-usb-0:2'
+       ports:
+         1: '1.1'
+         2: '1.2'
+         # ...
+
+Each hub has a ``base`` path (the PCI path up to the hub's root port) and
+a ``ports`` mapping from logical port number to the USB path suffix for
+that port.  The port suffixes depend on the hub's internal topology and
+must be determined for each hub model (for example by plugging a device
+into each port and checking ``udevadm info``).
+
+Resources reference a hub port using ``hub`` and ``port`` in their match
+dict.  For resources that need an ancestor match (like serial ports),
+add ``iface`` to specify the USB interface number:
+
+.. code-block:: yaml
+
+   board1:
+     USBSerialPort:
+       match:
+         hub: a
+         port: 3
+         iface: '1.0'
+
+     HIDRelay:
+       index: 4
+       match:
+         hub: a
+         port: 14
+
+When ``iface`` is present, the expansion produces an ``@ID_PATH`` (ancestor
+match) with the interface appended after a colon:
+``@ID_PATH: pci-0000:00:14.0-usb-0:10.2.3:1.0``.
+
+When ``iface`` is absent, the expansion produces a plain ``ID_PATH`` (direct
+match) with no interface suffix:
+``ID_PATH: pci-0000:00:14.0-usb-0:10.1.2``.
+
+Other match keys (such as ``ID_SERIAL_SHORT``) can be used alongside
+``hub``/``port`` and are preserved in the expanded match.  Resources that
+do not use hub/port matching (e.g. those matched by serial number) are
+unaffected.
+
+The ``hubs`` section is removed from the configuration data after
+expansion and does not appear as a resource group.
+
 Templating the Exporter Configuration
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 To reduce the amount of repeated declarations when many similar resources
--- a/doc/usage.rst
+++ b/doc/usage.rst
@@ -839,3 +839,49 @@ like this:
   $ labgrid-client -p example allow sirius/john
 
 To remove the allow it is currently necessary to unlock and lock the place.
+
+Simplifying USB device matching with hubs
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Labs with many USB devices often use multi-port USB hubs, and the raw
+``ID_PATH`` strings needed to match each device can be long and error-prone.
+The exporter supports a ``hubs`` section in the configuration file that maps
+logical hub names and port numbers to USB paths, so that resources can be
+described more concisely.
+
+For example, instead of writing::
+
+   board1:
+     USBSerialPort:
+       match:
+         '@ID_PATH': 'pci-0000:00:14.0-usb-0:10.2.3:1.0'
+
+you can define the hub once and reference it by name::
+
+   hubs:
+     a:
+       base: 'pci-0000:00:14.0-usb-0:10'
+       ports:
+         1: '2.1'
+         2: '2.2'
+         3: '2.3'
+
+   board1:
+     USBSerialPort:
+       match:
+         hub: a
+         port: 3
+         iface: '1.0'
+
+The ``iface`` field controls both the USB interface suffix and the match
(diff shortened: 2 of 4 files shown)


Emantor:
We already provide jinja2 templating in the exporter configuration, so a lot of USB resources can be added at once:
```
{% for idx, sysfs in [
        ('1', 'pci-0000:0d:00.4-usb-0:1.3'),
        ('2', 'pci-0000:0d:00.4-usb-0:1.2'),
        ('3', 'pci-0000:0d:00.4-usb-0:1.1'),
        ('4', 'pci-0000:0d:00.4-usb-0:1.4.4'),
        ('5', 'pci-0000:0d:00.4-usb-0:1.4.3'),
        ('6', 'pci-0000:0d:00.4-usb-0:1.4.2'),
        ('7', 'pci-0000:0d:00.4-usb-0:1.4.1')] %}

{% for if in [ '1.0', '1.1', '1.2', '1.3'] %}
lb-dlink-7-p{{idx}}-{{if}}:
  location: "7 Port Hub auf dem Schreibtisch"
  USBSerialPort:
    match:
      '@ID_PATH': '{{sysfs}}:{{if}}'
{% endfor %}

lb-usb3-7-p{{idx}}:
  location: "7 Port Hub auf dem Schreibtisch"
  IMXUSBLoader:
    match:
        'ID_PATH': '{{sysfs}}'
  AndroidFastboot:
    match:
        'ID_PATH': '{{sysfs}}'
  USBSDMuxDevice:
    match:
        '@ID_PATH': '{{sysfs}}'
  USBVideo:
    match:
        '@ID_PATH': '{{sysfs}}'
{% endfor %}
```

I'd rather properly document this than adding a special case for group parsing.

sjg20:
> We already provide jinja2 templating in the exporter configuration, so a lot of USB resources can be added at once:
> 
> ```
> {% for idx, sysfs in [
>         ('1', 'pci-0000:0d:00.4-usb-0:1.3'),
>         ('2', 'pci-0000:0d:00.4-usb-0:1.2'),
>         ('3', 'pci-0000:0d:00.4-usb-0:1.1'),
>         ('4', 'pci-0000:0d:00.4-usb-0:1.4.4'),
>         ('5', 'pci-0000:0d:00.4-usb-0:1.4.3'),
>         ('6', 'pci-0000:0d:00.4-usb-0:1.4.2'),
>         ('7', 'pci-0000:0d:00.4-usb-0:1.4.1')] %}
> 
> {% for if in [ '1.0', '1.1', '1.2', '1.3'] %}
> lb-dlink-7-p{{idx}}-{{if}}:
>   location: "7 Port Hub auf dem Schreibtisch"
>   USBSerialPort:
>     match:
>       '@ID_PATH': '{{sysfs}}:{{if}}'
> {% endfor %}
> 
> lb-usb3-7-p{{idx}}:
>   location: "7 Port Hub auf dem Schreibtisch"
>   IMXUSBLoader:
>     match:
>         'ID_PATH': '{{sysfs}}'
>   AndroidFastboot:
>     match:
>         'ID_PATH': '{{sysfs}}'
>   USBSDMuxDevice:
>     match:
>         '@ID_PATH': '{{sysfs}}'
>   USBVideo:
>     match:
>         '@ID_PATH': '{{sysfs}}'
> {% endfor %}
> ```
> 
> I'd rather properly document this than adding a special case for group parsing.

Thanks for looking at this!

I did look at the templating. It is very powerful but also a bit fiddly for hubs. My proposal is to build in hubs as first-class citizens.

Here is an example, showing how this approach is very easy to follow. It is also self-documenting. The important distinction is that I can look at the board, immediately see which hub has the USB, if it needs a replug, etc.

```
hubs:
  a:
    base: 'pci-0000:00:14.0-usb-0:10'
    ports:
      1: '2.1'
      2: '2.2'
      3: '2.3'
      4: '2.4'
      5: '3.1'
      6: '3.2'
      7: '3.3'
      8: '3.4'
      9: '4.1'
      10: '4.2'
      11: '4.3'
      12: '4.4'
      13: '1.1'
      14: '1.2'
      15: '1.3'
      16: '1.4'

olimex-a20:
  USBSerialPort:
    match:
      hub: a
      port: 2
      iface: '1.0'

  HIDRelay:
    index: 4
    match:
      hub: a
      port: 14

  SunxiUSBLoader:
    match:
      hub: a
      port: 12
      iface: '1.0'

**Comment by sjg20** [comment]
Just to sketch where this can go in future, since the `hubs` section gives labgrid a name for each hub and a logical number for each port.

Once the topology is described in one place, the exporter can report on it. In a local branch I have two commands built on top of this section.

`labgrid-exporter --list-hubs` lists every hub with its base path, how many ports it defines, and whether it is currently present. Run on the exporter host it also fills in live data from the USB tree: which bus and controller each hub sits on, and, for each xHCI controller, how many device slots are in use versus the maximum. That last part matters because the slots are shared per controller, so a hub can stop working simply because the controller is full - the output flags this with a '*FULL*' marker rather than leaving you to guess. Here is part of the output for my lab:

```
Hubs defined in kea_export_hubs.yaml:

HUB     BASE / ORIGIN                 PORTS  PRESENT  BUS  CONTROLLER
a       pci-0000:06:00.2-usb-0:2      16     yes      002  0000:06:00.2
b       pci-0000:04:00.0-usb-0:2      16     yes      009  0000:04:00.0
c       pci-0000:03:00.2-usb-0:5      16     yes      001  0000:03:00.2
...
ykush0  pci-0000:03:00.2-usb-0:5.1.2  3      yes      001  0000:03:00.2

USB controllers (xHCI device slots are shared per controller):

CONTROLLER    DRIVER    DEVICES  MAX  FREE  HUBS
0000:00:14.0  xhci_hcd  46       ?    ?     cg, cg2, g
0000:03:00.2  ehci-pci  113      127  14    c, d, e, f, ykush0, ykush2
0000:04:00.0  xhci_hcd  26       ?    ?     b, ykush1
0000:06:00.2  ehci-pci  31       127  96    a
```

The ehci controller on 0000:03:00.2 is down to 14 free slots with six hubs on it, so it is a candidate for trouble as more boards are added.

`labgrid-exporter --hub-tree` walks the live USB tree and prints, for each labelled port, the device plugged into it and the labgrid resource(s) it backs, recursing into any sub-hubs. So instead of a bare '@ID_PATH' you see the board and resource against the physical port it is on, along with the device's VID:PID, product name and serial. Ports with nothing on them show up as free (shown as '--'). A fragment for one hub:

```
hub a  (pci-0000:06:00.2-usb-0:2)
    1  olimex-a20      USBSDWireDevice  04e8:6001  sd-wire  [sdwireda2]
    2  olimex-a20      USBSerialPort    067b:2303  PL2303 Serial Port
    ...
    8  jerry           Servo            18d1:5002  servo  [730422-00045]
    9                                   0483:5750  LED badge -- 11x44
   10  tk1             recovery 7       16c0:05df  HID device
       edison          reset 5
       lcdk            HIDRelay 8
       pine64          HIDRelay 4
   11  --
   12  olimex-a20      SunxiUSBLoader   18d1:0003  AllWinner SoC  [20080411]
```

Port 11 is free, port 9 holds a device backing no resource (a stray LED badge), and port 10 is a single HID relay shared by several boards - exactly the kind of thing that is invisible in a list of path strings.

The reason this helps with bad connections is that it turns that wall of strings into a map of the lab. When a board's UART drops out you can run --hub-tree, find which hub and port it should be on, and see immediately whether the device is missing, has moved to a different port, or is present but backing the wrong resource. --list-hubs then tells you whether the whole hub or controller has gone away or run out of slots.

None of this is in this PR - it needs the diagnostic commands plus the per-port and controller-slot probing - but it all rests on the `hubs` abstraction here, which is why I wanted to note the direction.

Emantor:
These commands are very suitable for `labgrid-suggest` and I'd prefer that `labgrid-suggest` both learns how to report the correct templating for USB hubs and display the same structured output as your commands implemented on top of the labgrid exporter.

The labgrid exporter binary utility contains the facilities to run the exporter and adding those features to labgrid suggest has benefits outside of the exporter:
- This also works for locally connected USB hubs to make local yaml device configurations easier to write
- Jinja templating can be an optional additional output for exporter usage
- The USB hub listing is useful either way to find which devices are connected where to the computer it is running on

PR #1862, opened 2026-05-06
Support logging of serial traffic in the exporter

sjg20:
**Description**

Add optional centralised serial-traffic logging on the exporter
host: when `LG_SERIAL_TRACE_DIR` is set, the exporter passes
`trace-both` to the per-acquire `ser2net` and the resulting log
files are named `<board>-<user>.log`.  This closes a visibility
gap for shared/multi-user labs where the per-client `--logfile`
only captures one user's session and the operator otherwise has
no independent record of what is happening on the console.

Three commits:

1. **protocol**: add a `user` field to `ExporterSetAcquiredRequest`
   so the exporter knows who acquired a place; needed by (2).
   Older coordinators that don't send the field continue to work.
2. **exporter**: the trace-logging feature itself, plus tests, a
   usage section in `doc/usage.rst`, an env-var entry in the
   exporter man page (regenerated under `man/`).
3. **tests**: omit `labgrid/remote/generated/*` from coverage so
   protoc regen noise stops dragging the patch-coverage threshold
   under on routine protocol changes.

Tested on my own lab (~50 boards exported via ser2net 4.x,
including BeagleBone Black, Raspberry Pi 3, Coral and a handful
of x86 systems) over tailscale.  Trace files correctly capture
every keystroke and reply for each acquire session, and append
on repeated acquires by the same user on the same board.

**Checklist**
- [x] Tests for the feature 

- [x] PR has been tested


Emantor:
I like the general idea, especially since we can make use of the existing ser2net configuration capabilities for that.
Users should have SSH access to the exporter anyway, so from a permission/access point of view this also makes sense as long as we expect all users to have the same permissions to access all hardware connected to an exporter.

The thing I worry about is how to make this accessible to users and how to handle log rotation and time stamping so users are able to correlate which part of the log file was generated by them. Or do you intend this to be a purely administrative aid in debugging failing lab hardware?

Edit: I haven't checked that time stamping is handled by ser2net, if you could add a sample log file that would be great.

For the user side we already have `--lg-log` for the pytest plugin and for library usage `ConsoleLoggingReporter(<directory>` can be created, so this may be a bit redundant?

sjg20:
> I like the general idea, especially since we can make use of the existing ser2net configuration capabilities for that. Users should have SSH access to the exporter anyway, so from a permission/access point of view this also makes sense as long as we expect all users to have the same permissions to access all hardware connected to an exporter.
> 
> The thing I worry about is how to make this accessible to users and how to handle log rotation and time stamping so users are able to correlate which part of the log file was generated by them. Or do you intend this to be a purely administrative aid in debugging failing lab hardware?
> 
> Edit: I haven't checked that time stamping is handled by ser2net, if you could add a sample log file that would be great.
> 
> For the user side we already have `--lg-log` for the pytest plugin and for library usage `ConsoleLoggingReporter(<directory>` can be created, so this may be a bit redundant?

Thanks for the review!

To clarify the intent: this is an **admin-only audit/debug aid**, not a user-facing logging mechanism. It complements rather than replaces `--lg-log` and `ConsoleLoggingReporter`:

- `--lg-log` and `ConsoleLoggingReporter` record what *the user* sees, in a location the user controls. Users can turn them off, change the filter, or just not enable them.
- `LG_SERIAL_TRACE_DIR` records *every* byte that flows through the serial console of every acquire, on the exporter host, where the lab admin can read it. The user can't affect what gets logged.

The new exporter trace is specifically for *"which user ran which commands on board X between Tue 14:00 and 17:00 to put it in this state"*-type questions, which the per-client logs don't answer.

On timestamps, the exporter does pass `trace-both-timestamp: true`, but actually
ser2net only timestamps the OPEN and CLOSE session markers in the default trace
mode, so not individual data lines:

```
2026/05/13 06:58:50 OPEN (ipv6,::ffff:192.168.4.7,60208)
...
 => eecchhoo  MMee  aaggaaiinn
Me again

=> 2026/05/13 06:59:06 CLOSE netcon (network read close)
2026/05/13 06:59:06 CLOSE port (All users disconnected)
```

(Both directions land in the same byte stream, which is why each typed character shows up twice - once from the user, once from the board's local echo.)

If per-line timestamps actually matter for a deployment, v2 adds `LG_SERIAL_TRACE_HEXDUMP=1` which switches ser2net into hex+ASCII mode. The output is no longer readable text but every line carries a timestamp and a direction tag (`term` from the board, `tcp` from the client):

```
2026/05/13 07:21:11 tcp  65                       |e|
2026/05/13 07:21:11 term 65                       |e|
2026/05/13 07:21:11 tcp  63                       |c|
2026/05/13 07:21:11 term 63                       |c|
...
2026/05/13 07:21:15 term 0d 0a 48 65 6c 6c 6f 20  |..Hello |
2026/05/13 07:21:15 term 41 67 61 69 6e 0d 0a 3d  |Again..=|
```

On log rotation: not handled in the exporter itself.  ser2net opens the trace file in append mode on each acquire, so `logrotate`'s `copytruncate` or `daily` modes on `LG_SERIAL_TRACE_DIR` work
unchanged.

I updated usage.rst to show examples and talk about log rotation.

Emantor:
The username propagation is also implemented in https://github.com/labgrid-project/labgrid/pull/1939 based on the client identity, so we can't merge this until we get the identity propagation in place.

sjg20:
Thanks - agreed, the user propagation here overlaps with #1939 and that PR's client-identity approach is the better foundation, since it authenticates the identity rather than trusting what the coordinator was told.

I would suggest this split: I drop the 'Pass acquiring user from coordinator to exporter' commit from this PR and rebase the serial-trace logging on top of #1939 once it lands, taking the username for the trace filename from the propagated client identity instead. The rest of the PR (the ser2net trace itself, the hexdump option and the docs) is independent of where the identity comes from, so it should rebase cleanly.

Happy to do that as soon as #1939 is merged - or if it is still some way off, I could land the trace logging now with the identity falling back to 'unknown', and switch it over in a follow-up. Whichever you prefer.

PR #1863, opened 2026-05-07
Allow `NetworkService` to be used without an explicit SSH username

ozan956:
Allow `NetworkService` to be used without an explicit SSH username.

Until now, `NetworkService.username` was required and `SSHDriver` always passed
it through to `ssh`, `scp`, etc. That prevented setups where the SSH user should
come from the local SSH configuration or the default SSH user resolution.

This change makes `NetworkService.username` optional and updates `SSHDriver` to
only pass `-l <username>` or `user@host` when a username is explicitly set.

The client-side fallback path in `labgrid.remote.client` also no longer forces
`username="root"`. While adding regression coverage for that path, this also
surfaced that the fallback resource creation needs `name=None`, which is now
passed explicitly.

I verified the change with some tests called `test_run_no_username`, `test_put_get_no_username` and  `test_get_ssh_no_username` for the no-username paths.

**Checklist**
- [x] Tests for the feature 
- [x] PR has been tested

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -698,6 +698,8 @@ A :any:`NetworkService` describes a remote SSH connection.
 
 The example describes a remote SSH connection to the computer
 ``example.computer`` with the username ``root``.
+If ``username`` is omitted, labgrid will let SSH resolve the username through
+the local SSH configuration or the default SSH user selection.
 Set the optional password password property to make SSH login with a password
 instead of the key file.
 
@@ -710,7 +712,7 @@ These and the sudo configuration needs to be prepared by the administrator.
 
 Arguments:
   - address (str): hostname of the remote system
-  - username (str): username used by SSH
+  - username (str, default=None): optional, username used by SSH
   - password (str, default=None): optional, password used by SSH
   - port (int, default=22): port used by SSH
 
@@ -2091,7 +2093,9 @@ Arguments:
     will explicitly use the SFTP protocol for file transfers instead of scp's default protocol
   - explicit_scp_mode (bool, default=False): if set to True, ``put()``, ``get()``, and ``scp()``
     will explicitly use the SCP protocol for file transfers instead of scp's default protocol
-  - username (str, default=username from `NetworkService`_): username used by SSH
+  - username (str, default=username from `NetworkService`_): optional, username used by SSH
+    If neither `SSHDriver`_ nor `NetworkService`_ specifies a username, SSH's
+    own username resolution is used.
   - password (str, default=password from `NetworkService`_): password used by SSH
 
 UBootDriver
--- a/labgrid/driver/sshdriver.py
+++ b/labgrid/driver/sshdriver.py
@@ -118,8 +118,10 @@ class SSHDriver(CommandMixin, Driver, CommandProtocol, FileTransferProtocol):
                  "-o", "ControlPersist=300", "-o",
                  "UserKnownHostsFile=/dev/null", "-o", "StrictHostKeyChecking=no",
                  "-o", "ServerAliveInterval=15", "-MN", "-S", control.replace('%', '%%'), "-p",
-                 str(self.networkservice.port), "-l", self._get_username(),
-                 self.networkservice.address]
+                 str(self.networkservice.port)]
+        if username := self._get_username():
+            args += ["-l", username]
+        args += [self.networkservice.address]
 
         # proxy via the exporter if we have an ifname suffix
         address = self.networkservice.address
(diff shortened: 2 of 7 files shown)


doc/configuration.rst
```diff
@@ -706,7 +708,7 @@ These and the sudo configuration needs to be prepared by the administrator.
 
 Arguments:
   - address (str): hostname of the remote system
-  - username (str): username used by SSH
+  - username (str, default=""): optional, username used by SSH
```
Emantor: The default should be `None`, not the empty string.

labgrid/driver/sshdriver.py
```diff
@@ -118,8 +118,10 @@ def _start_own_master_once(self, timeout):
                  "-o", "ControlPersist=300", "-o",
                  "UserKnownHostsFile=/dev/null", "-o", "StrictHostKeyChecking=no",
                  "-o", "ServerAliveInterval=15", "-MN", "-S", control.replace('%', '%%'), "-p",
-                 str(self.networkservice.port), "-l", self._get_username(),
-                 self.networkservice.address]
+                 str(self.networkservice.port)]
+        if self._get_username():
```
Emantor:
```suggestion
        if username := self._get_username():
```

labgrid/driver/sshdriver.py
```diff
@@ -118,8 +118,10 @@ def _start_own_master_once(self, timeout):
...
+            args += ["-l", self._get_username()]
```
Emantor:
```suggestion
            args += ["-l", username]
```

labgrid/driver/sshdriver.py
```diff
@@ -214,9 +216,10 @@ def _run(self, cmd, codec="utf-8", decodeerrors="strict", timeout=None):
             raise ExecutionError("Keepalive no longer running")
 
         complete_cmd = [self._ssh, "-x", *self.ssh_prefix,
-                        "-p", str(self.networkservice.port), "-l", self._get_username(),
-                        self.networkservice.address
-                        ] + cmd.split(" ")
+                        "-p", str(self.networkservice.port)]
+        if self._get_username():
```
Emantor: same as above

labgrid/driver/sshdriver.py
```diff
@@ -513,13 +520,17 @@ def put(self, filename, remotepath=''):
     @Driver.check_active
     @step(args=['filename', 'destination'])
     def get(self, filename, destination="."):
+        source = f"{self.networkservice.address}:{filename}"
+        if self._get_username():
```
Emantor: same as above

labgrid/driver/sshdriver.py
```diff
@@ -560,7 +574,10 @@ def _cleanup_own_master(self):
 
     def _start_keepalive(self):
         """Starts a keepalive connection via the own or external master."""
-        args = [self._ssh, *self.ssh_prefix, self.networkservice.address, "cat"]
+        args = [self._ssh, *self.ssh_prefix]
+        if self._get_username():
```
Emantor: same as above

Emantor (commented):
This is going to be a breaking change for the remote infrastructure, older labgrid will expect the now optional username parameter. IMO we should still use `None` instead of `""` which works around that.

ozan956:
> This is going to be a breaking change for the remote infrastructure, older labgrid will expect the now optional username parameter. IMO we should still use `None` instead of `""` which works around that.

Thanks for the review! I resolved them all now. Sorry for late response.

jluebbe:
We use `NetworkService` for SSH daemons on the target. What would be the use-case for having the username depend on client side configuration? I'd normally prefer to have test suites behave identically for different users.

ozan956:
> We use `NetworkService` for SSH daemons on the target. What would be the use-case for having the username depend on client side configuration? I'd normally prefer to have test suites behave identically for different users.

That is a fair point. For a NetworkService declared in a test environment, I agree that the username should normally be explicit so that test suites behave identically for different users.

My use case is shared lab usage rather than making test suites depend on client-side configuration. Several users consume the same environment for ad-hoc ssh/scp operations, and using one fixed Unix account forces them to share the same remote home directory. This can lead to interference between users.

Making the username optional allows this shared environment to leave account selection to each user's SSH configuration. Explicitly configured usernames continue to behave exactly as before.

Also with #1889, we are going to sharing the envs remotely. So not having a username there would be better I guess. Users gonna anyways having ssh configs at their end.

jluebbe:
So in your case, each developer/lab user has their own user account on the target/device-under-test?

ozan956:
> So in your case, each developer/lab user has their own user account on the target/device-under-test?

Now we are using a shared user, especially since the ci is the main consumer of it. But once many more people start to use the boards, we cannot control what these all consumers doing with all those available commands. And something like this can happen: user#1 sends u-boot binares to home directory, working on board#1. at the same time user#2 sends their binaries to same home directory for board#2. So they going to be overriden. Since we dont wanna "lock' the directories, I believe the easiest thing is to seperate users.

I am open to any other suggestions to get rid of that race condition as well.

PR #1864, opened 2026-05-07
remote/client: support OpenOCD bootstrap without environment

ozan956:
The bootstrap fallback path was too narrow for OpenOCD-based setups without a
local environment.

Previously, `labgrid-client bootstrap` only auto-created `OpenOCDDriver` for
`NetworkAlteraUSBBlaster`. If a place exposed only `NetworkUSBDebugger`, the
client still failed with `target has no compatible resource available` even
though `OpenOCDDriver` can bind to that resource type.

The existing `bootstrap_args` path was also too limited. It only handled raw
`key=value` strings, which was not sufficient for `OpenOCDDriver` arguments
such as `load_commands`, `search`, or `config`, where structured values like
lists are often required.

This change:
- extends the bootstrap fallback to `NetworkUSBDebugger`
- parses bootstrap driver arguments as YAML values
- allows structured OpenOCD driver arguments to be passed directly from the CLI

This makes CLI-only OpenOCD bootstrap flows usable in more cases without
requiring a local environment file.

I verified the change with focused regression tests:

```bash
pytest -q tests/test_client.py -k 'bootstrap_network_usb_debugger or bootstrap_openocd_args'
```

**Checklist**
- [x] Tests for the feature 
- [x] PR has been tested

### DIFF
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -28,6 +28,7 @@ from typing import Any, Dict
 
 import attr
 import grpc
+import yaml
 
 # TODO: drop if Python >= 3.11 guaranteed
 from exceptiongroup import ExceptionGroup  # pylint: disable=redefined-builtin
@@ -932,6 +933,19 @@ class ClientSession:
                 target.activate(drv)
             return drv
 
+    def _parse_driver_args(self, args):
+        parsed = {}
+        for arg in args:
+            try:
+                key, value = arg.split("=", 1)
+            except ValueError:
+                raise UserError(f"invalid bootstrap argument '{arg}', expected key=value")
+            try:
+                parsed[key] = yaml.safe_load(value)
+            except yaml.YAMLError as e:
+                raise UserError(f"invalid value for bootstrap argument '{key}': {e}") from e
+        return parsed
+
     def power(self):
         place = self.get_acquired_place()
         action = self.args.action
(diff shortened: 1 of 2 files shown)


Emantor:
Do we need a complete yaml definition and loader for the bootstrap arguments? Can't we instead supply an openocd native target configuration instead or limit this to a list of key-value items?

ozan956:
> Do we need a complete yaml definition and loader for the bootstrap arguments? Can't we instead supply an openocd native target configuration instead or limit this to a list of key-value items?

Thanks @Emantor,

As you see I am trying to solve two problems. First, extending the bootstrap fallback path so labgrid-client bootstrap can also instantiate OpenOCDDriver for NetworkUSBDebugger when no local environment is available. Second, making it possible to pass the structured OpenOCD driver arguments needed by that CLI-only path.

Regarding your suggestion, I can say that this can work for some cases, but I do not think its the best option to have a config wrapper like that. The main reason is that the arguments here are not only static target configuration. They also describe the bootstrap workflow itself, for example custom `source ...`, `init`, `autoboot_elf`, and `shutdown` sequences. Those are valid OpenOCD commands, so they can certainly be moved into a script file, but requiring a separate OpenOCD file for every target or case-specific variation would add maintenance overhead ( since there can be multiple config/command combinations for multiple boards) and keep the CLI-only path less practical.

I do agree that the generic YAML parsing may be overkill. The important part for me is the ability to pass the OpenOCD-specific structured arguments needed by this bootstrap flow, not to introduce a fully generic typed argument mechanism for bootstrap in general. So, I think we can maybe restrict this path to a small whitelist of OpenOCD-specific keys such as config, search, load_commands, interface_config, etc.

jluebbe:
Would #1889 work for your use-case as well? Then you'd have the `load_commands` in the remote place config?

ozan956:
Exactly. We can close this once #1889 is merged. I did some improvements on top of Bastian's work in our fork (mentioned in #1889) to test it. Seemed fine and ultimately removed the need for this PR.

Issue #1375, opened 2024-04-26
Selecting the target in pytest


sjg20:
The pytest code for selecting a target currently looks like this (fixtures.py):
```
@pytest.fixture(scope="session")
def target(env):
    """Return the default target `main` configured in the supplied
    configuration file."""
    target = env.get_target()
    if target is None:
        raise UserError("Using target fixture without 'main' target in config")

    return target
```

This works well with a single place, called 'main'. But when there are multiple places, we need a way to select the correct one.

I propose adding a `--lg-target` option to fixtures.py so that this can be specified.

Then my script to check a particular commit (use with 'git bisect run') can be something like this:

pytest --lg-env env_rpi_try.cfg \
	--lg-target ${board} \
	--lg-log --lg-colored-steps \
	-k test_uboot_smoke

Does this seems reasonable?

Emantor:
The intention here is for your pytest files to provide fixtures for individual devices. Something like:
```python
@pytest.fixture(scope="session")
def devicea(env):
    return env.get_target("devicea")

@pytest.fixture(scope="session")
def deviceb(env):
    return env.get_target("deviceb")
```

And than use those as test fixtures.

sjg20:
That's fine, but in my case I have a lab of 33 boards and the tests for each are the same...so the above becomes quite unwieldy. I would need to duplicate all the tests 33 times just so they can use a different board

Emantor:
In this case you'll want to pass a different environment configuration per test. You can keep the same tests, but the target changes per environment file.

sjg20:
Again, this becomes fiddly. I would like to have the env for the lab in one file, rather than editing so many files.

So you are not keen on the feature I propose?

Emantor:
> Again, this becomes fiddly. I would like to have the env for the lab in one file, rather than editing so many files.
> 
> So you are not keen on the feature I propose?

No, unfortunately it does not fit the locking model for labgrid. If you use the pytest plugin for tests, labgrid requires all places which are used inside the environment file to be locked to your user. This works well if you are the only one using the lab, however does not work at all if you have multiple users accessing different hardware in the same infrastructure. That is why different environment files are required if the same tests are going to run on different hardware.

sjg20:
Oh dear I didn't know that. Can that be changed? To me it seem better to lock a place when it is needed, rather than locking all places mentioned in the environment file?

I am starting to see that each command-line program has its own file...is that right?
1. export.yaml for the exporter, contains the devices
[...]

PR #1866, opened 2026-05-12
Fetch env file from coordinator

sjg20:
**Description**

Adds a way for `labgrid-client` to fetch the env file from the coordinator instead of requiring every user to keep a local copy in sync.

- The coordinator gains a new `GetEnvironment` RPC and an `--environment FILE` flag pointing at the YAML env file to serve. The file is read fresh on each request, so admins can edit it in place.
- The client treats the literal string `coordinator:` as a special `LG_ENV` / `--config` value. When set, it issues a one-shot `GetEnvironment` RPC, writes the YAML under `$XDG_CACHE_HOME/labgrid` (or `~/.cache/labgrid`), and loads it like any other env file. Per-user paths inside the served env are still resolvable via the existing `LG_*` template substitution.
- Default behaviour is unchanged: with no `--environment` flag on the coordinator, `GetEnvironment` returns an empty string and clients keep loading env from a local file.

*What is the feature used for?* Triage and console access from a laptop, scaling to larger labs with many clients, and removing the out-of-band env-file distribution step (shared FS, git checkout, scp, etc).

*How does labgrid benefit?* Lower friction for casual remote use - a remote user only needs `labgrid`, network access to the coordinator and three env vars (`PATH`, `LG_COORDINATOR`, `LG_ENV=coordinator:`) to run `labgrid-client console -p <board>` from anywhere.

*How was it verified?* Ran the coordinator with `--environment <file>` and a client with `LG_ENV=coordinator:` against it locally; confirmed the YAML is fetched, cached under `~/.cache/labgrid`, and that subsequent commands resolve `LG_*` template paths correctly. Confirmed that with no `--environment` flag, behaviour is unchanged.

No hardware-specific changes.

**Checklist**
- [x] Tests for the feature
- [x] PR has been tested


flxzt:
from the experience of using labgrid, this change sounds great. Reducing complexity on user-side is something I would welcome very much.

Bastian-Krause:
I haven't looked in detail at this, but I'm currently experimenting with a rather similar change at https://github.com/Bastian-Krause/labgrid/commits/bst/coordinator-remote-env/. It's still a prototype, things might change, but I think that's the general direction we'd like to head.

I'll open a RFC PR with this next week.

Emantor:
Both of these approaches lock the environment down to one target, which limits the usefulness to single target tests. While this is definitely useful on it's own, we should rethink the concept to make multi-target environments possible as well.

sjg20:
> Both of these approaches lock the environment down to one target, which limits the usefulness to single target tests. While this is definitely useful on it's own, we should rethink the concept to make multi-target environments possible as well.

Thanks for looking at this.

I am actually using this approach with about 30 targets. Do you mean a single exporter?

I have some local changes which allow each exporter to send an environment and have the coordinator merge them. I haven't sent it yet since I thought this PR was a good starting point. Let me know if I should send that too.

But in any case this PR is likely just a starting point.

Emantor:
You are very correct, sorry, that was my fault. I will try to find time to take a closer look some time next week.

Bastian-Krause:
> Both of these approaches lock the environment down to one target, which limits the usefulness to single target tests. While this is definitely useful on it's own, we should rethink the concept to make multi-target environments possible as well.

That's not the case for [my approach](https://github.com/Bastian-Krause/labgrid/commits/bst/coordinator-remote-env/): The remote env is applied per place (optionally combined with a local env), so multi target envs still work. For pytest, a minimal example (relying solely on the remote envi) would be:
```yaml
targets:
  test1:
    resources:
      RemotePlace:
        name: test1
  test2:
    resources:
      RemotePlace:
        name: test2
```
But let's continue the discussion in the draft PR I'm going to open.

Bastian-Krause:
@sjg20 Could you give an example of an environment you give the exporter? Does it contain all your targets (including drivers and resources)? Looking at the [env skeleton in the docs](https://labgrid.readthedocs.io/en/latest/configuration.html#environment-configuration), do you also store options, features, paths, images, tools and imports in there?

sjg20:
> @sjg20 Could you give an example of an environment you give the exporter? Does it contain all your targets (including drivers and resources)? Looking at the [env skeleton in the docs](https://labgrid.readthedocs.io/en/latest/configuration.html#environment-configuration), do you also store options, features, paths, images, tools and imports in there?

Yes, please see here for exporter and environment:

[asc.txt](https://github.com/user-attachments/files/28438478/asc.txt)
[asc2.txt](https://github.com/user-attachments/files/28438491/asc2.txt)

jluebbe:
> [asc2.txt](https://github.com/user-attachments/files/28438491/asc2.txt)

This contains many different targets in a single environment file.
Each environment is intended to describe a single *test environment*, not a whole lab.
It abstracts different test environments from the perspective of a test suite/client program.
As such it doesn't make sense to have only one, as then you lose the ability to cleanly switch between test environments.

I think I explained this to @sjg20 before, but don't have time to dig it up at the moment.

To be explicit: I'm going to NACK any change which is motivated only by not using labgrid's abstractions as they are intended to be used.

I see the need to share more of the test environment's configuration between users and @Bastian-Krause's approach will give us a way to do that, while not breaking with existing abstractions.

sjg20:
> > [asc2.txt](https://github.com/user-attachments/files/28438491/asc2.txt)
> 
> This contains many different targets in a single environment file. Each environment is intended to describe a single _test environment_, not a whole lab. It abstracts different test environments from the perspective of a test suite/client program. As such it doesn't make sense to have only one, as then you lose the ability to cleanly switch between test environments.
> 
> I think I explained this to @sjg20 before, but don't have time to dig it up at the moment.

Yes, it is here:

https://github.com/labgrid-project/labgrid/issues/1375

Bastian-Krause:
> I haven't looked in detail at this, but I'm currently experimenting with a rather similar change at https://github.com/Bastian-Krause/labgrid/commits/bst/coordinator-remote-env/. It's still a prototype, things might change, but I think that's the general direction we'd like to head.
> 
> I'll open a RFC PR with this next week.

Here it is: #1889

PR #1872, opened 2026-05-20
remote/coordinator: run reservation scheduling under lock

alextercete:
**Description**
Reservation scheduling mutates shared coordinator state (`reservations`, `places`, and the derived `place.reservation` view), but the periodic scheduler path was performing these updates without holding the coordinator lock.

This can lead to races between the poll-based scheduler and concurrent RPC handlers (e.g. acquire/release/reservation operations), potentially resulting in inconsistent state or incorrect reservation transitions.

We now run reservation scheduling under the coordinator lock by making it async and awaiting it in the poll path. This ensures that all mutations of coordinator state are serialized consistently with the rest of the coordinator operations.

Renamed `schedule_reservations` to `_schedule_reservations` to reflect that it is now an internal async helper. All usages in the coordinator were updated.  This method is not part of a public API, so renaming it is not expected to impact external users.

_Testing_

- Ran existing unit tests covering reservation and acquire/release flows.
- Exercised reservation creation/cancel and place acquire/release scenarios manually.
- No deadlocks were observed in these runs.
- This patch does not add a dedicated stress test, and the current tests do not prove the absence of deadlocks.

_Performance_

- Measured `_schedule_reservations` runtime under load. With 100 places and 200 reservations created in parallel, observed runtime was ~0.0013s, with worst case ~0.0036 s.
- RPC paths were already invoking reservation scheduling while holding the coordinator lock; the primary change is that the poll-based scheduler path is now also serialized under the same lock.

This patch focuses on fixing the scheduler locking model and does not change coordinator scheduling behavior.

**Checklist**
- [x] PR has been tested

### DIFF
--- a/labgrid/remote/coordinator.py
+++ b/labgrid/remote/coordinator.py
@@ -241,8 +241,9 @@ class Coordinator(labgrid_coordinator_pb2_grpc.CoordinatorServicer):
 
     async def _poll_step_schedule(self):
         # update reservations
-        with warn_if_slow("schedule reservations"):
-            self.schedule_reservations()
+        async with self.lock:
+            with warn_if_slow("schedule reservations"):
+                self.schedule_reservations()
 
     async def poll(self, step_func):
         while not self.loop.is_closed():
@@ -956,6 +957,7 @@ class Coordinator(labgrid_coordinator_pb2_grpc.CoordinatorServicer):
         # only have a copy for convenience.
 
         # expire reservations
+        assert self.lock.locked()
         for res in list(self.reservations.values()):
             if res.state is ReservationState.acquired:
                 # acquired reservations do not expire


labgrid/remote/coordinator.py
```diff
@@ -950,11 +951,12 @@ async def GetPlaces(self, unused_request, unused_context):
         except Exception:
             logging.exception("error during get places")
 
-    def schedule_reservations(self):
+    async def _schedule_reservations(self):
```
Emantor: What exactly do we gain from turning thie into an `async` function? As far as I can tell this function does not employ an async primitive so it could be parallelized or am I missing something?

Emantor (commented):
I agree with the locking, however I do not see what we gain from turning `_schedule_reservation()` into an async function.

labgrid/remote/coordinator.py
alextercete: Apologies, those changes were leftover from a previous iteration of this fix. I've updated the commit to remove them.

Emantor (approved):

PR #1878, opened 2026-06-01
resource&driver: add mtd resource, flashcp driver

flxzt:
**Description**

This change adds a resource&driver pair for MTD partitions, commonly found in embedded board utilizing flash storage.

A resource is added that identifies MTD partitions by their number as they appear in the devfs,
and a driver that flashes these partitions using the `flashscp` tool, which is part of `mtdutils` on popular distributions.

**Checklist**

- [x] PR has been tested

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -682,6 +682,27 @@ Arguments:
 Used by:
   - `GpioDigitalOutputDriver`_
 
+DevfsMTDPartition
++++++++++++++++++
+
+A :any:`DevfsMTDPartition` resource describes a MTD partition accessible through the devFS.
+
+.. code-block:: yaml
+
+   DevfsMTDPartition:
+     index: 0
+
+Arguments:
+  - index (int): index of the MTD partition
+
+Used by:
+  - `FlashcpDriver`_
+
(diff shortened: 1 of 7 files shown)


Emantor:
How does this work? MTD devices are usually only accessible on the target itself, what value is added by abstracting this in the remote infrastructure? For on device helpers there is the [labgridhelper](https://github.com/labgrid-project/labgridhelper) which is the intended destination for helper functions that work on the target.

flxzt:
I see, then the MTDPartition is not really necessary.

Is the FlashcpDriver something that provides value in your view, since it integrates with labgrid's abstractions like `BootstrapProtocol`?

Emantor:
Unfortunately I don't think so. The `BoostrapProtocol` drivers are usually used to start a board over some sort of download mode to "bootstrap" into a bootloader and then start linux. The FlashcpDriver instead expects to already be running Linux.

jluebbe:
If you want to configure the commands to execute on the target during bring up, attributes on the strategy are a way to do that.

flxzt:
That makes sense, thanks for the answers and pointers.

PR #1885, opened 2026-06-03
Remote server tls authentication

asher-pem-arm:
This MR adds support for gRPC SSL/TLS server authentication to `labgrid-coordinator`, `labgrid-exporter` and `labgrid-client`.

Enabling a secure channel on all three components is done by adding the `--secure` argument. Paths to the certificate and key (`labgrid-coordinator`-only) can be specified with the `--cert` and `--key` arguments.

Testing has been added to verify that, given a secure-enabled `labgrid-coordinator`, both `labgrid-client` and `labgrid-exporter` can connect successfully and perform an operation.

#### Note on use of `labgrid-client` and `labgrid-exporter` without an explicit `--cert` specified

On Linux, this currently reads the Debian/Ubuntu CA bundle at:
`/etc/ssl/certs/ca-certificates.crt`

On macOS, this reads certificates from the system Keychain using:
`security find-certificate -a -p`

If the coordinator certificate is not trusted by the host, or the platform's system roots cannot be loaded, pass the coordinator certificate explicitly with `--cert`.

Closes: #1541

### DIFF
--- a/doc/getting_started.rst
+++ b/doc/getting_started.rst
@@ -446,14 +446,14 @@ Refer to the ``labgrid-coordinator`` man page for details.
 
 When you are connecting with ``labgrid-client`` or ``labgrid-exporter`` to a
 ``labgrid-coordinator`` that has TLS gRPC channels enabled you need to pass
-the ``--tls`` option. If ``--cert`` is not set, labgrid uses the host CA
-certificates to verify the coordinator certificate. Use ``--cert`` to provide
-a specific CA certificate instead.
+the ``--tls`` option. If ``--cacert`` is not set, labgrid uses the host CA
+certificates to verify the coordinator certificate. Use ``--cacert`` to provide
+a specific CA certificate or CA bundle instead.
 Refer to the ``labgrid-client`` and ``labgrid-exporter`` man pages for details.
 For ``RemotePlace`` connections from an environment config, set the
-``coordinator_tls`` option or ``LG_COORDINATOR_TLS``. If ``coordinator_cert``
-is not set, labgrid uses the host CA certificates. Set ``coordinator_cert``
-to provide a specific CA certificate instead.
+``coordinator_tls`` option or ``LG_COORDINATOR_TLS``. If ``coordinator_cacert``
+is not set, labgrid uses the host CA certificates. Set ``coordinator_cacert``
+to provide a specific CA certificate or CA bundle instead.
 
 Using a Strategy
 ----------------
--- a/doc/getting_started.rst
+++ b/doc/getting_started.rst
@@ -435,6 +435,26 @@ Follow these instructions to install the systemd files on your machine(s):
 
       # usermod -a -G labgrid <user>
 
+Enabling gRPC connection security
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+It is encouraged to use TLS for gRPC channels in a production environment.
+
+This can be enabled on the ``labgrid-coordinator`` by adding the ``--tls``,
+``--cert`` and ``--key`` options.
+Refer to the ``labgrid-coordinator`` man page for details.
+
+When you are connecting with ``labgrid-client`` or ``labgrid-exporter`` to a
+``labgrid-coordinator`` that has TLS gRPC channels enabled you need to pass
+the ``--tls`` option. If ``--cert`` is not set, labgrid uses the host CA
+certificates to verify the coordinator certificate. Use ``--cert`` to provide
+a specific CA certificate instead.
+Refer to the ``labgrid-client`` and ``labgrid-exporter`` man pages for details.
+For ``RemotePlace`` connections from an environment config, set the
+``coordinator_tls`` option or ``LG_COORDINATOR_TLS``. If ``coordinator_cert``
+is not set, labgrid uses the host CA certificates. Set ``coordinator_cert``
+to provide a specific CA certificate instead.
+
 Using a Strategy
 ----------------
 
--- a/doc/man/client.rst
+++ b/doc/man/client.rst
@@ -54,6 +54,11 @@ LG_COORDINATOR
 This variable can be used to set the default coordinator in the format
 ``HOST[:PORT]`` (instead of using the ``-x`` option).
 
+LG_COORDINATOR_TLS
+~~~~~~~~~~~~~~~~~~
+This variable can be set to enable TLS gRPC channels without using the
+``--tls`` option.
+
 LG_PROXY
 ~~~~~~~~
 This variable can be used to specify a SSH proxy hostname which should be used
@@ -127,6 +132,12 @@ Add all resources with the group "example-group" to the place example-place:
 .. code-block:: bash
 
    $ labgrid-client -p example-place add-match */example-group/*/*
+
+Retrieve a list of places on a ``labgrid-coordinator`` that uses TLS gRPC channels:
+
+.. code-block:: bash
+
+    $ labgrid-client --tls [--cert PATH] places
 
 See Also
 --------
--- a/doc/man/client.rst
+++ b/doc/man/client.rst
@@ -137,7 +137,7 @@ Retrieve a list of places on a ``labgrid-coordinator`` that uses TLS gRPC channe
 
 .. code-block:: bash
 
-    $ labgrid-client --tls [--cert PATH] places
+    $ labgrid-client --tls [--cacert PATH] places
 
 See Also
 --------
--- a/doc/man/coordinator.rst
+++ b/doc/man/coordinator.rst
@@ -28,7 +28,7 @@ OPTIONS
 --tls
     enable TLS gRPC channel
 --cert
-    path to TLS certificate (in PEM format)
+    path to server TLS certificate (in PEM format)
 --key
     path to TLS key (in PEM format)
 -d, --debug
--- a/doc/man/coordinator.rst
+++ b/doc/man/coordinator.rst
@@ -25,8 +25,40 @@ OPTIONS
     display command line help
 -l ADDRESS, --listen ADDRESS
     make coordinator listen on host and port
+--tls
+    enable TLS gRPC channel
+--cert
+    path to TLS certificate (in PEM format)
+--key
+    path to TLS key (in PEM format)
 -d, --debug
     enable debug mode
+
+TLS WITH A REVERSE PROXY
+------------------------
+
+Instead of enabling TLS in ``labgrid-coordinator`` directly, a reverse proxy can
+terminate TLS and forward cleartext gRPC to the coordinator. For example, with
+``nginx``:
+
+.. code-block:: nginx
+
+   server {
+       listen 20407 ssl http2;
+       server_name labgrid.example.com;
+
+       ssl_certificate     /etc/ssl/labgrid-coordinator.crt;
+       ssl_certificate_key /etc/ssl/labgrid-coordinator.key;
+
+       location / {
+           grpc_pass grpc://127.0.0.1:20408;
+       }
+   }
+
+In this setup, start ``labgrid-coordinator`` without ``--tls`` and point
+``labgrid-client`` and ``labgrid-exporter`` at the reverse proxy using
+``--tls``.
+
 
 SEE ALSO
 --------
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -41,6 +41,7 @@ from .common import (
     TAG_KEY,
     TAG_VAL,
     queue_as_aiter,
+    get_client_credentials,
 )
 from .. import Environment, Target, target_factory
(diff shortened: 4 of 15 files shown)


doc/man/client.rst
```diff
@@ -128,6 +128,12 @@ Add all resources with the group "example-group" to the place example-place:
 
    $ labgrid-client -p example-place add-match */example-group/*/*
 
+Retrieve a list of places on a ``labgrid-coordinator`` that uses secure gRPC channels:
+
+.. code-block:: bash
+
+    $ labgrid-client --secure [--cert PATH] places
```
Emantor: Feels very much like bikeshedding to write this, but should we rename `--secure` to `--tls` instead? This makes it clear that we are using TLS for gRPC secure channels instead of a generic `--secure` which does not carry much meaning.

Emantor (dismissed):
Otherwise looks good

doc/getting_started.rst
```diff
@@ -435,6 +435,21 @@ Follow these instructions to install the systemd files on your machine(s):
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+It is encouraged to secure gRPC channels in a production environment.
+
+This can be enabled on the ``labgrid-coordinator`` by adding the ``--secure``,
+``--cert`` and ``--key`` options.
+Refer to the ``labgrid-coordinator`` man page for details.
+
+When you are connecting with ``labgrid-client`` or ``labgrid-exporter`` to a
+``labgrid-coordinator``that has secure gRPC channels enabled you need to pass
```
gastmaier:
```suggestion
``labgrid-coordinator`` that has secure gRPC channels enabled you need to pass
```

doc/man/coordinator.rst
```diff
@@ -25,6 +25,12 @@ OPTIONS
     display command line help
 -l ADDRESS, --listen ADDRESS
     make coordinator listen on host and port
+--secure
+    enable TLS gRPC channel
+--cert
+    path to TLS certificate (in PEM format)
+--key
+    path to TLS key (in PEM format)
```
gastmaier: On the coordinator side, the tls at the reverse proxy is also an option, right?
At servers, I would rather have nginx handling the certificates.

If so, could you add an alternative, such as
```

Or use a reverse proxy to add TLS, for example with ``nginx``:

.. code-block:: nginx
    server {
        listen 20407 ssl http2;
        server_name labgrid.example.com;
        ssl_certificate     /etc/ssl/labgrid-coordinator.crt;
        ssl_certificate_key /etc/ssl/labgrid-coordinator.key;
        location / {
            grpc_pass grpc://127.0.0.1:20408;
        }
    }
```

https://blog.nginx.org/blog/nginx-1-13-10-grpc

labgrid/remote/common.py
```diff
@@ -58,6 +62,59 @@ def build_dict_from_map(m):
...
+        if p.returncode != 0 or not p.stdout:
+            return None
+        return p.stdout
+    except Exception:
+        logging.exception("unexpected error when fetching certificates from macOS Keychain")
+
+    return None
+
+
+def _fetch_root_certificates_linux():
```
gastmaier, line 80: On why this is needed.

the certificate precedence is (src/core/credentials/transport/tls/ssl_utils.cc @ ComputePemRootCerts) 
```
GRPC_DEFAULT_SSL_ROOTS_FILE_PATH
UseSystemRootsOverLanguageCallback()
ssl_roots_override_cb <= never reached
LoadSystemRootCerts()
installed_roots_path fallback
```
the cb is set at src/python/grpcio/grpc/_cython/cygrpc.pyx
added at github.com/grpc/grpc/commit/fa6cad701c7993aa6e5746824931efbfca84ca24

The only options are indeed GRPC_DEFAULT_SSL_ROOTS_FILE_PATH or as the argument value.

the grpc defaults for linux are (https://github.com/grpc/grpc/blob/master/src/core/credentials/transport/tls/load_system_roots_supported.cc#L48-L62)
```
    "/etc/ssl/certs/ca-certificates.crt", "/etc/pki/tls/certs/ca-bundle.crt",
    "/etc/ssl/ca-bundle.pem", "/etc/pki/tls/cacert.pem",
    "/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem"
```
I wouldn't have one of the default here, but just document that it will default to the bundled roots.pem if not explicitly set through the env or as the argument value.

doc/getting_started.rst
```diff
@@ -435,6 +435,21 @@ Follow these instructions to install the systemd files on your machine(s):
...
+the ``--secure`` (and ``--cert`` if the certificate is not trusted by the host
+machine) option.
+Refer to the ``labgrid-client`` and ``labgrid-exporter`` man pages for details.
+
```
gastmaier, line 457: explain roots.pem and GRPC_DEFAULT_SSL_ROOTS_FILE_PATH  precedence here, please

gastmaier (commented):
Looks good! Some minor comments

labgrid/remote/common.py
asher-pem-arm: Hi @gastmaier I think the important Python bit here is that grpcio installs a roots override callback which loads its bundled `roots.pem`, so on Python the normal default is not the Linux system trust store. That means if we do not pass `root_certificates`, `--tls` may ignore CAs installed on the host.

So I think the correct ordering for Labgrid should be:
- explicit `--cert`
- system trust store loaded by Labgrid
- gRPC/Python default only as fallback if we cannot load system roots

I'll update the getting_started docs to represent this

asher-pem-arm:
@Emantor I have added 3 commits for the comments in the PR, if you've happy with these I'll squash all 3 into the first commit to keep the PR tidy

doc/getting_started.rst
asher-pem-arm: updated

doc/man/coordinator.rst
asher-pem-arm: included

labgrid/remote/common.py
gastmaier: `--tls may ignore CAs installed on the host.`
Due to how the gprc source code is structured, it *will* ignore CAs installed on the host (never read) if roots.pem is present.
and roots.pem is always present with the pip package install:
```
find . -name 'roots.pem'
./lib/python3.14/site-packages/grpc/_cython/_credentials/roots.pem
```
Hacking `rm ./lib/python3.14/site-packages/grpc/_cython/_credentials/roots.pem` as a post-install step for grpcio is not an option.

gastmaier (approved):
Sounds good!

Users have the option to pass explicitly or read gprc doc for the env alternative `GRPC_DEFAULT_SSL_ROOTS_FILE_PATH`.

Emantor:
> @Emantor I have added 3 commits for the comments in the PR, if you've happy with these I'll squash all 3 into the first commit to keep the PR tidy

Looks good, please squash.

doc/man/coordinator.rst
```diff
@@ -25,9 +25,41 @@ OPTIONS
...
 -d, --debug
     enable debug mode
 
+TLS WITH A REVERSE PROXY
+------------------------
+
+Instead of enabling TLS in ``labgrid-coordinator`` directly, a reverse proxy can
+terminate TLS and forward cleartext gRPC to the coordinator. For example, with
+``nginx``:
```
jluebbe, line 42: In which cases would a reverse proxy preferred over having the coordinator do TLS itself?

labgrid/remote/common.py
Emantor: I would like the outcome of this whole investigation to be documented in the commit message.

doc/getting_started.rst
```diff
@@ -435,6 +435,26 @@ Follow these instructions to install the systemd files on your machine(s):
...
+This can be enabled on the ``labgrid-coordinator`` by adding the ``--tls``,
+``--cert`` and ``--key`` options.
+Refer to the ``labgrid-coordinator`` man page for details.
+
+When you are connecting with ``labgrid-client`` or ``labgrid-exporter`` to a
+``labgrid-coordinator`` that has TLS gRPC channels enabled you need to pass
+the ``--tls`` option. If ``--cert`` is not set, labgrid uses the host CA
+certificates to verify the coordinator certificate. Use ``--cert`` to provide
+a specific CA certificate instead.
```
jluebbe: For the client side, the usual approach is to use `--cacert` (like curl) to set the trusted CA certificates to check against. This leaves `--cert` free to allow future extension to client certificates.

labgrid/resource/remote.py
```diff
@@ -16,13 +17,47 @@ def __attrs_post_init__(self):
...
+
+    def _get_credentials(self):
+        from ..remote.common import get_client_credentials
+
+        tls = os.environ.get("LG_COORDINATOR_TLS") is not None
+        cert = None
+
+        if self.env:
+            config = self.env.config
+            tls = config.get_option("coordinator_tls", tls)
```
Emantor, line 27: The conversion to boolean should happen here, not in a separate `_is_tls_option_enabled` function.

Emantor (changes requested):

doc/man/coordinator.rst
asher-pem-arm: I think that depends on the deployment environment. I would expect the default option to be the direct coordinator TLS but some orgs might standardise TLS at the ingress layer, even then they might want to use the TLS enabled coordinator on top of that

labgrid/remote/common.py
asher-pem-arm: updated

doc/getting_started.rst
asher-pem-arm: updated

labgrid/resource/remote.py
asher-pem-arm: updated

asher-pem-arm:
updated with a new commit for supporting external clients with the client stream, will squash the latest 2 when you're both happy with the content

Emantor (changes requested):
IMO this looks good, two things:
1. The first commit should have a proper commit message instead of the commit message for the work done in the second commit.
2. The second commit should have the commit message of the first.

Also please squash, the `kw_only=True` commit for compatibility makes sense to me.

Emantor (approved):

labgrid/remote/common.py
```diff
@@ -58,6 +62,59 @@ def build_dict_from_map(m):
...
+    if platform.system() == "Darwin":
+        return _fetch_root_certificates_darwin()
+
+    if platform.system() == "Linux":
+        return _fetch_root_certificates_linux()
+
+    return None
+
+
+def get_client_credentials(args: Namespace) -> Optional[grpc.ChannelCredentials]:
```
Emantor: This should take `tls` and `cacert` as explicit arguments instead of taking a `Namespace`. This will become API and will be harder to replace later.

labgrid/resource/remote.py
```diff
@@ -16,13 +17,44 @@ def __attrs_post_init__(self):
...
+                cacert = config.resolve_path(str(cacert))
+            else:
+                cacert = None
+
+        args = Namespace(
+            tls=tls,
+            cacert=cacert,
+        )
+
+        return get_client_credentials(args)
```
Emantor: Not taking a `Namespace` here will make the `get_client_credentials()` easier to understand.

Emantor (changes requested):

labgrid/remote/client.py
```diff
@@ -2346,7 +2378,9 @@ def main():
             logging.debug('Starting session with "%s"', coordinator_address)
             loop = asyncio.new_event_loop()
             asyncio.set_event_loop(loop)
-            session = start_session(coordinator_address, extra=extra, debug=args.debug, loop=loop)
+            session = start_session(
+                coordinator_address, extra=extra, credentials=get_client_credentials(args), debug=args.debug, loop=loop
+            )
```
Bastian-Krause, line 2403: We need to look up `env.config.get_option("coordinator_tls")` and `env.config.get_option("coordinator_cacert")` here (similar to `coordinator_address` above) to start the session with/without tls.

Bastian-Krause (changes requested):

jluebbe (commented):
With the changes requested by @Emantor and @Bastian-Krause fid, I'd be fine with this.

labgrid/remote/client.py
asher-pem-arm: updated

labgrid/resource/remote.py
asher-pem-arm: updated

labgrid/remote/common.py
asher-pem-arm: updated

Emantor (approved):

Bastian-Krause (approved):

PR #1886, opened 2026-06-04
Exporter error propagation

asher-pem-arm:
Forward acquire/release failure reasons through the coordinator to the client.
This makes coordinator and exporter failures visible to users instead of returning only a generic lock/acquire failure.
The MR is split into focused commits:

report coordinator-side acquire/release failures to the client
preserve exporter failure reasons
report exporter command timeouts
reduce exported resource details in user-facing errors

Without this, it is hard to tell whether a failure happened in the client, coordinator, exporter.

### DIFF
--- a/labgrid/remote/coordinator.py
+++ b/labgrid/remote/coordinator.py
@@ -205,7 +205,11 @@ class ExporterCommand:
                 self.expired = True
 
 
-class ExporterError(Exception):
+class CoordinatorError(Exception):
+    pass
+
+
+class ExporterError(CoordinatorError):
     pass
 
 
@@ -650,9 +654,12 @@ class Coordinator(labgrid_coordinator_pb2_grpc.CoordinatorServicer):
         request.place_name = place.name
         cmd = ExporterCommand(request)
         self.get_exporter_by_name(resource.path[0]).queue.put_nowait(cmd)
-        await cmd.wait()
+        try:
+            await cmd.wait()
+        except asyncio.TimeoutError as e:
+            raise ExporterError("timed out waiting for exporter while acquiring resource") from e
         if not cmd.response.success:
-            raise ExporterError(f"failed to acquire {resource} ({cmd.response.reason})")
+            raise ExporterError(cmd.response.reason or "exporter returned failure without a reason")
         if resource.acquired != place.name:
             logging.warning("resource %s not acquired by this place after acquire request", resource)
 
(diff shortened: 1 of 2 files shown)


Emantor (changes requested):
The second commit message will need an actual message instead of the title line and it may be beneficial to squash some commits.

I would like some test infrastructure for these new changes, but this requires rework of the coordinator/exporter/client testing in the testsuite first, which is out of scope for this PR.

Implementation wise this looks good and should improve our error reporting to the user.

asher-pem-arm:
@Emantor happy to squash down to one commit if you think fits better?

Emantor:
Yes, squashing to one commit should be fine.

Emantor (dismissed):

jluebbe:
Could you update the `test_place_acquire_broken` to check that we actually see the propagated reason in the output from `labgrid-client`?

Bastian-Krause:
I guess we could then drop this completely?
https://github.com/labgrid-project/labgrid/blob/cb21e4b9b02011cf587a8959255ac867e8f05fbc/labgrid/remote/client.py#L769-L784

asher-pem-arm:
@jluebbe/ @Bastian-Krause have updated to drop the client failures and updated the `test_place_acquire_broken`

labgrid/remote/coordinator.py
```diff
@@ -665,33 +672,38 @@ async def _acquire_resources(self, place, resources):
         # all resources need to be free
         for resource in resources:
             if resource.acquired:
-                return False
+                raise CoordinatorError(f"{resource} is already acquired by {resource.acquired}")
```
Emantor, line 675:
```suggestion
                raise CoordinatorError(f"{resource} is already acquired by place '{resource.acquired}'")
```

Maybe we can also add the exporter groupname & resource name here? That way the error message does not change as much on the client side.

Emantor (changes requested):

labgrid/remote/coordinator.py
asher-pem-arm: @Emantor updated

asher-pem-arm:
FYI that we have found a edge case while testing this, we will update with commits to fix that

asher-pem-arm:
`remote/coordinator: propagate exporter failures` made exporter errors visible to clients, but the release path also needed to leave state consistent and allow retries. 
Before these 2 commits, resources were removed from the place before the exporter confirms release. If release failed, the exporter may still hold the resource, but a retry no longer included it.

The two follow-up commits address this:
- `remote/coordinator: publish place on release failure` publishes the updated place and schedules saving before returning the error, so clients see the changes even when only some resources were released.

- `remote: fix release resource state handling` keeps failed releases attached to the place for retry. It also restores the exporter’s acquisition state if cleanup fails, while preserving any broken state set during cleanup. Releasing an already-free resource succeeds and republishes its state, allowing retries after a timeout where release actually completed. Orphaned, broken, or disconnected resources can be removed from the place without blocking release.

These changes make the reported errors actionable, users can see the remaining resources and retry their release.

PR #1887, opened 2026-06-04
Fix issues that prevent using RawNetworkInterfaceDriver's network namespace feature

nick-potenski:
**Description**

Fixes several issues encountered when trying to use RawNetworkInterfaceDriver's setup_netns() method.

First, older version of iproute2 do not include the `-echo` option.  When trying to use `labgrid-raw-interface` on a exporter running Ubuntu 22, this would lead to failures when trying to create the macvtap.

Second, the `-m` mounts option was still being passed to the `nsenter` call by the netns agent script even though mount namespace support was dropped by bcfed98fbc2618a356fedc4513a152cca734e800.  This would result in failure like the following when trying to setup the namespace:
```
nsenter: reassociate to namespace 'ns/mnt' failed: Operation not permitted
``` 

**Checklist**
- [X] PR has been tested

### DIFF
--- a/helpers/labgrid-raw-interface
+++ b/helpers/labgrid-raw-interface
@@ -58,17 +58,28 @@ def open_tap(name, device="/dev/net/tun"):
     return fd
 
 
+def find_unused_macvtap_name():
+    """Find an unused macvtap interface name by checking existing interfaces."""
+    output = subprocess.check_output(["ip", "-j", "link", "show"])
+    existing = {iface["ifname"] for iface in json.loads(output)}
+    for i in range(0, 1000):
+        name = f"macvtap{i}"
+        if name not in existing:
+            return name
+    raise RuntimeError("Could not find an unused macvtap interface name")
+
+
 def handle_ns_macvtap(options):
     # Use macvtap in bridge mode. This is more consistent than using hairpin
     # mode and relying on a switch to relay packets back into the interface.
-    output = subprocess.check_output(
-        ["ip", "-j", "-echo", "link", "add", "link", options.ifname, "type", "macvtap", "mode", "bridge"]
-    )
-    output = json.loads(output)
-    assert len(output) == 1
-    ifindex_new = output[0]["ifindex"]
-    ifname_new = output[0]["ifname"]
-    assert ifname_new.startswith("macvtap")
+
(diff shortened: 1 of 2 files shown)


nick-potenski:
With these changes, I am now able to setup and use a network namespace in our Labgrid system.

I didn't file bug reports for these, but I can do so for tracking if you would like.

helpers/labgrid-raw-interface
```diff
@@ -58,17 +58,26 @@ def open_tap(name, device="/dev/net/tun"):
...
+    for i in range(0, 1000):
+        name = f"macvtap{i}"
+        if name not in existing:
+            return name
+    raise RuntimeError("Could not find an unused macvtap interface name")
+
+
 def handle_ns_macvtap(options):
     # Use macvtap in bridge mode. This is more consistent than using hairpin
     # mode and relying on a switch to relay packets back into the interface.
```
Emantor, line 74: We should add a TODO here to remind us to simplify the function again.

Emantor (dismissed):
Looks good to me aside from the nit.

helpers/labgrid-raw-interface
nick-potenski: Done.

Emantor (approved):

nick-potenski:
I would appreciate if these changes could be added to the 26.x branch as well.

nick-potenski:
@jluebbe - Did you have any feedback on this change?

Emantor (commented):
Version detection would make this more complicated and the change makes it racy (another user could use the found index before we create the device), but we could merge this as a holdover for the next 8 months. Afterwards Ubuntu 22.04 is EOL (unless you use extended LTS support), so we would revert this then. Does this match your expectations?

nick-potenski:
Yes, we have extended LTS support, but I think reverting once 22 is EOL makes sense for the Labgrid project.

nick-potenski:
Hi @Emantor  - is there any timeline for when we can have this picked to stable-26 branch?  Thank you for your consideration.

Emantor:
I'll try to get the picks done tomorrow, but will be on vacation next week and might not manage to draft the release notes for a release tomorrow, we'll see.

nick-potenski:
Release notes aren't a blocker for me; just the change on a stable branch. :)

Emantor:
The `stable-26.0` branch has been updated.

PR #1888, opened 2026-06-04
Auth: Limit places administration to local

AiyionPrime:
**Description**

**Checklist**
- [ ] Tests for the feature 

- [ ] PR has been tested

### DIFF
--- a/labgrid/remote/coordinator.py
+++ b/labgrid/remote/coordinator.py
@@ -5,6 +5,8 @@ import asyncio
 import traceback
 from enum import Enum
 from functools import wraps
+from ipaddress import ip_address
+from os import environ
 import time
 from contextlib import contextmanager
 import copy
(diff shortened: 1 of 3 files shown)


AiyionPrime:
@Emantor is there a branch I could take for inspiration to get this reviewed?

labgrid/remote/coordinator.py
```diff
@@ -182,6 +198,20 @@ async def wrapper(self, *args, **kwargs):
     return wrapper
 
 
+def auth_required(func):
+    @wraps(func)
+    async def decorated(self, request, context):
+        if environ.get("AUTH") == "LOCAL_ADMIN":
```
jluebbe, line 204: We should really stop adding more and more env variables to configure the exporter and reuse the existing configuration yaml for non-resource configuration as well.

@Emantor I think you mentioned that you had an Idea how that could look like?

AiyionPrime: For config related to deployment environments I tend to follow https://12factor.net/config

But if you'd rather have this in a config, I'd follow the wish :)

PR #1892, opened 2026-06-09
Add QEMUDriver.interact() and qemu-run.py example

Bastian-Krause:
**Description**
Since the QEMUDriver can't be used with labgrid-client (because it does not use labgrid's remote infrastructue), there is currently no good way of accessing its console interactively, e.g. for debugging.

Add a `QEMUDriver.interact()` method that creates a PTY, let microcom open it and relay the QEMUDriver's input and output to that PTY.

Add an example helper to run QEMU, either by activating the QEMUDriver directly or by transitioning into a strategy state first, then start an interactive console.

**Checklist**
- [x] PR has been tested

### DIFF
--- a/examples/qemu-run.py
+++ b/examples/qemu-run.py
@@ -1,0 +1,59 @@
+#!/usr/bin/env python3
+
+import argparse
+import logging
+import os
+
+from labgrid.logging import basicConfig, StepLogger
+from labgrid import Environment
+
+
+def main(args):
+    basicConfig(level=logging.DEBUG if args.verbose else logging.INFO)
+    StepLogger.start()
+
+    env = Environment(config_file=args.config)
+    target = env.get_target()
+    qemu = target.get_driver("QEMUDriver", activate=False)
+
+    if args.state:
+        strategy = target.get_driver("Strategy")
+        strategy.transition(args.state)
+    else:
+        print("No state given, skipping Strategy state transition")
+        target.activate(qemu)
+        qemu.on()
+
+    qemu.interact()
+
+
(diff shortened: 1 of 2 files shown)


Emantor (approved):

a3f:
I just gave it a try and wondering if there is a good way to terminate microcom when qemu exits?
Also, it's not possible to just spawn qemu with this, unless pytest has been called with -s in the first place. Could that be solved somehow? FWIW, here is what I did before:

```python
cmd = self.qemu.get_qemu_base_args()

with open("/dev/tty", "r+b", buffering=0) as tty:
    tty.write(bytes("Run manually with: \n{} -serial mon:stdio\n".format(quote_cmd(cmd)), "UTF-8"))
    if state == "qemu_dry_run":
        pytest.exit('Dry run. Terminating.')
    subprocess.run(cmd, stdin=tty, stdout=tty, stderr=tty)
```

Bastian-Krause:
> I just gave it a try and wondering if there is a good way to terminate microcom when qemu exits?

How/why does QEMU exit? The QEMUDriver starts QEMU on driver activation and stops it on deactivation.

> Also, it's not possible to just spawn qemu with this, unless pytest has been called with -s in the first place. Could that be solved somehow?

Unfortuntely, no. `-s`/`--capture=no` is the only way of enabling stdin with pytest, see https://docs.pytest.org/en/latest/how-to/capture-stdout-stderr.html#default-stdout-stderr-stdin-capturing-behaviour.

This seems to be a design decision: https://github.com/pytest-dev/pytest/issues/2189

See also #447.

PR #1791, opened 2025-12-30
util/agentwrapper: Implement _labgrid-agent

AiyionPrime:
> a helper to execute agents on the exporter host.

**Description**

This hasn't been tested yet, but is a mere discussion point.

The thesis: On a host where labgrid has been installed and the exporter is supposed to run 
we already have all the dependencies installed (at least in an environment, not necessarily an isolated one).

This means it should not only not be necessary to copy over the agent file, as its already present on the exporter host,
but not necessary to install dependencies globally on it either.

This introduces a new command `_labgrid-agent`, which due to the leading underscore should not confuse users using autocompletion.

**Considerations**

The approach has a considered drawback:
Currently the client uses the `agentwrapper` and the `agents/` shipped with the client matching its version perfectly.
With this change it would only use the latter and potential upcoming incompatibilities in the wrapper could show up, in case somebody installed different versions of labgrid on the client- and the exporterhost.

I think the risk is low though, as the wrapper does not change much anymore and a version check could be introduced, once its behaviour is changed in a breaking way.

**Checklist**
- [ ] Documentation for the feature
- [ ] Tests for the feature 

- [ ] The arguments and description in doc/configuration.rst have been updated

- ~Add a section on how to use the feature to doc/usage.rst~

- [ ] Add a section on how to use the feature to doc/development.rst

- [ ] PR has been tested

- [ ] Man pages have been regenerated

Fixes #1790

PR #1894, opened 2026-06-10
agentwrapper: try labgrid-python3 for agent environment

Emantor:
**Description**
Similar to the prefix, allow setup of the python interpreter to allow users to use their own python environments.

**Checklist**
- [x] PR has been tested

### DIFF
--- a/labgrid/util/agentwrapper.py
+++ b/labgrid/util/agentwrapper.py
@@ -60,8 +60,14 @@ class AgentWrapper:
                 ['rsync', '-e', ' '.join(ssh_opts), '-tq', agent,
                  f'{host}:{agent_remote}'],
             )
+            agent_python = "labgrid-python3" if not subprocess.call(ssh_opts + [host, '--', 'which', 'labgrid-python3']) else "python3"
+            if agent_python == "python3":
+                self.logger.debug(
+                    "labgrid-python3 on %s not found, using python3 which requires system installed python modules for agents",
+                    host
+                )
             self.agent = subprocess.Popen(
-                ssh_opts + [host, '--', 'python3', agent_remote],
+                ssh_opts + [host, '--', agent_python, agent_remote],
                 stdin=subprocess.PIPE,
                 stdout=subprocess.PIPE,
                 start_new_session=True,
--- a/tests/test_agent.py
+++ b/tests/test_agent.py
@@ -1,6 +1,7 @@
 import os
 import socket
 import subprocess
+import shutil
 
 import pytest
 from py.path import local
@@ -18,7 +19,7 @@ def subprocess_mock(mocker):
     agent = local(labgrid.util.agentwrapper.__file__).dirpath('agent.py')
 
     def run(args, **kwargs):
-        assert args[0] in ['rsync', 'ssh']
+        assert args[0] in ['rsync', 'ssh', 'which']
         if args[0] == 'rsync':
             src = local(args[-2])
             assert src == agent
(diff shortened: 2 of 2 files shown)


jluebbe:
For context, other related work is in #1782 & #1791 and another issue is #1839.

jluebbe:
The comments to https://github.com/labgrid-project/labgrid/pull/1782 also apply to this PR.

How about we probe whether `labgrid-python3` is in path on the exporter, and if so, use that? Then a user just needs to set up the venv (or whatever else) to point to the proper interpreter **on that exporter**.

Emantor:
Updated to probe for `labgrid-python3` and use that with fallback and debug message if plain `python3` is used

labgrid/util/agentwrapper.py
```diff
@@ -60,20 +61,31 @@ def __init__(self, host=None):
...
-                ssh_opts + [host, '--', 'python3', agent_remote],
+                ssh_opts + [host, '--', agent_python, agent_remote],
                 stdin=subprocess.PIPE,
                 stdout=subprocess.PIPE,
                 start_new_session=True,
             )
         else:
             # run locally
             self.fdpass, remote_fdpass = socket.socketpair()
+            agent_python = "labgrid-python3" if not shutil.which("labgrid-python3") else "python3"
```
jluebbe:
```suggestion
            agent_python = "labgrid-python3" if shutil.which("labgrid-python3") else "python3"
```

jluebbe: Is this actually needed? A virtualenv usually ensures that subprocesses still execute in the same venv. Looking for `labgrid-python3` would running a different interpreter.

jluebbe (changes requested):

labgrid/util/agentwrapper.py
Emantor: Yes and that will make it easier to support NixOS. Agents in parts use different modules (udisks i.e. uses `gi`) than required for plain labgrid and this way this can be unified across multiple machines.

jluebbe: The agents running locally in setups without coordinator/exporters need the same dependencies locally as the exporter would. So, the "client"/"library" side labgrid dependencies should also include all of those needed by agents.

I would find it very confusing, if when I'm working in a venv, agents would use a different python interpreter/installation.

Emantor: They would use it only if it exists, but this may break on systems where it is available and than NOT using the local venv. Right, I reverted that change.

I also peeked at the required dependencies for the agent, which are currently already required by labgrid (`pyusb`) or require coordination with distribution packages (`gobject-introspection`), so there is not need to modify the requirements list.

jluebbe (approved):

nvincent-vossloh:
Hello, I was notified of this PR through #1418.

I am not sure to understand how I should use this new feature.
Should I create a new python virtual env (`/opt/labgrid-python3-venv/` for instance) and then add a symlink to `/opt/labgrid-python3-venv/bin/python3` to a folder in the `PATH` ? 
Or is it something else ?

Thanks in advance.

Emantor:
You will need a virtual environment and than make that available to the user who connects to the exporter under the `labgrid-python3` name in the users `PATH`. So yes, `/usr/bin/labgrid-python3` → `/opt/labgrid-python3-venv/bin/python3` with a symlink should work. If it does not, debug logging should give you an indication whether `labgrid-python3` was correctly found.

PR #1897, opened 2026-06-17
doc: Rockchip: fix rk-usb-loader reference

saschahauer:
The rk-usb-loader option incorrectly references the barebox rk-usb-loader tool. The RKUSBDriver is not compatible with the barebox tool and instead works with the rkdeveloptool. Update the reference accordingly.

**Description**

**Checklist**
- [ ] Tests for the feature 

- [ ] PR has been tested


saschahauer:
Created PR to add a new driver which actually uses rk-usb-loader which adjusts the reality to the documentation

PR #1899, opened 2026-06-18
util/proxy: allow overriding LG_PROXY from the environment config

saschahauer:
The coordinator_address can already be set in the environment config, overriding the LG_COORDINATOR environment variable. But LG_COORDINATOR is only one part of the connection settings needed to reach a coordinator: the LG_PROXY variable, used to tunnel that connection over SSH, had no config counterpart and could only be set from the environment or the --proxy command line argument.

Add a 'proxy' option to the environment config that overrides LG_PROXY, mirroring coordinator_address. The precedence is --proxy command line argument > 'proxy' config option > LG_PROXY environment variable.

A present but empty/false 'proxy' option forces a direct connection without a proxy, even when LG_PROXY is set, which previously was not possible at all.

Assisted-by: Claude Opus 4.8

**Description**

**Checklist**
- [ ] Tests for the feature 

- [ ] PR has been tested

### DIFF
--- a/doc/man/device-config.rst
+++ b/doc/man/device-config.rst
@@ -32,6 +32,8 @@ previous driver.
 For a list of available resources and drivers refer to
 https://labgrid.readthedocs.io/en/latest/configuration.html.
 
+.. _labgrid-device-config-options:
+
 OPTIONS
 -------
 The ``options:`` top key configures various options such as the coordinator_address.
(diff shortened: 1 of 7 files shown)


Bastian-Krause:
An environment config describes the test environment. This includes the place name and (with some imagination) also the coordinator which holds the corresponding place definition.

I don't think a proxy should be part of the test environment. For the test environment, it should not matter from where you execute the tests and whether you need a proxy or what machine should act as a proxy.

saschahauer:
> An environment config describes the test environment. This includes the place name and (with some imagination) also the coordinator which holds the corresponding place definition.

Having the coordinator in the config is quite useful once you have multiple coordinators that you want to connect to on a single machine.

> 
> I don't think a proxy should be part of the test environment. For the test environment, it should not matter from where you execute the tests and whether you need a proxy or what machine should act as a proxy.

The mere presence of a LG_PROXY variable prevented me from thinking about what I would do without it. I helped myself with a ssh LocalForward now, so no need for LG_PROXY anymore for me

PR #1901, opened 2026-06-19
util/helper: fix process output getting truncated

zachdeibert:
**Description**

A process that logs a lot of data before exiting could have its output truncated because ProcessWrapper used to stop reading the pipe once the process exits, but instead should always read to the end of the pipe.

**Checklist**

- [x] PR has been tested


Bastian-Krause (approved):
Thanks for the PR, looks good to me!

PR #1905, opened 2026-06-29
tests: disable gRPC fork handlers to fix flaky fork() abort

Bastian-Krause:
**Description**
The test suite intermittently fails in CI due to (pexpect spawned) labgrid-client being killed during teardown by SIGABRT and gRPC C-core aborting after a fork:

```
fork_posix.cc:71] Other threads are currently calling into gRPC, skipping fork() handlers
ev_epoll1_linux.cc:1125] Check failed: next_worker->state == KICKED
```

gRPC documents that fork() is only safe if gRPC objects are created in the child after the fork [1]. The tests do the opposite: the pytest process spawns the labgrid-client commands via pexpect (using forkpty) while gRPC is already initialized in it (imported at collection, and used in-process by the coordinator tests). To partially support this, gRPC installs a pthread_atfork handler that tries to quiesce its poller across the fork. When another thread is in the C-core at fork time the handler loses that race, skips the reset, and the forked child aborts on the inconsistent poller before it can exec the client.

The problem could never be reproduced locally; it likely occurs more often in CI because scheduling pressure makes the race more likely to trigger.

Set GRPC_ENABLE_FORK_SUPPORT=0 to disable the handler. This is the documented remedy for callers that always exec() after fork() and never reuse a channel across fork(), which is what the test suite does. Set it in conftest.py before pytest's collection imports grpc.

[1] https://github.com/grpc/grpc/blob/master/doc/fork_support.md

**Checklist**
- [x] PR has been tested

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,4 +1,5 @@
 import logging
+import os
 from signal import SIGTERM
 import sys
 import threading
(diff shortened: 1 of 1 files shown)


Emantor (approved):

Emantor:
I looked at this briefly on friday too when wondering about the test failures, thanks for the deep dive and fixing the tests :+1:

Bastian-Krause:
Although this is not really material for "stable", having a stable CI (pun intended) would be worth it :)

PR #1916, opened 2026-06-30
tests: enable subprocess coverage

Emantor:
Subprocess coverage used to be provided by default by pytest-cov, however that was removed in newer versions in favor of using the coverage patch functionality. Use the coverage patch functionality to re-enable coverage of subprocesses, re-enabling coverage for the remote tests.

**Checklist**
- [x] PR has been tested

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,6 +1,7 @@
 [run]
 omit =
     labgrid/remote/generated/*
+patch = subprocess
 
 [report]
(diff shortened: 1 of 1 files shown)


Emantor:
`+15,2%`, not bad :grin:

Bastian-Krause (approved):
Very nice, thank you for looking into that!

PR #1918, opened 2026-07-02
Metadata identity

asher-pem-arm:
### Description
This is the first part of the coordinator unary-RPC rework. It introduces metadata-based identity for clients and exporters, reducing the coordinator’s dependency on startup messages sent through the long-lived streams.

### This PR:

- Adds client and server gRPC identity interceptors.
- Sends client/exporter identity and user-agent information as gRPC metadata.
- Initializes ClientStream and ExporterStream sessions from that metadata.
- Deprecates stream startup identity messages while retaining them as a fallback for older clients and exporters.
- Updates AcquirePlace, AllowPlace, and CreateReservation to infer caller identity from metadata or an existing stream session.
- No new coordinator RPCs are introduced in this part.

### Motivation
ClientStream currently acts both as the ordered live-update channel and as the source of caller identity for independent RPCs. This forces short-lived tools and services to establish and synchronise a stream before making simple request/response calls, adding startup latency and state-management overhead.

Carrying identity in each RPC’s metadata allows these callers to invoke the RPC, emit the result, and exit. Stateful clients still use ClientStream for ordered place and resource updates. This separation also provides call-scoped identity for future per-operation authorization; the metadata itself is not authentication.

### Compatibility note
Old clients and exporters remain compatible with the new coordinator through the startup-message fallback.
However, new clients and exporters stop sending startup identity messages and are therefore incompatible with coordinators predating this change. Those coordinators cannot establish the required stream session. 

### See also
Discussion on the gRPC unary rework: https://github.com/labgrid-project/labgrid/discussions/1881
Follow-up RPC changes: https://github.com/labgrid-project/labgrid/pull/1919
The List RPC changes: https://github.com/labgrid-project/labgrid/pull/1933
Follow on capabilities work: https://github.com/labgrid-project/labgrid/pull/1955


Emantor:
I took the liberty to reformatted for the PR template, this should make the description more readable.

labgrid/remote/grpc/interceptor/client.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def _inject(self, client_call_details: ClientCallDetails):
+        if self.username:
+            client_call_details.metadata.add(USERNAME_KEY, self.username)
+        client_call_details.metadata.add(HOSTNAME_KEY, self.hostname)
+        if self.user_agent:
+            client_call_details.metadata.add(USER_AGENT_KEY, self.user_agent)
+
+
+class IdentityClientUnaryUnaryInterceptor(UnaryUnaryClientInterceptor, BaseIdentityClientInterceptor):
```
Emantor, line 23: Do we really need to child classes here? To me it looks like these only differ in parameters, and this won't matter for python as long as the number of parameters remain the same.

It would be nice if we can simplify this to one class that implements the logic.

Emantor (changes requested):

labgrid/remote/grpc/interceptor/client.py
asher-pem-arm: `grpc.aio.Channel` sorts interceptors into per-RPC-type lists using `isinstance()`, so we would still need separate unary-unary and stream-stream interceptor classes.

Emantor: But we could just have a `IdentityClientInterceptor(UnaryUnaryClientInterceptor, StreamStreamClientInterceptor)` or is there something baring that on the gRPC side?

Edit: Does not look like it, the gRPC example uses a similar pattern: https://github.com/grpc/grpc/blob/b86a3460af2160cd7a02bd51014bc9fa036b91ff/examples/python/interceptors/default_value/default_value_client_interceptor.py#L48

labgrid/remote/coordinator.py
```diff
@@ -30,6 +35,10 @@
 from .generated import labgrid_coordinator_pb2_grpc
 from ..util import atomic_replace, labgrid_version, yaml, Timeout
 
+client_identity_context: contextvars.ContextVar[Optional[ClientIdentity]] = contextvars.ContextVar(
+    "client_identity", default=None
+)
```
Emantor, line 41: Binding the context globally means that the `client_identity_context` will have the value of the last call before it is changed by the next RPCs interceptor. IMO this should at least be bound to the `Coordinator` class or be an ephemeral object that only exists for the RPC call.

Emantor (changes requested):

labgrid/remote/coordinator.py
asher-pem-arm: `client_identity_context` is a `ContextVar`; gRPC’s AsyncIO interceptor docs say each RPC starts with an empty contextvars.Context, and that same context is used for the final handler. https://docs.python.org/3/library/contextvars.html#context-variables also recommend creating  `ContextVar` at module scope

Emantor: Yes, each RPC starts with an empty `self.context` in the grpc `_RPCState` class, however this does not clear the explicitly passed in context variable we passed in from the global coordinator module level, so shouldn't any state saved in that variable be retained across RPCs since we don't explicitly clear it?

asher-pem-arm: I think the key distinction is that the module level `ContextVar` object is not the stored identity value. It is the key used to look up a value in the current `contextvars.Context`. 
i.e. try testing:
```python3
var = ContextVar("id", default=None)

ctx_a = Context()
ctx_a.run(var.set, "test")

ctx_b = Context()
assert ctx_b.run(var.get) is None
assert ctx_a.run(var.get) == "test"
```

Emantor: But as far as I can see none of the code is using a specific created `Context()` object to access the contextvariable, To me this looks more like this:
```python3
>>> from contextvars import *
>>> cvar = ContextVar('c')
>>> cvar.set("cookie")
<Token var=<ContextVar name='c' at 0x7e75d07cf010> at 0x7e75d07fb640>
>>> def func_with_context():
...     c = Context()
...     print(cvar.get())
...
>>> func_with_context()
cookie
>>>
```
So my question would be where the gRPC code is using its created context in conjunction with the contextvariable we introduce here.

asher-pem-arm: Yeah so the missing bit in that example is `Context.run()`. If you add `Context().run(cvar.get)` you won't see the outer `cookie` value. In our code we do not call `Context.run()` directly because gRPC owns the RPC dispatch.

Emantor: Yeah, what was not clear to me is that I couldn't find any usage of `Context().run()` inside of gRPC. But contextvars are "magical" in conjunction with async tasks (which is rare in python), so no explicit context management is needed for the context variable, the following code makes that clear:
```python
import asyncio
import contextvars

var = contextvars.ContextVar('var', default={})

async def sub2():
    print(f'in sub2, {var.get()=}')
    var.set('sub1 set')

async def sub1():
    print(f'in sub1, {var.get()=}')
    var.set('sub1 set')
    await sub2()

async def main():
    var.set('main set')
    await sub1()
    print(f'in main, {var.get()=}')

asyncio.run(main())
print(f'after main {var.get()=}')
```
with as the output
```
in sub1, var.get()='main set'
in sub2, var.get()='sub1 set'
in main, var.get()='sub1 set'
after main var.get()={}
```
from [here](https://valarmorghulis.io/tech/202408-the-asyncio-tasks-and-contextvars-in-python/)

Emantor:
During the review of this PR yesterday it came up that we are missing a good justification for separate unary RPC instead of extending the `ClientStream` with filtering. The unary RPCs introduced in this PR duplicate the information on the stream and applications should be able to use the ClientStream for the same information.

asher-pem-arm:
Hey @Emantor ,

Thanks for sharing the concern. The primary motivation for this came out of our work on the Labgrid fleeting plugin, and the more we thought about it the more sense it made.

At the moment ClientStream is doing two jobs: it is both a watch/update mechanism and part of the session/identity for independent RPCs. That coupling is what we are trying to remove. A service, CI worker, web backend, or short-lived tool should be able to make a bounded RPC with identity/auth metadata on that RPC, get a response, and finish.
As we found on our end with our fleeting plugin implementation, requiring those clients to maintain a bidirectional stream means they also need reconnect logic, caching, sync handling, readiness state, and failure handling to safely perform simple operations. On the Labgrid fleeting plugin we managed to remove ~1000 lines of code/tests when we moved to using the unary RPC implementation.

This also matters for production/distributed deployments. If every client has to establish and maintain a ClientStream session, then coordinator endpoint changes, load balancer migration, reconnects, and horizontally scaled services all need to manage stream/session recovery. In many cases those clients gain no benefit from a long-lived session. The intent here is to decouple those concerns: clients that need ClientStream for live updates can still use it.

A stream also makes auth harder as the coordinator has to continuously filter updates for each client and ensure no unauthorised state is ever sent. For unary RPCs, the auth decision is scoped to one operation and one response.

If you feel this is useful context I can add this to the discussion as needed. Thanks

asher-pem-arm:
@Emantor as discussed I've updated this PR to only include the client stream decoupling

labgrid/remote/coordinator.py
```diff
@@ -326,9 +326,17 @@ async def ClientStream(self, request_iterator, context):
         assert peer not in self.clients
         out_msg_queue = asyncio.Queue()
 
+        identity = client_identity_context.get()
+        if identity:
+            logging.debug("client identity provided in gRPC metadata")
+            logging.debug(identity)
+            self.clients[peer] = ClientSession(self, peer, identity.id, out_msg_queue, identity.user_agent)
+
```
Emantor, line 334: The `ClientSession` creation can be moved into the `startup` handling, new clients will sent both for a release (old coordinators use the old `StartupDone` message, new ones the identity). The coordinator uses the identity for new clients and warns for connections of older clients.

With the introduction of auth we can than abort for older clients, which are not supported with auth.

Emantor (changes requested):

labgrid/remote/coordinator.py
asher-pem-arm: updated in latest commit

labgrid/remote/grpc/interceptor/client.py
Emantor: This simplification is still missing.

Emantor (changes requested):
Looking at the test output for this PR I see that the tests seem not to use the client identity:
```bash
pytest --cov-report=term-missing --cov=labgrid tests -k coordinator -s -vv
current_level: None
============================================================================================================ test session starts =============================================================================================================
platform linux -- Python 3.13.13, pytest-9.1.1, pluggy-1.6.0 -- /home/phoenix/work/ptx/labgrid/.venv/bin/python3
cachedir: .pytest_cache
benchmark: 5.2.3 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
rootdir: /home/phoenix/work/ptx/labgrid
configfile: pytest.ini (WARNING: ignoring pytest config in pyproject.toml!)
plugins: mock-3.15.1, isort-4.0.0, dependency-0.6.1, cov-7.1.0, benchmark-5.2.3, asyncio-1.4.0, labgrid-26.1.dev27
asyncio: mode=Mode.AUTO, debug=False, asyncio_default_fixture_loop_scope=None, asyncio_default_test_loop_scope=function
collected 496 items / 478 deselected / 1 skipped / 18 selected

tests/test_coordinator.py::test_startup coordinator: INFO:root:loaded 0 place(s)
coordinator: INFO:root:Enabled channelz support
coordinator: INFO:root:Coordinator ready
coordinator: listening on [::]:20408
PASSED
------------------------------------------------------------------------------------------------------------- live log teardown --------------------------------------------------------------------------------------------------------------
INFO               root: stopping Coordinator pid=27317
coordinator: INFO:root:Starting graceful shutdown...
coordinator: INFO:root:Starting graceful shutdown...

tests/test_coordinator.py::test_coordinator_add_place coordinator: INFO:root:loaded 0 place(s)
coordinator: INFO:root:Enabled channelz support
coordinator: INFO:root:Coordinator ready
coordinator: listening on [::]:20408
coordinator: INFO:root:client connected: ipv4:127.0.0.1:42882
coordinator: WARNING:root:client ipv4:127.0.0.1:42882 did not provide identity metadata; using deprecated startup identity
```
But since this is with the client interceptor, shouldn't these use the new identity metadata?

Edit:Scratch that, the tests use a bespoke channel stub which does not add the metadata.

labgrid/remote/grpc/interceptor/client.py
asher-pem-arm: sorry missed this comment, so the example uses synchronous gRPC but we're using grpc.aio which handles interceptors differently. If one interceptor class implements both unary and streaming interfaces, grpc.aio checks them with if/elif. It recognises only the first interface, so identity is added to unary calls but not from stream-stream calls.
See:
https://github.com/grpc/grpc/blob/b86a3460af2160cd7a02bd51014bc9fa036b91ff/src/python/grpcio/grpc/aio/_channel.py#L358-L372
and from their tests:
https://github.com/grpc/grpc/blob/b86a3460af2160cd7a02bd51014bc9fa036b91ff/src/python/grpcio_tests/tests_aio/unit/client_unary_unary_interceptor_test.py#L54-L71
https://github.com/grpc/grpc/blob/b86a3460af2160cd7a02bd51014bc9fa036b91ff/src/python/grpcio_tests/tests_aio/unit/client_stream_stream_interceptor_test.py#L32-L36

Emantor: The choices of gRPC astound me every time.

Emantor (changes requested):
The code looks good to me, all thats left in my opinion is to squash the last commit.
It was nice for reviewing, but it should be cleaner to not remove and re-add `StartupDone` support from a history point of view.

jluebbe:
What is the meaning of
```
Signed-off-by: Asher Pemberton <asher.pemberton@arm.com>
Reviewed-by: Asher Pemberton <asher.pemberton@arm.com> # gatekeeper
```
? You can't add a Reviewed-by for code you've written yourself.

labgrid/remote/coordinator.py
```diff
@@ -317,9 +326,16 @@ async def ClientStream(self, request_iterator, context):
         assert peer not in self.clients
         out_msg_queue = asyncio.Queue()
 
+        identity = client_identity_context.get()
+        if identity:
+            logging.debug("client identity provided in gRPC metadata")
+            logging.debug(identity)
```
Emantor: This should be one `logging.debug()` line.

labgrid/remote/coordinator.py
Emantor: We still create a `ClientSession` outside of the `elif kind == "startup"` here, we should move it into the `elif`. `ClientSession` should only be created in one single place, this makes reasoning for remote RPC issues vastly easier.

labgrid/remote/coordinator.py
```diff
@@ -409,9 +435,22 @@ async def ExporterStream(self, request_iterator, context):
         out_msg.hello.version = labgrid_version()
         yield out_msg
 
+        identity = client_identity_context.get()
+        if identity:
+            logging.debug("exporter identity provided in gRPC metadata")
+            logging.debug(identity)
```
Emantor: This should be a single `logging.debug` line.

labgrid/remote/coordinator.py
```diff
@@ -409,9 +435,22 @@ async def ExporterStream(self, request_iterator, context):
...
+            if existing := self.get_exporter_by_name(identity.id):
+                await context.abort(
+                    grpc.StatusCode.ALREADY_EXISTS,
+                    f"startup failed: exporter with name '{identity.id}' is already connected from {existing.peer}",
+                )
+            self.exporters[peer] = ExporterSession(self, peer, identity.id, command_queue, identity.user_agent)
```
Emantor, line 443: Same as for `ClientSession` above, should be created in a single place.

labgrid/remote/identity.py
```diff
@@ -0,0 +1,65 @@
...
+    user = identity_contextvar.get()
+    if user:
+        logger.debug("identity sourced from metadata")
+        return user.id
+
+    try:
+        logger.debug("identity sourced from self.clients")
+        return clients[context.peer()].name
+    except KeyError:
+        raise
```
Bastian-Krause: No-op, drop that.

pyproject.toml
```diff
@@ -84,6 +84,7 @@ dev = [
 
     # additional dev dependencies
     "psutil>=5.8.0",
+    "pytest-asyncio==1.3.0",
```
Bastian-Krause: Why do we need to pin to that exact version?

Emantor (changes requested):

labgrid/remote/coordinator.py
jluebbe, line 41: I've been trying to understand why you chose to use a `ContextVar` instead of using the gRPC [Server Side `context` ](https://grpc.github.io/grpc/python/grpc.html#service-side-context) object passed to each API method to inspect the [`invocation_metadata`](https://grpc.github.io/grpc/python/grpc.html#grpc.ServicerContext.invocation_metadata)? It feels like the `ContextVar` adds a mechanism similar to what is already provided by the `context` parameter.

Later, on the authorization side, a function decorator could also access the `ServicerContext` parameter of the call to decide whether to allow/deny the call.

Without a `ContextVar`, we don't have easy access to the `ClientIdentity`, but as auth checks should be done early, I'm skeptical if that access is needed at all. Without the `ContextVar`, the complexity of the Interceptors wouldn't be needed either.

I'm not saying we must avoid `ContextVar` and Interceptors, but I want to understand why they are the right choice.

jluebbe (commented):
The "why" of each change should also be explained in the commit messages.

> ### Motivation
> Decouple ClientStream’s live-update role from identity and independent RPCs, allowing services and short-lived tools to perform bounded authenticated operations without maintaining stream state. This simplifies client implementations, reconnection handling, horizontal scaling, and per-operation authorization.

I'm not fully convinced by the simplification arguments.

"client implementations" might be correct for *other* implementations, but labgrid's own client becomes more complex through the introduction of the Interceptors.

"reconnection handling" only applies to stateless clients, right? Anything using labgrid to access resources **needs** a ordered and loss-less stream of resource events to keep a consistent view of a Place. Supporting reconnection would mean buffering and replaying resource events upon reconnection, and I'm not sure if the complexity required for that is worth it.

"horizontal scaling" don't seem correct to me. As long as the state is held in the coordinator process (which I don't see us changing), there can only be a single instance as this ensures consistency.

"per-operation authorization" sounds right, as we no longer need to look up the `ClientSession` by peer.

So, I'd prefer a motivation description which make sense (primarily) from the perspective of using labgrid with the in-tree exporter and client.

pyproject.toml
asher-pem-arm: updated to remove pin

labgrid/remote/identity.py
asher-pem-arm: updated

labgrid/remote/coordinator.py
asher-pem-arm: updated

labgrid/remote/coordinator.py
asher-pem-arm: updated

labgrid/remote/coordinator.py
asher-pem-arm: updated

asher-pem-arm:
> What is the meaning of
> 
> ```
> Signed-off-by: Asher Pemberton <asher.pemberton@arm.com>
> Reviewed-by: Asher Pemberton <asher.pemberton@arm.com> # gatekeeper
> ```
> 
> ? You can't add a Reviewed-by for code you've written yourself.

@jluebbe the original code was written by Luke and then between Luke and myself we've been making the changes for the PR comments, when either one of us makes an update the other reviews it but as part of our internal process I'm the gatekeeper that decides when it gets upstreamed. Hopefully doesn't confuse things too much

asher-pem-arm:
> The "why" of each change should also be explained in the commit messages.
> 
> > ### Motivation
> > Decouple ClientStream’s live-update role from identity and independent RPCs, allowing services and short-lived tools to perform bounded authenticated operations without maintaining stream state. This simplifies client implementations, reconnection handling, horizontal scaling, and per-operation authorization.
> 
> I'm not fully convinced by the simplification arguments.
> 
> "client implementations" might be correct for _other_ implementations, but labgrid's own client becomes more complex through the introduction of the Interceptors.
> 
> "reconnection handling" only applies to stateless clients, right? Anything using labgrid to access resources **needs** a ordered and loss-less stream of resource events to keep a consistent view of a Place. Supporting reconnection would mean buffering and replaying resource events upon reconnection, and I'm not sure if the complexity required for that is worth it.
> 
> "horizontal scaling" don't seem correct to me. As long as the state is held in the coordinator process (which I don't see us changing), there can only be a single instance as this ensures consistency.
> 
> "per-operation authorization" sounds right, as we no longer need to look up the `ClientSession` by peer.
> 
> So, I'd prefer a motivation description which make sense (primarily) from the perspective of using labgrid with the in-tree exporter and client.

Hey @jluebbe , thanks for the reply.

By client simplification, we primarily meant implementations such as the fleeting plugin, CI workers, and web services that need short-lived request/response operations but do not consume live place or resource updates. The in-tree client still needs `ClientStream`, but we do not think the additional interceptor plumbing makes it overly complex.

There is also a practical startup-time benefit for commands such as `labgrid-client` resources. Currently, the client establishes `ClientStream`, receives the initial state, and waits for synchronization before producing output. A client that could simply call an RPC, emit the response, and exit would be significantly quicker.

You’re right that this PR does not implement reconnection for streaming consumers. A client could rebuild its current state from a fresh snapshot followed by ordered updates, so replay is not necessarily required. However, robust stream reconnection is a separate problem and should not be claimed as a benefit here.

Likewise, horizontal scaling was intended to refer to simplifying the use of multiple caller-service instances, not multiple stateful coordinator instances. Although decoupling RPC identity from an in-process stream session could be an enabling step towards scaling coordinator instances if their state were externalised in the future.

The concrete benefit for labgrid is the separation of responsibilities: `ClientStream` remains the ordered place/resource update channel, while unary RPC identity no longer depends on a `ClientSession` looked up using the client’s network connection. This also provides call-scoped identity as input for per-operation authorization. The metadata alone, however, should not be described as authentication.

I've updated the motivation and commit messages to include this

labgrid/remote/coordinator.py
asher-pem-arm: Currently the RPCs use the caller identity within RPC handlers, eg. for assigning acquired on a Place.

We expect performing authorisation in a central location (server interceptor) is beneficial, especially as more complex auth is supported.

Storing the identity (and in the future, ideally Capabilities) in a `ContextVar` means that the identity is accessible without needing a call the parse the metadata within each RPC.

asher-pem-arm:
note:
have updated the latest commit to support metadata only clients and tested all 3 possible startup methods:
- Metadata only
- StartupDone only
- Metadata + StartupDone

jluebbe:
> > What is the meaning of
> > ```
> > Signed-off-by: Asher Pemberton <asher.pemberton@arm.com>
> > Reviewed-by: Asher Pemberton <asher.pemberton@arm.com> # gatekeeper
> > ```
> > ? You can't add a Reviewed-by for code you've written yourself.
> 
> @jluebbe the original code was written by Luke and then between Luke and myself we've been making the changes for the PR comments, when either one of us makes an update the other reviews it but as part of our internal process I'm the gatekeeper that decides when it gets upstreamed. Hopefully doesn't confuse things too much

We're following the kernel conventions here. When you develop a commit together, use:
https://docs.kernel.org/process/submitting-patches.html#when-to-use-acked-by-cc-and-co-developed-by
```
Co-developed-by: Luke ... <...>
Signed-off-by: Luke ... <...>
Signed-off-by: Asher Pemberton <asher.pemberton@arm.com>
```

Reviewed-by has no place here, as you can't review the part of the commit yourself.

jluebbe:
> You’re right that this PR does not implement reconnection for streaming consumers. A client could rebuild its current state from a fresh snapshot followed by ordered updates, so replay is not necessarily required. However, robust stream reconnection is a separate problem and should not be claimed as a benefit here.

A client which rebuilds state from a fresh snapshot after reconnecting cannot be sure that it has not missed a resource event. That would break the fundamental assumption that there are no gaps in the resource event stream. So, for the foreseeable future, any client using labgrid drivers will need a long-duration session.

This is just something I wanted to clarify and not an argument against having metadata identity.

asher-pem-arm:
@jluebbe I have updated the commit messages and rebased in the latest changes. @jluebbe/ @Emantor are there any other changes requested/review comments for this PR? Would be useful to work out what we want to do with it to save time with rebases to keep it up to date

labgrid/remote/coordinator.py
```diff
@@ -31,6 +36,10 @@
 from .generated import labgrid_coordinator_pb2_grpc
 from ..util import atomic_replace, labgrid_version, yaml, Timeout
 
+client_identity_context: contextvars.ContextVar[Optional[ClientIdentity]] = contextvars.ContextVar(
```
Emantor, line 39: We would like a removal of a context variable & server interceptor and a replacement of that with a decorator. The decorator should look like this:
```python
def identity():
    def decorator(func):
        @wraps(func)
        async def wrapper(self, request, context):
            identity = ClientIdentity.from_metadata(context.invocation_metadata())
            return await func(self, request, context, identity=identity)

        return wrapper

    return decorator
```

and reworking the function signature for gRPC calls, i.e:
```python
    @identity
    @locked
    async def AddPlace(self, request, context, *, identity):
```

See [here](https://github.com/Emantor/labgrid/blob/topic/capabilities/labgrid/remote/coordinator.py) how this looks like with capabilities on top.

The capability work will than rename the decorator and add the capability.
This makes the code more pythonic and easier to follow, since the context-variable can than be passed explicitly within functions.

Emantor (changes requested):

labgrid/remote/coordinator.py
asher-pem-arm: hey @Emantor we thought you were going to include this refactor as part of the capabilities work? Think that makes sense as you have the best context here on how the code should look

Emantor: I would like to avoid introducing the interceptor & contextvar only to refactor them out later.
I could do the rework tomorrow on top of your branch if you don't mind.

asher-pem-arm: Hey @Emantor yeah think that would be better here, thanks :)

Emantor: It looks like I am not allowed to push into your branch, see the top commit on https://github.com/Emantor/labgrid/tree/detach-streams-from-unary-rpcs-master-pr1 and please review/test.

asher-pem-arm (commented):
PR will be closed, new decorator changes covered in https://github.com/labgrid-project/labgrid/pull/1973

PR #1920, opened 2026-07-02
Add Joulescope energy analyzer driver and EnergyAnalyzerProtocol

mliberty1:
Add support for the Joulescope energy analyzer (JS110/JS220/JS320) via pyjoulescope_driver:

- New EnergyAnalyzerProtocol (get_statistics/start/stop/capture)
- New JoulescopeDevice resource: a USBResource selected via udev matching; serial and model are derived from udev (ID_SERIAL_SHORT / ID_MODEL_ID)
- New JoulescopeDriver implementing EnergyAnalyzerProtocol and PowerProtocol (downstream power switch: JS110 via s/i/range/select, JS220/JS320 via s/i/range/mode). Statistics use the documented per-model base sample rate (h/fs reads back None on JS220/JS320).
- Remote/distributed support: exporting a JoulescopeDevice makes it available to clients as a NetworkJoulescopeDevice. pyjoulescope_driver runs on the exporting host through the labgrid agent (util/agents/joulescope.py), so the driver binds to both the local and network resource and works unchanged over the coordinator/exporter infrastructure; captures are recorded on the exporting host and copied back.
- Optional dependency extra labgrid[joulescope]
- Unit tests (mocked) for the driver and the agent, plus a hardware-in-the-loop suite gated behind --joulescope, configuration docs and changelog entry

Verified hardware-in-the-loop against a JS320, both locally and via a remote coordinator/exporter.

---

## Motivation

One of our customers uses labgrid, and requested Joulescope integration.  I was previously unaware of labgrid.  On investigation, it looked easy enough to add within the existing labgrid architecture.   This customer has already confirmed that this PR works for them in local mode.

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -765,6 +765,39 @@ Arguments:
 Used by:
   - `SigrokDriver`_
 
+JoulescopeDevice
+~~~~~~~~~~~~~~~~
+A :any:`JoulescopeDevice` resource describes a *Joulescope* energy analyzer
+(JS110, JS220 or JS320).  It is a USB resource, so a specific device is selected
+via udev matching when more than one Joulescope is connected; with a single
+Joulescope an empty match is sufficient.  The device is then addressed through
+``pyjoulescope_driver``.
+
+.. code-block:: yaml
+
+   JoulescopeDevice:
+     match:
+       ID_SERIAL_SHORT: 'S3C8'
+
+You need to ensure proper udev permissions:
+
+.. code-block:: bash
+
+   wget https://raw.githubusercontent.com/jetperch/joulescope_driver/refs/heads/main/72-joulescope.rules
+   sudo cp 72-joulescope.rules /etc/udev/rules.d/
+   sudo udevadm control --reload-rules
+
+Arguments:
+  - match (dict): key and value pairs for a udev match, see `udev Matching`_
+
+Used by:
+  - `JoulescopeDriver`_
+
+NetworkJoulescopeDevice
+~~~~~~~~~~~~~~~~~~~~~~~~~
+A :any:`NetworkJoulescopeDevice` resource describes a `JoulescopeDevice`_ resource
+available on a remote computer.
+
 IMXUSBLoader
 ~~~~~~~~~~~~
 An :any:`IMXUSBLoader` resource describes a USB device in the imx loader state.
@@ -3190,6 +3223,50 @@ samples is an iterable of samples.
 
 This driver relies on buffering of the subprocess call.
 Reading a few samples will very likely work - but obtaining a lot of samples may stall.
+
+JoulescopeDriver
+~~~~~~~~~~~~~~~~
+The :any:`JoulescopeDriver` uses a `JoulescopeDevice`_ or
+`NetworkJoulescopeDevice`_ resource to measure current, voltage and power,
+accumulate charge and energy, capture high-rate samples to a JLS file, and switch
+downstream power by connecting/disconnecting the device current path.
+
+Binds to:
+  device:
+    - `JoulescopeDevice`_
+    - `NetworkJoulescopeDevice`_
+
+Implements:
+  - :any:`EnergyAnalyzerProtocol`
+  - :any:`PowerProtocol`
+
+.. code-block:: yaml
+
+   JoulescopeDriver:
+     frequency: 10.0
+     delay: 2.0
+
+Arguments:
+  - frequency (float, default=2.0): statistics update frequency in Hz
+  - delay (float, default=2.0): delay in seconds between off and on during a
+    power cycle
+
+The latest measurement is read with ``get_statistics()``, which returns a dict
+with ``current``, ``voltage`` and ``power`` sub-dicts (each with ``avg``,
+``std``, ``min`` and ``max``) plus the accumulated ``charge_C`` (Coulombs) and
+``energy_J`` (Joules).
+``start()`` and ``stop()`` bracket a charge/energy accumulation window;
+``stop()`` returns the accumulated ``energy_J``, ``charge_C`` and the
+``duration_s`` of the window.
+``capture(filename, signals=None, duration=..., frequency=None)`` records
+high-rate samples to a JLS file for the requested duration.  ``frequency`` (in
+Hz) sets the device sample rate and is sticky: once set it stays in effect for
+later captures on the same activated driver until changed again, rather than
+reverting to the device default.
+
+Power switching via ``on()``, ``off()`` and ``cycle()`` connects and
+disconnects the device current path, controlling downstream power to the
+device under test.
 
 USBSDMuxDriver
 ~~~~~~~~~~~~~~
--- a/labgrid/driver/joulescopedriver.py
+++ b/labgrid/driver/joulescopedriver.py
@@ -1,0 +1,130 @@
+import os
+import time
+import uuid
+
+import attr
+
+from ..factory import target_factory
+from ..protocol import EnergyAnalyzerProtocol, PowerProtocol
+from ..resource.joulescope import JoulescopeDevice
+from ..resource.remote import NetworkJoulescopeDevice
+from ..step import step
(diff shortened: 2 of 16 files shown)


doc/configuration.rst
```diff
@@ -765,6 +765,55 @@ Arguments:
...
+Joulescope an empty match is sufficient.  The device is then addressed through
+``pyjoulescope_driver``.
+
+.. code-block:: yaml
+
+   JoulescopeDevice:
+     match:
+       ID_SERIAL_SHORT: 'S3C8'
+
+Arguments:
```
LinjingZhang, line 790: It’s worth mentioning here that user need to ensure proper udev permissions: 
```
$ wget https://raw.githubusercontent.com/jetperch/pyjoulescope/master/99-joulescope.rules
$ sudo cp 99-joulescope.rules /etc/udev/rules.d/
$ sudo udevadm control --reload-rules
```
copied from: https://joulescope.readthedocs.io/en/latest/user/install.html#linux

doc/configuration.rst
```diff
@@ -765,6 +765,55 @@ Arguments:
...
+  - match (dict): key and value pairs for a udev match, see `udev Matching`_
+
+Used by:
+  - `JoulescopeDriver`_
+
+NetworkJoulescopeDevice
+~~~~~~~~~~~~~~~~~~~~~~~~~
+A :any:`NetworkJoulescopeDevice` resource describes a `JoulescopeDevice`_ that is
```
LinjingZhang: The network resources should be self-explanatory. The note is good, I'm not sure if this is the right place to do so.

See also the documentation for other network USB resources: https://labgrid.readthedocs.io/en/latest/configuration.html#usbtmc

doc/configuration.rst
```diff
@@ -3191,6 +3240,55 @@ samples is an iterable of samples.
...
 Reading a few samples will very likely work - but obtaining a lot of samples may stall.
 
+JoulescopeDriver
+~~~~~~~~~~~~~~~~
+The :any:`JoulescopeDriver` uses a `JoulescopeDevice`_ or
+`NetworkJoulescopeDevice`_ resource to measure current, voltage and power,
+accumulate charge and energy, capture high-rate samples to a JLS file, and switch
+downstream power by connecting/disconnecting the device current path.
+
+``pyjoulescope_driver`` runs on the host the Joulescope is attached to (via the
```
LinjingZhang: same here, redundant information for network resource.

LinjingZhang (approved):
I tested it in my Labgrid environment, and it worked perfectly! However, it's currently limited to local connections—I haven't set up any actual distributed remote access yet.

doc/configuration.rst
Emantor: This is true for all of the USB devices and I am aware that this is a problem with our documentation in general and not specific to this driver.

LinjingZhang: I see. 
I was just referring to the documentation for another similar device:
https://labgrid.readthedocs.io/en/latest/configuration.html#usbtmc

labgrid/driver/joulescopedriver.py
```diff
@@ -0,0 +1,130 @@
...
+    attached device and for one shared over the distributed infrastructure via
+    a :class:`~labgrid.resource.remote.NetworkJoulescopeDevice`.  Only the host
+    with the device attached needs the ``joulescope`` extra installed.
+
+    Power switching (``on``/``off``/``cycle``) controls downstream power to the
+    device under test: the JS110 uses the current range ``select`` and the JS220
+    and JS320 use the current range ``mode``.
+
+    Args:
+        frequency (float): statistics update frequency in Hz
```
LinjingZhang, line 38: After analyzing the output data, I realized that this should actually be in 1MHz.
Could you verify this?

doc/configuration.rst
mliberty1: Removed in 564d8c901ac7afde0dad282540ba8546945ce758

doc/configuration.rst
mliberty1: Added instructions in 564d8c901ac7afde0dad282540ba8546945ce758

doc/configuration.rst
mliberty1: Removed in 564d8c901ac7afde0dad282540ba8546945ce758

mliberty1:
I just updated this PR to integrate the review feedback and squashed.

@LinjingZhang : you can find the exact changes here
https://github.com/labgrid-project/labgrid/commit/564d8c901ac7afde0dad282540ba8546945ce758

doc/configuration.rst
```diff
@@ -3194,6 +3224,50 @@ samples is an iterable of samples.
...
+
+.. code-block:: yaml
+
+   JoulescopeDriver:
+     frequency: 10.0
+     delay: 2.0
+
+Arguments:
+  - frequency (float, default=2.0): statistics update frequency in Hz
+  - delay (float, default=2.0): delay in seconds between off and on during a
```
LinjingZhang, line 3251: The parameter documentation needs to be adjusted. Are the units here really Hz?
It is worth noting that the `delay `is used only for PowerProtocol.

labgrid/util/agents/joulescope.py
```diff
@@ -0,0 +1,234 @@
...
+
+    # -- sample capture -----------------------------------------------------
+
+    def capture(self, filename, signals=None, duration=None, frequency=None):
+        if duration is None:
+            raise ValueError("capture() requires a duration in seconds")
+        if frequency is not None:
+            # Note: this changes the device sample rate for the rest of the
+            # session; it is not restored to the default after the capture.
+            self._jsdrv.publish(self._path + "/h/fs", int(frequency))
```
LinjingZhang, line 155: If the frequency is changed, will the device reboot?
In my case, if I call capture()?

labgrid/util/agents/joulescope.py
```diff
@@ -0,0 +1,234 @@
...
+
+    def _configure_statistics(self):
+        dev = self._path
+        if self._model == "js110":
+            self._jsdrv.publish(dev + "/s/i/range/select", "auto")
+            # host-side statistics honor the requested frequency and report std
+            self._jsdrv.publish(dev + "/s/i/ctrl", "on")
+            self._jsdrv.publish(dev + "/s/v/ctrl", "on")
+            self._jsdrv.publish(dev + "/s/p/ctrl", "on")
+            base = 2_000_000  # JS110 host-side statistics sample rate
```
LinjingZhang, line 75: Is the samplw rate fixed? For each device.
Then what does the `frequency` I set in labgrid yaml file apply to?

labgrid/driver/joulescopedriver.py
mliberty1: It is in Hz.  The calculation is in util.agents.JoulescopeSession._configure_statistics.  What are you seeing that make think otherwise?  Note that statistics frequency (provided here) and sample freqeucny (provided to `capture`) are two different things.

doc/configuration.rst
mliberty1: Yes, Hz..

labgrid/util/agents/joulescope.py
mliberty1: You can change frequency with each call to capture.  The device should not reboot.

labgrid/util/agents/joulescope.py
mliberty1: The sample rate frequency is not fixed, but the base frequency is.  You are looking at how the scnt value is calculated to set the statistics rate.  `h/fs` configures the sample rate in `capture`.

labgrid/driver/joulescopedriver.py
LinjingZhang: Thank you for your explanation. I think I got confused between the sample rate frequency (the parameter passed to `capture()`)
and the driver update frequency (defined in the YAML file).

StefanHri:
When can we expect this to get merged? Support for Joulescope is highly needed on our side.

PR #1926, opened 2026-07-04
migrate siglent power backend vxi11 -> pyvisa, add power show/config options

Rubusch:
Hi! 

First this will migrate the driver/power/siglent.py from vxi11 to pyvisa. We use different Siglents in our lab, and could verify it locally. Second is a proposal, add basic configuration options to the power command as an optional add-on for power supplies which offer that. The idea is, at least to set and verify remotely voltage and current.

Usage example:
$ labgrid-client power config --voltage 5.0 --amps 1.5
    Selected role main from configuration file
    set voltage to 5.0V
    set amps to 1.5A
    
$ labgrid-client power show
    Selected role main from configuration file
    Voltage: 0.0V [5.0V], Current: 0.0A [1.5A]
    
$ labgrid-client power on
    Selected role main from configuration file
    
$ labgrid-client power show
    Selected role main from configuration file
    Voltage: 5.001V [5.0V], Current: 0.209A [1.5A]

Note, I did neither add any testes, nor documentation. Just let me know if you're interested to elaborate this further, or in case split the commits, or whatever to do with it.

**Checklist**
- [ ] Tests for the feature 

- [ ] PR has been tested

### DIFF
--- a/labgrid/driver/powerdriver.py
+++ b/labgrid/driver/powerdriver.py
@@ -7,7 +7,7 @@ import attr
 
 from ..exceptions import InvalidConfigError
 from ..factory import target_factory
-from ..protocol import PowerProtocol, DigitalOutputProtocol, ResetProtocol
+from ..protocol import PowerProtocol, DigitalOutputProtocol, ResetProtocol, ProgrammablePowerProtocol
 from ..resource import NetworkPowerPort
 from ..step import step
 from ..util.proxy import proxymanager
@@ -232,6 +232,21 @@ class NetworkPowerDriver(Driver, PowerResetMixin, PowerProtocol):
     @Driver.check_active
     def get(self):
         return self.backend.power_get(self._host, self._port, self.port.index)
+
+    @Driver.check_active
+    @step()
+    def show(self):
+        return self.backend.power_show(self._host, self._port, self.port.index)
+
+    @Driver.check_active
+    @step(args=['voltage'])
+    def voltage(self, voltage):
+        return self.backend.power_voltage(self._host, self._port, self.port.index, voltage)
+
+    @Driver.check_active
+    @step(args=['amps'])
+    def amps(self, amps):
+        return self.backend.power_amps(self._host, self._port, self.port.index, amps)
 
 @target_factory.reg_driver
 @attr.s(eq=False)
--- a/labgrid/protocol/powerprotocol.py
+++ b/labgrid/protocol/powerprotocol.py
@@ -13,3 +13,17 @@ class PowerProtocol(abc.ABC):
     @abc.abstractmethod
     def cycle(self):
         raise NotImplementedError
+
+
+class ProgrammablePowerProtocol(PowerProtocol):
+    @abc.abstractmethod
+    def show(self, index):
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def voltage(self, index, voltage):
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def amps(self, index, amps):
+        raise NotImplementedError
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -247,8 +247,12 @@ Currently available are:
   for details.
 
 ``siglent``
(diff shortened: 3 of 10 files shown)


labgrid/protocol/powerprotocol.py
```diff
@@ -13,3 +13,17 @@ def off(self):
...
+    @abc.abstractmethod
+    def show(self, index):
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def voltage(self, index, voltage):
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def amps(self, index, amps):
```
vermaete, line 28: Is 'current' not more suiteable i.s.o. amps?

labgrid/driver/powerdriver.py
```diff
@@ -233,6 +233,21 @@ def cycle(self):
     def get(self):
         return self.backend.power_get(self._host, self._port, self.port.index)
 
+    @Driver.check_active
+    @step()
+    def show(self):
+        return self.backend.power_show(self._host, self._port, self.port.index)
+
+    @Driver.check_active
```
vermaete, line 241: Could power (W) also be added?  Some  PowerSupplies (also Siglent) do allow this.
Reading Voltage and Current and multiply it in software will introduce and error because the time difference between the two reads.

Rubusch: @vermaete : I'll try to have a look into it this weekend and push something, after that I'm travelling for around four weeks. So, in general "amps" -> "current" and "power" in [W]. Let me know if something else is missing. I have to have a look into our siglent material here, let's see. ty

@Bastian-Krause : There seems to be some interest in this. Pls, what should I additionally need to provide as *test* and *documentation* here? Can you point me to something similar/driver you'd like to see? If not, I might figure out and come up with anything.

vermaete: @Rubusch Thomas-Vreys did some work on your fork and has created a PR with some of the remarks solved.
Thomas will be available at Wednesday to continue with it, if needed.
We do have some power-supplies to test this.
Thanks for the useful Labgrid feature.

Rubusch: @vermaete I merged this. I had a look at my hardware, unfortunately I cannot verify power readings. So, I appreaciate the effort. Thanks.

I still agree in "current" being more suitable than "amps". Now, I did not like to change before merging and currently busy/travelling. Shall I still add a commit to rename it, or can we live with that as is? I think to user interface it exposes "current".

vermaete: thanks for merging.  Naming as it is now is fine for me.

Thomas-Vreys:
Came across #1835, they've built a generic pyvisa/SCPI layer with PSU backends for the DP832, RND 320-KD3305P and ITECH IT-M3632. Same underlying idea as what we're doing for the Siglent, just at a more general level. The transport is a config option so you could also connect over USB. But no Siglent backend there yet but maybe worth knowing whether programmable PSUs are meant to head that way.

Sabolik:
> Came across #1835, they've built a generic pyvisa/SCPI layer with PSU backends for the DP832, RND 320-KD3305P and ITECH IT-M3632. Same underlying idea as what we're doing for the Siglent, just at a more general level. The transport is a config option so you could also connect over USB. But no Siglent backend there yet but maybe worth knowing whether programmable PSUs are meant to head that way.

I believe it should exactly fits to your use case, we also use power protocol on top of Laboratory power supply, rigol in our case. There is  [TMInstrumentPower](https://github.com/labgrid-project/labgrid/pull/1835/changes#diff-5e3dd36d509111b80e62344e1e91e35be10f1a23bb5b7184aa29305e513a4b89R111) driver to support this. Appropriate backed implementation should be the only missing part, with [backend ](https://github.com/labgrid-project/labgrid/pull/1835/changes#diff-5e3dd36d509111b80e62344e1e91e35be10f1a23bb5b7184aa29305e513a4b89R33) list to be extended.

PR #1931, opened 2026-07-13
Add FTDI bit-bang GPIO output support

ozan956:
This adds `FTDIGPIO` resource, an `FTDIGPIODriver`, and a small USB agent. The feature is intended for boards where FTDI data-bus pins are used as simple digital outputs, for example to control straps, enables, reset-like signals, or other board control lines during tests.

This targets FTDI asynchronous bit-bang GPIO on the data bus pins. It does not target CBUS GPIO, EEPROM-configured ACBUS functions, or generic MPSSE GPIO.

Supported devices are:

- FT2232-style devices, `0403:6010`, interfaces `1..2`
- FT4232H, `0403:6011`, interfaces `1..4`
- FT232H, `0403:6014`, interface `1`

Each interface exposes an independent 8-bit data bus. `index` selects bit `0..7` on that interface.

For writes, the agent first reads the current pin byte, modifies only the requested bit, and writes the byte back. This avoids unintentionally changing other bits on the same FTDI interface. `get()` reads the current level of the configured output line.

I also would like to support input pins with `direction` feature of those chips in future, so https://github.com/labgrid-project/labgrid/pull/1458 would be very helpful.

I tested the driver with a FT232H device (`0403:6014`).

**Checklist**
- [x] Tests for the feature
- [x] PR has been tested

### DIFF
--- a/labgrid/util/agents/ftdigpio.py
+++ b/labgrid/util/agents/ftdigpio.py
@@ -1,0 +1,149 @@
+# SPDX-License-Identifier: GPL-2.0-or-later
+"""Agent for controlling FTDI data-bus GPIOs via bit-bang mode."""
+
+import threading
+
+import usb.core
+import usb.util
+
+SIO_SET_BITMODE = 11
+SIO_READ_PINS = 12
+BITMODE_ASYNC_BITBANG = 1
+OUT_REQTYPE = 0x40
+IN_REQTYPE = 0xC0
+
+USB_TIMEOUT = 1000
+GPIO_MASK = 0xFF
+SUPPORTED_DEVICES = {
+    0x6010: 2,  # FT2232C/D/H Dual UART/FIFO IC
+    0x6011: 4,  # FT4232H Quad UART/MPSSE IC
+    0x6014: 1,  # FT232HL/Q
+}
+
+
+class FTDIGPIO:
+    def __init__(self, vendor_id, model_id, busnum, devnum, interface):
+        self._validate_device(vendor_id, model_id, interface)
+        self._interface = interface - 1
+        self._index = interface
+        self._lock = threading.Lock()
+
+        self._dev = self._find_device(vendor_id, model_id, busnum, devnum)
+        self._detach_kernel_driver()
+        try:
+            cfg = self._dev.get_active_configuration()
+        except usb.core.USBError:
+            self._dev.set_configuration()
+            self._detach_kernel_driver()
+            cfg = self._dev.get_active_configuration()
+
+        intf = cfg[(self._interface, 0)]
+        usb.util.claim_interface(self._dev, self._interface)
+        self._ep_out = usb.util.find_descriptor(
+            intf,
+            custom_match=lambda ep: usb.util.endpoint_direction(ep.bEndpointAddress) == usb.util.ENDPOINT_OUT,
+        )
+        if self._ep_out is None:
+            raise ValueError("FTDI output endpoint not found")
+
+    def close(self):
+        try:
+            try:
+                usb.util.release_interface(self._dev, self._interface)
+            except usb.core.USBError:
+                pass
+        finally:
+            usb.util.dispose_resources(self._dev)
+
+    def _detach_kernel_driver(self):
+        if self._dev.is_kernel_driver_active(self._interface):
+            self._dev.detach_kernel_driver(self._interface)
+
+    @staticmethod
+    def _validate_device(vendor_id, model_id, interface):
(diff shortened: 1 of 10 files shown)


ozan956:
Just pushed input support as well. 

We’ve been using the driver in our setups for the past few months and haven’t run into any issues so far.

cc @CalebEthridgeADI

labgrid/util/agents/ftdigpio.py
```diff
@@ -0,0 +1,184 @@
...
+_directions_lock = threading.Lock()
+
+
+class FTDIGPIO:
+    def __init__(self, vendor_id, model_id, busnum, devnum, interface):
+        self._validate_device(vendor_id, model_id, interface)
+        self._interface = interface - 1
+        self._index = interface
+        self._key = (busnum, devnum, interface)
+        self._lock = threading.Lock()
```
Emantor: What does this lock protect? Agents keep running through the session, but we only ever access them serially from the client, unless you are using threading within labgrid which is not supported.

labgrid/util/agents/ftdigpio.py
```diff
@@ -0,0 +1,184 @@
...
+    0x6010: 2,  # FT2232C/D/H Dual UART/FIFO IC
+    0x6011: 4,  # FT4232H Quad UART/MPSSE IC
+    0x6014: 1,  # FT232HL/Q
+}
+
+# Per-interface output direction mask, keyed by (busnum, devnum, interface). A
+# set bit marks the corresponding line as an output. Written only by setup() at
+# driver activation; get()/set() read it but never reprograms the bit-mode.
+_directions = {}
+_directions_lock = threading.Lock()
```
Emantor, line 28: Same as for the lock below.

labgrid/util/agents/ftdigpio.py
```diff
@@ -0,0 +1,184 @@
...
+SUPPORTED_DEVICES = {
+    0x6010: 2,  # FT2232C/D/H Dual UART/FIFO IC
+    0x6011: 4,  # FT4232H Quad UART/MPSSE IC
+    0x6014: 1,  # FT232HL/Q
+}
+
+# Per-interface output direction mask, keyed by (busnum, devnum, interface). A
+# set bit marks the corresponding line as an output. Written only by setup() at
+# driver activation; get()/set() read it but never reprograms the bit-mode.
+_directions = {}
```
Emantor: Why do the directions need to be locked in the first place?

labgrid/util/agents/ftdigpio.py
```diff
@@ -0,0 +1,184 @@
...
+        intf = cfg[(self._interface, 0)]
+        usb.util.claim_interface(self._dev, self._interface)
+        self._ep_out = usb.util.find_descriptor(
+            intf,
+            custom_match=lambda ep: usb.util.endpoint_direction(ep.bEndpointAddress) == usb.util.ENDPOINT_OUT,
+        )
+        if self._ep_out is None:
+            raise ValueError("FTDI output endpoint not found")
+
+    def close(self):
```
Emantor, line 57: Is this required? Does the GPIO state change if the device is released? Otherwise this can use the same handling as the USB HID relay agent.

labgrid/util/agents/ftdigpio.py
```diff
@@ -0,0 +1,184 @@
...
+
+def _run_with_device(vendor_id, model_id, busnum, devnum, interface, callback):
+    device = FTDIGPIO(vendor_id, model_id, busnum, devnum, interface)
+    try:
+        return callback(device)
+    finally:
+        device.close()
+
+
+def handle_get(vendor_id, model_id, busnum, devnum, interface, index):
```
Emantor, line 136: Instead of pinholing these functions through the same function with a callback, it should instead keep a list of known endpoints around similar to the usb hid relay agent.

Emantor (changes requested):

labgrid/util/agents/ftdigpio.py
ozan956: >  but we only ever access them serially from the client

right, will be removed :+1:

labgrid/util/agents/ftdigpio.py
ozan956: Direction state must be retained because SIO_SET_BITMODE programs the direction mask for the entire interface. Configuring another pin must preserve the direction of all previously configured pins.

labgrid/util/agents/ftdigpio.py
ozan956: Nope, will follow the USB HID.

PR #1932, opened 2026-07-13
feat: rename LG_TOKEN to LG_RESERVATION

asher-pem-arm:
**Description**

**Checklist**
- [x] Tests for the feature 

- [x] PR has been tested

# Motivation
Reservation identifiers are not authentication tokens. Calling them IDs makes their purpose clearer and avoids confusion with credentials.

# Summary
Rename the reservation identifier from token to reservation ID.

- Replace LG_TOKEN with LG_RESERVATION.
- Rename Python Reservation.token to Reservation.id.
- Update CLI output, help text, documentation, and tests.
- Add id support to reservation filters.
- Retain LG_TOKEN input and the token filter as legacy fallbacks.
- Keep protobuf token fields unchanged for wire compatibility.

# Testing

## Automated

The following tests have been updated and are all passing.

```shell
pytest tests/test_coordinator.py
pytest tests/test_client.py
pytest tests/test_pb2.py
```

### Reserve, Lock, Unlock (`LG_RESERVATION_ID`)

```python
$ labgrid-client reserve name=p1 --wait --shell
export LG_RESERVATION_ID=YR0SUORW5C
$ LG_RESERVATION_ID=YR0SUORW5C labgrid-client -p + lock
acquired place p1
$ LG_RESERVATION_ID=YR0SUORW5C labgrid-client -p + unlock
released place p1
```

### Reserve, Lock, Unlock (`LG_TOKEN`)

```python
$ labgrid-client reserve name=p1 --wait --shell
export LG_RESERVATION_ID=T9GRNKWWA7
$ LG_TOKEN=T9GRNKWWA7 labgrid-client -p + lock
acquired place p1
$ LG_TOKEN=T9GRNKWWA7 labgrid-client -p + unlock
released place p1
```

### Reserve, Lease, Extend, Release (`LG_RESERVATION_ID`)

```python
$ labgrid-client reserve name=p1 --wait --shell
export LG_RESERVATION_ID=EHA0E0L9OF
LG_RESERVATION_ID=EHA0E0L9OF labgrid-client -p + lease
leased place p1
$ LG_RESERVATION_ID=EHA0E0L9OF labgrid-client extend
extended lease EHA0E0L9OF until 2026-07-07 13:25:00.422546
$ LG_RESERVATION_ID=EHA0E0L9OF labgrid-client -p + release
released place p1
```

### Reserve, Lease, Extend, Release (`LG_TOKEN`)

```python
$ labgrid-client reserve name=p1 --wait --shell
export LG_RESERVATION_ID=HD22S7BBAN
$ LG_TOKEN=HD22S7BBAN labgrid-client -p + lease
leased place p1
$ LG_TOKEN=HD22S7BBAN labgrid-client extend
extended lease HD22S7BBAN until 2026-07-07 13:26:25.733619
$ LG_TOKEN=HD22S7BBAN labgrid-client -p + release
released place p1
```

### Reserve, Wait (`LG_RESERVATION_ID`)

```python
$ labgrid-client reserve name=p1 --wait --shell
export LG_RESERVATION_ID=AOCT0KXTOG
$ LG_RESERVATION_ID=AOCT0KXTOG labgrid-client wait
owner: <machine>/<user>
...
```

### Reserve, Wait (`LG_TOKEN`)

```python
$ labgrid-client reserve name=p1 --wait --shell
export LG_RESERVATION_ID=FDH6EL58TI
$ LG_TOKEN=FDH6EL58TI labgrid-client wait
owner: <machine>/<user>
...
```

### DIFF
--- a/doc/usage.rst
+++ b/doc/usage.rst
@@ -67,7 +67,7 @@ tags).
   $ labgrid-client reserve board=imx6-foo
   Reservation 'SP37P5OQRU':
     owner: rettich/jlu
-    token: SP37P5OQRU
+    id   : SP37P5OQRU
     state: waiting
     filters:
       main: board=imx6-foo
@@ -76,7 +76,7 @@ tags).
 
 As soon as any matching place becomes free, the reservation state will change
 from ``waiting`` to ``allocated``.
-Then, you can use the reservation token prefixed by ``+`` to refer to the
+Then, you can use the reservation id prefixed by ``+`` to refer to the
 allocated place for locking and usage.
 While a place is allocated for a reservation, only the owner of the reservation
 can lock that place.
@@ -86,7 +86,7 @@ can lock that place.
 
   $ labgrid-client wait SP37P5OQRU
   owner: rettich/jlu
-  token: SP37P5OQRU
+  id:    SP37P5OQRU
   state: waiting
   filters:
     main: board=imx6-foo
@@ -94,7 +94,7 @@ can lock that place.
   timeout: 2019-08-06 12:58:14.900621
   …
   owner: rettich/jlu
-  token: SP37P5OQRU
+  id:    SP37P5OQRU
   state: allocated
   filters:
     main: board=imx6-foo
@@ -107,7 +107,7 @@ can lock that place.
   $ labgrid-client reservations
   Reservation 'SP37P5OQRU':
     owner: rettich/jlu
-    token: SP37P5OQRU
+    id:    SP37P5OQRU
     state: acquired
     filters:
       main: board=imx6-foo
@@ -117,20 +117,31 @@ can lock that place.
     timeout: 2019-08-06 12:59:11.840780
   $ labgrid-client -p +SP37P5OQRU console
 
+
+.. note::
+
+   Reservation tokens are now called reservation IDs.
+   As part of this terminology change, ``LG_TOKEN`` has been renamed to
+   ``LG_RESERVATION``, and ``labgrid-client reserve --shell`` now exports
+   ``LG_RESERVATION``.
+   Command output now uses ``id`` instead of ``token``.
+   The legacy ``LG_TOKEN`` variable remains accepted as input for backwards
+   compatibility, but scripts should migrate to ``LG_RESERVATION``.
+
 When using reservation in a CI job or to save some typing, the ``labgrid-client
 reserve`` command supports a ``--shell`` command to print code for evaluating
 in the shell.
-This sets the ``LG_TOKEN`` environment variable, which is then automatically
+This sets the ``LG_RESERVATION`` environment variable, which is then automatically
 used by ``wait`` and expanded via ``-p +``.
 
 .. code-block:: bash
 
   $ eval `labgrid-client reserve --shell board=imx6-foo`
-  $ echo $LG_TOKEN
+  $ echo $LG_RESERVATION
   ZDMZJZNLBF
   $ labgrid-client wait
   owner: rettich/jlu
-  token: ZDMZJZNLBF
+  id:    ZDMZJZNLBF
   state: waiting
   filters:
     main: board=imx6-foo
@@ -138,7 +149,7 @@ used by ``wait`` and expanded via ``-p +``.
   timeout: 2019-08-06 13:06:44.629736
   …
   owner: rettich/jlu
-  token: ZDMZJZNLBF
+  id:    ZDMZJZNLBF
   state: allocated
   filters:
     main: board=imx6-foo
--- a/doc/usage.rst
+++ b/doc/usage.rst
@@ -67,7 +67,7 @@ tags).
   $ labgrid-client reserve board=imx6-foo
   Reservation 'SP37P5OQRU':
     owner: rettich/jlu
-    token: SP37P5OQRU
+    id   : SP37P5OQRU
     state: waiting
     filters:
       main: board=imx6-foo
@@ -76,7 +76,7 @@ tags).
 
 As soon as any matching place becomes free, the reservation state will change
 from ``waiting`` to ``allocated``.
-Then, you can use the reservation token prefixed by ``+`` to refer to the
+Then, you can use the reservation id prefixed by ``+`` to refer to the
 allocated place for locking and usage.
 While a place is allocated for a reservation, only the owner of the reservation
 can lock that place.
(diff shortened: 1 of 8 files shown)


jluebbe:
The downside is that users will likely find old examples/code/talks using `LG_TOKEN` and become confused because this is no longer mentioned anywhere. Are we sure enough that the new name is so much clearer that it justifies the cost?

I see the possible confusion with authentication tokens, tough. I'm not sure the `_ID` suffix adds much; `LG_RESERVATION` would more concise and perhaps clear enough (it could still be called `.id` internally in the code)?

@Emantor @Bastian-Krause

Emantor:
I do agree on removing the `_ID` suffix. Is there anything else we might want to add after `LG_RESERVATION`? Otherwise we can just use `LG_RESERVATION` without the ID suffix.

labgrid/remote/client.py
```diff
@@ -438,19 +438,21 @@ def _match_places(self, pattern):
...
         if pattern.startswith("+"):
-            token = pattern[1:]
-            if not token:
-                token = os.environ.get("LG_TOKEN", None)
-            if not token:
+            reservation_id = pattern[1:]
+            if not reservation_id:
+                reservation_id = os.environ.get("LG_RESERVATION_ID", None)
+            if not reservation_id:
+                reservation_id = os.environ.get("LG_TOKEN", None)  # For backwards compatibility
```
Emantor: We should raise a `DeprecationWarning` to alert users if the old environment variable `LG_TOKEN` is used.

Emantor (changes requested):

labgrid/remote/client.py
jluebbe: Note that `DeprecatingWarning`s are [hidden by default in many cases](https://docs.python.org/3/library/exceptions.html#DeprecationWarning).

We could have a log message, but I'd like to avoid warnings which are not actionable. Perhaps we should only add the warning in a later release when it's more likely that all components support the new variable.

jluebbe:
The documentation should have a note mentioning the rename, so people have a change to understand what's happening. :)

labgrid/remote/client.py
asher-pem-arm: Leaving as is for now then. Once `LG_RESERVATION` has been available for a full release and there's a planned removal version then can emit a log warning on stderr when `LG_TOKEN` is used.

asher-pem-arm:
updated with rename and docs

labgrid/remote/client.py
Emantor: > We could have a log message, but I'd like to avoid warnings which are not actionable. Perhaps we should only add the warning in a later release when it's more likely that all components support the new variable.

I would not log, we should instead send a message to stderr to tell them to use the new environment variable instead of the old one. This seems very actionable to me and shortens the cycle with which we can remove the old variable.

asher-pem-arm: yeah good idea, updated

Emantor: @jluebbe are you okay with these changes?

Emantor (approved):

Emantor:
This looks good to me, I think we should squash the fixes which should also fix the DCO errors.

asher-pem-arm:
@Emantor thanks, updated

doc/usage.rst
```diff
@@ -117,28 +117,39 @@ can lock that place.
...
+
+.. note::
+
+   Reservation tokens are now called reservation IDs.
+   As part of this terminology change, ``LG_TOKEN`` has been renamed to
+   ``LG_RESERVATION``, and ``labgrid-client reserve --shell`` now exports
+   ``LG_RESERVATION``.
+   Command output now uses ``id`` instead of ``token``.
+   The legacy ``LG_TOKEN`` variable remains accepted as input for backwards
+   compatibility, but scripts should migrate to ``LG_RESERVATION``.
```
Emantor, line 131: This section should export **why** we are forcing people to change to `LG_RESERVATION` in all workflows/scripts/CLI usage. IMO this is needed because `LG_TOKEN` implies a security property (which is not present here) since the meaning of `TOKEN` has changed over the last few years. Additional explanation or links are welcome as well.

Emantor (changes requested):

labgrid/remote/client.py
```diff
@@ -2210,11 +2221,11 @@ def get_parser(auto_doc_mode=False) -> "argparse.ArgumentParser | AutoProgramArg
     subparser.set_defaults(func=ClientSession.create_reservation)
 
     subparser = subparsers.add_parser("cancel-reservation", help="cancel a reservation")
-    subparser.add_argument("token", type=str, nargs="?")
+    subparser.add_argument("reservation-id", type=str, nargs="?", help="the reservation id (previously called token)")
```
Emantor:
```suggestion
    subparser.add_argument("reservation_id", type=str, nargs="?", help="the reservation id (previously called token)", metavar="reservation-id")
```

and we can replace the various `getattr(args, "reservation-id")` calls with `args.reservation_id`.

labgrid/remote/client.py
```diff
@@ -1572,28 +1583,28 @@ async def create_reservation(self):
 
         res = Reservation.from_pb2(response.reservation)
         if self.args.shell:
-            print(f"export LG_TOKEN={res.token}")
+            print(f"export LG_RESERVATION={res.id}")
```
Emantor, line 1593: We need to export the old `LG_TOKEN` here as well at least for a release. Also please add a `TODO` to eventually remove the `LG_TOKEN` export print after that.

Emantor (changes requested):
The commit message explains what is done, but not why it is done (security implications of the token name) please expand on that.

doc/usage.rst
asher-pem-arm: updated

labgrid/remote/client.py
asher-pem-arm: updated

labgrid/remote/client.py
asher-pem-arm: updated. 
FYI exporting both variables could cause problems:

Say I:

Use eval `lc reserve abc=def --wait --shell`
`LG_TOKEN` set to A
`LG_RESERVATION` set to A

Some old script/command is run to set `LG_TOKEN`
`LG_TOKEN` set to B

Use labgrid-client
You will access A, not B, because `LG_ RESERVATION` takes precedence

Emantor: We could warn about that in `_get_reservation_id_from_env`, this should make it more clear.

asher-pem-arm: good idea! updated

Bastian-Krause (approved):

PR #1935, opened 2026-07-13
Add input path for sysfs GPIOs

ozan956:
This is a follow-up to the GPIO input work discussed in labgrid-project/labgrid#1458. Instead of adding a separate input-only sysfs agent, this PR keeps the existing `sysfsgpio` agent and extends it so a GPIO line can be configured either as output or input by the active driver.

The change adds:

- `DigitalInputProtocol`, for drivers which can read a digital signal without
  necessarily supporting writes.
- `GpioDigitalInputDriver`, binding to `SysfsGPIO`, `MatchedSysfsGPIO`, and
  `NetworkSysfsGPIO`.
- `labgrid-client io get <name>` support for explicitly configured
  `DigitalInputProtocol` drivers.
- Sysfs GPIO agent support for input reads, while preserving output as the
  default behavior for existing users.

`DigitalOutputProtocol` now derives from `DigitalInputProtocol`, matching the existing behavior of digital output drivers which already implement `get()`. This keeps existing output users compatible while allowing input-only drivers to use the same CLI command for reads.

The client-side behavior is intentionally conservative: `io get` first looks for a configured `DigitalInputProtocol` driver. If no such driver exists, the existing resource-only fallback still creates a `GpioDigitalOutputDriver` for `NetworkSysfsGPIO`, so existing configurations continue to work as before.

The sysfs agent caches GPIO line objects, so this PR also makes cached lines reconfigure their direction before reuse. This avoids stale direction state if the same GPIO line is accessed through different configured drivers during a test setup.

The implementation still uses the existing sysfs GPIO backend. Moving this to the newer gpiochip/libgpiod interface would be useful, since sysfs GPIO is deprecated, but that is intentionally left as future work to keep this PR small and focused.

**Checklist**
- [x] Tests for the feature
- [x] PR has been tested

cc @Emantor @flxzt

### DIFF
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -991,7 +991,10 @@ class ClientSession:
 
         drv = None
         try:
-            drv = target.get_driver("DigitalOutputProtocol", name=name)
+            if action == "get":
+                drv = target.get_driver("DigitalInputProtocol", name=name)
+            else:
+                drv = target.get_driver("DigitalOutputProtocol", name=name)
         except NoDriverFoundError:
             for resource in target.resources:
                 if name and resource.name != name:
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -651,6 +651,7 @@ Arguments:
   - invert (bool, default=False): optional, whether the logic level is inverted (active-low)
 
 Used by:
+  - `GpioDigitalInputDriver`_
   - `GpioDigitalOutputDriver`_
 
 NetworkSysfsGPIO
@@ -684,6 +685,7 @@ Arguments:
   - invert (bool, default=False): optional, whether the logic level is inverted (active-low)
 
 Used by:
+  - `GpioDigitalInputDriver`_
   - `GpioDigitalOutputDriver`_
 
 NetworkService
@@ -2558,6 +2560,30 @@ Implements:
 
 Arguments:
   - delay (float, default=2.0): delay in seconds between off and on
+
+GpioDigitalInputDriver
+~~~~~~~~~~~~~~~~~~~~~~
+The :any:`GpioDigitalInputDriver` reads a digital signal from a GPIO line.
+
+This driver configures GPIO lines via
+`the sysfs kernel interface <https://www.kernel.org/doc/html/latest/gpio/sysfs.html>`__
+as an input.
+
+Binds to:
+  gpio:
+    - `SysfsGPIO`_
+    - `MatchedSysfsGPIO`_
+    - `NetworkSysfsGPIO`_
+
+Implements:
+  - :any:`DigitalInputProtocol`
+
+.. code-block:: yaml
+
+   GpioDigitalInputDriver: {}
+
+Arguments:
+  - None
 
 GpioDigitalOutputDriver
 ~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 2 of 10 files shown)


Emantor:
Why are you taking authorship for most commits instead of working on top of @flxzt commits?

ozan956:
Hi @Emantor,

There was only one original commit containing all the changes. I built on top of it, added my own changes, and reworked parts of the implementation.

I included a `Reported-by` tag, but I can change it to `Signed-off-by` or `Co-developed-by`, depending on which one best matches labgrid’s contribution guidelines.

I’m happy to update the commit metadata accordingly. Let me know which format you prefer.

labgrid/remote/client.py
```diff
@@ -991,7 +991,10 @@ def digital_io(self):
 
         drv = None
         try:
-            drv = target.get_driver("DigitalOutputProtocol", name=name)
+            if action == "get":
+                drv = target.get_driver("DigitalInputProtocol", name=name)
```
flxzt, line 995: I believe this implementation in the client has the same issue as I had when I attempted to implement this.
Have you checked whether https://github.com/labgrid-project/labgrid/pull/1458#issuecomment-2265361505 also happens for you?

flxzt:
I would appreciate if attribution is kept in some way.

Emantor:
> Hi @Emantor,
> 
> There was only one original commit containing all the changes. I built on top of it, added my own changes, and reworked parts of the implementation.
> 
> I included a `Reported-by` tag, but I can change it to `Signed-off-by` or `Co-developed-by`, depending on which one best matches labgrid’s contribution guidelines.
> 
> I’m happy to update the commit metadata accordingly. Let me know which format you prefer.

I usually retain the original authorship in the commit and add my `Co-developed-by` for commits that I changed, but ultimately this is up to @flxzt.

flxzt:
`Co-developed-by` me would be fine for me

labgrid/remote/client.py
ozan956: Yes, tested on a Raspberry Pi 2 Model B running Raspbian with kernel `6.12.47+rpt-rpi-v7`.

I tested whether `labgrid-client io get` changes the sysfs direction from `out` to `in` after setting the line high:

```bash
$labgrid-client -p rpi-gpio io high
cat /sys/class/gpio/gpio529/direction
out
cat /sys/class/gpio/gpio529/value
1

$labgrid-client -p rpi-gpio io get
digital IO for place rpi-gpio is high

cat /sys/class/gpio/gpio529/direction
out
cat /sys/class/gpio/gpio529/value
1
```
So I could not reproduce the behavior from the linked comment with this branch. After io get, the line remains configured as out and the value remains high.

This should be because the resource-only NetworkSysfsGPIO fallback still uses GpioDigitalOutputDriver for labgrid-client io get. The new DigitalInputProtocol path is only used when an input driver is explicitly configured.

ozan956:
Thanks @flxzt and @Emantor, I updated the commit trailers/sign-offs accordingly. :+1:

Emantor (approved):

Bastian-Krause (approved):

labgrid/remote/client.py
Emantor: This is correct for backwards compatibility for this PR, long term we could think about the introduction of new client subcommands to distinguish these cases.

jluebbe (approved):

PR #1939, opened 2026-07-16
Propagate client identity to exporter

asher-pem-arm:
**Description**

**Checklist**
- [x] Tests for the feature 

- [x] PR has been tested

### Description
Exporters currently receive only the logical place name when acquiring or leasing a resource. They cannot identify which client initiated the operation in exporter-local logs. This improves per-resource diagnostics and traceability on distributed exporter hosts. The identity is used only for logging; authorization and resource ownership semantics are unchanged.

### This PR:
- Adds client identity fields to acquire and lease protobuf requests.
- Forwards the existing hostname/username identity from the coordinator.
- Logs successful acquisitions and leases at the exporter.

Builds on:
https://github.com/labgrid-project/labgrid/pull/1884


labgrid/remote/coordinator.py
```diff
@@ -1101,14 +1347,103 @@ async def PollReservation(self, request: labgrid_coordinator_pb2.PollReservation
...
+
+        reservation = self.reservations[token]
+
+        if reservation.owner != username:
+            await context.abort(
+                grpc.StatusCode.PERMISSION_DENIED,
+                f"Reservation {token} is owned by a different user",
+            )
+
+        if reservation.state is not ReservationState.leased:
```
ozan956: Could we also check reservation.expired before extending here? The reservation state is only changed to expired by the periodic scheduler. This means a lease can already be past its timeout while still having the leased state. During that interval, extend would move the timeout forward and effectively revive the expired lease. Rejecting it here would keep the API behavior aligned with the configured timeout. A small regression test would be helpful too.

doc/getting_started.rst
```diff
@@ -311,6 +311,14 @@ To interact with this place, it needs to be acquired first, this is done by
 
     labgrid-venv $ labgrid-client -p example-place acquire
 
+For short-lived access, a place can also be leased using a reservation:
+
+.. code-block:: bash
+
+    labgrid-venv $ labgrid-client reserve board=example --shell
+    labgrid-venv $ labgrid-client wait
+    labgrid-venv $ labgrid-client -p + lease
```
ozan956: I guess the second command going to fail, since we are not exposing the token

ozan956: maybe you can wrap the first command with eval?

labgrid/remote/coordinator.py
```diff
@@ -691,6 +749,72 @@ async def _acquire_resources(self, place, resources):
...
+        request = labgrid_coordinator_pb2.ExporterLeaseExtendedRequest()
+        request.group_name = resource.path[1]
+        request.resource_name = resource.path[3]
+        if place is not None:
+            request.place_name = place.name
+        request.duration = duration
+
+        cmd = ExporterCommand("lease_extended_request", request)
+        self.get_exporter_by_name(resource.path[0]).queue.put_nowait(cmd)
+        await cmd.wait()
```
ozan956: Could we handle the timeout from cmd.wait() as an ExporterError here, or catch TimeoutError in ExtendLease()? 

ExporterCommand.wait() raises asyncio.TimeoutError when the exporter does not respond within 10 seconds, but the cleanup path currently catches only ExporterError. As a result, an unresponsive exporter skips the lease-cancellation cleanup and the caller receives an unexpected RPC error. _lease_resource() already wraps this timeout, so applying the same pattern here may work well.

PR #1942, opened 2026-07-25
driver: don't assume target is set in __str__

Sanjays2402:
**Description**

`Driver.__attrs_post_init__` temporarily sets `self.target = None` while binding is in progress, so a driver's `__str__` must not dereference it. When `Target.bind_driver` raised `BindingError(f"{client} got unexpected bindings: ...")`, formatting that message called `SerialDriver.__str__` and blew up with `AttributeError: 'NoneType' object has no attribute 'name'`, hiding the real cause. `QemuDriver.__str__` had the same pattern, so both were fixed (as suggested in the issue).

Tested locally: the new `test_unexpected_binding_reports_binding_error` fails with the reported `AttributeError` without the source change and passes with it; `tests/test_serialdriver.py` and `tests/test_target.py` are green (52 passed).

**Checklist**
- [ ] Documentation for the feature
- [x] Tests for the feature
- [ ] The arguments and description in doc/configuration.rst have been updated
- [ ] Add a section on how to use the feature to doc/usage.rst
- [ ] Add a section on how to use the feature to doc/development.rst
- [x] PR has been tested
- [ ] Man pages have been regenerated

Fixes #1414

PR #1943, opened 2026-07-25
remote/client: let -p restrict env-wide acquire/release

### LINKED ISSUE #1868: -p/--place is silently ignored by acquire/release when -c env.yaml is also given
# Issue: `-p`/`--place` is silently ignored by `acquire`/`release` when `-c env.yaml` is also given

## Summary

Since commit `a9b59e4e` ("remote/client: make acquire/release operate on all RemotePlaces in env"), invoking `labgrid-client` with **both** `-c env.yaml` and `-p NAME`  silently ignores `-p` and operates on every `RemotePlace` in the env config. There is no way for the user to opt out of the env-wide behavior without dropping `-c` entirely.

## Affected version
- `master` at commit `a9b59e4e` and later.

## Reproducer
Coordinator with three empty places `test-place-1`, `test-place-2`, `test-place-3`.

`multi.yaml`:
```yaml
targets:
  main:
    resources:
      RemotePlace:
        name: test-place-1
  second:
    resources:
      RemotePlace:
        name: test-place-2
  third:
    resources:
      RemotePlace:
        name: test-place-3
```

```console
$ labgrid-client -c multi.yaml -p test-place-2 acquire
Selected role second from configuration file
acquired place test-place-1
acquired place test-place-2
acquired place test-place-3

$ labgrid-client who
User     Host     Place         Changed
labgrid  labgrid  test-place-1  ...
labgrid  labgrid  test-place-2  ...
labgrid  labgrid  test-place-3  ...
```

Expected: only `test-place-2` acquired.
Actual: all three places acquired; the `-p test-place-2` argument has no effect on which places are touched. `release` exhibits the same behavior.

## Root cause

`labgrid/remote/client.py`:

```python
async def acquire(self):
    errors = []
    places = self.get_place_names_from_env() if self.env else [self.args.place]
    for place in places:
        ...
```

## Impact

- CI pipelines and ad-hoc users that historically used `labgrid-client -c env.yaml -p NAME acquire` (a documented and idiomatic combination prior to `a9b59e4e`) silently acquire every place in the env, contending with other users / pipelines and locking resources that should remain free.
- The behavior change is silent: `-p` is parsed and even printed (`Selected role second from configuration file`), suggesting it took effect. There is no warning or error.

## Expected behavior

When the user explicitly supplies a place via `-p`/`--place`, `LG_PLACE`, or the legacy `PLACE` env var, `acquire`/`release` should operate on that single place only, even if `-c env.yaml` is also provided. 

## Proposed fix

Patch attached.

[0001-remote-client-let-p-restrict-env-wide-acquire-releas.patch](https://github.com/user-attachments/files/27699376/0001-remote-client-let-p-restrict-env-wide-acquire-releas.patch)

Sanjays2402:
**Description**

Since `acquire`/`release` started operating on all `RemotePlace`s in the environment config, passing both `-c env.yaml` and `-p NAME` silently ignored `-p` and acquired every place in the env — contending with other users and CI jobs that historically used that combination to manage one device at a time.

An explicitly requested place (`-p`/`--place`, `LG_PLACE`, or the legacy `PLACE` variable) now restricts `acquire`/`release` to that single place. Without one, the env-wide behaviour is unchanged.

**Checklist**
- [x] Tests for the feature
- [x] PR has been tested

Tested locally: the new `test_place_acquire_with_env_and_place` fails on master (both places acquired) and passes with the fix; `test_place_acquire`, `test_place_acquire_multiple` and `test_place_acquire_enforce` still pass.

Fixes #1868

### DIFF
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -506,6 +506,17 @@ class ClientSession:
 
         return places
 
+    def get_place_names_to_operate_on(self):
+        """Returns the place names acquire/release should operate on.
+
+        An explicitly requested place (via -p/--place, LG_PLACE or PLACE) always
+        wins, even when an environment config is given. Only without one do we
+        operate on every RemotePlace of the environment.
+        """
+        if self.env and not getattr(self.args, "explicit_place", False):
+            return self.get_place_names_from_env()
+        return [self.args.place]
+
     def get_idle_place(self, place=None):
         place = self.get_place(place)
         if place.acquired:
@@ -710,7 +721,7 @@ class ClientSession:
 
     async def acquire(self):
         errors = []
-        places = self.get_place_names_from_env() if self.env else [self.args.place]
+        places = self.get_place_names_to_operate_on()
         for place in places:
             try:
                 await self._acquire_place(place)
@@ -770,7 +781,7 @@ class ClientSession:
 
     async def release(self):
         errors = []
-        places = self.get_place_names_from_env() if self.env else [self.args.place]
+        places = self.get_place_names_to_operate_on()
         for place in places:
             try:
                 await self._release_place(place)
@@ -2274,6 +2285,10 @@ def main():
 
     if args.place is None:
         args.place = place
+
+    # remember whether the user explicitly selected a place, so that env-wide
+    # commands can be restricted to it
+    args.explicit_place = args.place is not None
 
     if args.state is None:
         args.state = state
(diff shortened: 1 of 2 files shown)


Emantor:
I don't think we want to support this case. Environment files describe the complete testing environment, if only a subset of places is required for the test run, this should be done in a new separate environment file.

Sanjays2402:
Fair enough, that's a reasonable line to hold. I read `-p` as already being a filter in the client, so extending it to acquire/release felt consistent, but if environment files are meant to describe the complete environment then a separate file is the cleaner answer.

Happy to close this. #1942 is unrelated and stands on its own.

PR #1944, opened 2026-07-25
Add support for PyOCD and labgrid-client reset subcommand

gumulka:
**Description**

This PR adds support for the [pyocd](https://pyocd.io/) command line tool. A Python based tool for programming Arm Cortex microcontrollers.

The command line tool was used instead of the API, because its integration was easier and it does not add a dependency to the project.

OpenOCD is great, but working with the config files, can become quite convoluted. PyOCD on the other hand tries to figure most stuff out on its own, but does take Hints like `--target` to define which MCU is targeted. This makes it much easier to use for programming devices using labgrids bootstrap mechanism.

`labgrid-client bootstrap <file>` has been extended to use pyocd, when a USBDebugging Interface is available on the exporter.

A new subcommand `labgrid-client reset` has been added to make use of the `ResetProtocol` to reset a target.

I'm currently using a single CMSIS DAP-Link controller for programming, power and serial. Since Serial and Power are done using the same USB-Port, the boot messages can only be read through a `reset` command and not with a power-cycle.

I did test some more complex commands like:
`labgrid-client bootstrap build/zephyr/zephyr.signed.hex target_name=nrf52832 load_commands="-e sector -a 0x6000"` and everything works fine.

**Checklist**
- [x] Tests for the feature 
- [ ] ~~Add a section on how to use the feature to doc/usage.rst~~
- [ ] ~~Add a section on how to use the feature to doc/development.rst~~
- [x] PR has been tested

PyOCD currently does not support distinguishing between adapters based on the USB-Path, but only based on serial. If there is no serial, a pseudo-serial is generated based on the path and some more data.
Maybe at some point it can be added, but not yet.

### DIFF
--- a/labgrid/driver/pyocddriver.py
+++ b/labgrid/driver/pyocddriver.py
@@ -52,7 +52,8 @@ class PyOCDDriver(Driver, BootstrapProtocol, ResetProtocol):
         # FIXME make sure we always have an environment or config
         if self.target.env:
             self.tool = self.target.env.config.get_tool("pyocd")
-            self.config = self.target.env.config.resolve_path(self.config)
+            if self.config:
+                self.config = self.target.env.config.resolve_path(self.config)
         else:
             self.tool = "pyocd"
 
@@ -60,9 +61,9 @@ class PyOCDDriver(Driver, BootstrapProtocol, ResetProtocol):
         cmd = [self.tool, subcommand]
         if self.serial:
             cmd += ["--uid", self.serial]
-        if self.target_name is not None and "--target" not in commands:
+        if self.target_name is not None and (not commands or "--target" not in commands):
             cmd += ["--target", self.target_name]
-        if self.frequency is not None and "--frequency" not in commands:
+        if self.frequency is not None and (not commands or "--frequency" not in commands):
             cmd += ["--frequency", self.frequency]
 
         if self.config is not None:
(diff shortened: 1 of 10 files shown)


labgrid/driver/pyocddriver.py
```diff
@@ -0,0 +1,108 @@
...
+            filename = self.target.env.config.get_image_path(self.image)
+
+        mf = ManagedFile(filename, self.interface)
+        mf.sync_to_resource()
+
+        if self.load_commands:
+            if isinstance(self.load_commands, str):
+                commands = self.load_commands.split()
+            else:
+                commands = self.load_commands
```
ozan956: I noticed this while testing two consecutive `load()` calls on the same `PyOCDDriver` instance with list-based `load_commands`.

For example:
```
    load_commands: ["--frequency", "100"]
```
Then, in a custom test script:
```
    driver.load("first.hex")
    driver.load("second.hex")
```
The first call mutates the configured list to:
```
    ["--frequency", "100", "first.hex"]
```
As a result, the second call invokes:
```
    pyocd load --frequency 100 first.hex second.hex
```
This does not occur across separate `labgrid-client bootstrap` invocations, since each CLI invocation creates a new process and driver instance. It can occur whenever the same driver object is reused programmatically.

ozan956 (commented):
Thanks for implementing this! LGTM overall, I left only one comment regarding `load_commands`.

labgrid/driver/pyocddriver.py
gumulka: Hey, thanks for pointing that out! I just [added a new commit](https://github.com/labgrid-project/labgrid/pull/1944/commits/f2abdc6868aae135558924bfa64e1102c547688f), that has a pytest and a fix for it.

jluebbe:
> PyOCD currently does not support distinguishing between adapters based on the USB-Path, but only based on serial. If there is no serial, a pseudo-serial is generated based on the path and some more data.
> Maybe at some point it can be added, but not yet.

Path-based matching is relatively important for labs with multiple places. Is there an upstream feature request for this? If not, please crate one and link to it here.

ozan956:
Hey @gumulka, I did not see any issues/PRs on PyOCD about that, are you going to open an issue on PyOCD to track this?

Emantor:
Hi, I looked at this with @Bastian-Krause and we were not entirely sure why the changes for reset are necessary, some more explanation would be great.

Issue #1945, opened 2026-07-27
Gradually add type annotations to the labgrid core


rpoisel:
I'd be willing to gradually add [type annotations](https://docs.python.org/3.10/library/typing.html) to the labgrid core package under [`labgrid/`](https://github.com/labgrid-project/labgrid/tree/master/labgrid).

Since labgrid now requires Python 3.10 or newer, modern annotation syntax could be used without introducing an additional compatibility dependency.

Improving type coverage would provide several benefits:

- Better IDE and [LSP](https://microsoft.github.io/language-server-protocol/) support, including navigation, completion, and inline documentation
- Earlier detection of incompatible arguments, return values, and attribute assignments
- Safer refactoring by making interfaces and assumptions more explicit
- Better support for static type checkers such as [ty](https://docs.astral.sh/ty/) and   [mypy](https://mypy.readthedocs.io/)
- Better information for downstream users calling labgrid as a Python library

I noticed that the repository already contains a basic [`mypy.ini`](https://github.com/labgrid-project/labgrid/blob/master/mypy.ini), although type coverage in the core package currently appears to be fairly minimal.

Before investing substantial effort, I'd appreciate the maintainers' feedback:

- Would gradually improving type coverage be welcome?
- Is there a preferred type checker or target configuration?
- Would you prefer small, module-specific pull requests?
- Are there particular parts of the public API that would be good starting points?

My intention would be to proceed incrementally, keeping individual pull requests reviewable and avoiding behavioral changes where possible.

Emantor:
We started of with type annotations in mind for labgrid, however at the time they were still new and it was hard to model some of the types correctly in conjunction with the `attrs` package if I remember correctly. If that has changed in the meantime I would be happy to review and merge type annotation PRs.

rpoisel:
> We started of with type annotations in mind for labgrid, however at the time they were still new and it was hard to model some of the types correctly in conjunction with the `attrs` package if I remember correctly. If that has changed in the meantime I would be happy to review and merge type annotation PRs.

Thanks for your quick and helpful response!

I'll soon prepare an initial PR to give you a better idea of my intended changes.

rpoisel:
@Emantor, in my view, the first PR (#1946) is ready. It should give you an idea what I meant by adding typing hints to the code base.

jluebbe:
@rpoisel Do you see a way to get helpful checks from CI without adding too much checker-specific ignore directives?

rpoisel:
[...]

PR #1946, opened 2026-07-28
driver: add initial type hints for common driver code

rpoisel:
**Description**

This PR starts adding type annotations incrementally, as discussed in #1945.

In later PRs, the [ty](https://docs.astral.sh/ty/) type checker will be added and activated for CI runs of this project.

Verified locally with:

``` shell
uv run --extra dev pytest tests/test_driver_common.py
uv run ruff check labgrid/driver/common.py tests/test_driver_common.py
uv run ruff format --check --diff labgrid/driver/common.py tests/test_driver_common.py
uv run ty check
```

**Checklist**

- [ ] Tests for the feature
- [x] PR has been tested

### DIFF
--- a/labgrid/driver/common.py
+++ b/labgrid/driver/common.py
@@ -1,8 +1,11 @@
 import logging
 import subprocess
+from typing import cast
+
 import attr
 
 from ..binding import BindingError, BindingMixin
+from ..resource import Resource
 from .exception import ExecutionError
 
 
@@ -21,7 +24,7 @@ class Driver(BindingMixin):
     - deactivate
     """
 
-    def __attrs_post_init__(self):
+    def __attrs_post_init__(self) -> None:
         super().__attrs_post_init__()
         if self.target is None:
             raise BindingError("Drivers can only be created on a valid target")
@@ -31,7 +34,7 @@ class Driver(BindingMixin):
             logger_name += f":{self.name}"
         self.logger = logging.getLogger(logger_name)
 
-    def get_priority(self, protocol):
+    def get_priority(self, protocol) -> int:
         """Retrieve the priority for a given protocol
 
         Arguments:
@@ -41,17 +44,17 @@ class Driver(BindingMixin):
             Int: value of the priority if it is found, 0 otherwise.
         """
         for cls in self.__class__.__mro__:
-            prios = getattr(cls, 'priorities', {})
+            prios = getattr(cls, "priorities", {})
             # we found a matching parent priorities attribute with the matching protocol
             if prios and protocol in prios:
-                return prios.get(protocol)
+                return cast(int, prios[protocol])
             # If we find the parent protocol, set the priority to 0
             if cls.__name__ == protocol.__name__:
                 return 0
 
         return 0
 
-    def get_export_name(self):
+    def get_export_name(self) -> str:
         """Get the name to be used for exported variables.
 
         Falls back to the class name if the driver has no name.
@@ -60,12 +63,12 @@ class Driver(BindingMixin):
             return self.name
         return self.__class__.__name__
 
-    def get_export_vars(self):
+    def get_export_vars(self) -> dict[str, str]:
         """Get a dictionary of variables to be exported."""
         return {}
 
     @property
-    def skip_deactivate_on_export(self):
+    def skip_deactivate_on_export(self) -> bool:
         """Drivers are deactivated on export by default.
 
         If the driver can handle external accesses even while active, it can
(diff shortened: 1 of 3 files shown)


Emantor (dismissed):
Changes look good to me, do you also want to add a `ty` check run to this PR for the CI or is it too early for that?

rpoisel:
> Do you also want to add a `ty` check run to this PR for the CI or is it too early for that?

Thanks for your approval!

I'd be happy to add such check to CI. If you don't mind, I'd do that in a separate commit.

EDIT: just re-pushed to get the commit of this PR verified.

labgrid/driver/common.py
```diff
@@ -60,29 +64,32 @@ def get_export_name(self):
...
+        res: set[Resource] = set()
         for supplier in self.suppliers:
             res |= supplier.get_bound_resources()
         return res
 
-def check_file(filename, *, command_prefix=[]):
-    if subprocess.call(command_prefix + ['test', '-r', filename]) != 0:
-        raise ExecutionError(f"File {filename} is not readable")
+
+def check_file(filename, *, command_prefix=None) -> None:
```
Bastian-Krause:
```suggestion
def check_file(filename, *, command_prefix: list[str] | None = None) -> None:
```

labgrid/driver/common.py
```diff
@@ -60,29 +64,32 @@ def get_export_name(self):
...
+    if command_prefix is None:
+        command_prefix: list[str] = []
+    if subprocess.call(command_prefix + ["test", "-r", filename]) != 0:
```
Bastian-Krause:
```suggestion
    if subprocess.call((command_prefix or []) + ["test", "-r", filename]) != 0:
```

labgrid/driver/common.py
```diff
@@ -1,7 +1,11 @@
 import logging
 import subprocess
+from typing import cast
+
 import attr
 
+from labgrid.resource.common import Resource
```
Bastian-Krause: Stick to relative imports like we do in the rest of the file:
```suggestion
from ..resource import Resource
```

labgrid/driver/common.py
```diff
@@ -21,17 +25,17 @@ class Driver(BindingMixin):
     - deactivate
     """
 
-    def __attrs_post_init__(self):
+    def __attrs_post_init__(self) -> None:
         super().__attrs_post_init__()
         if self.target is None:
-            raise BindingError("Drivers can only be created on a valid target")
+            raise BindingError("Drivers can only be created on a valid target")  # ty: ignore[too-many-positional-arguments]
```
Bastian-Krause: So we need to mark each instantiation of an attr.s class like that? I don't think that scales..

Doesn't mypy work better with that?

Bastian-Krause (changes requested):
> I'd be happy to add such check to CI. If you don't mind, I'd do that in a separate commit.

You didn't to that, did you?

Maybe we should drop the mypy.ini with this PR? (edit: depending on the mypy question below)

labgrid/driver/common.py
Emantor: AFAIR `mypy` had the same issues, thats why we never got anywhere with the initial type hints.

rpoisel:
> > I'd be happy to add such check to CI. If you don't mind, I'd do that in a separate commit.
> You didn't to that, did you?

No, I didn't because I want to do that in a [separate commit](https://github.com/labgrid-project/labgrid/pull/1946#issuecomment-5141368264).

> Maybe we should drop the mypy.ini with this PR?

I would leave it for now.

Now that you mention that: I think I will even remove the `ty` dependency entirely in this PR and add it in the (separate) PR that also adds the CI typing checks.

labgrid/driver/common.py
rpoisel: I'll accept your suggestions using Git directly.

Then, my commits will be digitally signed and "Verified" in the GitHub UI.

labgrid/driver/common.py
rpoisel: In the separate PR that will add `ty` as a dependency incl. CI check, I'll disable the `too-many-positional-arguments` check for the whole repo until there is clean support for attrs classes.

At the moment, `ty` does not really support attrs. See astral-sh/ty#2404 if you want to learn more.

jluebbe:
The current result looks good. Please squash the updates into the initial commit and update the commit message.

rpoisel:
> The current result looks good. Please squash the updates into the initial commit and update the commit message.

Thanks, it's done.

Bastian-Krause (approved):

labgrid/driver/common.py
Bastian-Krause: ok

PR #1949, opened 2026-07-31
client: fix header row placement in 'who' output

atline:
The 'who' command stored the header row as the first element of the result list and then called result.sort(), which sorted the header together with the data rows. Because string comparison is case sensitive (uppercase letters sort before lowercase), the 'User' header could end up between uppercase and lowercase user names instead of staying on top.

Sort only the data rows (result[1:]) so the header always remains the first line.

**Description**

Currently:
```
$ labgrid-client who
Freenix   nxa18884-linux-1                 imx8qm-mek-bu07               2026-07-28 21:48:50.118257
User      Host                             Place                         Changed
bamboo    lsv052263.swis.nl-cdc01.nxp.com  imx8qm-mek-bu06               2026-07-27 18:06:45.687974
```

After:
```
$ labgrid-client who
User      Host                             Place                         Changed
Freenix   nxa18884-linux-1                 imx8qm-mek-bu07               2026-07-28 21:48:50.118257
bamboo    lsv052263.swis.nl-cdc01.nxp.com  imx8qm-mek-bu06               2026-07-27 18:06:45.687974
```

**Checklist**
- [✓] PR has been tested

### DIFF
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -419,7 +419,7 @@ class ClientSession:
             if self.args.show_exporters:
                 exporters = {resource_path[0] for resource_path in place.acquired_resources}
                 result[-1].append(", ".join(sorted(exporters)))
-        result.sort()
+        result[1:] = sorted(result[1:])
 
         widths = [max(map(len, c)) for c in zip(*result)]
         layout = []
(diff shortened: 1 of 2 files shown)


atline:
I think using `with pexpect.spawn('python -m labgrid.remote.client who') as spawn` won't help with coverage reporting since it spawns an external process rather than running code in-process? That said, this seems to be the convention used throughout these tests.

Emantor:
I fixed subinterpreter coverage in a recent commit, so this should work while looking unintuitive.

Emantor (approved):

PR #1950, opened 2026-08-03
Use dedicated loggers for ClientSession/Place

Bastian-Krause:
**Description**
labgrid produces a lot of logs. Try to get rid of logs emitted by the root logger, so proper filtering and attribution is possible.

**Checklist**
- [x] PR has been tested

### DIFF
--- a/labgrid/remote/client.py
+++ b/labgrid/remote/client.py
@@ -132,6 +132,8 @@ class ClientSession:
         self.sync_id = itertools.count(start=1)
         self.sync_events = {}
 
+        self.logger = logging.getLogger("ClientSession")
+
     async def start(self):
         """Starts receiving resource and place updates from the coordinator."""
         self.resources = {}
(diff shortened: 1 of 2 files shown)


labgrid/remote/client.py
```diff
@@ -132,6 +132,8 @@ def __attrs_post_init__(self):
         self.sync_id = itertools.count(start=1)
         self.sync_events = {}
 
+        self.logger = logging.getLogger("ClientSession")
```
Emantor, line 135: Should we add a session identifier into the longer? f"ClientSession.{address}"?

Bastian-Krause: I can't think of a case with multiple ClientSessions in a single run, so I guess it's not required.

Emantor: Ah, I somehow thought this was on the coordinator side.

Emantor (approved):

pengutronix-mirror (dismissed):
Permission test

PR #1952, opened 2026-08-06
driver/rawnetworkinterfacedriver: improve debugging experience

Bastian-Krause:
**Description**
Small collection of patches allowing to see what commands are actually executed (via `processwrapper` and via the `labgrid-raw-interface` sudo helper).

**Checklist**
- [x] PR has been tested

/cc @JoshuaWatt

### DIFF
--- a/helpers/labgrid-raw-interface
+++ b/helpers/labgrid-raw-interface
@@ -9,6 +9,7 @@
 
 import argparse
 import os
+import shlex
 import string
 import sys
 import subprocess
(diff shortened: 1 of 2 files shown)


Emantor (approved):

Emantor:
@nick-potenski you might also want to take a look.

Bastian-Krause:
To limit the log messages on info level, we should add a `processwrapper.check_output(.., loglevel=)` and turn the "internal" calls to DEBUG level.

PR #1954, opened 2026-08-10
doc/usage: document fixtures on top of state-modifying fixtures as an anti pattern

Bastian-Krause:
**Description**
Add a warning to the strategy fixture section to avoid fixtures on top of state-modifying fixtures.

### DIFF
--- a/doc/usage.rst
+++ b/doc/usage.rst
@@ -699,6 +699,20 @@ For this example, you should get a report similar to this:
 
   ========================== 3 passed in 29.77 seconds ===========================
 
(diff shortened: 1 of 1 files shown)


Emantor (approved):

SmithChart (approved):
Thanks! This describes exactly the problem I ran into while debugging the LXA TAC tests for https://github.com/linux-automation/labgrid-lxatac/pull/37 .

In that case it seems that the fixture execution order must have changed at some point: Initially the modified test worked, but than failed after some time.

PR #1955, opened 2026-08-12
Coordinator Capability implementation

Emantor:
**Description**
Implement capabilities and checks on the coordinator side.
This relies on the existing identity pull request and will require changes.

Happy for any feedback on the implementation, I know that documentation is still missing, but at this point we should start discussing the implementation.

The Tests currently only really test the happy path, testing the denied path will become easier once we have policies, but I can try to figure this out for this PR if desired.

**Checklist**
- [x] Tests for the feature 
- [x] PR has been tested


labgrid/remote/coordinator.py
```diff
@@ -1244,6 +1337,7 @@ def main():
     parser.add_argument(
         "--pystuck-port", metavar="PORT", type=int, default=6666, help="use a different pystuck port than 6666"
     )
+    parser.add_argument("--capabilities", action="store_true", default=False, help="enable using capabilities, which also enforces client identities")
```
asher-pem-arm: what's the plan for this flag? is this intended as a migration mechanism before full auth?

labgrid/remote/coordinator.py
```diff
@@ -190,6 +191,23 @@ async def wrapper(self, *args, **kwargs):
 
     return wrapper
 
+async def check_capability(cls, identity, capability, context):
+    if not identity and cls.use_capabilities == True:
+        await context.abort(grpc.StatusCode.UNAUTHENTICATED, "Client identity is required when using capabilities")
+    if identity and (capability not in identity.capabilities):
+        await context.abort(grpc.StatusCode.PERMISSION_DENIED, f"Capability {req_cap} not in client capabilities {ctx.capabilities}")
```
asher-pem-arm: don't we want undefined here not `PERMISSION_DENIED`?

labgrid/remote/coordinator.py
```diff
@@ -1137,15 +1212,32 @@ async def CreateReservation(self, request: labgrid_coordinator_pb2.CreateReserva
...
+        try:
+            owner = infer_peer_identity(self.clients, context, client_identity_context)
+        except KeyError:
+            await context.abort(grpc.StatusCode.FAILED_PRECONDITION, f"Peer {peer} does not have a valid session")
+
+        owned = self.reservations[token].owner == owner
+        if self.use_capabilities:
+            if not identity:
+                await context.abort(grpc.StatusCode.UNAUTHENTICATED, "Client identity is required when using capabilities")
+            if owned and not Capability.cancel_reservation_owned in identity.capabilities:
```
asher-pem-arm: should that not be :
```
if owned and not (
    Capability.cancel_reservation_owned in identity.capabilities
    or Capability.cancel_reservation_any in identity.capabilities
):
```

labgrid/remote/coordinator.py
```diff
@@ -980,10 +1043,20 @@ async def AllowPlace(self, request, context):
...
+        owned = place.acquired == username
+        if not owned and not self.use_capabilities:
             await context.abort(
                 grpc.StatusCode.FAILED_PRECONDITION, f"Place {placename} is not acquired by {username}"
             )
+
+        if self.use_capabilities:
+            if not identity:
+                await context.abort(grpc.StatusCode.UNAUTHENTICATED, "Client identity is required when using capabilities")
+            if owned and not Capability.allow_place_owned in identity.capabilities:
```
asher-pem-arm: should that not be:
```
if owned and not (
    Capability.allow_place_owned in identity.capabilities
    or Capability.allow_place_any in identity.capabilities
):
```

labgrid/remote/coordinator.py
```diff
@@ -954,6 +998,22 @@ async def ReleasePlace(self, request, context):
         if fromuser and place.acquired != fromuser:
             return labgrid_coordinator_pb2.ReleasePlaceResponse()
 
+        try:
+            username = infer_peer_identity(self.clients, context, client_identity_context)
+        except KeyError:
+            await context.abort(grpc.StatusCode.FAILED_PRECONDITION, f"Peer {peer} does not have a valid session")
```
asher-pem-arm: peer undefined

labgrid/remote/coordinator.py
```diff
@@ -1137,15 +1212,32 @@ async def CreateReservation(self, request: labgrid_coordinator_pb2.CreateReserva
...
         token = request.token
         if not isinstance(token, str) or not token:
             await context.abort(grpc.StatusCode.INVALID_ARGUMENT, f"Invalid token {token}")
         if token not in self.reservations:
             await context.abort(grpc.StatusCode.FAILED_PRECONDITION, f"Reservation {token} does not exist")
+
+        try:
+            owner = infer_peer_identity(self.clients, context, client_identity_context)
+        except KeyError:
+            await context.abort(grpc.StatusCode.FAILED_PRECONDITION, f"Peer {peer} does not have a valid session")
```
asher-pem-arm: peer undefined

asher-pem-arm (commented):
picked up a few small issues but overall looks good. It would be helpful to document what is planned to be included in this PR/future ones to help us clarify the review scope

labgrid/remote/coordinator.py
Emantor: Yes, with this flag we indicate that the new identities are required to enable capabilities.

asher-pem-arm: Any reason why we would want a flag? If capabilities are populated by the configured authentication plugin, the coordinator could always enforce them. The default/legacy plugin can preserve current behaviour by granting all capabilities, while secure plugins return restricted sets. With a flag we could get into issues such as auth enabled but capability checks accidentally disabled.

labgrid/remote/coordinator.py
Emantor: I can't find an undefined error code in https://grpc.io/docs/guides/status-codes/, and `PERMISSION_DENIED` should be a correct code for a user that wants to run an RPC call for which he does not have the capability.

labgrid/remote/coordinator.py
Emantor: I was debating that, but it probably makes more sense and will not leave the user wondering if he has `_any` and not `_owned`.

Fixed in fixup commit.

labgrid/remote/coordinator.py
Emantor: Same as below, fixed in fixup commit.

labgrid/remote/coordinator.py
```diff
@@ -182,6 +192,28 @@ async def wrapper(self, *args, **kwargs):
     return wrapper
 
 
+async def check_capability(cls, identity, capability, context):
+    if not identity and cls.use_capabilities == True:
+        await context.abort(grpc.StatusCode.UNAUTHENTICATED, "Client identity is required when using capabilities")
+    if identity and (capability not in identity.capabilities):
+        await context.abort(
+            grpc.StatusCode.PERMISSION_DENIED, f"Capability {req_cap} not in client capabilities {ctx.capabilities}"
```
asher-pem-arm, line 194: `req_cap` and  `ctx` undefined

asher-pem-arm (commented):
Can you clarify the intended plugin-wiring boundary? The coordinator, client, and exporter currently instantiate identity interceptors and the server interceptor parses metadata directly into `ClientIdentity`. Is this deliberately temporary, or should the interceptors accept injected `ClientAuthPlugin`/`ServerAuthPlugin` implementations now with plugin discovery and configuration added later?

Emantor:
Yes, I deliberatly added the capabilities to the identity in the interceptor to have something to test against, the intention is to revert that change and have the plugins handle the capability fillout for the identity.

Your suggestion to take this a step further and introduce dummy plugins to prevent any confusion on how the handling is supposed to be done makes sense if I understand it correctly, I will look into it.

labgrid/remote/coordinator.py
Emantor: I think the naming here may be wrong, in essence this enforces the new authentication method and disables the fallback since only the new client identities can have capabilities and we require this feature switch for backwards compatibility at least in the initial release.

labgrid/remote/coordinator.py
```diff
@@ -183,6 +186,29 @@ async def wrapper(self, *args, **kwargs):
...
+        )
+
+
+def require_capability(req_cap):
+    def decorator(func):
+        @wraps(func)
+        async def wrapper(self, request, context):
+            identity = ClientIdentity.from_metadata(context.invocation_metadata())
```
asher-pem-arm, line 202: We don't check if identity is set here.
Ideally we would use a shared helper for all these calls that catches `NoIdentityPresent` and use `identity = None`

labgrid/remote/coordinator.py
```diff
@@ -410,9 +463,22 @@ async def ExporterStream(self, request_iterator, context):
         out_msg.hello.version = labgrid_version()
         yield out_msg
 
+        identity = ClientIdentity.from_metadata(context.invocation_metadata())
```
asher-pem-arm, line 466: identity assigned above in line 453

labgrid/remote/coordinator.py
```diff
@@ -1078,25 +1227,58 @@ async def CreateReservation(self, request: labgrid_coordinator_pb2.CreateReserva
...
         if not isinstance(token, str) or not token:
             await context.abort(grpc.StatusCode.INVALID_ARGUMENT, f"Invalid token {token}")
         if token not in self.reservations:
             await context.abort(grpc.StatusCode.FAILED_PRECONDITION, f"Reservation {token} does not exist")
+
+        try:
+            owner = infer_peer_identity(self.clients, context, identity)
+        except KeyError:
+            await context.abort(
+                grpc.StatusCode.FAILED_PRECONDITION, f"Peer {context.peer} does not have a valid session"
```
asher-pem-arm, line 1252: should be context.peer()

asher-pem-arm (commented):
@Emantor have added some review comments on some small things that might have been missed, have left the tests alone as I think they're still WIP but let me know if you would like those reviewed as well

PR #1687, opened 2025-07-08
pyproject.toml: use non-deprecated license specification

Bastian-Krause:
**Description**
Fixes these deprecation warnings during `python -m build`:
```
env/lib/python3.11/site-packages/setuptools/config/_apply_pyprojecttoml.py:82: SetuptoolsDeprecationWarning: `project.license` as a TOML table is deprecated
!!
        ********************************************************************************
        Please use a simple string containing a SPDX expression for `project.license`. You can also use `project.license-files`. (Both options available on setuptools>=77.0.0).
        By 2026-Feb-18, you need to update your project and remove deprecated calls
        or your builds will no longer be supported.
        See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
        ********************************************************************************
!!
```
```
env/lib/python3.11/site-packages/setuptools/config/_apply_pyprojecttoml.py:61: SetuptoolsDeprecationWarning: License classifiers are deprecated.
!!
        ********************************************************************************
        Please consider removing the following classifiers in favor of a SPDX license expression:
        License :: OSI Approved :: GNU Lesser General Public License v2 or later (LGPLv2+)
        See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
        ********************************************************************************
!!
```

**Checklist**
- [ ] PR has been tested

Closes #1956

PR #1956, opened 2026-08-14
pyproject.toml: migrate to PEP 639 license expressions

ejoerns:
**Description**

Make `license` a proper SDPX license expression.
Use `license-files` for license files.
Replace the also-deprecated `License ::` classifier.

See https://setuptools.pypa.io/en/latest/userguide/license_migration.html

Fixes:

```
SetuptoolsDeprecationWarning: `project.license` as a TOML table is deprecated
  !!

  ********************************************************************************
  Please use a simple string containing a SPDX expression for `project.license`. You can also use `project.license-files`. (Both options available on setuptools>=77.0.0).

  By 2027-Feb-18, you need to update your project and remove deprecated calls
  or your builds will no longer be supported.

  See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
  ********************************************************************************
```
```
SetuptoolsDeprecationWarning: License classifiers are deprecated.
  !!

  ********************************************************************************
  Please consider removing the following classifiers in favor of a SPDX license expression:

  License :: OSI Approved :: GNU Lesser General Public License v2 or later (LGPLv2+)

  See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
  ********************************************************************************
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,7 +15,8 @@ authors = [
 ]
 description = "embedded systems control library for development, testing and installation"
 readme = "README.rst"
-license = { file="LICENSE" }
+license = "LGPL-2.1-or-later"
+license-files = [ "LICENSE" ]
 requires-python = ">=3.10"
 classifiers = [
     "Intended Audience :: Developers",
@@ -29,7 +30,6 @@ classifiers = [
     "Programming Language :: Python :: 3.12",
     "Programming Language :: Python :: 3.13",
     "Programming Language :: Python :: 3.14",
-    "License :: OSI Approved :: GNU Lesser General Public License v2 or later (LGPLv2+)",
 ]
 dependencies = [
     "attrs>=21.4.0",


Emantor (approved):

Bastian-Krause (changes requested):
Duplicate of #1687. Needs investigation if the situation changed since then, see the comments in that PR.

Emantor:
Looks like the current recommendation for our case is to use `license-files` exclusively and drop `license`, see https://github.com/pypa/setuptools/commit/a9ac94417b2533689a27a6ce5a68cb12987efa99#diff-6d257eba84cbc108227977885d3a690248c666c3bf8c612fc3fcfd38264d547dR104

Bastian-Krause:
Looking at https://layers.openembedded.org/layerindex/recipe/37586/, this would likely break scarthgap.

Bastian-Krause:
> Looking at https://layers.openembedded.org/layerindex/recipe/37586/, this would likely break scarthgap.

I don't think we should wait until that's EOL, but maybe a little bit longer so everybody had enough time to switch to wrynose?

I think a certain someone will need meta-labgrid on scarthgap with a recent enough version of labgrid soon. Bumping setuptools there might cause fallout for other Python packages, so I guess I would need to revert this patch then if we merge it now.

Emantor:
I thought license-files would be supported for every version of setup tools that supports project.toml?

PR #1958, opened 2026-08-16
OpenOCDDriver: disable default server ports

ozan956:
**Description**

OpenOCD starts GDB, telnet, and TCL servers on fixed default ports. When
multiple boards on the same exporter bootstrap concurrently, the second
OpenOCD instance fails to bind these ports.

Disable the server ports before loading OpenOCD configuration files.
OpenOCDDriver does not provide a lifecycle for using or stopping these
servers, while disabling them allows parallel bootstrap operations.

Configurations that intentionally require a server interface can re-enable
the required port before `init`.

**Checklist**

- [ ] Tests for the feature
- [x] PR has been tested

Tested with:

```
pytest -q tests/test_openocd.py
ruff check labgrid/driver/openocddriver.py
ruff format --check labgrid/driver/openocddriver.py
```

### DIFF
--- a/labgrid/driver/openocddriver.py
+++ b/labgrid/driver/openocddriver.py
@@ -55,6 +55,14 @@ class OpenOCDDriver(Driver, BootstrapProtocol):
         cmd = [self.tool]
         cmd += chain.from_iterable(("--search", path) for path in self.search)
         cmd += self._get_usb_path_cmd()
+        cmd += [
+            "--command",
+            "gdb_port disabled",
+            "--command",
+            "telnet_port disabled",
+            "--command",
+            "tcl_port disabled",
+        ]
 
         managed_configs = []
         for config in self.config:
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -2318,6 +2318,12 @@ OpenOCDDriver
 An :any:`OpenOCDDriver` controls *OpenOCD* to bootstrap a target with a
 bootloader.
 
+The driver disables OpenOCD's GDB, telnet and TCL server ports. It does not
+provide a lifecycle to expose or stop these server interfaces, and disabling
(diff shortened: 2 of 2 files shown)


Emantor (commented):
That change makes sense to me, a different option would be to rewrite the exporter to control the openOCD invocation similar to how we handle serial ports with ser2net.

Emantor (dismissed):

Emantor:
This currently breaks tests and there is a DCO complaint.

ozan956:
> That change makes sense to me, a different option would be to rewrite the exporter to control the openOCD invocation similar to how we handle serial ports with ser2net.

Thanks, that is a good point! However, that would be a broader architectural change. Since OpenOCDDriver currently invokes OpenOCD directly and does not use those server interfaces, I would prefer to keep this PR focused and consider exporter-managed OpenOCD invocation in a separate one.

labgrid/driver/openocddriver.py
```diff
@@ -1,11 +1,12 @@
 from itertools import chain
+
 import attr
 
 from ..factory import target_factory
 from ..protocol import BootstrapProtocol
 from ..step import step
-from ..util.managedfile import ManagedFile
 from ..util.helper import processwrapper
+from ..util.managedfile import ManagedFile
```
Emantor, line 7: Why the resorting of the imports? This should be a separate commit or be dropped completely.

Background is that we might want to exclude these cleanup commits from git blame to retain more history.

Emantor (changes requested):

labgrid/driver/openocddriver.py
ozan956: Understood 👍 Thanks for explanation. Addressed it.

Emantor (approved):

jluebbe (approved):

PR #1959, opened 2026-08-21
resource/remote: handle unknown resource

Svendsen991:
**Description**

Instead of failing to create a RemotePlaceManager print out a warning on unknown remote resources and continue.

This will allow use of standard tools like labgrid-client's console together with a remote place where some of its resources are unknown/custom.

This will add value for projects using not yet upstreamed resources or if they're simply too custom for upstream. So useful labgrid-client tools can still be used without any workarounds.

Current workaround is importing labgrid client main in custom project where the custom resources are registered and create custom command to run labgrid-client.

**Checklist**

- [x] PR has been tested locally

It was tested by adding an unknown resource to an exporter and adding it to a place with a serial connection like
```yaml
# Exporter yaml
DUT-X:
  linux:
    cls: 'USBSerialPort'
    match:
      ID_PATH: 'X'
    speed: 115200
  wifi:
    cls: 'WifiService'
    address: 'x'
    username: 'root'
    ssid: 'ssidX'
```

### DIFF
--- a/labgrid/resource/remote.py
+++ b/labgrid/resource/remote.py
@@ -3,6 +3,7 @@ import os
 import attr
 
 from ..factory import target_factory
+from ..exceptions import InvalidConfigError
 from .common import NetworkResource, ManagedResource, ResourceManager
 
 
@@ -47,8 +48,13 @@ class RemotePlaceManager(ResourceManager):
         resource_entries = self.session.get_target_resources(place)  # pylint: disable=no-member
         expanded = []
         for (resource_name, _), resource_entry in resource_entries.items():
-            new = target_factory.make_resource(
-                remote_place.target, resource_entry.cls, resource_name, resource_entry.args)
+            try:
+                new = target_factory.make_resource(
+                    remote_place.target, resource_entry.cls, resource_name, resource_entry.args)
+            except InvalidConfigError:
+                self.logger.warning("unknown resource cls: %s, name: %s", resource_entry.cls, resource_name)
+                continue
+
             new.parent = remote_place
             new.avail = resource_entry.avail
             new.extra = resource_entry.extra


jluebbe:
This would mean that if you run the same testsuite in different environments (perhaps some outdated and missing new drivers), you'd just get a warning and then surprising errors when it tries to bind the drivers.

So, I don't think simply warning and continuing there is a viable approach.

Svendsen991:
> This would mean that if you run the same testsuite in different environments (perhaps some outdated and missing new drivers), you'd just get a warning and then surprising errors when it tries to bind the drivers.
> 
> So, I don't think simply warning and continuing there is a viable approach.
 
In an outdated environment. The most normal case where only new resources has been added. It should still work as expected.
If you modify or delete resources in your coordinator you can get surprising errors regardless and it's not guaranteed to be in the remote `on_resource_added` anyway.

Also I see these possible errors[1]? And I find them as informative if not more than `InvalidConfigError(f"unknown resource class {resource}")`.


[1]
When cls is a string
https://github.com/labgrid-project/labgrid/blob/master/labgrid/target.py#L123
https://github.com/labgrid-project/labgrid/blob/master/labgrid/factory.py#L166
```
raise KeyError(f"No driver/resource/protocol of type '{string}' in factory, perhaps not registered?")
```

When having access to cls (I guess here it will not be outdated because you have access to the class)
https://github.com/labgrid-project/labgrid/blob/master/labgrid/target.py#L147
```
raise NoResourceFoundError(
     f"no {cls.__name__} resource{name_msg} found in {self}"
)
```

Emantor:
> > This would mean that if you run the same testsuite in different environments (perhaps some outdated and missing new drivers), you'd just get a warning and then surprising errors when it tries to bind the drivers.
> > So, I don't think simply warning and continuing there is a viable approach.
> 
> In an outdated environment. The most normal case where only new resources has been added. It should still work as expected. If you modify or delete resources in your coordinator you can get surprising errors regardless and it's not guaranteed to be in the remote `on_resource_added` anyway.
> 
> Also I see these possible errors[1]? And I find them as informative if not more than `InvalidConfigError(f"unknown resource class {resource}")`.
> 
> [1] When cls is a string https://github.com/labgrid-project/labgrid/blob/master/labgrid/target.py#L123 https://github.com/labgrid-project/labgrid/blob/master/labgrid/factory.py#L166
> 
> ```
> raise KeyError(f"No driver/resource/protocol of type '{string}' in factory, perhaps not registered?")
> ```

This very much indicates that we expect you to register your driver/resources with labgrid, which can be done by importing them in the environment file and supplying that to the client.

> When having access to cls (I guess here it will not be outdated because you have access to the class) https://github.com/labgrid-project/labgrid/blob/master/labgrid/target.py#L147
> 
> ```
> raise NoResourceFoundError(
>      f"no {cls.__name__} resource{name_msg} found in {self}"
> )
> ```

This is also solved by registering your own resource via the imports key in the environment file.

Issue #1851, opened 2026-04-10
`Steps` stack is not thread-safe


hamztouiz:
### Description

The global `Steps._stack` in `labgrid/step.py` is a plain Python list shared across all threads. I'm hitting an `AssertionError` in `Steps.pop()` intermittently, and I believe it is caused by a race condition between threads.

I'm not sure if my usage of `USBVideoDriver` is the intended one — I'm running a continuous background image capture (OpenCV/GStreamer) while simultaneously calling other `@step`-decorated drivers from the main thread. It's possible this use case was not anticipated, and that `Steps` was designed assuming single-threaded usage only.

That said, if other scenarios can also trigger background threads (e.g. async callbacks, GStreamer internals), this could affect more users.

### My setup

- `USBVideoDriver` is used to capture frames continuously in a background thread (via OpenCV/GStreamer)
- While the capture is running, the main thread calls `USBSDMuxDriver.set_mode()` (or any other `@step`-decorated method)
- Both threads end up pushing/popping on the same shared `_stack`, corrupting its order

### Traceback

```
File ".../labgrid/step.py", line 222, in wrapper
    step.stop()
File ".../labgrid/step.py", line 186, in stop
    steps.pop(self)
File ".../labgrid/step.py", line 29, in pop
    assert self._stack[-1] is step
AssertionError
```

### Environment

- labgrid version: 25.0.1
- Python: 3.13
- OS: Ubuntu
- Test runner: Robot Framework

### Question

Is it intentional that `Steps` is not thread-safe? Is my usage of `USBVideoDriver` in a background capture thread considered unsupported?

If making `_stack` thread-local is acceptable, I'd be happy to submit a PR:

```python
import threading

class Steps:
    def __init__(self):
        self._local = threading.local()
        self._subscribers = []

    @property
    def _stack(self):
        if not hasattr(self._local, 'stack'):
            self._local.stack = []
        return self._local.stack
```

mbodmer:
I have multiple targets which I use in parallel (e.g. reboot all targets into u-boot, or execute an ssh command on all of them) and also would need a threadsafe Steps stack for this.

For now I use a monkeypatch in conftest.py based on your proposal, like this:

```py
import importlib
import threading
import warnings

try:
    _lg_step = importlib.import_module("labgrid.step")
    Steps = getattr(_lg_step, "Steps", None)
    if Steps is None:
        raise AttributeError("labgrid.step module has no attribute 'Steps'")

[...]

PR #1960, opened 2026-08-24
step: make Steps._stack thread-local for thread safety

cnabers:
Replace the shared stack list with a threading.local()-backed property so that each thread maintains its own step stack, preventing a fatal race condition in multi-threaded usage. 

I first surfaced this issue when attempting to boot two QEMU targets in parallel using the `QEMUDriver`.
Essentially doing this:

```
import threading

def _run(target):
    driver = target.get_driver(QEMUDriver)
    driver.on()   # @step-decorated -- pushes/pops onto the shared Steps._stack

threads = [threading.Thread(target=_run, args=(t,)) for t in targets.values()]
for t in threads:
    t.start()
for t in threads:
    t.join()

```

When I tried running qemu boot - each on one thread - it crashed since the threads were modifying the same stack list at the same time for separate targets. This applies to all `@step` decorated functions throughout the repo.

By applying this patch we will essentially have a stack-per-thread setup and the concurrently executing threads do not step on eachother via `@step`. Current single threaded behavior remains unaffected.

**Checklist**
- [ ] Tests for the feature 

- [ ] PR has been tested

### DIFF
--- a/labgrid/step.py
+++ b/labgrid/step.py
@@ -1,5 +1,6 @@
 import inspect
 import os
+import threading
 import warnings
 from functools import wraps
 from time import monotonic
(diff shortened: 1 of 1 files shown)


Bastian-Krause:
Sorry, we won't add threading support for labgrid. It would complicate the code (and maintenance effort) and we don't see a convincing use case for it.

Duplicate of #1294, #1851, #980.

PR #1961, opened 2026-08-25
protocol: add PortForwardProtocol

alextercete:
**Description**

Introduce `PortForwardProtocol` and implement it in `SSHDriver` and `ADBDriver`.

`SSHDriver` now uses automatic allocation for remote ports and reads the allocated port from `ssh -O forward`. The existing methods are kept as deprecated compatibility wrappers.

`ADBDriver` uses `adb forward` and `adb reverse`. For `RemoteUSBADBDevice`, it chains ADB and SSH port forwards through the exporter.

`labgrid-client forward` now uses a configured `PortForwardProtocol` driver, with implicit SSH forwarding retained as a compatibility fallback.

**Checklist**

- [x] Tests for the feature
- [x] The description in `doc/configuration.rst` has been updated
- [x] PR has been tested

### DIFF
--- a/labgrid/driver/sshdriver.py
+++ b/labgrid/driver/sshdriver.py
@@ -1,4 +1,4 @@
-"""The SSHDriver uses SSH as a transport to implement CommandProtocol and FileTransferProtocol"""
+"""The SSHDriver uses SSH as a transport to implement CommandProtocol, FileTransferProtocol and PortForwardProtocol"""
 import contextlib
 import os
 import re
@@ -8,12 +8,13 @@ import shutil
 import subprocess
 import tempfile
 import time
+import warnings
 from functools import cached_property
 
 import attr
 
 from ..factory import target_factory
-from ..protocol import CommandProtocol, FileTransferProtocol
+from ..protocol import CommandProtocol, FileTransferProtocol, PortForwardProtocol
 from .commandmixin import CommandMixin
 from .common import Driver
 from ..step import step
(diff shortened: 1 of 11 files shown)


labgrid/util/ssh.py
```diff
@@ -381,18 +381,24 @@ def add_remote_port_forward(self, remote_port, local_port, remote_bind=None):
...
-        self._r_forwards.add(forward)
+        stdout = self._run_socket_command("forward", [forward])
+        if remote_port == 0:
+            try:
+                remote_port = int(stdout.strip())
+            except ValueError:
+                self._run_socket_command("cancel", [forward])
+                raise
+
+        self._r_forwards[remote_port, local_port, remote_bind] = forward
```
jluebbe, line 392: This needs a comment why we need to store the original forward string.

labgrid/driver/sshdriver.py
```diff
@@ -291,43 +292,71 @@ def forward_local_port(self, remoteport, localport=None):
...
         with self._forward(forward):
-            yield localport
+            yield local_port
 
     @Driver.check_active
     @contextlib.contextmanager
-    def forward_remote_port(self, remoteport, localport):
+    def forward_local_port(self, remoteport, localport=None):
+        warnings.warn(
+            "SSHDriver.forward_local_port() is deprecated, use local_forward() instead",
```
jluebbe, line 317: Why do we need to deprecate the existing functions, instead of just making them explicit in the new Protocol?

alextercete: I wanted the protocol to support auto allocation when port `0` is requested, and make that the default. But the parameters in the existing `SSHDriver` methods didn't allow for that without breaking backwards compatibility.

I considered keeping the old method names and inspecting the parameters to make the method satisfy both the new protocol and the old signature, but I figured that was unnecessary complexity for what's probably niche functionality. So deprecation seemed the best route.

Happy to change the approach.

alextercete: @jluebbe Any thoughts on how to move forward?

labgrid/util/proxy.py
```diff
@@ -25,6 +26,46 @@ def force_proxy(cls, force_proxy):
         assert isinstance(force_proxy, str)
         cls._force_proxy = force_proxy
 
+    @classmethod
+    @contextlib.contextmanager
+    def local_forward(cls, res, remote_host, remote_port, *, local_port=0):
```
ozan956, line 31: Should these helpers also account for [_force_proxy](https://github.com/labgrid-project/labgrid/pull/1961/changes#diff-380de7c22836d01d8da1aaafc15ed0504540327b74c891ddbbfd1a0266358d2aR22) (the `LG_PROXY`) ? `get_host_and_port()` falls back to `LG_PROXY` when no resource-specific proxy is required, but `local_forward()` and `remote_forward()` seem to connect directly to `res.host` in that case. For a `RemoteUSBADBDevice`, would not this make the ADB connection work through `LG_PROXY` while port forwarding still requires direct SSH access to the exporter?

alextercete: Good catch, I hadn't considered `LG_PROXY`. We could potentially use the proxy as a jump host (`-J` flag)?

PR #1963, opened 2026-08-30
build: add uv support

rpoisel:
Add uv.lock and document the workflow in doc/development.rst so `uv sync`/`uv run` work out of the box for development.

[uv](https://docs.astral.sh/uv/) improves on the existing virtualenv/pip workflow:

- Reproducible builds: uv.lock pins exact resolved versions, so every contributor and CI run installs the same packages.
- Auditability: uv.lock's exact, hash-pinned dependency graph (including transitive deps) makes CVE scanning accurate and complete, e.g. by using [uv audit](https://astral.sh/blog/uv-audit) (supply chain security).
- Speed: cached packages reinstall in milliseconds instead of seconds, improving developer experience (DX).

This is backwards-compatible: `pip install -e ".[dev]"` still works.

CI now also runs the full lint/format/test suite through uv (reusable-unit-tests-uv.yml), using `--locked` to fail on lockfile issues and a uv.lock-keyed cache to demonstrate the speed improvement.

xenavalkyrie (xena extra) pins pypacker==4.9, whose invalid python_requires metadata fails to build under modern setuptools. Override pypacker to >=5.4 to fix `uv sync` for everyone.

**Checklist**
- [x] Tests for the feature
- [x] PR has been tested

### DIFF
--- a/.github/workflows/reusable-unit-tests-uv.yml
+++ b/.github/workflows/reusable-unit-tests-uv.yml
@@ -1,0 +1,51 @@
+name: reusable unit tests (uv)
+
+on:
+  workflow_call:
+    inputs:
+      branch:
+        type: string
+        required: false
+
+jobs:
+  build:
+    runs-on: ubuntu-24.04
+    continue-on-error: false
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        ref: ${{ inputs.branch }}
+    - name: Install uv
+      uses: astral-sh/setup-uv@v10.0.1
+      with:
+        enable-cache: true
+        cache-dependency-glob: "uv.lock"
+    - name: Install system dependencies
+      run: |
+        sudo apt-get install -yq libow-dev openssh-server openssh-client graphviz openocd
+        sudo mkdir -p /var/cache/labgrid/runner && sudo chown runner /var/cache/labgrid/runner
+    - name: Prepare local SSH
+      run: |
+        # the default of 777 is too open for SSH
+        chmod 755 ~
+        ssh-keygen -f ~/.ssh/id_ed25519.local -t ed25519 -N ""
+        cat ~/.ssh/id_ed25519.local.pub >> ~/.ssh/authorized_keys
+        echo -e "Host localhost ip6-localhost\n  Hostname 127.0.0.1\n  IdentityFile ~/.ssh/id_ed25519.local\n  UserKnownHostsFile ~/.ssh/known_hosts.local" >> ~/.ssh/config
+        ssh -o StrictHostKeyChecking=no localhost echo OK
+    - name: Enable unprivileged user namespaces
+      run: |
+        sudo sysctl kernel.unprivileged_userns_clone=1
+        unshare -Un true
+    - name: Install labgrid with uv
+      run: |
+        uv sync --extra dev --locked
+    - name: Lint with pylint
+      run: |
+        uv run pylint --list-msgs-enabled
+        uv run pylint labgrid
+    - name: Format with ruff
+      run: |
+        uv run ruff format --check --diff
+    - name: Test with pytest
+      run: |
+        uv run pytest -r a --local-sshmanager --ssh-username runner -k "not test_docker_with_daemon"
--- a/tests/test_coordinator.py
+++ b/tests/test_coordinator.py
@@ -1,8 +1,13 @@
-import pytest
+import asyncio
+import logging
+from unittest.mock import AsyncMock, MagicMock
 
 import grpc
-import labgrid.remote.generated.labgrid_coordinator_pb2_grpc as labgrid_coordinator_pb2_grpc
+import pytest
+
 import labgrid.remote.generated.labgrid_coordinator_pb2 as labgrid_coordinator_pb2
+import labgrid.remote.generated.labgrid_coordinator_pb2_grpc as labgrid_coordinator_pb2_grpc
+from labgrid.remote.coordinator import Coordinator, ExporterError, ExporterSession, warn_if_slow
 
 
 @pytest.fixture(scope="function")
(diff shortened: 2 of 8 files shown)


.github/workflows/reusable-unit-tests-uv.yml
ozan956: I believe this workflow is to only show the diff and going to be removed, right?

uv.lock
ozan956: this is something nasty to maintain 😄

tests/test_coordinator.py
ozan956, line 1: I would separate this into another PR, with more explanatory description in the commit

ozan956 (commented):
I like uv (it being fast, easy to use, etc.)

With the locked env, everybody using the existing pip based flow may end up with a different dependency set than the ci. Then there is a chance that this may lead to tests passing locally but failing in ci, or vice versa, right?

Just out of curiosity, did you run into a dependency issue on your side where locking the environment was necessary?

tests/test_coordinator.py
rpoisel: This test was necessary to not decrease the test coverage - that's mentioned in the commit message. Would you expect this explanation elsewhere?

If you are willing to accept some decrease in test coverage (less than 0.5%), we can move this test to a dedicated PR.

rpoisel:
Thanks, that is a good point.

> With the locked env, everybody using the existing pip based flow may end up with a different dependency set than the ci. Then there is a chance that this may lead to tests passing locally but failing in ci, or vice versa, right?

Yes. The current `pip install -e ".[dev]"` workflow resolves the version ranges from `pyproject.toml` at installation time. The resulting dependency set may therefore change when new versions are published, i.e. people will get different results then the CI of this GitHub project.

This PR does not replace the existing pip-based CI matrix, though. It adds the locked `uv` job alongside it. I think the two approaches provide complementary incentives:

- The locked `uv` job provides a <ins>reproducible</ins> baseline for development and CI.
- The existing pip jobs test labgrid against the dependency versions currently resolved from its declared constraints.

The latter is particularly useful for a library, since testing only the locked environment could hide incompatibilities with newer dependency versions. Using the committed `uv.lock` in turn ensures the same resolved package versions for the applicable Python version and platform.

> Just out of curiosity, did you run into a dependency issue on your side where locking the environment was necessary?

Yes, I have encountered dependency incompatibilities due to missing lock files several times in other projects. The intention of this PR was to establish a reproducible baseline.

While adding it, `uv` did expose the existing `xenavalkyrie`/`pypacker` metadata issue, although that is not by itself the reason for introducing a lockfile.

ozan956:
Thanks, I got the idea. 

I initially thought we were also going to replace some of the existing CI with uv to gain some speed there as well.

I am still wondering a bit about the value and maintenance of the locked baseline specifically for labgrid though.

For a regular user who is not using uv, the main advantage essentially that maintainers have a known-good dependency snapshot to reproduce when debugging dependency regressions, right?

If that is the main use case, could we not simply store the resolved dependency set from the existing CI with `pip freeze` and use that as the reference when a previously passing job starts failing? That would let us see exactly which dependency versions changed without introducing another locked CI path. And since we have scheduled jobs, we can keep it as artifact or within the log.

It would also avoid having to maintain an additional lockfile and decide when and how that baseline should be updated. If the main goal is diagnostic, keeping the resolved dependency set as part of the CI output/artifacts seems simpler to me.

tests/test_coordinator.py
ozan956: > This test was necessary to not decrease the test coverage 

I think that is actually part of my concern here :)

I would not use keeping the coverage percentage unchanged as the reason for adding a test. 

Coverage is useful for pointing us towards potentially untested behavior, but ideally the test itself should exist because there is a meaningful behavior or regression we want to protect.

If these "edge cases" are worth testing, then absolutely, but I would prefer having them in a focused PR explaining which behavior they protect. If the only reason for adding them here is to keep Codecov from dropping slightly, I would rather accept the small coverage change than couple unrelated tests to the PR.

rpoisel:
> If that is the main use case, could we not simply store the resolved dependency set from the existing CI with `pip freeze` and use that as the reference when a previously passing job starts failing? That would let us see exactly which dependency versions changed without introducing another locked CI path. And since we have scheduled jobs, we can keep it as artifact or within the log.
> 
> It would also avoid having to maintain an additional lockfile and decide when and how that baseline should be updated. If the main goal is diagnostic, keeping the resolved dependency set as part of the CI output/artifacts seems simpler to me.

Your suggestion would certainly help with diagnosing dependency changes, but [uv.lock](https://docs.astral.sh/uv/concepts/projects/layout/#the-lockfile) serves a different use case.

`pip freeze` records what happened to be installed in one particular Python/platform environment. It does not provide the same reproducible baseline: it is not a solver-generated lockfile and does not contain artifact hashes or a cross-platform resolution.

A committed `uv.lock` instead defines the dependency resolution associated with the source revision, allowing developers and CI to recreate a known, tested environment.

The lockfile does require periodic updates, but that is also part of its value: dependency changes become explicit and reviewable instead of happening implicitly during installation.

tests/test_coordinator.py
rpoisel: I just removed this test from this PR to add it in a separate PR if needed.

Bastian-Krause:
I don't see us maintaining pinned dependencies. Back in the autobahn/crossbar days we had to pin and that was always a pain to maintain. Since labgrid is mainly a library, I consider minimum requirements the correct approach.

Also I don't consider reproducibility, auditability and cached package installs as goals for labgrid itself. We're already struggling with maintenance resources. If you need those things, just maintain such a lock file yourself.

rpoisel:
> I don't see us maintaining pinned dependencies. Back in the autobahn/crossbar days we had to pin and that was always a pain to maintain. Since labgrid is mainly a library, I consider minimum requirements the correct approach.

Okay, I'll close this PR then.

> We're already struggling with maintenance resources. 

Please let me know if I can help. Now that I know the labgrid framework quite well and use it professionally in my work with clients, I see this as an opportunity to give something back. I’d be willing to dedicate some of my professional time to helping maintain the framework.

Bastian-Krause:
> Please let me know if I can help. Now that I know the labgrid framework quite well and use it professionally in my work with clients, I see this as an opportunity to give something back. I’d be willing to dedicate some of my professional time to helping maintain the framework.

Reviewing PRs with the overall labgrid architecture in mind would certainly help. We'll try to write some guidelines, so it's easier to see if a PR aligns with labgrid's general direction.

rpoisel:
> Reviewing PRs with the overall labgrid architecture in mind would certainly help. We'll try to write some guidelines, so it's easier to see if a PR aligns with labgrid's general direction.

That sounds good. Thanks! If you want me to help here, just let me know.

PR #1965, opened 2026-09-03
fix: parse LG_COORDINATOR_TLS value explicitly

asher-pem-arm:
**Description**
Parse LG_COORDINATOR_TLS explicitly so only true enables TLS. Values such as false, 0, or an empty string now correctly leave TLS disabled.
Updates documentation and adds regression tests for both labgrid-client and RemotePlace.

**Checklist**
- [x] Tests for the feature 

- [x] PR has been tested


Emantor (commented):
Changes look good to me, I wondering whether we should also support `LG_COORDINATOR_TLS=1`.

asher-pem-arm:
@Emantor updated to support LG_COORDINATOR_TLS=1 as well

Emantor (approved):

PR #1971, opened 2026-09-14
driver/shelldriver: disable line editing before injecting the run helper

belascholtes:
**Description**

On shells with line editing enabled (e.g. bash on a serial console with `TERM=linux`), command lines longer than the terminal width are echoed with a forced-wrap redraw: the character at the wrap boundary is followed by a bare CR and repeated. Log viewers that split on the bare CR then show the boundary character twice:

```
root@ector:~# MARKER='FXSU''AQJBOV' run '/usr/bin/zmqClientApp -t 30000 ipc:///va
ar/run/hgp/hgp-nm-wlan /tmp/hgp_shell_rpc_req.json'
```

This reads as `ipc:///vaar/run/...` although the executed command correctly contained `ipc:///var/run/...`. It is only cosmetic (the marker protocol is unaffected), but the duplicated character regularly misleads readers of the logs into chasing typos that don't exist.

This PR disables the shell's line editing (`set +o emacs; set +o vi`) once per session in `_inject_run()`, so long command lines echo as plain tty line-discipline output. This only changes how the shell echoes its input; unlike alternatives such as widening the terminal with `stty cols`, it is not observable by programs running on the target (a wider winsize changes e.g. unpiped `ps` output truncation, which could affect existing tests that match on it). The only residual effect is that a human attaching interactively to the same shell session afterwards has line editing off until the next login (`set -o emacs` restores it).

Verified on real hardware (i.MX6 board, bash on an 80-column serial console): long command echoes come back as clean single lines, and `run()` output parsing and exit codes are unaffected. `pytest`, `pylint` and `ruff format --check` pass locally.

**Checklist**

- [x] PR has been tested (on-target verification on an i.MX6 serial console as described above, plus the unit test suite)


ozan956 (commented):
Hey @belascholtes,

Thanks for the PR!

I understand the problem you are trying to solve here, but I am not convinced this needs to become the default behavior of `ShellDriver`.

Would not it be cleaner to handle this in the test suite which needs it? For example, you could disable line editing when entering the relevant test/session and restore it again when leaving. That way, the workaround is fully controlled by the consumer and the shell is also left in a known good state for whoever uses it afterwards.

In general, I think keeping `ShellDriver` as plain as possible is valuable. Users can always adjust the shell behavior for their particular setup, but making this the default means every consumer inherits that policy whether they need it or not.

What is the advantage for your use case of having this inside labgrid itself, compared to doing the same setup/cleanup explicitly in the script or test suite which needs it?

belascholtes:
Hi @ozan956, thanks for taking a look!

Some context: a test session for us involves many reboots, and the shell state resets with each one. The ShellDriver already injects its helper on every login, so it felt like the natural place to cover every fresh shell automatically. I didn't see a clean way to do this from the test suite without tracking every reboot. Could you give me a hint where you'd hook it in?

From my understanding the change doesn't affect testing behavior, only the serial log, which is why I thought it was worth sharing as a default. But I may be missing use cases. Would it make sense to make this opt-in behind an argument instead?

belascholtes:
To answer your question about the advantage directly: the duplicated characters in the serial logs are gone for everyone with a standard shell setup, not just for our suite.

ozan956:
Hey @belascholtes,

I get the idea. Just out of curiosity, why do you need to reboot so often during testing? Is this something like a boot-time test or boot power measurement, where the DUT resets in a way that you cannot reliably detect from software? Or is it like all-in-one test?

My main concern is that `ShellDriver` is used in several different workflows, not only automated test execution.

For example, CLI users may use labgrid strategies, and I think there is an example case where making this part of `_inject_run()` changes more than just the log output.

Consider the standard `ShellStrategy` together with:

```
labgrid-client -c env.yaml -p <place> -s shell console
```

`ShellStrategy.transition("shell")` eventually does:

```
self.target.activate(self.shell)
self.shell.run("systemctl is-system-running --wait")
```

so activating the `ShellDriver` would also execute the proposed:

```
set +o emacs
set +o vi
```

After the strategy transition, `labgrid-client` explicitly deactivates the active `ConsoleProtocol`:

```
# deactivate console drivers so we are able to connect with microcom later
con = target.get_active_driver("ConsoleProtocol")
target.deactivate(con)
```

and the `console` command then attaches the user.

So with the standard `ShellStrategy`, the flow can already be:

```
ShellStrategy -> shell
              -> ShellDriver activation
              -> _inject_run()
              -> line editing disabled
              -> strategy transition completes
              -> console handed to microcom
              -> user continues interactively
```

The same applies to the other strategies we have like `UBootStrategy` and `BareboxStrategy` as well.

In that case, the user attaching to the console inherits the modified shell state. I think thats the reason why we should keep the default `ShellDriver` behavior neutral.

For your test suite, I would still prefer applying this after entering the shell, in the fixture or helper that performs the reboot or state transition. If that is difficult with your test architecture, an explicit opt-in option would seem safer to me than changing the behavior unconditionally for every `ShellDriver` activation.

We also recently merged sysfs input GPIO support, so if the main difficulty is detecting whether the board was power-cycled, that might provide another way to track the reset/power-cycle state when it cannot be determined reliably from the DUT itself.

ozan956:
> To answer your question about the advantage directly: the duplicated characters in the serial logs are gone for everyone with a standard shell setup, not just for our suite.

Yeah, I agree with that. But sometimes we also need to think about the trade-offs :smile:

Do we really want to change the state of every shell just to fix a cosmetic logging issue? That feels like a fairly large side effect for something which is mainly about how the command echo looks in the logs.

I am glad you brought this issue up. I am just not convinced that changing the default state of every shell activated through ShellDriver is the right place to solve it.

belascholtes:
I completely agree, if this potentially breaks someone's workflow, it shouldn't be the default.

For us, both ways work. We can either solve it completely on our side, or I can adapt the PR to make it an opt-in for the ShellDriver (e.g. a `disable_lineediting` argument, defaulting to off). Which one would you prefer?

> Just out of curiosity, why do you need to reboot so often during testing?

It's a platform test suite for our firmware, so many tests involve reboots by design: some conditionally, e.g. to apply configurations or to test the RTC, others unconditionally, e.g. to test the watchdog.

ozan956:
Thanks for the context!

I would prefer keeping this on your side rather than adding another option to the driver.

For example, you could either have a small custom strategy which applies the shell setup whenever you transition to the console, or more simply keep it in your test suite and run it whenever a fresh shell is entered after a reboot.

----

My concern with adding `disable_lineediting` as an option is that this can easily become a growing list of shell-specific knobs, and the driver itself should not be concerning about it. So lets keep ShellDriver generic :smile:

belascholtes:
OK, thanks for taking the time to think it through!

ozan956:
Thanks again for bringing this up.

Even when a particular implementation does not end up being merged, raising the problem and discussing the trade-offs is still very valuable.

Keep the patches coming!

